PR #2, opened 2015-10-28
Get symbol names from documents

robrix:
A minor convenience.

### DIFF
--- a/src/runtime/language.c
+++ b/src/runtime/language.c
@@ -1,4 +1,4 @@
-#include "runtime/parser.h"
+#include "tree_sitter/parser.h"
 
 size_t ts_language_symbol_count(const TSLanguage *language) {
   return language->symbol_count;
--- a/src/runtime/language.c
+++ b/src/runtime/language.c
@@ -1,5 +1,9 @@
 #include "runtime/parser.h"
 
+size_t ts_language_symbol_count(const TSLanguage *) {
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
I'm thinking that in light of the issues you raise in https://github.com/maxbrunsfeld/tree-sitter/issues/3, maybe this method should be on the language, i.e. `ts_language_symbol_name`. What do you think?

robrix:
Good call :+1:

robrix:
This now implements `ts_language_symbol_name` and `ts_language_symbol_count`.

src/runtime/language.c
```diff
@@ -0,0 +1,9 @@
+#include "runtime/parser.h"
```
maxbrunsfeld: I think this should be `#include "tree_sitter/parser.h"` (the header in the `include` directory), since `struct TSLanguage` is defined there. Sorry, it's a bit confusing to have two files called `parser.h`. I think this happens to work because `tree_sitter/parser.h` gets transitively included through a couple of steps.

robrix: Good catch!

robrix: :white_check_mark:

maxbrunsfeld:
Thanks!

robrix:
:bow: thank you!

PR #6, opened 2015-11-08
Fix handling of changing precedence within lexical rules.

maxbrunsfeld:
A precedence annotation wrapping a sequence of characters now only affects how
tightly those characters bind to *each other*, not how tightly they bind to the
preceding character.

This bug surfaced because a generated lexer was failing to recognize a `\n` character as a token, instead treating it as ubiquitous whitespace. It made this error because, even though anonymous ubiquitous tokens have the lowest possible precedence, the character immediately *after* the `\n` was part of a normal token, which had *normal* precedence (0). So the action of advancing into that following token was incorrectly prioritized above the correct action: accepting the line-break token.

Refs https://github.com/robrix/tree-sitter-swift/pull/9


robrix:
Oh wow, this was way subtler than I understood. Thank you @maxbrunsfeld, I don’t think I would’ve stood a chance of diagnosing this one.

PR #8, opened 2015-11-20
Handle ambiguities where each interpretation expects different lexical tokens

maxbrunsfeld:
Refs https://github.com/robrix/tree-sitter-swift/pull/11


robrix:
:tada:

PR #9, opened 2015-11-30
Track row/column ranges in TSNodes

joshvera:
Adds `ts_node_start_point` and `ts_node_end_point` to extract row and column information from TSNodes.


maxbrunsfeld:
Looks like CI is failing because `length.h` uses C struct literal syntax (`(TSPoint){ .row = row, .column = column }`, which doesn't work when that header is included in a C++ file unless you're using clang, which adds that syntax to C++ as an extension. I have travis building with gcc, because I make this type of error all the time. I think if you convert it to something like this it'll be fine (even though it's a lot less pretty):

```c
TSPoint result;
result.row = row;
result.column = column;
return result;
```

maxbrunsfeld:
Actually, I think the `ts_point` functions should probably be in a separate header file, `point.h`, since they aren't related to `TSLength`

src/runtime/lexer.c
```diff
@@ -53,6 +53,10 @@ static void ts_lexer__start(TSLexer *self, TSStateId lex_state) {
 static void ts_lexer__start_token(TSLexer *self) {
   LOG("start_token chars:%lu", self->current_position.chars);
   self->token_start_position = self->current_position;
+
+  LOG("start_token row:%lu", self->current_point.row);
+  LOG("start_token column:%lu", self->current_point.column);
```
maxbrunsfeld: Could you put these in the same log message, like `start_token chars:%lu, row:%lu, column:%lu`?

spec/runtime/node_spec.cc
```diff
@@ -43,6 +43,12 @@ describe("Node", []() {
       AssertThat(ts_node_pos(array_node).bytes, Equals<size_t>(2));
       AssertThat(ts_node_size(array_node).bytes, Equals<size_t>(25));
 
+      AssertThat(ts_node_start_point(array_node).row, Equals<size_t>(0));
+      AssertThat(ts_node_start_point(array_node).column, Equals<size_t>(2));
+
+      AssertThat(ts_node_end_point(array_node).row, Equals<size_t>(0));
+      AssertThat(ts_node_end_point(array_node).column, Equals<size_t>(27));
+
```
maxbrunsfeld: I think that since each `ts_node` function does a little bit of work to keep track of nodes' rows and columns, we need to add assertions about `ts_node_start_point` and `ts_node_end_point` to all of these tests.

Could you change the input string on line 15 to contain some line breaks, so that there'll be some non-trivial row/column computations? Maybe turn it into a pretty-formatted multi-line JSON array, like this:

```c
ts_document_set_input_string(
  document,
  "\n"
  "\n"
  "[\n"
  "  123,\n"
  "  false,\n"
  "  {\n"
  "    \"x\": null\n"
  "  }\n"
  "]"
);
```

This will require changing the `ts_node_pos` assertions.

src/runtime/lexer.c
joshvera: :+1:

spec/runtime/node_spec.cc
joshvera: :+1:

maxbrunsfeld:
:fireworks: Awesome! Thanks for diving in and adding this @joshvera!

robrix:
:clap:

PR #12, opened 2015-12-05
Clean up API a bit after the addition of row/column tracking

maxbrunsfeld:
* [x] Make 1D and 2D node accessor functions more consistent with each other.
  * Replace `ts_node_pos` with `ts_node_start_char` and `ts_node_start_byte`, to be more consistent with `ts_node_start_point`.
  * Replace `ts_node_size` with `ts_node_end_char` and `ts_node_end_byte`, to be more consistent with `ts_node_end_point`.
* [x] Eliminate the `TSLength` struct from the API. Bytes and characters are now treated separately everywhere.
* [x] Correct the spelling of the word 'descendant' in the API.
* [x] Remove duplication in the internal logic for handling 1D and 2D coordinates.

/cc @joshvera @robrix - The minor API changes in the first bullet point will affect you. I think the new functions will be more convenient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,18 +137,18 @@ int main() {
 
   TSNode root_node = ts_document_root_node(document);
   printf(
-    "Root name: %s, position: %lu, size: %lu\n",
+    "Root name: %s, start: %lu, end: %lu\n",
     ts_node_name(root_node, document),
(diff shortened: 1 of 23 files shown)


robrix:
Yeah, this will tidy things up a little bit. Thank you!

PR #14, opened 2015-12-09
Fix more errors in incremental subtree reuse

maxbrunsfeld:
This makes the randomized edits in the corpus specs truly random, and fixes some additional incremental parsing bugs revealed by the random edits.


robrix:
:tada:

PR #20, opened 2015-12-28
Support UTF16 directly

maxbrunsfeld:
This makes the API easier to use from javascript.

### DIFF
--- a/include/tree_sitter/runtime.h
+++ b/include/tree_sitter/runtime.h
@@ -8,10 +8,16 @@ extern "C" {
 #include <stdlib.h>
 #include <stdbool.h>
 
+typedef enum {
+  TSInputEncodingUTF8,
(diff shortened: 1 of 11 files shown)


maxbrunsfeld:
@robrix If you're using `ts_document_set_input_string`, that API hasn't changed; it still expects a UTF8 string. But the more general `ts_document_set_input` function now allows you to specify either UTF8 or UTF16 as the encoding.

robrix:
:+1: Yeah, wasn’t thinking it’d break us, but more that we have an extra strategy available for handling UTF16 inputs.

maxbrunsfeld:
Ah, sorry for misconstruing. Yeah, If UTF16 support is useful for you guys, let me know if you think we should add an `encoding` argument to `set_input_string` as well.

PR #25, opened 2016-01-29
Fix memory leaks

maxbrunsfeld:
:warning: WIP :non-potable_water: :warning:

### DIFF
--- a/src/runtime/tree.c
+++ b/src/runtime/tree.c
@@ -62,13 +62,21 @@ TSTree *ts_tree_make_copy(TSTree *self) {
 }
 
 void ts_tree_assign_parents(TSTree *self) {
-  TSLength offset = ts_length_zero();
+  TSLength offset;
+
+recur:
+  offset = ts_length_zero();
   for (size_t i = 0; i < self->child_count; i++) {
     TSTree *child = self->children[i];
     if (child->context.parent != self) {
       child->context.parent = self;
       child->context.index = i;
       child->context.offset = offset;
+      if (i == self->child_count - 1) {
+        self = child;
+        goto recur;
+      }
+
       ts_tree_assign_parents(child);
     }
     offset = ts_length_add(offset, ts_tree_total_size(child));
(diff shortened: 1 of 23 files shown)


robrix:
:metal:

src/runtime/tree.c
```diff
@@ -132,13 +141,25 @@ void ts_tree_retain(TSTree *self) {
...
+      for (size_t i = 0; i < self->child_count - 1; i++)
+        ts_tree_release(self->children[i]);
+      TSTree *last_child = self->children[self->child_count - 1];
       ts_free(self->children);
+      ts_free(self);
+
+      self = last_child;
+      goto recur;
+    }
+
```
maxbrunsfeld: This manual tail call optimization, and the one above in `ts_tree_assign_parents`, make the stack depth a lot more predictable when parsing large files:

before:
before-tail-call-opt

after:
after-tail-call-opt

For reasonable inputs, the stack depth is now effectively constant, since there is an upper bound on the depth of real programs' concrete syntax trees. It is still possible for a malicious user to create a pathological file consisting only of thousands of nested parentheses, or something like that. In that case, the stack could overflow. To guard agains that, I think it would be sufficient to track the depth explicitly, and just error out if an unreasonably deep tree is encountered. I'll leave that to a later PR though.

robrix:
Fantastic visualization of the improvement. Nice one @maxbrunsfeld!

maxbrunsfeld:
OK. No more leaks, and no more green builds in the presence of leaks.

```
Test run complete. 12099 tests run. 12099 succeeded.
==9067== 
==9067== HEAP SUMMARY:
==9067==     in use at exit: 44,826 bytes in 229 blocks
==9067==   total heap usage: 4,317,565 allocs, 4,317,336 frees, 826,652,402 bytes allocated
==9067== 
==9067== LEAK SUMMARY:
==9067==    definitely lost: 0 bytes in 0 blocks
==9067==    indirectly lost: 0 bytes in 0 blocks
==9067==      possibly lost: 0 bytes in 0 blocks
==9067== 
==9067== For counts of detected and suppressed errors, rerun with: -v
==9067== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 18 from 18)
```

PR #26, opened 2016-02-17
Overhaul error recovery

maxbrunsfeld:
* [x] Switch to the more powerful 'forward-move'-based error recovery algorithm described in [1] and [2]
* [x] When consuming an invalid token, change lexer to just return the unexpected token, not a special error token. This will reduce the amount of unnecessary trees that are allocated and then discarded as errors are encountered on some versions of the parse stack.
* [x] When multiple equally-good recoveries exist, make the choice as predictably as possible.
* [x] Fix edge cases in error lexing - when an error is encountered, rewind back to the start position and re-tokenize from the error state.
* [x] Fix test failures due to incremental tree reuse bugs during error recovery.
* [x] Compress the parse table to make up for the additional 'recover states', which would otherwise make the parse table even larger (https://github.com/tree-sitter/tree-sitter/pull/32).
* [x] Switch to a state-matching based algorithm for incremental node reuse. With this new recovery algorithm, state matching is the only sufficient criteria that I know of for reusing a subtree containing errors. The fact that the parse tables will now be significantly compressed should make the state-matching approach effective enough to just use all the time.
* [x] During 'condensation phase', perform all possible reductions, not just one set of reductions. This will make recoveries better in some cases, but could probably be done in a separate PR.
* [x] Add the ability to skip tokens at the end of the file.
* [x] Improve the logic for abandoning erroneous parse version. Currently, too many versions are retained for too long, causing poor performance for certain types of parse errors.

[1] [Error Detection and Recovery in LR Parsers](http://what-when-how.com/compiler-writing/bottom-up-parsing-compiler-writing-part-13/)
[2] [Error recovery for LR Parsers](http://www.dtic.mil/dtic/tr/fulltext/u2/a043470.pdf)


maxbrunsfeld:
/cc @robrix @joshvera 

This PR changes the format of the generated parsers such that the old parsers won't work with the new tree-sitter and vise versa. It's also just a big change in general; there could be bugs. But now errors should be handled much more gracefully and the parse tables will have a lower memory footprint.

robrix:
Nice! Looking forward to playing with this.

PR #29, opened 2016-02-19
Expose nodes' wrapper symbols

maxbrunsfeld:
/cc @robrix

### DIFF
--- a/include/tree_sitter/runtime.h
+++ b/include/tree_sitter/runtime.h
@@ -8,6 +8,10 @@ extern "C" {
 #include <stdlib.h>
 #include <stdbool.h>
 
+typedef unsigned short TSSymbol;
+typedef struct TSLanguage TSLanguage;
+typedef struct TSDocument TSDocument;
+
 typedef enum {
   TSInputEncodingUTF8,
   TSInputEncodingUTF16,
@@ -46,9 +50,11 @@ typedef struct {
   size_t offset[3];
 } TSNode;
 
-typedef unsigned short TSSymbol;
-typedef struct TSLanguage TSLanguage;
-typedef struct TSDocument TSDocument;
+typedef struct {
+  TSSymbol value;
+  bool done;
+  void *data;
+} TSSymbolIterator;
 
 size_t ts_node_start_char(TSNode);
 size_t ts_node_start_byte(TSNode);
@@ -57,6 +63,8 @@ size_t ts_node_end_char(TSNode);
 size_t ts_node_end_byte(TSNode);
 TSPoint ts_node_end_point(TSNode);
 TSSymbol ts_node_symbol(TSNode);
+TSSymbolIterator ts_node_symbols(TSNode);
+void ts_symbol_iterator_next(TSSymbolIterator *);
 const char *ts_node_name(TSNode, const TSDocument *);
 char *ts_node_string(TSNode, const TSDocument *);
 bool ts_node_eq(TSNode, TSNode);
@@ -90,7 +98,7 @@ TSNode ts_document_root_node(const TSDocument *);
 size_t ts_document_parse_count(const TSDocument *);
 
 size_t ts_language_symbol_count(const TSLanguage *);
-const char *ts_language_symbol_name(TSLanguage *, TSSymbol);
+const char *ts_language_symbol_name(const TSLanguage *, TSSymbol);
 
 #define ts_builtin_sym_error 0
 #define ts_builtin_sym_end 1
--- a/spec/helpers/tree_helpers.cc
+++ b/spec/helpers/tree_helpers.cc
@@ -1,4 +1,6 @@
 #include "helpers/tree_helpers.h"
+#include "runtime/document.h"
+#include "runtime/node.h"
 #include <ostream>
 
 using std::string;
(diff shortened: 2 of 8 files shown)


robrix:
Nice! So when we’re parsing within `_expression → _operation → _comparison → greater-than`, we can walk upwards from `greater-than` and build the chain for this particular node (`greater-than` is a `comparison` is an `operator` is an `expression`). Very cool :smile:

maxbrunsfeld:
Yeah, exactly. I was thinking the JS binding to this API would return an array `['_expression', '_operation', '_comparison', '_greater_than']`, but w/ the iterator we can avoid performing any intermediate allocations.

maxbrunsfeld:
@robrix Now that we have this concept, I think it might make sense to drop the leading underscores from the names of productions like [`_statement`](https://github.com/tree-sitter/tree-sitter-javascript/blob/c51c8d4172e50201a56f4390b7c59c0ca58967b9/grammar.js#L43) and [`_expression`](https://github.com/tree-sitter/tree-sitter-javascript/blob/c51c8d4172e50201a56f4390b7c59c0ca58967b9/grammar.js#L218), and tweak the `TSNode` API such that fixed-arity nodes with only one child are always *combined* in this same way.

In other words, for this grammar subset:

```js
rules: {
  expression: $ => choice(
    $.sum,
    $.identifier,
    // ...
  ),

  sum: $ => seq($.expression, '*', $.expression),

  identifier: $ => (/\w+/),

  // ...
}
```

and the input: `x + y`, you'd get this AST:

```
(expression.sum
  (expression.identifier)
  (expression.identifier))
```

instead of this AST:

```
(expression (sum
  (expression (identifier))
  (expression (identifier))))
```

That way, the 'private rule' concept would be reserved for only rules that are truly hidden from the API. What do you think? Can you see any problems with just *always* squashing fixed singleton nodes like that?

maxbrunsfeld:
I think that I will merge this in the meantime, with just the addition of `ts_node_symbols` (no changes to the way the s-expressions are generated in `ts_node_string`, so that existing tests don't need to be updated yet).

robrix:
> That way, the 'private rule' concept would be reserved for only rules that are truly hidden from the user of the parser, and you can create a nice clean syntax tree by default, without needing to use the underscore. What do you think? Can you see any problems with just always squashing fixed singleton nodes like that?

I think this is a really cool idea! :metal:

The only thing that comes to mind is that e.g. in a grammar with statements and expressions, we might (or might not?) want to exclude expressions from being judged as statements:

```js
rules: {
  statement: $ => choice(
    $.if,
    …,
    $.expression
  ),
  if: $ => seq("if", $.condition, '{', repeat(seq($.statement, ';')), '}'),
  condition: $ => $.expression,
  expression: $ => choice(
    $.identifier,
    $.sum,
    …
  ),
  sum: $ => seq($.expression, '+', $.expression),
  …
}
```

Since is-a relationships are related exclusively by the structure of the (internal) parse tree, we’d have `expression`s with different qualified names depending on where they originated:

```
if shouldAdd {
  a + (b + c)
}
```

→

```
(program
  (statement.if (condition.expression.identifier)
    (statement.expression.sum (expression.identifier) (expression.sum (expression.identifier) (expression.identifier)))))
```

It’s clear that `sum`s and `identifier`s _are_ `expression`s, and that `if`s are `statement`s. But it seems like `condition`s _have_ `expression`s.

Maybe a slight tweak to the approach & another to the grammar would resolve this:

1. The wrapping of a single production in another forms a _has-a_ relationship, not an _is-a_ one. So in this case the `condition` rule would be surfaced as a node:

        ```js
        (statement.if (condition (expression.identifier)) …)
        ```

2. Using that, we can then tweak the grammar to introduce an `expression_statement` rule. We can further make it private if we don’t want it to surface in the parse tree either as part of the _is-a_ chain for the `expression` it contains or as a node in its own right.

I think this would resolve the issue adequately, as it gives one a proper escape hatch.

As an alternative, we might be able to trace the paths from which a given production could arise and then judge them to be _is-a_ iff they’re unique. However, I think that could be problematic in cases where e.g. the left hand side of an assignment can only be a restricted subset of expressions:

```js
rules: {
  expression: $ => choice(
    $.identifier,
    $.assignment,
    …
  ),
  assignment: $ => seq($._lhs, '=', $.expression),
  _lhs: $ => choice($.identifier, $.member, $.subscript),
  …
}
```

We would still want to recognize `identifier`, `member`, and `subscript` as being expressions; they’re just not used as such in `_lhs`.

----

All that aside, I wonder if it’ll be sufficiently apparent to end users that nodes in this format are specifying an _is-a_ relationship rather than a _has-a_ relationship. For the sake of looking at a concrete alternative, an explicit API might look like this:

```js
rules: {
  statement: $ => isA(choice(
    $.if,
    …,
    $.expression_statement
  )),
  …
  condition: $ => $.expression,
  expression: $ => isA(choice(
    $.identifier,
    $.sum,
    …
  )),
}
```

So: default to _has-a_, with the _is-a_ combinator taking a single argument and wrapping it appropriately.

Another explicit-API alternative:

```js
rules: {
  expression: $ => {
    identifier: (/\w+/),
    member: seq($.expression, '.', $.expression.identifier),
    subscript: seq($.expression, '[', $.expression, ']'),
    …
  },
  …
  _lhs: $ => choice($.expression.identifier, $.expression.member, $.expression.subscript),
  …
}
```

This is perhaps the most explicit, while not requiring the addition of any new API, only changes to extant API.

PR #30, opened 2016-02-24
Fix bugs in ambiguity management

maxbrunsfeld:
Fixes an issue discovered in https://github.com/tree-sitter/tree-sitter-ruby/pull/16

#### Debugging

This adds a `ts_document_print_debugging_graphs` function. When this boolean flag is set on the document, it dumps a [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) representation of the graph-structured parse stack after every parse action. This output can be piped directly to `dot`to produce a sequence of graphs and parse actions, like this:

screen shot 2016-02-23 at 5 41 04 pm

#### Bug fix

This fixes a bug wherein ambiguities were merged incorrectly, producing nonsensical parse trees.


robrix:
:metal: that’s pretty fantastic work dude!

PR #32, opened 2016-07-05
Parse table compression

maxbrunsfeld:
The goal here is to significantly reduce the number of states in the generated parsers. This will shrink the the static memory footprints of the parsers and make them compile faster.

The approach is to merge similar parse states by allowing reductions to take place in the presence of lookahead tokens which are invalid. This preserves the behavior that errors are detected as soon as they occur; it just means that the parser will proceed to build up more structure to the left of an error before beginning error recovery.

I held off doing this optimization for a long time because I wasn't sure how it would affect the error recovery algorithm, but now that https://github.com/tree-sitter/tree-sitter/pull/26 is mostly fleshed out, I can tell that it won't be a problem.

Currently, this makes parser generation even slower than it already is, because the states are merged in a post-processing step. It's probably possible to do this in a way that actually makes it *faster*, but it would be more complicated, and I'm less concerned about parser generation speed than I am about memory footprint and compile time, so I'm going to worry about optimizing parser generation later.

### DIFF
--- a/spec/helpers/random_helpers.cc
+++ b/spec/helpers/random_helpers.cc
@@ -1,0 +1,35 @@
+#include <string>
+#include <stdlib.h>
+
+using std::string;
+
+static string random_string(char min, char max) {
+  string result;
+  size_t length = random() % 12;
+  for (size_t i = 0; i < length; i++) {
+    char inserted_char = min + (random() % (max - min));
+    result += inserted_char;
+  }
+  return result;
+}
+
+static string random_char(string characters) {
+  size_t index = random() % characters.size();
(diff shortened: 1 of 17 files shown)


maxbrunsfeld:
There are still some optimizations that could be done to the parse tables, but this more than makes up for the cost of the "recovery" parse states that I've added to the parse table for https://github.com/tree-sitter/tree-sitter/pull/26.

Here are some results for the JavaScript parser:

**Master** - 1.4 MB, 4287 states
**New error recovery, no compression** - 1.9 MB, 5689 states
**New error recovery, compression** - 934 kb, 2433 states

PR #33, opened 2016-08-23
Add an API to allow multiple grammar rules to produce the same type of node

maxbrunsfeld:
It's sometimes desirable to have two different grammar rules produce the same type of syntax tree node. I'm generally hesitant to complicate the grammar API in any way, so I'll justify this feature with a few examples that have come up recently:

1. Special keys in JavaScript objects literals:

  The keywords `get` and `set` can be used as keys in JS objects literals. But in the same position, they can also have special syntactic meaning as part of a [method definition](http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions). So whether the string `get` should be interpreted as an `identifier` token or as a specific `get` token depends on what comes *next* in the text.

  The way I've dealt with this for right now is to [redefine `identifier`](https://github.com/tree-sitter/tree-sitter-javascript/commit/cbb743a3dff45635b21d5a80c7d72922f92e3f0d#diff-95fc33185b7215010b60f43469000a7aL471) as a syntactic rule that includes the `get` and `set` tokens in addition to the usual identifier regex token. But this has a small performance cost when parsing: every regular identifier has be first tokenized as a `normal_identifier`, then reduced via the `identifier -> normal_identifier` production. It also is not as 'local' of a solution as I would prefer. The original problem was about object literals, and I had to go change `identifier` to deal with it.

  What I would prefer to do instead is redefine `pair` (the rule that defines key-value pairs in object literals), such that the `get` and `set` tokens can be used as keys:

  ```js
    pair: $ => seq(
      choice($.string, $.number, $.identifier, $.get_set_identifier),
      ':',
      $._expression
    ),

    get_set_identifier: $ => choice('get', 'set'),
  ```

  This would perform better, is more 'localized' to object literals, and it would work fine today. The problem with this solution is that `{foo: 5}` and `{get: 5}` would then produce two different syntax trees: one would have `identifier`, and one would have `get_set_identifier`. Applications could just deal with it, but it feels like an implementation detail of the parser that's unnecessarily leaked.

  #### Enter aliases
  
  After this PR, we'll be able to adopt the solution above, and add this to the grammar:

  ```js
  aliases: {
    get_set_identifier: $ => $.identifier,
  },
  ```
  
  Then, the word `get` in `{get: 5}` will appear as in `identifier` in the tree, even though the parser handled it differently than a normal identifier.

2. JS statements with omitted semicolons.

  In the JS grammar, certain types of statements are required to end with semicolons (or newlines), except when they are the last statement in a block. This requires some rules to be mostly duplicated (e.g. `return_statement` vs `trailing_return_statement`). Using `aliases`, we could hide the distinction between these two rules, making them both appear as `return_statement`.

3. Special array-of-string / array-of-symbol rules in Ruby

  See https://github.com/tree-sitter/tree-sitter-ruby/issues/6. Using aliases, we could make the 'bare strings' in these literals appear as normal `string` nodes in the syntax tree.

/cc @robrix @joshvera @rewinfrey


robrix:
I _think_ I follow this, but I’d like to be sure. So:

- When should one use an alias vs. mark it up as an expected conflict?
- Are aliases applicable to nonterminal productions as well?

maxbrunsfeld:
> When should one use an alias vs. mark it up as an expected conflict?

So the alias thing is purely cosmetic; it just means, for example "I want to rename `bare_string` to `string`; so that it appears as a `string` in the syntax tree".

Expected conflicts are more related to the parsing algorithm; they let you enumerate cases where it's ok that more than one token of lookahead is required in order to distinguish between certain rules.

> Are aliases applicable to nonterminal productions as well?

Yeah; they let you rename any named rule.

After sleeping on this, I'm already leaning towards thinking it's too complicated to be worth the trouble.

maxbrunsfeld:
I think I just now understood your question about expected conflicts better. The problem of `get` and `set` as object literal keys might seem related to expected conflicts because it constitutes a kind of ambiguity: is "get" an identifier, or a keyword? Expected conflicts allow you to opt in to resolving ambiguities at runtime.

But because this 'ambiguity' occurs at the lexical level (which kind of token to produce) rather than at the syntactic level, "get" is immediately tokenized as the word `get`, rather than as an identifier; we can't defer this decision like we can with syntactic ambiguity.

Of course, it would probably be possible for tree-sitter to handle this type of ambiguity automatically, by detecting lexical ambiguities like this. But these types of lexical conflicts are extremely common, and the vast majority of the time, the solution is to definitely treat the token as a keyword. AFAICT the case of `get` and `set` is somewhat uncommon, so there would be an undue performance cost if we were to always treat keywords as possibly being identifiers.

Issue #34, opened 2016-09-03
Thoughts on detecting invalidated ranges


nathansobo:
I've been thinking all morning about the problem of detecting invalidated ranges. A key realization I've had is that the problem is very different from computing a diff because we don't need to describe the nature of the change, but just its range. That means we can walk the trees linearly and record invalidations whenever we encounter differences between the children of a given node, even if we don't know the cause of said differences. A slight wrinkle is that we only want to consider differences in the subset of nodes that affect syntax highlighting for purposes of invalidation. I think the following pseudocode might get us close:
### `recordInvalidatedRanges(oldNode, newNode, ranges)`

Starting at the root of the old and new trees:
- Are these nodes structurally equivalent (both nodes are non-null, have the same type and spatial footprint)?
  - Yes:
    - Does the node in the old tree contain a textual change?
      - No:
        - Terminate the recursion.
      - Yes:
        - Call `recordInvalidatedRanges` recursively on corresponding children of both nodes. If the number of children differ, pass null to represent the missing child from the shorter list.
  - No:
    - Do either of the nodes have a type that affects highlighting?
      - Yes:
        - Union the spatial footprint of both nodes and add it to the invalidation list, then terminate the recursion.
      - No:
        - Collect the descendants of both nodes that affect highlighting and call `recordInvalidatedRanges` recursively on the collected lists. If the length of these lists differ, pass null to represent the missing child from the shorter list.

nathansobo:
My thoughts have evolved, here's a new sketch.

Define an iterator over syntax trees that implements two operations:

* `next` Moves to the next sibling node. If no next sibling exists, ascends to the parent and go to its next sibling, recursively. Returns false at the end of the tree, otherwise true.
* `descendLeft` Go to the first child of the current node, if one exists. Returns false if no first child exits, otherwise true.

Our task is to fully consume both iterators starting at the root node in the following way.

Compare the current node of both iterators.

* If the nodes have the same type and range and the old tree's node has no changes, call `next` on both iterators and repeat the loop.
* Otherwise an invalidation may occur at or within these nodes:
  * Call `descendLeft` on both iterators until encountering a node with a type that affects syntax highlighting or reaching a leaf node.
  * Compare nodes from either tree. If equivalent, call `next` on both iterators and repeat the loop.
  * If the nodes are different, record an invalidation for the range of one or both of the nodes, depending on whether their type affects syntax highlighting. Call `next` on the iterator whose current node ends first and repeat the loop.

[...]

PR #38, opened 2016-09-09
Report text ranges whose syntactic meaning have changed after re-parsing

maxbrunsfeld:
This will allow the [`TreeSitterDecorationLayer`](https://github.com/atom/tree-sitter-syntax/blob/66695c8605a5b2872d6706b9ab4b467b6f7708a7/lib/tree-sitter-decoration-layer.js) to implement the `DisplayLayer's` [`getInvalidatedRanges()`](https://github.com/atom/text-buffer/blob/b7a82b80afd6e1b8cd9d37288849982ad2204731/src/display-layer.coffee#L244) method more intelligently, so that we can avoid recomputing every onscreen line after an edit.

Refs https://github.com/tree-sitter/tree-sitter/issues/34

### DIFF
--- a/include/tree_sitter/parser.h
+++ b/include/tree_sitter/parser.h
@@ -17,8 +17,7 @@ typedef unsigned short TSStateId;
 typedef struct {
   size_t bytes;
   size_t chars;
-  size_t rows;
-  size_t columns;
+  TSPoint extent;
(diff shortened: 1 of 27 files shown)


joshvera:
👍

PR #37, opened 2016-09-07
API cleanup

maxbrunsfeld:
* Rename `TSDebugger`, -> `TSLogger`
* Rename `ts_node_descendant_for_range` -> `ts_node_descendant_for_char_range`
* Introduce `ts_node_descendant_for_point_range` and `ts_node_descendant_for_byte_range`
* Rename `ts_document_make` -> `ts_document_new`
* Rename `ts_node_name` -> `ts_node_type`

PR #39, opened 2016-10-05
s/TSDebugger/TSLogger in spec name

tclem:
Really minor cosmetic fix in specs that got left behind from the API clean up in #37.

cc @maxbrunsfeld


maxbrunsfeld:
Thanks!

PR #42, opened 2016-10-18
Mergeable symbols

tclem:
In getting https://github.com/tree-sitter/tree-sitter-ruby to build and pass tests again we discovered that one of the parse table optimizations is being overly aggressive in collapsing up states. This adds one additional check to skip merging states unless the symbol on either side is in our set of recovery symbols.

🎩  @maxbrunsfeld for 🍐 ing with me and writing the first pass at this code.

### DIFF
--- a/src/compiler/parse_table.h
+++ b/src/compiler/parse_table.h
@@ -97,6 +97,8 @@ class ParseTable {
 
   std::vector<ParseState> states;
   std::map<rules::Symbol, ParseTableSymbolMetadata> symbols;
+
+  std::set<rules::Symbol> mergeable_symbols;
 };
 
 }  // namespace tree_sitter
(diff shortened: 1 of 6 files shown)


src/compiler/parse_table.h
```diff
@@ -97,6 +97,8 @@ class ParseTable {
 
   std::vector<ParseState> states;
   std::map<rules::Symbol, ParseTableSymbolMetadata> symbols;
+
+  std::set<rules::Symbol> mergeable_symbols;
```
tclem: Wondering about this name a bit. Aren't these really **un**-mergeable?

maxbrunsfeld: I think it does make sense to call them **mergeable**, because when considering which symbols to merge, if we encounter any symbol that is **not** in this set, we abort merging:

```cpp
if (mergeable_symbols.count(symbol) == 0)
  return false;
```

Thus, it functions as a whitelist of symbols that *can* be merged.

joshvera:
👍

PR #47, opened 2016-11-14
Represent byte, char and tree counts as 32 bit numbers

maxbrunsfeld:
The parser spends the majority of its time allocating and freeing trees and stack nodes.
Also, the memory footprint of the AST is a significant concern when using tree-sitter
with large files. This library is already unlikely to work very well with source files
larger than 4GB, so representing rows, columns, byte lengths and child indices as
unsigned 32 bit integers seems like the right choice.


maxbrunsfeld:
On my machine, this speeds up parsing jQuery (242k) by about 8ms (about 16%).

PR #50, opened 2016-11-20
Partially precompute item set transitive closures

maxbrunsfeld:
This speeds up parser generation significantly. Generating the JavaScript parser now takes 6 seconds instead of 13 seconds. And generating the Ruby parser now takes 2 minutes instead of 3 minutes 😬 .

🎩 @nathansobo


nathansobo:
Curious to see the profile.

maxbrunsfeld:
screen shot 2016-11-19 at 10 43 25 pm

It looks like 20% lex table generation, 30% parse table generation (most of which is still spent computing transitive closures), and 37% parse state deduping.

PR #51, opened 2016-12-01
Update JavaScript test

rewinfrey:
After updating behavior in `tree-sitter-javascript`, a JavaScript test case failed.

This updates the spec fixture so that the no longer failure node in the tree is removed.

Additionally this adds an instruction on how to run the tests.

/cc @maxbrunsfeld

### DIFF
--- a/spec/fixtures/error_corpus/javascript_errors.txt
+++ b/spec/fixtures/error_corpus/javascript_errors.txt
@@ -53,7 +53,7 @@ if ({a: 'b'} {c: 'd'}) {
     (object (pair (identifier) (string)))
     (ERROR (object (pair (identifier) (string))))
     (statement_block
-      (ERROR (function
+      (function
         (formal_parameters (identifier))
         (statement_block (expression_statement (identifier)))))
       (expression_statement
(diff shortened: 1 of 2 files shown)


spec/fixtures/error_corpus/javascript_errors.txt
```diff
@@ -53,13 +53,13 @@ if ({a: 'b'} {c: 'd'}) {
...
     (statement_block
-      (ERROR (function
+      (function
         (formal_parameters (identifier))
         (statement_block (expression_statement (identifier)))))
       (expression_statement
         (function
           (formal_parameters (identifier))
-          (statement_block (expression_statement (identifier))))))))
+          (statement_block (expression_statement (identifier)))))))
```
rewinfrey: @maxbrunsfeld I saw your commit on master that updates this test 👍 🙇 

I'm curious though why the build for this branch is causing 119 build failures? From what I can see, it looks like this build should be fine, but I'm missing something.

rewinfrey:
Closing due to fix already committed: https://github.com/tree-sitter/tree-sitter/commit/d627042fa6ea69d2676ea1549f99eacaa6756408

PR #52, opened 2016-12-01
Disallow syntax rules that match the empty string (for now)

maxbrunsfeld:
These *epsilon rules* are actually somewhat complicated to handle; they've never worked 100% correctly. For now, I'm just going to disallow them.

### DIFF
--- a/src/compiler/build_tables/build_parse_table.cc
+++ b/src/compiler/build_tables/build_parse_table.cc
@@ -177,7 +177,7 @@ class ParseTableBuilder {
             parse_table.add_terminal_action(state_id, lookahead, action);
           } else {
             ParseAction &existing_action = entry.actions[0];
-            if (allow_any_conflict) {
+            if (existing_action.type == ParseActionTypeAccept || allow_any_conflict) {
               entry.actions.push_back(action);
             } else {
               int existing_precedence = existing_action.precedence();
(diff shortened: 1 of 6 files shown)


joshvera:
👍 do you think this might affect any of our parsers?

maxbrunsfeld:
Just ruby, but it's a trivial adjustment to make it work.

src/compiler/build_tables/build_parse_table.cc
```diff
@@ -177,7 +177,7 @@ class ParseTableBuilder {
             parse_table.add_terminal_action(state_id, lookahead, action);
           } else {
             ParseAction &existing_action = entry.actions[0];
-            if (allow_any_conflict) {
+            if (existing_action.type == ParseActionTypeAccept || allow_any_conflict) {
```
joshvera: What does this change have to do with `TSCompileErrorTypeEpsilonRule` or the `flatten_grammar` function?

maxbrunsfeld: Pretty much nothing; it's fixing an unrelated bug. But the integration test that I wrote for epsilon rules crashed because it triggered this edge case, so I'm just including this as a drive-by fix.

joshvera: 👌

PR #54, opened 2016-12-03
External scanners

maxbrunsfeld:
This adds the ability for a tree-sitter parser to tokenize input using arbitrary hand-written code. 

~The API for external scanners will need to grow and/or change in order to allow the scanner to maintain its state properly during incremental parsing, but I think it's worth shipping this with only batch parsing support, so that we can try it out with the Ruby grammar, and maybe start work on a Python grammar.~

I have created [a Python grammar](https://github.com/tree-sitter/tree-sitter-python) using this API, and I am going to use this grammar in the core test suite to verify that incremental parsing works in the presence of external scanners. I want to get incremental parsing *basically* working before merging this, so that I can feel confident in the external scanner API.

* [x] Verify at generate time that external token names do not match any internal rule names unless the internal rule is a token.


tclem:
🤘

src/runtime/parser.c
```diff
@@ -1112,6 +1151,9 @@ void parser_destroy(Parser *self) {
     array_delete(&self->tree_path1);
   if (self->tree_path2.contents)
     array_delete(&self->tree_path2);
+  if (self->external_scanner_payload)
+    if (self->language->external_scanner.destroy)
```
joshvera: Maybe combine this with the above `if` statement?

src/runtime/parser.c
```diff
@@ -1103,6 +1129,19 @@ bool parser_init(Parser *self) {
   return true;
 }
 
+void parser_set_language(Parser *self, const TSLanguage *language) {
```
joshvera: Could `parser_set_language` just call into `parser_destroy`?

src/runtime/parser.c
```diff
@@ -208,24 +208,54 @@ static bool parser__condense_stack(Parser *self) {
...
-  while (!self->language->lex_fn(&self->lexer.data, current_state)) {
-    if (current_state != ERROR_STATE) {
+  while (!parser__try_lex(self, lex_mode)) {
+    if (!found_error) {
       LOG("retry_in_error_mode");
-      current_state = ERROR_STATE;
+      found_error = true;
+      lex_mode = self->language->lex_modes[ERROR_STATE];
       ts_lexer_reset(&self->lexer, start_position);
-      ts_lexer_start(&self->lexer, current_state);
```
joshvera: Why were we previously calling `ts_lexer_start` here, and why don’t we anymore?

src/compiler/parse_table.cc
```diff
@@ -150,9 +151,7 @@ bool ParseState::has_shift_action() const {
 set<Symbol> ParseState::expected_inputs() const {
   set<Symbol> result;
   for (auto &entry : terminal_entries)
-    result.insert(Symbol(entry.first, true));
-  for (auto &entry : nonterminal_entries)
```
joshvera: Is this handled somewhere else now?

src/runtime/parser.c
```diff
@@ -239,34 +269,27 @@ static Tree *parser__lex(Parser *self, TSStateId parse_state) {
...
-    result =
-      ts_tree_make_leaf(symbol, padding, size,
-                        ts_language_symbol_metadata(self->language, symbol));
+    Length padding = length_sub(self->lexer.token_start_position, start_position);
+    Length size = length_sub(self->lexer.current_position, self->lexer.token_start_position);
+    TSSymbolMetadata metadata = ts_language_symbol_metadata(self->language, symbol);
+    result = ts_tree_make_leaf(symbol, padding, size, metadata);
   }
 
-  if (!result)
```
maxbrunsfeld: Yeah, this conditional was leftover from before https://github.com/tree-sitter/tree-sitter/pull/43, when the system tried to recover from malloc failure. I've removed that logic across the board, since malloc never even actually returns null on most systems, and apps that use tree-sitter aren't likely to have a fine-grained OOM recovery strategy anyway.

src/compiler/parse_table.cc
maxbrunsfeld: This dates back from a time when the parse table was constructed differently. Prior to this commit, the only caller of `ParseState::expected_inputs()` would immediately filter out the nonterminals.

src/runtime/parser.c
maxbrunsfeld: Yeah, this conditional was leftover from before https://github.com/tree-sitter/tree-sitter/pull/43, when the system tried to recover from malloc failure. I've removed that logic across the board, since malloc never even actually returns null on most systems, and apps that use tree-sitter aren't likely to have a fine-grained OOM recovery strategy anyway.

src/runtime/parser.c
maxbrunsfeld: In order to handle the two-step process of lexing with the external scanner and then lexing with the regular scanner function, I've extracted a helper function `parser__try_lex`. The call to `ts_lexer__start` has been moved into that function.

src/runtime/parser.c
maxbrunsfeld: I see what you're saying: to remove the duplication about destroying the external scanner. So `parser_destroy` does some additional teardown that we wouldn't want to do here, but maybe we could to the reverse: `parser_destroy` could call `parser_set_language(self, NULL)`.

src/runtime/parser.c
maxbrunsfeld: :+1:

src/runtime/parser.c
joshvera: 👍

src/runtime/parser.c
joshvera: 👍

src/compiler/parse_table.cc
joshvera: 👍

tclem:
🎉

PR #55, opened 2017-01-31
Allow \S for negated whitespace regex shorthand

tclem:


maxbrunsfeld:
Awesome! Want to add `\W` and `\A` while you're here?

maxbrunsfeld:
✨

PR #59, opened 2017-02-10
Input string with length

robrix:
This PR adds runtime API to construct an input string with an explicit length.

### DIFF
--- a/spec/runtime/document_spec.cc
+++ b/spec/runtime/document_spec.cc
@@ -116,6 +116,17 @@ describe("Document", [&]() {
       AssertThat(spy_input->strings_read, Equals(vector<string>({" [null, 2" })));
     });
 
+    it("allows setting input string with length", [&]() {
+      const char content[] = { '1' };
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
Actually, @robrix could you add a small test for this, just so we have some call to this method in the test suite? Maybe [here](https://github.com/tree-sitter/tree-sitter/blob/master/spec/runtime/document_spec.cc#L138)?

spec/runtime/document_spec.cc
```diff
@@ -116,6 +116,17 @@ describe("Document", [&]() {
       AssertThat(spy_input->strings_read, Equals(vector<string>({" [null, 2" })));
     });
 
+    it("allows setting input string with length", [&]() {
```
tclem: How's this look for a basic happy path @maxbrunsfeld?

maxbrunsfeld: Perfect.

PR #60, opened 2017-02-18
[Experimental] Compute error recovery states dynamically

maxbrunsfeld:
⚠️ WIP ⚠️ 

This is an attempt to further reduce the branching of the parse stack during error recovery, as mentioned in #58.


maxbrunsfeld:
#101 Accomplished something similar to the goal of this PR.

PR #71, opened 2017-03-21
Run randomized tests against the latest Javascript grammar

maxbrunsfeld:
The latest Javascript grammar revealed a few bugs in the runtime:
* [x] The JS corpus now includes some examples with non-ascii text. Random edits to these examples now create invalid unicode, which we weren't handling properly.
* [x] Automatic semicolon tokens often have zero size. This new case revealed some bugs in the runtime:
  * [x] Error recovery was infinite looping in a couple of places. 
  * [x] Ranges whose syntax had changed after an edit are failing to be reported


maxbrunsfeld:
Ok, I believe the most obvious problems revealed by the recent change to `tree-sitter-javascript` have been fixed.

PR #72, opened 2017-03-24
Make external scanner symbol map unique

joshvera:


src/compiler/generate_code/c_code.cc
```diff
@@ -295,7 +295,10 @@ class CCodeGenerator {
   }
 
   void add_external_scanner_symbol_map() {
-    line("TSSymbol ts_external_scanner_symbol_map[EXTERNAL_TOKEN_COUNT] = {");
+    string language_function_name = "tree_sitter_" + name;
+    string external_scanner_name = language_function_name + "_external_scanner";
+
+    line("TSSymbol " + external_scanner_name + "_symbol_map[EXTERNAL_TOKEN_COUNT] = {");
```
maxbrunsfeld: I think that rather than giving this value a unique name, we can give it `static` linkage, so that it's only visible within this translation unit. I think that's what I do for all other fields on the `TSLanguage` struct fields, I just forgot to do it on this one.

So we just insert `static` before the declaration.

maxbrunsfeld:
:+1:

PR #73, opened 2017-04-12
Symbol types

robrix:
This PR defines a symbol type enum & API to get the type for any symbol in a language.


maxbrunsfeld:
:+1: Makes a lot of sense.

PR #76, opened 2017-06-09
OOB-read in remove_duplicate_parse_states

philipturnbull:
Some ParseActions have `state_id == -1` which can cause an out-of-bounds read when removing duplicated states. Also fix a think-o in `each_referenced_state`

### DIFF
--- a/src/compiler/build_tables/build_parse_table.cc
+++ b/src/compiler/build_tables/build_parse_table.cc
@@ -394,8 +394,7 @@ class ParseTableBuilder {
       } else {
         ParseState &state = *iter;
         state.each_referenced_state([&new_state_ids](ParseStateId *state_index) {
-          if (*state_index != (ParseStateId)(-1))
-            *state_index = new_state_ids[*state_index];
+          *state_index = new_state_ids[*state_index];
         });
         ++iter;
       }
--- a/src/compiler/build_tables/build_parse_table.cc
+++ b/src/compiler/build_tables/build_parse_table.cc
@@ -394,7 +394,8 @@ class ParseTableBuilder {
       } else {
         ParseState &state = *iter;
         state.each_referenced_state([&new_state_ids](ParseStateId *state_index) {
-          *state_index = new_state_ids[*state_index];
+          if (*state_index != (ParseStateId)(-1))
+            *state_index = new_state_ids[*state_index];
         });
         ++iter;
       }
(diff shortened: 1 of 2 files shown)


src/compiler/build_tables/build_parse_table.cc
```diff
@@ -394,7 +394,8 @@ class ParseTableBuilder {
       } else {
         ParseState &state = *iter;
         state.each_referenced_state([&new_state_ids](ParseStateId *state_index) {
-          *state_index = new_state_ids[*state_index];
+          if (*state_index != (ParseStateId)(-1))
+            *state_index = new_state_ids[*state_index];
```
maxbrunsfeld: My intention was to for `each_referenced_state` to only call this callback with valid state indices, so we shouldn't have to check them here.

I think all we need is your other fix to `each_referenced_state` itself, plus the one additional check I mention below.

src/compiler/parse_table.cc
```diff
@@ -151,7 +151,7 @@ bool ParseState::has_shift_action() const {
 void ParseState::each_referenced_state(function<void(ParseStateId *)> fn) {
   for (auto &entry : terminal_entries)
     for (ParseAction &action : entry.second.actions)
-      if (action.type == ParseActionTypeShift || ParseActionTypeRecover)
+      if (action.type == ParseActionTypeShift || action.type == ParseActionTypeRecover)
```
maxbrunsfeld: 😬  Oh wow. Thanks for catching that.

We need to check one other condition here to ensure that `action` will have a meaningful `state_index` field:

```cpp
if ((action.type == ParseActionTypeShift && !action.extra) || action.type == ParseActionTypeRecover)
  fn(&action.state_index)
```

**Background** - There's a special 'shift-extra' action that tells the parser to consume an 'extra token' (e.g. a comment). These actions don't change the parser's state, so they don't have a `state_index` field. I can't remember why I modeled these actions using a separate boolean `extra` field; it seems like there should just be a separate `ParseActionTypeShiftExtra` value for the `type` field. But anyway, with the current setup, only shift actions that are *not* `extra` will have a meaningful `state_index`.

src/compiler/parse_table.cc
```diff
@@ -154,7 +154,8 @@ void ParseState::each_referenced_state(function<void(ParseStateId *)> fn) {
       if (action.type == ParseActionTypeShift || action.type == ParseActionTypeRecover)
         fn(&action.state_index);
   for (auto &entry : nonterminal_entries)
-    fn(&entry.second);
+    if (entry.second != (ParseStateId)(-1))
```
maxbrunsfeld: I don't think we need this conditional; the only things we write into this map are valid state numbers.

philipturnbull: Ah! 😬  I mis-read "we shouldn't have to check them here as "we shouldn't check them here"

maxbrunsfeld:
:zap:

PR #77, opened 2017-06-14
Fix errors found by scan-build

philipturnbull:
It didn't find anything too scary, the worst thing is two NULL pointer derefs in `runtime/parser.c`. I haven't got enough domain knowledge to know if they would be hit in normal use, or if they are just pathological edge-cases.

The final missing part from this PR is to run `scan-build` automatically in the CI build. I have it working locally but doesn't work correctly on travis-ci.org 😢  I will update this branch once I have it properly working!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+dist: trusty
 language: cpp
 compiler:
 - gcc
@@ -9,10 +10,11 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-5
+    - clang
 
 install:
 - export CXX="g++-5"
-- script/configure
+- scan-build script/configure
 
 script:
 - script/ci
--- a/script/ci
+++ b/script/ci
@@ -2,6 +2,9 @@
 
 set -e
 
+. script/lib.sh
+
 script/fetch-fixtures
 script/check-mallocs
-script/test
+scan_build make
+script/test -b
--- a/script/ci
+++ b/script/ci
@@ -6,5 +6,4 @@ set -e
 
 script/fetch-fixtures
 script/check-mallocs
-scan_build make -j2
 script/test -b
--- a/script/ci
+++ b/script/ci
@@ -6,5 +6,5 @@ set -e
 
 script/fetch-fixtures
 script/check-mallocs
-scan_build make
+scan_build make -j2
 script/test -b
--- a/script/scan-build
+++ b/script/scan-build
@@ -1,7 +1,0 @@
-#!/usr/bin/env bash
-
-set -e
-
-. script/lib.sh
-
-scan_build make "$@"
--- a/script/scan-build
+++ b/script/scan-build
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+. script/lib.sh
+
+scan_build make "$@"
--- a/script/test
+++ b/script/test
@@ -2,6 +2,8 @@
 
 set -e
 
+. script/lib.sh
+
 function usage {
   cat <<-EOF
 USAGE
@@ -11,6 +13,8 @@ USAGE
 OPTIONS
 
   -h  print this message
+
+  -b  run make under scan-build static analyzer
 
   -d  run tests in a debugger (either lldb or gdb)
 
@@ -26,6 +30,7 @@ OPTIONS
 
   -z  pipe tests' stderr to \`dot(1)\` to render an SVG log
 
+
 EOF
 }
 
@@ -37,8 +42,9 @@ args=()
 target=tests
 export BUILDTYPE=Test
 cmd="out/${BUILDTYPE}/${target}"
+run_scan_build=
 
-while getopts "df:s:gGhpvS" option; do
+while getopts "bdf:s:gGhpvS" option; do
   case ${option} in
     h)
       usage
@@ -69,6 +75,9 @@ while getopts "df:s:gGhpvS" option; do
     S)
       mode=SVG
       ;;
+    b)
+      run_scan_build=true
+      ;;
   esac
 done
 
@@ -78,7 +87,11 @@ else
   args+=("--reporter=singleline")
 fi
 
-make $target
+if [[ ! -z "$run_scan_build" ]]; then
+	scan_build make $target
+else
+	make $target
+fi
 args=${args:-""}
 
 if [[ -n $profile ]]; then
--- a/script/test
+++ b/script/test
@@ -87,10 +87,10 @@ else
   args+=("--reporter=singleline")
 fi
 
-if [[ ! -z "$run_scan_build" ]]; then
-	scan_build make -j2 $target
+if [[ -n "$run_scan_build" ]]; then
+  scan_build make -j2 $target
 else
-	make -j2 $target
+  make -j2 $target
 fi
 args=${args:-""}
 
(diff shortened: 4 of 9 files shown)


src/runtime/parser.c
```diff
@@ -839,7 +839,7 @@ static void parser__accept(Parser *self, StackVersion version,
 
     if (parser__select_tree(self, self->finished_tree, root)) {
       ts_tree_release(self->finished_tree);
-      assert(root->ref_count > 0);
+      assert((root == NULL) || (root->ref_count > 0));
```
maxbrunsfeld: `root` should never be `NULL` here; would it make `scan-build` happy if we changed this to this?

```c
assert(root && root->ref_count > 0);
```

maxbrunsfeld:
Nice! It'll be great to have static analysis running on CI.

philipturnbull:
This is ready for review. The tricky thing was configuring a travis environment with a version of clang that supports `-std=c++14`. `./script/ci` and `./script/scan-build` also work locally on OS X too 👌

script/test
```diff
@@ -78,7 +87,11 @@ else
   args+=("--reporter=singleline")
 fi
 
-make $target
+if [[ ! -z "$run_scan_build" ]]; then
```
maxbrunsfeld: I think the combination `! -z` can be written `-n` (non-empty).

script/test
```diff
@@ -78,7 +87,11 @@ else
...
+	scan_build make -j2 $target
+else
+	make -j2 $target
```
maxbrunsfeld: Could you adjust this indentation to match the rest of the file?

script/scan-build
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+. script/lib.sh
+
+scan_build make "$@"
```
maxbrunsfeld: When would you use this script as opposed to running `script/test -b`?

.travis.yml
```diff
@@ -9,10 +10,11 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-5
+    - clang
```
maxbrunsfeld: Is `g++-5` still used, or does using `scan-build` require us to build with clang?

philipturnbull: Both compilers are used. For each target`scan-build` will build the object file as normal with`g++-5` then statically-analyze the target with `clang++`. `clang++` in analysis mode doesn't create any object files.

.travis.yml
```diff
@@ -9,10 +10,11 @@ addons:
...
 
 install:
 - export CXX="g++-5"
-- script/configure
+- scan-build script/configure
```
philipturnbull: `scan-build` sets `$CC` and `$CXX` to be the `ccc-analyzer` script which `gyp` will then pick up. When `make` runs, it invokes `ccc-analyzer` for each object, which will then invoke `g++-5` and `clang++`.

It's kind ugly but [is the recommended way](https://clang-analyzer.llvm.org/scan-build.html#recommended_autoconf)  😐

.travis.yml
maxbrunsfeld: :+1: Ah, that makes sense. I use gcc and libstdc++ on CI just because I normally use clang when I develop locally, and I want to be sure the library works with both toolchains.

script/ci
```diff
@@ -2,6 +2,9 @@
 
 set -e
 
+. script/lib.sh
+
 script/fetch-fixtures
 script/check-mallocs
-script/test
+scan_build make -j2
+script/test -b
```
philipturnbull: Ah, this is a remnant of an older version of this patch. I will remove that `scan_build make -j2` step

.travis.yml
maxbrunsfeld: This might only be necessary if we were using autoconf:

> Running `configure` typically generates makefiles that have hardwired paths to the compiler, and by running `configure` through `scan-build` that path is set to `ccc-analyzer`.

Although `gyp` does insert a default value for `CC` and `CXX` into the makefile, you can still override it when running `make`, at least on my system:

```sh
$ script/configure
$ CXX=non-existent-compiler make

make: non-existent-compiler: No such file or directory
make: *** Error 1
```

So I'd think we could just run `script/test -b`. Does this work for you?

script/scan-build
philipturnbull: 🤔 This was useful for quick testing but you are right, it isn't particularly useful now.

.travis.yml
philipturnbull: Without running `./script/configure` under `scan-build` I get different results on my setup and travis. On my setup (with some extra debugging) `ccc-analyzer` is invoked correctly:

```
$ CC=`which gcc` CXX=`which g++` ./script/test -b
...
+ scan-build --use-analyzer=/usr/local/bin/clang --use-cc=/usr/bin/gcc --use-c++=/usr/bin/g++ --status-bugs make -j2 tests
scan-build: Using '/usr/local/bin/clang' for static analysis
  /usr/local/Cellar/scan-build/279/bin/../libexec/ccc-analyzer '-DTREE_SITTER_WRAP_MALLOC=true' -Iinclude -Isrc -Iexternals/utf8proc  -fasm-blocks -mpascal-strings -O0 -gdwarf-2 -arch x86_64 -Wall -Wextra -Wno-unused-parameter  -MMD -MF out/Test/.deps/out/Test/obj.target/runtime/src/runtime/document.o.d.raw  -c -o out/Test/obj.target/runtime/src/runtime/document.o src/runtime/document.c
  /usr/local/Cellar/scan-build/279/bin/../libexec/ccc-analyzer '-DTREE_SITTER_WRAP_MALLOC=true' -Iinclude -Isrc -Iexternals/utf8proc  -fasm-blocks -mpascal-strings -O0 -gdwarf-2 -arch x86_64 -Wall -Wextra -Wno-unused-parameter  -MMD -MF out/Test/.deps/out/Test/obj.target/runtime/src/runtime/error_costs.o.d.raw  -c -o out/Test/obj.target/runtime/src/runtime/error_costs.o src/runtime/error_costs.c
```
but the analyzer isn't invoked on travis:
```
+scan-build --use-analyzer=/usr/local/clang-3.5.0/bin/clang --use-cc=gcc --use-c++=g++-5 --status-bugs make -j2 tests
scan-build: Using '/usr/local/clang-3.5.0/bin/clang' for static analysis
  gcc '-DTREE_SITTER_WRAP_MALLOC=true' -Iinclude -Isrc -Iexternals/utf8proc  -Wall -Wextra -Wno-unused-parameter -g -std=c99 -MMD -MF out/Test/.deps/out/Test/obj.target/runtime/src/runtime/document.o.d.raw  -c -o out/Test/obj.target/runtime/src/runtime/document.o src/runtime/document.c
  gcc '-DTREE_SITTER_WRAP_MALLOC=true' -Iinclude -Isrc -Iexternals/utf8proc  -Wall -Wextra -Wno-unused-parameter -g -std=c99 -MMD -MF out/Test/.deps/out/Test/obj.target/runtime/src/runtime/error_costs.o.d.raw  -c -o out/Test/obj.target/runtime/src/runtime/error_costs.o src/runtime/error_costs.c
```

which is annoying. I assume this is a difference between the scan-build versions(?)

maxbrunsfeld: Ah, dang. That's ok; including `scan-build` before the `configure` seems fine then.

maxbrunsfeld:
📈  Thanks @philipturnbull!

PR #83, opened 2017-06-29
Increase size of ref_counts and add assertions

philipturnbull:
This bumps the size of the reference counts from 16- to 32-bit counters to make
it less likely to overflow. Also assert in the retain function that the
reference count didn't overflow.

32-bits seems big enough for non-pathological examples but a more fool-proof
fix may be to bump it to 64-bits.


maxbrunsfeld:
Thanks!

PR #89, opened 2017-07-12
Add a way to automatically inline rules

maxbrunsfeld:
Sometimes when writing a grammar, it's helpful to extract rule common structures into named 'helper' rules, even if you don't want those rules to exist in the parser. Currently, we do this in some grammars by creating functions like [this](https://github.com/tree-sitter/tree-sitter-javascript/blob/034f8bd2efdb2ed72a9ec7b76394813efed28910/grammar.js#L813-L815) and calling them like [this](https://github.com/tree-sitter/tree-sitter-javascript/blob/034f8bd2efdb2ed72a9ec7b76394813efed28910/grammar.js#L177).

This PR adds a built-in way of marking certain rules to be 'inlined', so that we can write these types of rules using the normal syntax. This implementation will also allow the parser to generate faster than with the technique shown above.

While I was here, I also found another opportunity for optimization of parser generation, so parser generation times should be lower than ever.


rewinfrey:
@maxbrunsfeld I missed this when it came through initially, but thank you for adding this! ❤️  👍

PR #90, opened 2017-07-14
Add an API for renaming nodes based on their context

maxbrunsfeld:
#### Motivation

Currently, most tree-sitter grammars have some kind of `identifier` token that is used to represent several different constructs in the language: references to variables, introductions of new variables, references to types, introductions of new types, object fields, etc. Using the current syntax tree, It isn't as easy as I would like to distinguish between these different meanings of `identifier`.

#### Solution

This PR adds a new type of rule called `rename`, which can be used like this:

```js
expression: $ => choice(
  $.call_expression,
  $.binary_expression,
  $.number,
  // ...
  rename($.identifier, 'variable_reference')
),

type: $ => choice(
  $.array_type,
  rename($.identifier, 'type_reference')
),

member_expression: $ => seq(
  $.expression,
  '.',
  rename($.identifier, 'property_name')
)
```

Unlike *wrapping* the `identifier` token in a `variable_reference` *parent* token, which increases the size of the syntax tree significantly, the `rename` rule has pretty much zero runtime cost.

/cc @nathansobo

### DIFF
--- a/doc/grammar-schema.json
+++ b/doc/grammar-schema.json
@@ -135,6 +135,23 @@
       "required": ["type", "members"]
     },
 
+    "rename-rule": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "pattern": "^RENAME$"
+        },
+        "value": {
+          "type": "string"
+        },
+        "content": {
+          "$ref": "#/definitions/rule"
+        }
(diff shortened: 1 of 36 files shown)


nathansobo:
This is a really cool feature. Nice work!

PR #91, opened 2017-07-14
Add support for fuzzing with libFuzzer

philipturnbull:
I'm not too happy with `script/build-fuzzers` but this seems to be the easiest way to build the binaries with the exact set of flags needed to enable fuzzing. I couldn't really figure out how to integrate this into the gyp file. It works well though, and can build fuzzers for all of the languages we currently support ✨ 

I wrote some (brief) docs on how to build and run the fuzzers on Linux but let me know if anything is unclear. As we talked about a couple of weeks ago, it doesn't really make sense to fuzz during CI so all the code is separate from `script/test`


maxbrunsfeld:
Wow, I like how you generate the fuzzing dictionary from the grammar.json. A shell script for building the fuzzers seems fine to me. Thanks so much for setting this up!

PR #92, opened 2017-07-17
Add test for UTF16 out-of-bound read

philipturnbull:
utf16_iterate does not check that 'length' is a multiple of two which leads to
an out-of-bound read:

```
==105293== Conditional jump or move depends on uninitialised value(s)
==105293==    at 0x54F014: utf16_iterate (utf16.c:7)
==105293==    by 0x539251: string_iterate(TSInputEncoding, unsigned char const*, unsigned long, int*) (encoding_helpers.cc:15)
==105293==    by 0x53939D: string_byte_for_character(TSInputEncoding, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, unsigned long) (encoding_helpers.cc:43)
==105293==    by 0x507BAD: SpyInput::read(void*, unsigned int*) (spy_input.cc:47)
==105293==    by 0x551049: ts_lexer__get_chunk (lexer.c:29)
==105293==    by 0x5515C2: ts_lexer_start (lexer.c:152)
==105293==    by 0x5469AB: parser(long,...)(long long) (parser.c:297)
==105293==    by 0x547896: parser__get_lookahead (parser.c:439)
==105293==    by 0x54B2DF: parser__advance (parser.c:1150)
==105293==    by 0x54C2B6: parser_parse (parser.c:1348)
==105293==    by 0x53F06F: ts_document_parse_with_options (document.c:136)
==105293==    by 0x53EF4F: ts_document_parse (document.c:107)
```

I _think_ returning '0' from `utf16_iterate` if 'length < 2' is the right fix


test/runtime/document_test.cc
```diff
@@ -72,6 +72,19 @@ describe("Document", [&]() {
         "(array (true) (false))");
     });
 
+    it("handles truncated UTF16 data", [&]() {
+      char *content = reinterpret_cast<char*>(malloc(1));
+
+      spy_input->content = string((const char *)content, 1);
```
maxbrunsfeld: I think that rather than allocating the content with `malloc`, you could just pass a string literal to the `std::string` constructor.

```c++
spy_input->content = string("x", 1);
```

This should behave identically with regard to triggering the bug, because either way, the `string` will copy 1 byte from the given content into its own internal buffer.

Plus, this way will avoid reading from uninitialized memory, which I would think would trigger an undefined behavior warning in its own right.

test/runtime/document_test.cc
```diff
@@ -72,19 +72,6 @@ describe("Document", [&]() {
...
-      spy_input->content = string((const char *)content, 1);
-      spy_input->encoding = TSInputEncodingUTF16;
-
-      ts_document_set_input(document, spy_input->input());
-      ts_document_invalidate(document);
-      ts_document_parse(document);
-
-      free(content);
-    });
-
```
maxbrunsfeld: I think there's some value in testing it through `ts_document_parse`. I'm guessing this test didn't trigger the OOB read because `std::string` appends a null byte? Maybe you could use `ts_document_set_input_string_with_length` here, with a manually-malloc'd  buffer (contrary to what I suggested earlier).

philipturnbull: You're totally right, much better to go through `ts_document_parse`. I thought it was some magic in `std::string` but I tracked it down to [the spy buffer logic](https://github.com/tree-sitter/tree-sitter/blob/20b664969e8ec2a5aa94754f41975e8792522fa4/test/helpers/spy_input.cc#L64) which is helpfully padding allocations and initialising memory, which totally defeats valgrind 😄

test/helpers/spy_input.cc
```diff
@@ -57,12 +56,19 @@ const char * SpyInput::read(void *payload, uint32_t *bytes_read) {
...
-  memset(spy->buffer, 0, spy->buffer_size);
-  memcpy(spy->buffer, result.data(), byte_count);
+  delete[] spy->buffer;
+  if (byte_count) {
+    spy->buffer = new char[byte_count];
+    memcpy(spy->buffer, result.data(), byte_count);
+  } else {
+    spy->buffer = nullptr;
+  }
+
```
maxbrunsfeld: Does calling `malloc` on every chunk slow down the tests at all? I'm wondering if we really need to do this everywhere, or if we can just do something custom for this one test, like calling `ts_document_set_input_string_with_length`.

philipturnbull: The cost of `new` and `delete` get lost in the noise afaict.

With `master`:
```
$ git rev-parse HEAD
afb499bf2e5c9aee78ae9a8611268008e1136359
$ for x in $(echo "1 2 3 4 5"); do time ./out/Test/tests >/dev/null; done
real    0m20.394s
user    0m20.068s
sys     0m0.148s

real    0m21.864s
user    0m21.492s
sys     0m0.168s

real    0m20.518s
user    0m20.136s
sys     0m0.168s

real    0m20.215s
user    0m19.908s
sys     0m0.148s

real    0m20.478s
user    0m20.116s
sys     0m0.192s
```

this branch:
```
$ git rev-parse HEAD
52cec9ed39a56f226d47499db6362a192138fd5a
$ for x in $(echo "1 2 3 4 5"); do time ./out/Test/tests >/dev/null; done

real    0m19.423s
user    0m19.084s
sys     0m0.160s

real    0m19.946s
user    0m19.588s
sys     0m0.188s

real    0m21.006s
user    0m20.672s
sys     0m0.172s

real    0m19.634s
user    0m19.304s
sys     0m0.160s

real    0m19.492s
user    0m19.188s
sys     0m0.128s
```

maxbrunsfeld: Ah ok, thanks for investigating!

PR #94, opened 2017-07-18
Large external scanner states

maxbrunsfeld:
### Motivation

Previously, external scanners were only allowed to store a fixed amount of state between invocations - the amount that would fit directly inside of a certain piece of tree-sitter's internal `Tree` struct; a piece that was going unused in the case of leaf nodes.

But in general, languages' scanners may need to store more state than that. For example, the Heredoc construct in Ruby and Bash requires that the scanner can store the entirety of a heredoc delimiter string. These are usually still quite small, but they are allowed to be as long as the programmer wants.

With the old external scanner API, in these cases, the serialization call would simply fail, which would cause unpredictable results in the presence of ambiguity or when re-parsing after an edit.

### New API

The `serialize` and `deserialize` hooks now have the following signatures:

```c
unsigned tree_sitter_bash_external_scanner_serialize(
  void *payload,
  char *buffer
);

void tree_sitter_bash_external_scanner_deserialize(
  void *payload,
  const char *buffer,
  unsigned length
);
```

The scanner is passed a buffer to which it can write however many bytes it wants, and it returns how many bytes were written. There is still a hard upper bound, but it is now a *kilobyte*, exposed to the scanners via the `TREE_SITTER_SERIALIZATION_BUFFER_SIZE` constant. I don't think a scanner will ever need to store more than this, but if the need arises, we can simply raise this limit, it's a one-time allocation per parser instance.

In addition, I've removed the `reset` hook. Now, the contract is that scanners should reset their state to the initial state when `deserialize` is called with `length` set to zero, which feels pretty natural for the scanners I've worked with so far.

### Implementation

When the state serialized by the external scanner is small enough, we'll still store it directly in the tree struct, avoiding any extra memory allocations. This will almost always be the case for python, and will usually be the case in bash and ruby. But when more space is needed, we'll allocate a buffer on the heap to hold the state. This is similar to what `std::string` does with the 'short string optimization'.

### Testing

In order to drive out these changes, I have added `bash` to the list of fixture grammars that I use in tree-sitter's core test suite. I could have used ruby, but the ruby parser still takes a long time to generate compared to the others, whereas the bash parser is simpler and generates very quickly.

/cc @nathansobo @iolsen - I think this is the last change I need to make to the library before starting work on MVP Atom integration for our initial set of languages: C, C++, Go, JSON, JavaScript, TypeScript, Python, Ruby, and now Bash.


nathansobo:
🎉

PR #103, opened 2017-09-13
Assertion failure in parser__advance

philipturnbull:
:wave: I found this via fuzzing. The python testcase decodes to:
```
00000000  35 63 6f 6e 88 2c 29 33  2c 2c 2c 2c 63 6f 6e 88  |5con.,)3,,,,con.|
00000010  2c 2a 2c 3a 35 63 6f 6e  2c                       |,*,:5con,|
```

which triggers:
```
Assertion failed: ((uint32_t)0 < (&reduction.slices)->size), function parser__advance, file src/runtime/parser.c, line 1202.
```

which looks like `parser__reduce` is returning a result with no slices.


maxbrunsfeld:
I believe that this is addressed by the latest changes to error recovery in #101. Re-running CI.

maxbrunsfeld:
Thanks @philipturnbull! The latest error recovery is much simpler than the old one, and I think we're making some progress towards error recovery that reliably does not hang or crash!

PR #125, opened 2018-01-16
WIP tree-sitter grammar development tips doc

aymannadeem:
Started a very WIP doc that aims to bring a bit more empiricism to the process of grammar development. This documentation is supposed to serve as a `parse tree style guide` as well as a `grammar development guide`. By pairing with Max (and others), I've learned how to exploit the mechanics of tree-sitter without really knowing a lot of underlying logic or when to make certain decisions. Since it's not immediately intuitive for me, I wanted to document what I've learned. 

@maxbrunsfeld, @tclem, @rewinfrey, @robrix, @joshvera — would love your expertise here. It would be nice to put some writing down and not simply rely on word of mouth as the means to transition this knowledge, especially as we hire. If someone wants to pair with me later on this, I'm happy to do that too.  Additionally, if folks have a good understanding of what the tree structure should be like for the purpose of abstract interpretation (looking at you @tclem) please chime in!

### DIFF
--- a/tree-sitter-tips.md
+++ b/tree-sitter-tips.md
@@ -1,0 +1,79 @@
+# Tree-sitter tips and tricks
+
+Language development using [Tree-Sitter](https://github.com/tree-sitter/tree-sitter) can be an arduous task. These tips are intended to help bring empiricism to the process by reducing trial and error (or at least introducing some guard rails).
+
+### :bulb: Rely on Language Specs, but not too much.
+There are language specifications (like [this one for Java](https://docs.oracle.com/javase/specs/jls/se9/html/index.html)) that comprehensively describe a language. Specs provide the lexical and syntactic structure of a language in addition to its components (ex., types, variables, declarations, etc.) and how they interconnect. This can serve as a great baseline for writing your grammar. However, mirroring the spec 100% is not always the best approach, especially when its structure does not align with the parse tree you'd like to generate using tree-sitter. In this situation, it's better to deviate from the spec and prioritize generating a good parse tree. One way to do this may be to find an LR parser for the language you're working on ([such as this one for Java](https://github.com/javaparser/javaparser)) to see how they've modeled their tree outputs. Studying tree outputs from other LR parsers can be a good way to guide your own grammar development. 
+
+### :bulb: What does a "good" parse tree look like?
+
+Here are some attributes of a "pretty" tree:
+
+:white_check_mark: Not overly verbose.
+
+:white_check_mark: Everything is outlined clearly.
+
+:white_check_mark: Communicates good information needed further along in the pipeline.
+
+![image](https://user-images.githubusercontent.com/875834/34972577-41bc12ac-fa37-11e7-9a12-a05d18b6f6e4.png)
+
+_Example of a good tree (WIP -- better examples welcome!):_
+
+```
+===
+module
+===
+
+module com.foo { }
+open module com.foo { }
+
+---
+
+(program
+  (module_declaration (module_identifier (identifier) (identifier)))
+  (module_declaration (module_identifier (identifier) (identifier))))
+```
+
+### :bulb: Improving a parse tree by making it more compact
+Here are some things that might help:
+
+**1. Aliasing** - by aliasing different things, you can maintain compact tree structure while marking differences in your grammar. _(Need better description)_
+
+**2. Making things hidden** - by putting an underscore before a rule (`_rule` vs `rule`), it will allow you to omit displaying it in the tree.
+
+### :bulb: Testing in the wild to help prioritize what to support next.
+
+Once you have developed a significant portion of the grammar, find a file from an active open source repository that follows the typical structure of the language and see how much of it can be successfully parsed by your system. Is it 40%? 60%? This can be a good way to benchmark how practical your grammar is in terms of popular usage of your language. It can also be an effective way to identify what to support and prioritize what's next.
+
+### :bulb: Knowing when to stop development
+Most languages have a long-tail of features that don't get utilized frequently. It is reasonable to prioritize developing 80% of a language and only supporting commonly used elements.
+
+### :bulb: Handling conflicts
+
+Conflicts may arise due to ambiguities in the grammar. This is when the parser can not decide what the next symbol in an input stream should be because there are multiple ways to parse some strings. There are three ways to handle conflict:
+1. Rearrange code.
+2. Specifying associativity and/or precedence.
+3. Adding a conflict.
+
+**1. Rearrange code.** There are ways to specify rules in a way that makes it easier for the lexer to digest. This can be done by using the following helper functions:
+
+_(WIP: Explain why the following help you avoid conflicts)_
+
+- `inline` - _(need Max's help here)_
+- `commaSep` - creates a repeating sequence of 0 or more tokens separated by a comma
+- `commaSep1` - creates a repeating sequence of 1 or more tokens separated by a comma
+- `sep1`- creates a repeating sequence of 0 or more tokens separated by the specified delimiter
+
+**2. Specifying associativity and/or precedence.**
+
+Another way of resolving a conflict is through associativity and precedence. Specifying precedence allows us to prioritize productions in the grammar. If there are two or more ways to proceed, the production with the higher precedence will get preference. Left and right associativity can also be used to reflect how to proceed. For instance, a left-associative evaluation is `(a Q b) Q c` vs. a right-associative evaluation would render `a Q (b Q c)`. In this way, associativity changes the meaning of the expression.
+
+_(WIP: Add examples)_
+
+**3. Add a conflict.** 
+
+Adding conflicts allows the parser to pursue two paths in parallel, and decide which one to proceed with further along the process. By specifying a conflict with one rule, you can prevent the parser from recursively descending.
+
+Adding a conflict to your grammar should be a last resort, and it's typically best to exhaust the first two options first.
+
+_(WIP: add example)_


aymannadeem:
Another set of tips that might be useful may include how to approach grammar development depending on the kind of language it is. 

I had this [discussion](https://github.com/github/semantic-diff/issues/1446) prior to starting on the Java grammar. I started with literals, and then expanded outward. And once I hit a level of ambiguity I reversed my approach: starting with class declarations, then modules, interfaces, methods and building "inward".

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+**1. Aliasing** - by aliasing different things, you can maintain compact tree structure while marking differences in your grammar. _(Need better description)_
+
+**2. Making things hidden** - by putting an underscore before a rule (`_rule` vs `rule`), it will allow you to omit displaying it in the tree.
+
+### :bulb: Testing in the wild to help prioritize what to support next.
+
+Once you have developed a significant portion of the grammar, find a file from an active open source repository that follows the typical structure of the language and see how much of it can be successfully parsed by your system. Is it 40%? 60%? This can be a good way to benchmark how practical your grammar is in terms of popular usage of your language. It can also be an effective way to identify what to support and prioritize what's next.
+
+### :bulb: Knowing when to stop development
+Most languages have a long-tail of features that don't get utilized frequently. It is reasonable to prioritize developing 80% of a language and only supporting commonly used elements.
```
robrix: I think we probably want to support 100% of a language—or, curiously, more, since we often want to support multiple versions and such—but not necessarily all in one go, if that makes sense. I.e. spread it out over multiple PRs, and/or accept that you’ll catch errors “in the wild” sometimes. Is that what you mean here?

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+### :bulb: Handling conflicts
+
+Conflicts may arise due to ambiguities in the grammar. This is when the parser can not decide what the next symbol in an input stream should be because there are multiple ways to parse some strings. There are three ways to handle conflict:
```
robrix: This is a great thing to document ✨

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+(program
+  (module_declaration (module_identifier (identifier) (identifier)))
+  (module_declaration (module_identifier (identifier) (identifier))))
+```
+
+### :bulb: Improving a parse tree by making it more compact
+Here are some things that might help:
+
+**1. Aliasing** - by aliasing different things, you can maintain compact tree structure while marking differences in your grammar. _(Need better description)_
```
robrix: An example of this would be especially great 👍

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+**2. Making things hidden** - by putting an underscore before a rule (`_rule` vs `rule`), it will allow you to omit displaying it in the tree.
```
robrix: Here, too, I think an example would help. I.e. “this grammar (with `rule`) gives you this tree, this grammar (with `_rule`) gives you _this_ tree.”

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
+# Tree-sitter tips and tricks
+
+Language development using [Tree-Sitter](https://github.com/tree-sitter/tree-sitter) can be an arduous task. These tips are intended to help bring empiricism to the process by reducing trial and error (or at least introducing some guard rails).
+
+### :bulb: Rely on Language Specs, but not too much.
```
robrix: 💯

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
+# Tree-sitter tips and tricks
+
+Language development using [Tree-Sitter](https://github.com/tree-sitter/tree-sitter) can be an arduous task. These tips are intended to help bring empiricism to the process by reducing trial and error (or at least introducing some guard rails).
```
robrix: That’s a _really_ good goal—and super valuable wisdom from the trenches.

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+### :bulb: Rely on Language Specs, but not too much.
+There are language specifications (like [this one for Java](https://docs.oracle.com/javase/specs/jls/se9/html/index.html)) that comprehensively describe a language. Specs provide the lexical and syntactic structure of a language in addition to its components (ex., types, variables, declarations, etc.) and how they interconnect. This can serve as a great baseline for writing your grammar. However, mirroring the spec 100% is not always the best approach, especially when its structure does not align with the parse tree you'd like to generate using tree-sitter. In this situation, it's better to deviate from the spec and prioritize generating a good parse tree. One way to do this may be to find an LR parser for the language you're working on ([such as this one for Java](https://github.com/javaparser/javaparser)) to see how they've modeled their tree outputs. Studying tree outputs from other LR parsers can be a good way to guide your own grammar development.
```
robrix: Might be worth noting a couple of concrete deviations we’d want to make:

- supporting multiple versions of a language
- producing a simpler tree by parsing a more lenient grammar than the spec allows
- some specs may encode semantics or context-sensitivity that we want to ignore, e.g. a spec for C might state (I’m spec…ulating, it’s been æons since I’ve read the C spec) that `foo.bar` is only valid when `bar` is a field of some `struct` and `foo` is known to have that type, but since those are both semantic rather than syntactic properties, we would want to parse it regardless
  - in some cases that might even imply ambiguity between different syntactic forms which we’d just have to live with by representing both with a single tree!

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+### :bulb: Knowing when to stop development
+Most languages have a long-tail of features that don't get utilized frequently. It is reasonable to prioritize developing 80% of a language and only supporting commonly used elements.
+
+### :bulb: Handling conflicts
+
+Conflicts may arise due to ambiguities in the grammar. This is when the parser can not decide what the next symbol in an input stream should be because there are multiple ways to parse some strings. There are three ways to handle conflict:
+1. Rearrange code.
+2. Specifying associativity and/or precedence.
+3. Adding a conflict.
```
robrix: Another tactic here is to take two rules that parse the same string and combine them into a single rule that gets used in two places. I think this could be part of “rearrange code,” tho I like the focus that item currently has on using combinators.

tree-sitter-tips.md
tclem: What do you think about phrasing this as "Grammar development" instead of "Language development"?

Another thought when I first read this is: should we host this kind of documentation on github/tree-sitter or github/tree-sitter-cli? I think tree-sitter-cli might be a better location as the API is a little higher level (you get to write grammars in JavaScript instead of JSON).

tree-sitter-tips.md
tclem: It would also be good to note the kinds of use cases for these parse trees (e.g. syntax highlighting in a text editor like Atom) as that often drives the level of structure and detail needed.

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+:white_check_mark: Not overly verbose.
+
+:white_check_mark: Everything is outlined clearly.
+
+:white_check_mark: Communicates good information needed further along in the pipeline.
+
+![image](https://user-images.githubusercontent.com/875834/34972577-41bc12ac-fa37-11e7-9a12-a05d18b6f6e4.png)
+
+_Example of a good tree (WIP -- better examples welcome!):_
```
tclem: Might be worth looking at the corpus of examples in something like tree-sitter-javascript for this.

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+### :bulb: Rely on Language Specs, but not too much.
+There are language specifications (like [this one for Java](https://docs.oracle.com/javase/specs/jls/se9/html/index.html)) that comprehensively describe a language. Specs provide the lexical and syntactic structure of a language in addition to its components (ex., types, variables, declarations, etc.) and how they interconnect. This can serve as a great baseline for writing your grammar. However, mirroring the spec 100% is not always the best approach, especially when its structure does not align with the parse tree you'd like to generate using tree-sitter. In this situation, it's better to deviate from the spec and prioritize generating a good parse tree. One way to do this may be to find an LR parser for the language you're working on ([such as this one for Java](https://github.com/javaparser/javaparser)) to see how they've modeled their tree outputs. Studying tree outputs from other LR parsers can be a good way to guide your own grammar development. 
+
+### :bulb: What does a "good" parse tree look like?
+
+Here are some attributes of a "pretty" tree:
```
tclem: :heart: the idea of clearly defining what a good parse tree looks like. This could be built into sort of a style guide for trees. Similar to an earlier comment, I think defining a "good" parse tree is going to be closely related to what you are going to use the tree for. Maybe we could start with stating some of the goals of tree-sitter and work back to a style guide.

tree-sitter-tips.md
tclem: I think this is the part of grammar development that still often confuses me the most: what do you do when there are conflicts? tree-sitter-cli does give some hints for the specific situation - maybe it would help to walk through a concrete example of the kind of conflicts you might run into?

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+3. Adding a conflict.
+
+**1. Rearrange code.** There are ways to specify rules in a way that makes it easier for the lexer to digest. This can be done by using the following helper functions:
+
+_(WIP: Explain why the following help you avoid conflicts)_
+
+- `inline` - _(need Max's help here)_
+- `commaSep` - creates a repeating sequence of 0 or more tokens separated by a comma
+- `commaSep1` - creates a repeating sequence of 1 or more tokens separated by a comma
+- `sep1`- creates a repeating sequence of 0 or more tokens separated by the specified delimiter
```
tclem: I believe these are all helper functions coming either from https://github.com/tree-sitter/tree-sitter-cli#rules or built up from those [primitives](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js#L887). Maybe link to documentation?

I'm not sure I would consider these functions the primary means of dealing with conflicts, maybe an example like you noted, would help. To me, rearranging code often means expressing the rules differently as there are lots of options in the design space. Generally if you have lots of different rules looking for a symbol like `>` you run the risk of conflicts as the presence of that symbol doesn't resolve easily to a single rule.

tree-sitter-tips.md
```diff
@@ -0,0 +1,79 @@
...
+
+**2. Specifying associativity and/or precedence.**
```
tclem: 👍

tclem: Might be good to note that resolving conflicts this way is a compile time solution as opposed to the "Add a conflict" section below which means the parser will try deal with the ambiguity at runtime.

tree-sitter-tips.md
aymannadeem: No, I guess I was trying to posit that while there's always more to do (given the evolving nature of languages), that for our purposes development can be paused. But you're right that I should instead say we can theoretically keep going. 

Would it be fair to say that it's worth knowing what you'll be using the output for, in order to constrain and place some guard rails around what you prioritize supporting?

robrix: Ah—that’s an excellent point! Calling out language evolution explicitly seems like a good idea 👍

> Would it be fair to say that it's worth knowing what you'll be using the output for, in order to constrain and place some guard rails around what you prioritize supporting?

Totally 👍

mads-hartmann:
Thanks a lot for writing this up. Even in its current state it's already very helpful 👍

maxbrunsfeld:
❤️ This was a very helpful reference when I finally got around to adding docs.

@aymannadeem Now that most of this has been incorporated into the Tree-sitter docs site and/or the docs in your app's repo, should we close this one out?

anhangaba:
I'm note sure if this is the place.

But it would be nice having some explanation on how to deal with case insensitive tokens.

In VHDL (and probably Ada) the primitives types can be written in many different ways:

```
bit_vector
Bit_Vector
BIT_VECTOR
```

The lower case or upper cases are more common. But the specs uses the capitalized version in the examples. I'm not sure how to approach this.

PR #145, opened 2018-03-14
small unit test protip

aymannadeem:
Hey @maxbrunsfeld – in writing up some stuff on the Java front, added a minor tweak to this. I created a separate README for the `tree-sitter-java` project in preparation for open sourcing it, but @tclem thought some of those things could roll up into a more general top-level tree-sitter doc.


maxbrunsfeld:
Thanks!

PR #147, opened 2018-03-16
Add API for indicating which external tokens can match the empty string

maxbrunsfeld:
### Problem

Tree-sitter tries to produce the smallest possible parse table for each language by *merging* parse states which are similar in certain ways (See #32). One concern when merging parse states is by adding to the set of tokens that are valid in any given state, we change the behavior of the lexer when it is run in that state. So we need to be careful not to merge states in such a way that introduces lexical conflicts.

The detection of lexical conflicts is made more complicated by the fact that some grammars' external tokens can match the empty string.

Previously, when evaluating whether parse state merges were legal, the ParseTableBuilder would assume that any external token could potentially match the empty string. This logic was added to fix some invalid parse state merging that was happening when generating the bash parser.

Unfortunately, it resulted in overly conservative parse state merging, which was especially undesirable in a language like Ruby with a large number of parse states and a lot of external tokens which *cannot*
actually match the empty string.

### Solution

Now, external tokens which match the empty string should explicitly be marked as such using syntax like this:

```js
externals: $ => [
  $.foo,
  $.bar,
  [$.concat, {blank: true}],
]
```

I wish it weren't necessary to expose this detail in the grammar API, but reducing the size of the ruby parser is very important and I don't see any way to achieve this without requiring the grammar author to explicitly indicate which external tokens have this property.

### DIFF
--- a/src/compiler/build_tables/parse_table_builder.cc
+++ b/src/compiler/build_tables/parse_table_builder.cc
@@ -491,9 +491,11 @@ class ParseTableBuilderImpl : public ParseTableBuilder {
           // Do not add a token if it conflicts with any token in the follow set
           // of an existing external token.
           if (existing_token.is_external()) {
-            const LookaheadSet &following_tokens = following_tokens_by_token[existing_token];
-            for (auto &incompatible_token : incompatible_tokens) {
-              if (following_tokens.contains(incompatible_token)) return false;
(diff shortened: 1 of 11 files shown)


maxbrunsfeld:
> I wish it weren't necessary to expose this detail in the grammar API

Actually, I've realized it's not necessary. Closing and reopening a different PR.

PR #149, opened 2018-03-17
Suggest the use of devDependencies in a parser

### LINKED ISSUE #146: Should tree-sitter-cli really be added as one of dependencies in parser?
Since Atom 1.25.0 has been released a few hours ago, I started learning tree-sitter and read the documentation. In the [Installing the tools](https://tree-sitter.github.io/tree-sitter/creating-parsers.html#installing-the-tools) section, I found the following statement.

> Add `tree-sitter-cli` to the `dependencies` section of `package.json`

In my understanding, however, `tree-sitter-cli` serves as kind of a build tool and is not of much use for the end users (e.g., authors of Atom's language packages). Therefore, I suppose it should be added as a `devDependencies` just like the libraries in this organization do.

I could prepare a PR regarding this point, but before that I want to know if I'm right or not.

yudai-nkt:
I believe the PR title is self-explanatory and doesn't need further explanation.

Fix #146.


maxbrunsfeld:
Thanks! Sorry for misunderstanding at first.

yudai-nkt:
Don't be sorry!

I can't figure out why this tiny change leads to failure in AppVeyor. :thinking: Is this a problem of `tree-sitter-cpp` as there is a contradiction between expected/actual S-expression in `real_grammars.cc`?

maxbrunsfeld:
Yeah, there is a large suite of randomized tests, and there are a couple of tests which are intermittently failing right now. I'm working on a fix, but unfortunately there's some other things I need to work on first. Sorry for the confusing experience.

PR #154, opened 2018-03-30
Add error message for TSCompileErrorTypeInvalidTokenContents

314eter:
I accidentally used a symbol inside a token. An error message could have saved some time.


maxbrunsfeld:
Oh that blank error message must have been very confusing! Thanks for the fix!

314eter:
The reason I did that, was that I wanted to allow escape sequences in strings to appear in the syntax tree, but no whitespace and comments (which are in `extras`). Is there another way to temporarily disable `extras`?

PR #157, opened 2018-04-04
make default extras explicit in docs

Pike:
Took me a while to find out that I could handle whitespace explicitly in my grammar.

Not sure if this is the best place to put it, but I had found it here.

Side note, https://tree-sitter.github.io/tree-sitter/ is hard to find, too. Could you add that as URL to the tree-sitter repo?


maxbrunsfeld:
Thanks for adding this! I have reworked the docs site a bit so it's better organized and finally works on mobile. Now I link to it from the README (and removed the redundant docs from the README).

PR #165, opened 2018-05-09
Remove all use of mutable state in syntax trees

maxbrunsfeld:
Supersedes #162

### Problem

In [Atom](https://github.com/atom/atom) and [Xray](https://github.com/atom/xray), we need to be able to re-parse documents on a background thread. Ideally, we'd also like to be able to *use* syntax trees from multiple threads simultaneously.

Currently, Tree-sitter's syntax trees are *not* cheaply copyable and cannot be shared between multiple threads if any of the threads are performing edits or re-parsing. This is because the syntax tree nodes contain [a small amount of contextual state](https://github.com/tree-sitter/tree-sitter/blob/d5cfc06fa2f03d68b1d2f307519200d35b960af2/src/runtime/tree.h#L30-L36) that is mutated when they are reused in a new tree. The mutable state includes a parent pointer and some other state that is derived from a node's relationship to its parent.

### Solution

By storing a little more data on the stack-allocated `TSNode` structs, we can remove the need for parent pointers in the syntax tree. This will allow us to 'copy' a syntax tree simply by incrementing its reference count. We can then use different copies of syntax trees for different purposes on different threads.

### New API

Syntax trees are no longer owned and encapsulated by a mutable `TSDocument`. The `TSTree` is now a public object of its own. It is created using a `TSParser`, which is now also public.

```c
ts_parser_set_language(parser, ts_language_javascript());

const char *source_code = "console.log(result);";
TSTree *tree = ts_parser_parse_string(parser, NULL, source_code, strlen(source_code));

TSNode node = ts_tree_root_node(tree);
assert(!strcmp(ts_node_type(node), "program"));
```

When a tree is edited, it can be *re-parsed*, which creates a new tree. Both trees are still valid:

```c
ts_tree_edit(tree, &edit);
TSTree *new_tree = ts_parser_parse_string(parser, tree, new_source_code, strlen(new_source_code));
```

Syntax trees can be easily passed to different threads by *copying* them, which is now very cheap. It's fine to edit and re-parse a tree while a copy of that tree is still in use.

```c
TSTree *tree2 = ts_tree_copy(tree);
spawn_background_task(tree2);
ts_tree_edit(tree, &edit);
```

### Tasks

* [x] Remove the use of parent pointers in the syntax tree. Reimplement node APIs to find a node's parent by searching from the root.
* [x] Add a new 'TreeCursor' API for traversing a tree optimally in the absence of parent pointers.
* [x] Use atomic reference counts in syntax trees so that trees can be safely retained and released from multiple threads simultaneously.
* [x] Rework the public runtime API:
  * [x] Make `ts_tree_edit` return a *new* tree instead of mutating the given tree.
  * [x] Remove the `Document` struct which currently owns and encapsulates the tree. Instead, expose the tree as its own public object.
  * [x] Create a public `Parser` struct that can be used in conjunction with an `Input` and optionally an old tree to construct new trees.
* [x] Update README
* [x] Add a unit test that uses syntax trees on multiple threads


rewinfrey:
🎉 👍

PR #177, opened 2018-06-15
Include the HTML parser in the randomized test suite, fix discovered bug

maxbrunsfeld:
It looks like the HTML parser is the first one where both of these are true:

1. The external scanner carries state
2. The external scanner produces `extra` tokens (comments)

There was a bug where we'd make a copy of a `Subtree` but neglect to copy its serialized external scanner state buffer, so we'd end up with two subtrees with pointers to the same heap-allocated buffer, resulting in a double free 😓.

Depends on https://github.com/tree-sitter/tree-sitter-html/pull/1

/cc @queerviolet

### DIFF
--- a/script/fetch-fixtures
+++ b/script/fetch-fixtures
@@ -30,3 +30,4 @@ fetch_grammar go         master
 fetch_grammar ruby       master
 fetch_grammar typescript master
 fetch_grammar bash       master
+fetch_grammar html       master
--- a/script/fetch-fixtures.cmd
+++ b/script/fetch-fixtures.cmd
@@ -9,6 +9,7 @@ call:fetch_grammar go         master
 call:fetch_grammar ruby       master
 call:fetch_grammar typescript master
 call:fetch_grammar bash       master
+call:fetch_grammar html       master
 EXIT /B 0
 
 :fetch_grammar
--- a/src/runtime/subtree.c
+++ b/src/runtime/subtree.c
@@ -67,20 +67,17 @@ bool ts_external_scanner_state_eq(const ExternalScannerState *a, const ExternalS
 
 // SubtreeArray
 
-bool ts_subtree_array_copy(SubtreeArray self, SubtreeArray *dest) {
-  const Subtree **contents = NULL;
-  if (self.capacity > 0) {
-    contents = ts_calloc(self.capacity, sizeof(Subtree *));
-    memcpy(contents, self.contents, self.size * sizeof(Subtree *));
-    for (uint32_t i = 0; i < self.size; i++) {
-      ts_subtree_retain(contents[i]);
-    }
-  }
-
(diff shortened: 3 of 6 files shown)


maxbrunsfeld:
Hmm, looks like the clang static analyzer just now started to take issue with some code that's been there for a while now 🤔 .

queerviolet:
Weird—I don’t see it in the build logs?

~ ashi.

maxbrunsfeld:
It says its a warning, but I think warnings from the static analyzer cause our build to fail:

```
src/runtime/subtree.c:496:48: warning: Dereference of null pointer (loaded from variable 'child')
      *child = ts_subtree_invalidate_lookahead(*child, edit_byte_offset - child_start_byte, pool);
                                               ^~~~~~
```

maxbrunsfeld:
Uh oh, I forgot. `std::unordered_map` doesn't even exist pre-c++11. I think we can just use `std::map` instead, but I'll compare their performance when parsing a big HTML file.

```
test/fixtures/grammars/html/src/tag.h:5:12: error: ‘std::unordered_map’ has not been declared
 using std::unordered_map;
```

queerviolet:
Seems like it shouldn’t matter too much—map uses an r-b tree, IIRC?

~ ashi.

thomasjo:
> It says it's a warning, but I think warnings from the static analyzer cause our build to fail

In my opinion, warnings should _always_ be treated as errors. Most warnings produced by GCC and Clang these days are indicators of potentially very serious problems (memory leaks, narrowing, overflows, etc).

If this was my project, I would add `-Werror` sooner rather than later.

maxbrunsfeld:
Yeah, I'd be fine adding `-Werror`. I think the only thing stopping me from adding it right now is this warning that only gcc seems to issue about initializing a union member: 

  ```
  src/runtime/subtree.c:160:5: warning: missing initializer for field ‘alias_sequence_id’ of ‘struct <anonymous>’ [-Wmissing-field-initializers]
     .alias_sequence_id = 0,
     ^
  In file included from src/runtime/subtree.c:9:0:
  src/runtime/subtree.h:61:16: note: ‘alias_sequence_id’ declared here
       uint16_t alias_sequence_id;
                ^
  ```

  It's showing me the line number where I initialize `alias_sequence_id` and saying `missing initializer for field 'alias_sequence_id'` 🤔. It might be catching some genuine problem, but generating the wrong message?

maxbrunsfeld:
FWIW, here's a flame graph of the time spent parsing this page's source:

tree-sitter-html-flame-graph

It takes 56ms on my laptop which I think is fine for a 481kb file. There is about 1ms spent looking up tags by name. In the flame graph, the function name is truncated, but it appears as (`allocator...`) within the `scan_implicit_end_tag` method on the left. It's small enough that I don't think it's worth worrying about.

PR #193, opened 2018-08-05
Added more detail explanation for Creating Parser section

wongjiahau:
## Changes
- Added more code snippet for Creating Parser section
- Removed the `$` symbols from shell command section, so that the user can copy the command more easily. 
- Break some sentences into several lines


maxbrunsfeld:
Hey, sorry for the lack of response here. There's a few changes you've proposed that I don't want. I did take your suggestion of removing the leading `$` (prompt) symbols from the shell command snippets; good point about that making it harder to copy and paste.

PR #200, opened 2018-09-12
Fix bug where missing token was inserted outside of any included range

maxbrunsfeld:
Fixes https://github.com/atom/language-javascript/issues/595

The parser that we use for [parsing regexes in JavaScript](https://github.com/tree-sitter/tree-sitter-regex) currently fails on certain valid regexes like `{ab`. For that particular regex, it currently tries to recover from the error by inserting a *missing* pattern character before the `{`.

There was previously a bug in the runtime where missing tokens were accidentally allowed to be inserted outside of `included_ranges`. That allowed the erroneous regex to "escape out" of its expected delimiters, causing syntax highlighting problems.

This PR adds a check to make sure that when we insert missing tokens, we position them within one of the current `included_ranges`.

/cc @t9md, @queerviolet

### DIFF
--- a/src/runtime/lexer.c
+++ b/src/runtime/lexer.c
@@ -256,6 +256,10 @@ void ts_lexer_advance_to_end(Lexer *self) {
   }
 }
 
+void ts_lexer_mark_end(Lexer *self) {
+  ts_lexer__mark_end(&self->data);
+}
+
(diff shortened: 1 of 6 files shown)


queerviolet:
Curses! Opened https://github.com/tree-sitter/tree-sitter-regex/issues/1

PR #204, opened 2018-10-04
Add a system for assigning properties to syntax nodes

maxbrunsfeld:
### Motivation

There are several use cases where it'd be nice to be able to specify declaratively that certain nodes have certain properties. For example, I'd like to build a cross-language tool like `ctags` that can be used to link definitions to their corresponding usages, so I'd like to indicate in a generic way:

* which nodes introduce new lexical scopes
* which nodes *define* local variables
* which nodes *reference* local variables
* which nodes *define* functions
* which nodes *reference* functions

However, I'd like to avoid providing this information directly *in* the grammar. I'd like these properties to be an optional feature that you could layer on top of an existing syntax tree.

One analogous technology that we use is CSS, which provides a declarative language for indicating that certain nodes in a tree have certain properties.

### Solution

This PR introduces the concept of a *Property Sheet*. A property sheet is like a stylesheet, but instead of specifying styling, you specify arbitrary properties that should be associated with particular syntax nodes.

Specifically, I've added one new public function to the compiler library (the library used by `tree-sitter-cli`). The function is called `ts_compile_property_sheet`. Together with some functionality that is going to live in `tree-sitter-cli`, this function will allow you to use the CSS language to write a property sheet like this:

```css
/*
 * Statement blocks define new lexical scopes.
 */
statement_block {
  define: scope;
}

/*
 * Variable declarators introduce new local variables.
 */
formal_parameters, variable_declarator {

 /*
  * Within a variable declarator, the name of the local variable
  * is given either by an identifier, or a shorthand property identifier.
  * Note the use of the ampersand operator for nested selectors.
  * This is now standardized.
  */
  & > identifier,
  & object_pattern > shorthand_property_identifier {
    define: local;
  }
}

/*
 * Other identifiers serve as references to existing local variables.
 */
identifier, shorthand_property_identifier {
  reference: local;
}
```

Then, when you run `tree-sitter generate`, Tree-sitter will compile this stylesheet into an **optimal state machine** for matching these selectors against syntax nodes. This state machine has a simple JSON format that can be easily used from any language.

### The State Machine

The state machine is designed to allow a given node's properties to be computed with negligible overhead by descending to the node from the root of the tree. And by maintaining a stack of states, you can walk the tree and compute each node's properties along the way. Representing the state machine as a simple JSON object ensures that bindings like `node-tree-sitter`, `rust-tree-sitter`, and `haskell-tree-sitter` can all easily use it and abstract over it.


maxbrunsfeld:
There currently aren't any tests for this in the core library. It's easier to test at a higher level, because the CSS parsing is done in `tree-sitter-cli`. So all of the tests are currently in `tree-sitter-cli`, added in https://github.com/tree-sitter/tree-sitter-cli/pull/45.

PR #209, opened 2018-10-25
Fix assetion of an example program

vzvu3k6k:
In my environment, [test-json-parser in the document](http://tree-sitter.github.io/tree-sitter/using-parsers#an-example-program) failed with the following message:

```
test-json-parser.c:42: int main(): Assertion `ts_node_child_count(array_node) == 4' failed.
```

It seems that `array_node` has five nodes: `[`, `number`, `,`, `null`, `]`.


maxbrunsfeld:
Thank you!

PR #214, opened 2018-10-28
Regex: hex (\xFF) and unicode (\uFFFF) escape sequence support

jmitchell:
Grammar specifications sometimes refer to code point ranges using a
hexadecimal representation. For example, RFC 5234 which specifies ABNF
includes the following rule in the self-describing ABNF grammar:

    char-val       =  DQUOTE *(%x20-21 / %x23-7E) DQUOTE
                           ; quoted string of SP and VCHAR
                           ;  without DQUOTE

Supporting these escapes in tree-sitter's regular expression grammar
simplifies translation to grammar.js.

### DIFF
--- a/src/compiler/prepare_grammar/parse_regex.cc
+++ b/src/compiler/prepare_grammar/parse_regex.cc
@@ -188,7 +188,7 @@ class PatternParser {
       }
 
       default: {
-        return {single_char(), CompileError::none()};
+        return single_char();
       }
     }
   }
@@ -203,12 +203,17 @@ class PatternParser {
     }
 
     while (has_more_input() && (peek() != ']')) {
-      auto characters = single_char();
+      auto sc_result = single_char();
+      if (sc_result.second.type) return sc_result;
+      auto characters = sc_result.first;
 
       if (peek() == '-') {
         next();
         if (!characters.includes_all && characters.included_chars.size() == 1 && peek() != ']') {
-          auto next_characters = single_char();
+          sc_result = single_char();
+          if (sc_result.second.type) return sc_result;
+          auto next_characters = sc_result.first;
+
           if (!next_characters.includes_all && next_characters.included_chars.size() == 1) {
             characters.include(
               *characters.included_chars.begin(),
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
This looks excellent! Two questions:

1. I noticed you return an error if too few hex digits are supplied after the ‘x’ or ‘u’. Does this match the behavior of instantiating a regex in JavaScript? Or does it just treat it as a literal character sequence like ‘x5’? Ideally, we’d match the behavior of JS there.

2. Did you get a chance to test this end to end, through JavaScript, via tree-sitter-cli?

jmitchell:
Glad you asked. As-is the PR _does not_ accurately reflect what JS does.

```
$ node --version
v8.12.0
$ node
> '0'.match(/\x30/)
[ '0', index: 0, input: '0' ]
> '0'.match(/\x3g/)
null
> 'x3g'.match(/\x3g/)
[ 'x3g', index: 0, input: 'x3g' ]
> 'x5'.match(/\x5/)
[ 'x5', index: 0, input: 'x5' ]
> 'u5'.match(/\u5/)
[ 'u5', index: 0, input: 'u5' ]
```

In case it takes a while for me to get back to this, some relevant parts I found in [ECMA-262](https://www.ecma-international.org/publications/standards/Ecma-262.htm) (9th ed.):

- [21.1 RegExp (Regular Expression) Objects](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-regexp-regular-expression-objects)
    - `CharacterEscape`
        - `HexEscapeSequence`
        - `RegExpUnicodeEscapeSequence`
        - `IdentityEscape`
- [B.1.4 Regular Expressions Patterns](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-regular-expressions-patterns)
    - "... each alternative is considered only if previous production alternatives do not match."

By the way, this patch is currently used to implement [tree-sitter-abnf](https://github.com/jmitchell/tree-sitter-abnf).

jmitchell:
Regex escape sequences seem to be working as expected since the Rust port. Thank you.

maxbrunsfeld:
Ah, thanks for following up! And sorry that I didn't help more to get this PR merged.

jmitchell:
You helped plenty! I care about the functionality, not the particular code :100:

PR #230, opened 2018-11-12
Fix incorrect node reuse for edits right at EOF

maxbrunsfeld:
This fixes a regression I just introduced in https://github.com/tree-sitter/tree-sitter/pull/225.

Due to a mistake in the randomized tests, we weren't testing for edits that occurred *right* at the end of the file (such as inserting a character at the end of the file). We don't often see this in Atom, because there is usually a newline character at the end of the file.

I've added an explicit unit test for this case, as well as generalized the randomized tests to make sure that they cover this.

Thanks to @pranaygp for catching this and providing an excellent reproduction script!


maxbrunsfeld:
Version 0.13.20 of the node module is out.

PR #232, opened 2018-11-13
runtime: make ts_language_symbol_for_name work for alias symbols

bfredl:
I noticed `ts_language_symbol_for_name(c_lang, "type_identifier")` didn't work, even if `ts_node_symbol()` works for such a node, and then `ts_language_symbol_name(c_lang, symbol)` returns `"type_identifier"`.

Is this an oversight or does alias symbols need to be treated differently somehow? The patch seems to work for my usecase. The test fails on master. I simply check the last node, but I could hardcode some node name instead, if we know of a symbol that always will be a alias symbol.

### DIFF
--- a/src/runtime/language.c
+++ b/src/runtime/language.c
@@ -50,7 +50,8 @@ const char *ts_language_symbol_name(const TSLanguage *language, TSSymbol symbol)
 TSSymbol ts_language_symbol_for_name(const TSLanguage *self, const char *name) {
   if (!strcmp(name, "ERROR")) return ts_builtin_sym_error;
 
-  for (TSSymbol i = 0; i < self->symbol_count; i++) {
+  uint32_t count = ts_language_symbol_count(self);
+  for (TSSymbol i = 0; i < count; i++) {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
The failure on Travis was due to a bug in `tree-sitter-python`. I just fixed that and restarted the build.

bfredl:
@maxbrunsfeld Seems to pass now. Does anything need to be changed?

bfredl:
> and a named node with the same name (e.g. the function rule vs the "function" string in the JavaScript grammar).

What would be the proper solution for this? Gradually nudge grammars to use separate names for named and anonymous nodes? (though it would break existing consumers of the grammar?) Or add a variant of `ts_language_symbol_for_name` to add a `bool` flag to prefer which version?

maxbrunsfeld:
> Or add a variant of `ts_language_symbol_for_name` to add a bool flag to prefer which version?

Yeah, that's probably the right solution. We could also just *change* the existing function; there isn't a lot of code that depends on it (that I know of).

PR #234, opened 2018-11-20
Remove `-s` option from `which` in test script

mpsijm:
When I tried running the tests with a debugger, the script failed with the following message:

    Illegal option -s
    Usage: /usr/bin/which [-a] args
    Illegal option -s
    Usage: /usr/bin/which [-a] args
    No debugger found

In case this `-s` flag is needed for other OSes than Linux (I'm working with Ubuntu 18.04), this could be based on a `uname` check (like on line 44).

### DIFF
--- a/script/test
+++ b/script/test
@@ -112,9 +112,9 @@ case ${mode} in
     ;;
 
   debug)
-    if which -s lldb; then
+    if which lldb; then
       lldb $cmd -- "${args[@]}"
-    elif which -s gdb; then
+    elif which gdb; then
       gdb $cmd -- "${args[@]}"
     else
       echo "No debugger found"
--- a/script/test
+++ b/script/test
@@ -112,9 +112,9 @@ case ${mode} in
     ;;
 
   debug)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
The `-s` option was just to silence the output, so we didn't get unwanted paths printed out from the script.

I believe we can achieve the same thing in a cross-platform way by just redirecting the stderr from `which` to `/dev/null`. Also, I've recently learned that it's better to use `hash` than `which`, because `hash` is a bash built-in, so it doesn't require spawning a child process.

thomasjo:
Alternatively `command` which is POSIX (and a built-in).

mpsijm:
Using either `hash` or `command` will cause output on stderr if the command does not exist, so in that case we'll have to do some piping anyway.
I do agree that `hash` is a better alternative to `which`, so I've updated my PR to use that :slightly_smiling_face:

maxbrunsfeld:
Thanks!

PR #236, opened 2018-11-21
Enable direct parsing of UTF-16 strings

zachwaugh:
This adds a new function that can accept an encoding so you can parse a UTF-16 string directly by using the `TSInputEncodingUTF16` encoding. I'm writing Swift bindings for tree-sitter and having the library provide this simplifies things quite a bit. I don't write much C, so I'm not sure if this is the best approach.

### DIFF
--- a/src/runtime/parser.c
+++ b/src/runtime/parser.c
@@ -87,7 +87,7 @@ typedef enum {
 } ErrorComparison;
 
 typedef struct {
-  const void *string;
+  const char *string;
   uint32_t length;
 } TSStringInput;
 
@@ -1647,7 +1647,7 @@ TSTree *ts_parser_parse_string(TSParser *self, const TSTree *old_tree,
 }
 
 TSTree *ts_parser_parse_string_encoding(TSParser *self, const TSTree *old_tree,
-                                        const void *string, uint32_t length, TSInputEncoding encoding) {
+                                        const char *string, uint32_t length, TSInputEncoding encoding) {
   TSStringInput input = {string, length};
   return ts_parser_parse(self, old_tree, (TSInput) {
     &input,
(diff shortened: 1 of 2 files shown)


src/runtime/parser.c
```diff
@@ -87,7 +87,7 @@ typedef enum {
 } ErrorComparison;
 
 typedef struct {
-  const char *string;
+  const void *string;
```
maxbrunsfeld: I think that this field should stay as a `char` pointer, rather than a `void` pointer. Code below this in `ts_string_input_read` expects to be able to do byte-wise pointer arithmetic on the field. It looks like that's why the windows CI is failing.

src/runtime/parser.c
```diff
@@ -1643,11 +1643,16 @@ TSTree *ts_parser_parse(TSParser *self, const TSTree *old_tree, TSInput input) {
 
 TSTree *ts_parser_parse_string(TSParser *self, const TSTree *old_tree,
                                const char *string, uint32_t length) {
+  return ts_parser_parse_string_encoding(self, old_tree, string, length, TSInputEncodingUTF8);
+}
+
+TSTree *ts_parser_parse_string_encoding(TSParser *self, const TSTree *old_tree,
+                                        const void *string, uint32_t length, TSInputEncoding encoding) {
```
maxbrunsfeld: Related to the above comment, I think it might actually be clearer to take a `const char *` pointer here. For applications that want to pass a UTF16 string represented as a different type like `const uint16_t *`, the caller would have to explicitly cast it to a `(const char *)`, but I think that's ok.

zachwaugh: Thanks! I wasn't sure the best way to handle that. I went with the `const void *` since it made it easier to deal with from Swift, but I'll update it here and above to use `const char *` everywhere and do the cast in Swift.

maxbrunsfeld:
Thanks! Very cool to hear that you’re using Tree-sitter on iOS.

PR #246, opened 2018-12-04
Introduce an EXCLUDE rule

maxbrunsfeld:
### Background

Tree-sitter uses context-aware tokenization - in a given parse state, Tree-sitter only recognizes tokens that are syntactically valid in that state. This is what allows Tree-sitter to tokenize languages correctly without requiring the grammar author to think about different lexer modes and states. In general, Tree-sitter tends to be permissive in allowing words that are keywords in some places to be used freely as names in other places.

Sometimes this permissiveness causes unexpected error recoveries. Consider this C syntax error:

```c
float // <-- error
int main() {}
```

Currently, when [tree-sitter-c](https://github.com/tree-sitter/tree-sitter-c) encounters this code, it doesn't detect an error until the word `main`, because it interprets the word `int` as a variable, declared with type `float`. It doesn't see `int` as a keyword, because the keyword `int` wouldn't be allowed in that position.

### Solution

In order improve this error recovery, the grammar author needs a way to explicitly indicate that certain keywords are *not allowed* in certain places. For example in C, primitive types like `int` and control-flow keywords like `while` are not allowed as variable names in declarators.

This PR introduces a new `EXCLUDE` rule to the underlying JSON schema. From JavaScript, you can use it like this:

```js
declarator: choice(
  $.pointer_declarator,
  $.array_declarator,
  $.identifier.exclude('if', 'int', ...etc)
)
```

Conceptually, you're saying "a declarator can match an identifier, but *not* these other tokens".

### Implementation

Internally, all Tree-sitter needs to do is to insert the excluded tokens (`if`, `int`, etc) into the set of valid lookahead symbols that it uses when tokenizing, in the relevant states. Then, when the lexer sees the string "if", it will recognize it as an `if` token, not just an identifier. Then, as always when there's an error, the parser will find that there are no valid parse actions for the token `if`.

### Alternatives

I could have instead introduced a new field on the entire grammar called `keywords`. Then, if we added `if` to the grammar's `keywords`, the word `if` would *always* be treated as its own token, in *every* parse state.

This is less general, and it wouldn't really work AFAICT. Even in C, there are states where the word `int` should *not* be treated as a keyword. For example, inside of a string (`"int"`), as the name of a macro definition `#define int`. And in other languages, there are many *more* cases like this than there are in C. For example in JavaScript, it's fine to have an object property named `if`.

### Relevant Issues

https://github.com/atom/language-c/issues/308

### DIFF
--- a/cli/src/generate/build_tables/build_lex_table.rs
+++ b/cli/src/generate/build_tables/build_lex_table.rs
@@ -32,15 +32,17 @@ pub(crate) fn build_lex_table(
         let tokens = state
             .terminal_entries
             .keys()
+            .copied()
+            .chain(state.exclusions.iter())
             .filter_map(|token| {
                 if token.is_terminal() {
                     if keywords.contains(&token) {
                         syntax_grammar.word_token
                     } else {
-                        Some(*token)
+                        Some(token)
                     }
                 } else if token.is_eof() {
-                    Some(*token)
+                    Some(token)
                 } else {
                     None
                 }
--- a/cli/src/generate/build_tables/build_lex_table.rs
+++ b/cli/src/generate/build_tables/build_lex_table.rs
@@ -33,7 +33,7 @@ pub(crate) fn build_lex_table(
             .terminal_entries
             .keys()
             .copied()
-            .chain(state.exclusions.iter())
+            .chain(state.keywords.iter())
             .filter_map(|token| {
                 if token.is_terminal() {
                     if keywords.contains(&token) {
--- a/cli/src/generate/build_tables/build_parse_table.rs
+++ b/cli/src/generate/build_tables/build_parse_table.rs
@@ -163,6 +163,7 @@ impl<'a> ParseTableBuilder<'a> {
                     external_lex_state_id: 0,
                     terminal_entries: IndexMap::default(),
                     nonterminal_entries: IndexMap::default(),
+                    exclusions: TokenSet::new(),
                     core_id,
                 });
                 self.parse_state_queue.push_back(ParseStateQueueEntry {
@@ -355,7 +356,7 @@ impl<'a> ParseTableBuilder<'a> {
             )?;
         }
 
-        // Finally, add actions for the grammar's `extra` symbols.
+        // Add actions for the grammar's `extra` symbols.
         let state = &mut self.parse_table.states[state_id];
         let is_end_of_non_terminal_extra = state.is_end_of_non_terminal_extra();
 
@@ -421,6 +422,12 @@ impl<'a> ParseTableBuilder<'a> {
             }
         }
 
+        for exclusion in item_set.exclusions() {
+            if !state.terminal_entries.contains_key(&exclusion) {
+                state.exclusions.insert(exclusion);
+            }
+        }
+
         Ok(())
     }
 
@@ -452,7 +459,7 @@ impl<'a> ParseTableBuilder<'a> {
                 if item.step_index > 0 {
                     if self
                         .item_set_builder
-                        .first_set(&step.symbol)
+                        .first_set_without_excluded(&step.symbol)
                         .contains(&conflicting_lookahead)
                     {
                         if item.variable_index != u32::MAX {
@@ -939,8 +946,8 @@ fn populate_following_tokens(
         .collect::<TokenSet>();
     for production in productions {
         for i in 1..production.steps.len() {
-            let left_tokens = builder.last_set(&production.steps[i - 1].symbol);
-            let right_tokens = builder.first_set(&production.steps[i].symbol);
+            let left_tokens = builder.last_set_with_excluded(&production.steps[i - 1].symbol);
+            let right_tokens = builder.first_set_with_excluded(&production.steps[i].symbol);
             for left_token in left_tokens.iter() {
                 if left_token.is_terminal() {
                     result[left_token.index].insert_all_terminals(right_tokens);
--- a/cli/src/generate/build_tables/build_parse_table.rs
+++ b/cli/src/generate/build_tables/build_parse_table.rs
@@ -163,7 +163,7 @@ impl<'a> ParseTableBuilder<'a> {
                     external_lex_state_id: 0,
                     terminal_entries: IndexMap::default(),
                     nonterminal_entries: IndexMap::default(),
-                    exclusions: TokenSet::new(),
+                    keywords: TokenSet::new(),
                     core_id,
                 });
                 self.parse_state_queue.push_back(ParseStateQueueEntry {
@@ -422,9 +422,9 @@ impl<'a> ParseTableBuilder<'a> {
             }
         }
 
-        for exclusion in item_set.exclusions() {
-            if !state.terminal_entries.contains_key(&exclusion) {
-                state.exclusions.insert(exclusion);
+        for keyword in item_set.keywords() {
+            if !state.terminal_entries.contains_key(&keyword) {
+                state.keywords.insert(keyword);
             }
         }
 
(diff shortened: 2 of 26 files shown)


JoeBlanchard:
Hey Max,

I have been working on developing Treesitter grammars for extensible languages developed by the Minnesota Extensible Language Tools Group. This group actually wrote the paper on context aware scanning that you mentioned as part of your underlying research for this project.

This rule seems useful, but also a hassle to list all the keywords that should be excluded wherever an identifier should appear. I was wondering if you had considered implementing a version of lexer classes where these classes of terminals can submit to or dominate other classes of terminals to specify precedence. This can be used to specify that one prefers keywords over identifiers. The [paper](https://www-users.cs.umn.edu/~evw/pubs/vanwyk07gpce/vanwyk07gpce.pdf) describes how this works fully.

The benefit of doing this is, as mentioned in the paper in section 2 under the lexical precedence relation header, that we can always view higher precedence terms as being a member of the valid lookahead set when parsing so we can fix the same problem as you are doing with the exclude rule. This could be done once using submits to and dominates instead of excluding the keywords every place the identifier occurs.

Also, using submits to and dominates with lexer classes defines a partial precedence on terminals that is less constraining and allows more flexibility than the total precedence currently required in Treesitter. This would help Treesitter support more languages more easily. I know it'd likely be a significant undertaking to implement this. So, I'm wondering your rational against doing this and why or why not you would consider implementing this in the future.

I'm not sure where the best place to post this was, but I guess this will do. Thanks for your time!

maxbrunsfeld:
Good to hear from you @JoeBlanchard! There's some things that I like about the approach the paper describes, but I don't fully understand how it addresses this problem.

> we can always view higher precedence terms as being a member of the valid lookahead set when parsing so we can fix the same problem as you are doing with the exclude rule. 

I don't think that's the behavior that I want though. By default, in any given parse state, I don't want any syntactically-invalid tokens to be added to the valid lookahead set. Rather, I want the default to be that we produce the *most permissive* possible parser-lexer pairing.

For example, in JavaScript you can use words like `if` and `else` as identifiers, as long as you're in a context where the `if` and `else` keywords wouldn't be valid.

```js
if (a) {
  const b = {
    if: 1,
    else: 2,
  };
} else {
  // ...
}
```

In Tree-sitter's JavaScript grammar, this 'Just Works', with no extra effort because:
* When an `if` keywords *is* valid, it "dominates" the identifier token because of Tree-sitter's default rule that String tokens dominate Regex tokens.
* Inside of the Object literal above, the `if` and `else` keywords are not valid lookaheads.

This even comes up in older languages like C, because you *can* do stuff like this:

```c
#define float double
#ifdef float
// ...
#endif
```

And in code like that☝️, we parse `float` as an identifier.

If we *implicitly* added keywords like `if` to every lookahead set containing `identifier`, it seems like we'd lose these behaviors.

maxbrunsfeld:
In other words, I want Tree-sitter's default to be highly permissive, treating all keywords as contextual, because most languages really *do* behave this way in many cases. Our number 1 priority is to successfully parse all code that's valid. If we successfully parse some code that's *invalid*, that's not a big problem to me.

Once you have a working parser, I *do* want to add a more niche API that lets you manually expand the valid lookahead set in specific contexts. That's what this `exclude` is for.

> This rule seems useful, but also a hassle to list all the keywords that should be excluded wherever an identifier should appear. 

The way I'm imagining it, we'd only call `exclude` in one place in `tree-sitter-c`: we'd *just* add it to the `_declarator` rule like I showed above, in order to improve Tree-sitter's error recovery in the specific case of code like

```c
float // not done typing yet
int foo;
```

maxbrunsfeld:
I might be misunderstanding the way `dominates` works in the paper though. Am I missing something?

Overall, my impression is that fundamentally, I *do* want this concept to be context-specific, rather than global, whereas `dominates` and `submitsTo` are global concepts.

JoeBlanchard:
Thanks for getting back @maxbrunsfeld 

> Overall, my impression is that fundamentally, I _do_ want this concept to be context-specific, rather than global, whereas `dominates` and `submitsTo` are global concepts.

Yes you are correct that `domiantes` and `submitsTo` are global concepts. Your reasoning makes a lot of sense for the way you have written grammars in Tree-sitter. The approach of keywords, such as `if` and `else`, always dominating identifiers for the purpose of error detection would cause parse errors in the examples you provided despite being correct programs.

Additionally, based on the development timeline of Tree-sitter it is logical that implementing this behavior was not a priority.

However, grammars _can_ be written such that `submitsTo` and `dominates` do work. For example, in the C example you provided.

> This even comes up in older languages like C, because you _can_ do stuff like this:
> 
> ```c
> #define float double
> #ifdef float
> // ...
> #endif
> ```
> 
> And in code like that☝️, we parse `float` as an identifier.
> 
> If we _implicitly_ added keywords like `if` to every lookahead set containing `identifier`, it seems like we'd lose these behaviors.

Here a grammar could be written that has a `macro_identifier` and `identifier` rule that both match the same regular expression but are separate rules. However, the keywords would only dominate the `identifier` and not the `macro_identifier`. Thus, since a `macro_identifier` is in the valid lookahead set while `identifier` is not, the keywords will not be added and this will be parsed correctly.

The main advantage of `submitsTo` and `dominates` is that it transforms precedence, which is a total ordering on all terminals, to a partial ordering only on all terminals where precedence is important and those terminals will appear in the same valid lookahead set. This feature is very useful when writing extensible programming languages which is what the MELT group does, but may not provide as much benefit to most languages used in Atom and on Github. 

Thanks for getting back to me with such a detailed response! By the way, the reason I have been looking into Tree-sitter is that we are attempting to auto-generate the `grammar.js` file for a Tree-sitter specification from a [Silver](http://melt.cs.umn.edu/silver/) grammar specification to generate highlighters for [ableC](http://melt.cs.umn.edu/ableC/) and extensions developed for ableC. I'll probably implement the use of `submitsTo` and `dominates` in Silver using the new EXCLUDE rule for now. When do you plan to merge the rule?

maxbrunsfeld:
> The main advantage of `submitsTo` and `dominates` is that it transforms precedence, which is a total ordering on all terminals, to a partial ordering only on all terminals where precedence is important

Yeah, I can see how that would help with maintainability in some cases, both for lexical disambiguation and for syntactic disambiguation as well. I think that numerical precedence is very *easy* to understand for people, but there are definitely cases where it's awkward to force things into a total ordering.

> When do you plan to merge the rule?

Unfortunately, I got distracted with some other issues, and I haven't had time to work on this. I never even fully got it working correctly, so there's still a bit of work to do.

maxbrunsfeld:
Ok, I have brought this PR back from the dead. So far, I'm keeping the grammar API the same way I originally proposed it.

smatlapudi:
@maxbrunsfeld,  I am checking on this PR as I have the same issue with following SQL code with invalid syntax ( `,`  after `col_one` should not be there). The parser I generated not throwing error as expected but it is taking `from` keyword as column name.
**Note:**   `from` clause is optional this SQL grammar and  just a `select expr, expr` statement is valid one.

```
select 
  col_one,
 from 
  table_one
```

Is this `exclude` solution work for case insensitive keywords ( like SQL grammar here) use-case also ?

JoostK:
Asking here since this was closed: is there still incentive to support something like this in the future?

maxbrunsfeld:
Yes, I’m going to work more on this, but give it a slightly different API and terminology. I’ll link to a new PR soon.

maxbrunsfeld:
Subscribe to https://github.com/tree-sitter/tree-sitter/pull/1635 for further updates on this. I plan to land that PR soon, but it's not quite done.

PR #249, opened 2018-12-13
Return TSTreeCursor by reference

albfan:
To create vala binding to tree-sitter API there're a series of conventions:

https://wiki.gnome.org/Projects/Vala/LegacyBindings

We can deal with most of them (like passing `TSNode` by value) except one: Return `TSTreeCursor` by value in `ts_tree_cursor_new`

This patch is a convenience modification to allocate `TSTreeCursor` struct and free it, allowing vala to use it.

https://gitlab.gnome.org/albfan/vala-tree-sitter/blob/master/.gitmodules#L3-4

I'm not a API expert, but taking with people who is, they suggest to contribute upstream to iron this little quirks. If you are open to API refinement just let us (GNOME) know.

I'm gonna directly use this binding into https://gitlab.gnome.org/albfan/gtk-code-folding-vala/issues/4 but I'm sure there're a bunch of IDEs (GNOME Builder, code from elementary) looking for AST using tree-sitter. As much standard the API is, the better and easier it will be to integrate. 

Don't get me wrong, I know tree-sitter is totally usable right now.

### DIFF
--- a/include/tree_sitter/runtime.h
+++ b/include/tree_sitter/runtime.h
@@ -134,7 +134,7 @@ TSNode ts_node_descendant_for_point_range(TSNode, TSPoint, TSPoint);
 TSNode ts_node_named_descendant_for_point_range(TSNode, TSPoint, TSPoint);
 void ts_node_edit(TSNode *, const TSInputEdit *);
 
-TSTreeCursor ts_tree_cursor_new(TSNode);
+TSTreeCursor *ts_tree_cursor_new(TSNode);
 void ts_tree_cursor_delete(TSTreeCursor *);
 void ts_tree_cursor_reset(TSTreeCursor *, TSNode);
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
> We can deal with most of them (like passing `TSNode` by value) except one: Return `TSTreeCursor` by value in `ts_tree_cursor_new`.

Just for my understanding, what is it about `TSTreeCursor` that makes it inconvenient to return it by value, unlike `TSNode`?

Overall, I'd like to avoid requiring an extra `malloc`/`calloc` call. It doesn't seem necessary, since you can already efficiently implement your desired pointer-based API by *wrapping* the current API, whereas the reverse is not true. If we allocate internally, there's no way for callers to avoid the extra `malloc`/`free` cost.

PR #257, opened 2018-12-28
Fix declarations of functions without arguments

george-hopkins:
When compiled with `-Wstrict-prototypes` warnings are emitted for functions which don't take declare the type of their arguments. By setting the argument list to `void`, we can explicitely specify that the functions take no arguments. The patch only affects C code and is backwards-compatible.

### DIFF
--- a/include/tree_sitter/parser.h
+++ b/include/tree_sitter/parser.h
@@ -92,7 +92,7 @@ struct TSLanguage {
   struct {
     const bool *states;
     const TSSymbol *symbol_map;
-    void *(*create)();
+    void *(*create)(void);
     void (*destroy)(void *);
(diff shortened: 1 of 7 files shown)


maxbrunsfeld:
Hey @george-hopkins, I'm sorry I didn't take the time to understand and merge this before. I actually was unaware of this syntax until recently. 

Since you opened this PR, there have been some major changes to the repo (mainly https://github.com/tree-sitter/tree-sitter/pull/260), so this conflicts now.

I believe I've taken care of the `-Wstrict-prototypes` warning in https://github.com/tree-sitter/tree-sitter/commit/5a59f19b694993dbe841220e6b56508322201236.

Thanks for the contribution, and apologies again that I didn't merge it at the time.

PR #270, opened 2019-02-06
Bytes/C Fixes

vmg:
Hi @maxbrunsfeld! You forgot to merge part of my patch in the `rust-tree-sitter` repo: the attributes walker must be fixed to use `&[u8]` too, so we can continue using it from Treelights.

While I was here, I also fixed two C compilation warnings.

Cheers,
vmg

### DIFF
--- a/cli/src/tests/properties_test.rs
+++ b/cli/src/tests/properties_test.rs
@@ -40,7 +40,7 @@ fn test_walk_with_properties_with_nth_child() {
     parser.set_language(language).unwrap();
     let tree = parser.parse(source_code, None).unwrap();
 
-    let mut cursor = tree.walk_with_properties(&property_sheet, source_code);
+    let mut cursor = tree.walk_with_properties(&property_sheet, source_code.as_bytes());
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
🤦‍♂️ Oh man, sorry about that @vmg.

In the meantime, just so you're not blocked, the result of your PR is still published as 0.3.6.

> get_changed_ranges.c(379): warning C4013: '__attribute__' undefined; assuming extern returning int
> get_changed_ranges.c(379): error C2065: 'fallthrough': undeclared identifier

It looks like our good old Visual Studio doesn't support that `attribute` syntax. I'll take a crack at wrapping that in an `#ifdef` or something.

maxbrunsfeld:
For some reason, the `_POSIX_SOURCE` macro caused `snprintf` to stop being defined in `stdio.h`, when I compile with clang on macOS.

```
lexer.c:102:5: warning: implicitly declaring library function 'snprintf'
```

Do you know what's going on there?

PR #273, opened 2019-02-11
binding: Add overflow checks when slicing in `parse`

vmg:
Because of the way TreeSitter parsers are generated, the parse callback
can ask for a byte offset at the exact end of the byte slice we're
processing, and slicing at the end of the slice in Rust causes a panic,
so we need to explicitly ward against this.

-----------

@maxbrunsfeld: You removed these checks when porting over my patches, but I'm afraid they're necessary. The Treelights test suite panics without them. :sweat:

### DIFF
--- a/lib/binding/lib.rs
+++ b/lib/binding/lib.rs
@@ -229,7 +229,8 @@ impl Parser {
 
     pub fn parse(&mut self, input: impl AsRef<[u8]>, old_tree: Option<&Tree>) -> Option<Tree> {
         let bytes = input.as_ref();
-        self.parse_with(&mut |i, _| &bytes[i..], old_tree)
+        let len = bytes.len();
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
🤔 Wow I'm surprised that the Tree-sitter test suite does not catch this, as it now uses the Rust `Parser::parse` method and parses hundreds of examples. I bet it has something to do with "embedded" languages within files, for which I have less exhaustive tests.

I'm going to merge this for now, but then circle back and fix whatever bug is causing the out-of-bounds offset to be passed.

maxbrunsfeld:
Published to crates.io: 0.3.8.

vmg:
> I bet it has something to do with "embedded" languages within files

Ayup, this is where Treelights panics.

PR #283, opened 2019-02-19
Add a `highlight` sub-crate and expose its functionality through the CLI

maxbrunsfeld:
This PR adds a new top-level directory containing a rust crate called `tree_sitter_highlight`. This crate provides functionality to perform syntax highlighting using a Tree-sitter parser. I've also added a `highlight` subcommand to the Tree-sitter CLI, which uses this new crate to perform syntax highlighting at the command line.

### Library

The `tree_sitter_highlight` crate exposes two public functions:
1. `highlight` returns an iterator that yields `ScopeStart` and `ScopeEnd` values, which can be used to render any kind of syntax highlighting.
2. `highlight_html` returns a vector of HTML strings - one for each line in the document.

### CLI

The `tree-sitter highlight` command renders ANSI colors to the terminal.

Screenshot (note the HTML within JavaScript within HTML):

screen shot 2019-02-19 at 1 11 23 pm

A few steps are required in order for this command to find parsers on your file system:

1. The parsers must be in a well-known location, with a directory name that matches a well-known pattern. Currently, they must match the glob pattern: `~/github/tree-sitter-*`. It'll be pretty easy to make this configurable.

2. The parsers must have some new fields in their `package.json` file. [Here's](https://github.com/tree-sitter/tree-sitter-javascript/blob/035a1101f96b2913085ef4242a50b3e1f3ffbd1c/package.json#L27-L35) an example. The new fields are partly based on how Atom grammars works, documented [here](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/#language-recognition).

The important new fields are:

* `file-types` - An array of filename suffixes. The grammar will be used for files whose names end with one of these suffixes. Note that the suffix may be an entire filename.
* `highlights` - The relative path to a compiled *Property Sheet* that should be used for syntax highlighting.
* `injection-regex` - An optional regex that is used to determine when this language should be used to parse a portion of a larger document. The regex will be tested against the text of a syntax node which is identified by the `injection-language` property of a node in the parent document. More docs on this to come.


highlight/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+    }
+}
+
+pub trait HTMLAttributeCallback<'a>: Fn(Scope) -> &'a str {}
+
+pub fn highlight<'a, F>(
+    source: &'a [u8],
+    language: Language,
+    property_sheet: &'a PropertySheet<Properties>,
+    injection_callback: &'a F,
```
maxbrunsfeld: The highlighting functionality is mostly exposed through this `highlight` function and the `highlight_html` wrapper.

They both take an `injection_callback` parameter that specifies how to retrieve other 'embedded' languages. This callback is necessary because that language retrieval works differently in different contexts. In the Tree-sitter CLI, the languages are loaded lazily and dynamically using `dlopen`, whereas in other apps, the languages may be statically linked into the app.

highlight/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+{
+    Highlighter::new(source, language, property_sheet, injection_callback)
+}
+
+pub fn highlight_html<'a, F1, F2>(
+    source: &'a [u8],
+    language: Language,
+    property_sheet: &'a PropertySheet<Properties>,
+    injection_callback: &'a F1,
+    attribute_callback: &'a F2,
```
maxbrunsfeld: The `highlight_html` function also takes a second callback - `attribute_callback` that specifies how to render a scope as HTML. In the case of the CLI, it uses an inline `style` attribute, whereas other apps may want to map scopes to their own classes.

cli/src/main.rs
```diff
@@ -64,14 +63,30 @@ fn run() -> error::Result<()> {
...
     let home_dir = dirs::home_dir().unwrap();
     let current_dir = env::current_dir().unwrap();
     let config_dir = home_dir.join(".tree-sitter");
+    let theme_path = config_dir.join("theme.json");
+    let parsers_dir = config_dir.join("parsers");
 
-    fs::create_dir_all(&config_dir).unwrap();
-    let mut loader = Loader::new(config_dir);
+    // TODO - make configurable
+    let parser_repo_paths = vec![home_dir.join("github")];
```
maxbrunsfeld: Here I've hard coded the CLI to look for parsers in the `~/github` folder. I could probably allow this list of paths to be specified in a config file in the `~/.tree-sitter` folder instead. I just haven't decided how I want that config file to look yet.

cli/src/highlight.rs
```diff
@@ -0,0 +1,309 @@
...
+              "number": {"color": 94, "bold": true},
+              "property": 124,
+              "operator": {"color": 239, "bold": true},
+              "punctuation.bracket": 239,
+              "punctuation.delimiter": 239,
+              "string.special": 30,
+              "string": 28,
+              "tag": {"color": 18},
+              "variable.builtin": {"bold": true}
+            }
```
maxbrunsfeld: The CLI will load an optional *theme* file at `~/.tree-sitter/theme.json` that looks like this. The keys are the scopes, and the values are ansi color numbers.

maxbrunsfeld:
I published this to crates.io as `tree_sitter_highlight`, version 0.1.0.

vmg:
Neato! If I'm looking at this right we should be able to drop all Tree-Sitter specific highlight code from Treelights and depend directly on this crate now. Is that correct?

maxbrunsfeld:
@vmg Yeah, exactly. I've just opened a PR to do that.

PR #293, opened 2019-03-08
Add a C API for syntax highlighting

maxbrunsfeld:
This PR adds a C API for the `tree-sitter-highlight` crate.

Now, when you run `cargo build`, a static library called `libtree_sitter_highlight.a` is produced, in addition to the other build artifacts.

My thought is that in a long-lived server, we'd have a single global `TSHighlighter` instance, and each thread (or just each request) would have its own `TSHighlightBuffer`.

```c
#include "tree_sitter/api.h"
#include "tree_sitter/highlight.h"
#include <assert.h>

// Specify what HTML attributes to add for each scope
const char *ATTRIBUTE_STRINGS[50] = {
  [TSHighlightScopeKeyword] = "style='color: purple;'",
  [TSHighlightScopeFunction] = "style='color: red;'",
};

// Declare our languages
const TSLanguage *tree_sitter_html();
const TSLanguage *tree_sitter_javascript();

int main() {
  // Build our highlighter and our output buffer
  TSHighlighter *highlighter = ts_highlighter_new(ATTRIBUTE_STRINGS);
  TSHighlightBuffer *buffer = ts_highlight_buffer_new();

  // Register the HTML language
  TSHighlightError error;
  error = ts_highlighter_add_language(
      highlighter,
      "text.html.basic",
      tree_sitter_html(),
      THE_HTML_PROPERTY_SHEET_JSON,
      NULL
  );
  assert(error == 0);

  // Register the JavaScript language
  error = ts_highlighter_add_language(
      highlighter,
      "source.js",
      tree_sitter_javascript(),
      THE_JS_PROPERTY_SHEET_JSON,
      "^javascript$"
  );
  assert(error == 0);

  // Highlight some source code, writing into the buffer
  const char *source_code = "<script>const x = y();</script>";
  error = ts_highlighter_highlight(
      highlighter,
      "text.html.basic",
      source_code,
      strlen(source_code),
      buffer
  );
  assert(error == 0);

  // Get the content and the line info from the buffer.
  uint32_t html_len = ts_highlight_buffer_len(buffer);
  uint32_t html_line_count = ts_highlight_buffer_line_count(buffer);
  const char *html = ts_highlight_buffer_content(buffer);
  const uint32_t *html_line_offsets = ts_highlight_buffer_line_offsets(buffer);

  printf("The HTML's length is %u\n", html_len);
  printf("The HTML has %u lines\n", html_line_count);
  printf("The HTML's second line starts at byte %u\n", html_line_offsets[1]);
  printf("Here's the HTML: %s", html);

  // Clean up
  ts_highlighter_delete(highlighter);
  ts_highlight_buffer_delete(buffer);
}
```

Compilation is pretty straightforward:

```sh
clang \
  -I lib/include \
  -I highlight/include \
  target/debug/libtree_sitter_highlight.a \
  test.c
```

/cc @tclem @vmg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -661,12 +661,12 @@ dependencies = [
  "spin 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.0.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "tree-sitter 0.3.9",
- "tree-sitter-highlight 0.1.4",
+ "tree-sitter-highlight 0.1.5",
 ]
 
 [[package]]
 name = "tree-sitter-highlight"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "regex 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -6,6 +6,7 @@ use serde::ser::SerializeMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::{json, Value};
 use std::collections::HashMap;
+use std::time::Instant;
 use std::{fmt, fs, io, path};
 use tree_sitter::{Language, PropertySheet};
 use tree_sitter_highlight::{highlight, highlight_html, HighlightEvent, Properties, Scope};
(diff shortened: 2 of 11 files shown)


vmg:
This looks great, Max. Having a persistent buffer to be reused between calls is the way I implemented this for PrettyLights and it makes a difference when it comes to allocation overhead.

I think the only requirement for integrating this into the Go service is having a separate build library that includes all the external TreeSitter grammars _and_ calls `ts_highlighter_add_language` to generate a fully initialized highlighter. Linking & initializing the highlighters from Go is gonna be a headache otherwise.

 I'm not sure whether that library should be implemented in C or Rust, but either approach would work.

Issue #303, opened 2019-03-14
Stale code: build fails when DEBUG_GET_CHANGED_RANGES is defined


jmitchell:
`get_changed_ranges.c` has some `#ifdef DEBUG_GET_CHANGED_RANGES ... #endif` code, some of which has gone stale and causes the build to fail when it's defined.

By default `DEBUG_GET_CHANGED_RANGES` [isn't defined](https://github.com/tree-sitter/tree-sitter/blob/2de54c101ebb7c65423be3e6a502875dc6db4b1f/lib/src/get_changed_ranges.c#L8), so this isn't a pressing issue.

After this is fixed perhaps `#ifdefs` like this should be covered in CI.

### Setup

```
$ git rev-parse HEAD
2de54c101ebb7c65423be3e6a502875dc6db4b1f

$ git diff
diff --git a/lib/src/get_changed_ranges.c b/lib/src/get_changed_ranges.c
index 8eb89d46..bf54e145 100644
--- a/lib/src/get_changed_ranges.c
+++ b/lib/src/get_changed_ranges.c
@@ -5,7 +5,7 @@
 #include "./tree_cursor.h"
 #include <assert.h>
 
-// #define DEBUG_GET_CHANGED_RANGES
+#define DEBUG_GET_CHANGED_RANGES
 
 static void ts_range_array_add(TSRangeArray *self, Length start, Length end) {
   if (self->size > 0) {
```

### Failure

```
$ ./script/test
...
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "include
" "-I" "utf8proc" "-Wall" "-Wextra" "-std=c99" "-Wno-unused-parameter" "-DTREE_SITTER_TEST=" "-o" "/home/jake/s/3/githu
b/tree-sitter/tree-sitter/target/debug/build/tree-sitter-9fe5396995803c33/out/src/lib.o" "-c" "src/lib.c"             
cargo:warning=In file included from src/lib.c:10:
cargo:warning=src/./get_changed_ranges.c:329:77: error: no member named 'symbol' in 'union Subtree'                   
cargo:warning=  const char *name = ts_language_symbol_name(self->language, entry.subtree->symbol);                    
cargo:warning=                                                             ~~~~~~~~~~~~~  ^                           
cargo:warning=1 error generated.
exit code: 1

--- stderr
thread 'main' panicked at '

[...]

PR #304, opened 2019-03-14
Use 1-indexed rows in CLI and log output (resolves #287)

### LINKED ISSUE #287: 0-indexed line numbers in CLI output
Parser output and error messages refer to rows in the input using 0-indexed values, but editors and many other tools conventionally using 1-indexing for row numbers.

Has switching to 1-indexing been considered? I don't necessarily think it's a good idea to change the internal representation in the generated C code, but it may be less surprising for users if all output and errors were 1-indexed instead.

jmitchell:
Switches the CLI/logging output from 0-indexed rows to 1-indexed.

I _think_ I caught all the relevant cases where rows are emitted, although it's possible some are missing.

## Automated test coverage

- [x] Verify `row:0` doesn't appear in debug logs
- [x] Verify `position: 0,` doesn't appear in graph logs when the `TREE_SITTER_TEST_ENABLE_LOG_GRAPHS` environment variable is manually set
- [ ] Verify `[0,` doesn't appear in `tree-sitter parse` standard output
  - checked manually
- [ ] Verify fixes in `get_changed_ranges.c`
  - not attempted, in part because of #303

### DIFF
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -56,6 +56,10 @@ fn test_parsing_with_logging() {
         "reduce sym:struct_item, child_count:3".to_string()
     )));
     assert!(messages.contains(&(LogType::Lex, "skip character:' '".to_string())));
+
+    for (_, m) in &messages {
+        assert!(!m.contains("row:0"));
+    }
 }
 
 #[test]
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -63,6 +63,29 @@ fn test_parsing_with_logging() {
 }
 
 #[test]
+#[cfg(unix)]
+fn test_parsing_with_debug_graph_enabled() {
+    use std::io::{BufRead, BufReader, Seek};
+
+    let has_zero_indexed_row = |s: &str| s.contains("position: 0,");
+
+    let mut parser = Parser::new();
+    parser.set_language(get_language("javascript")).unwrap();
+
+    let mut debug_graph_file = tempfile::tempfile().unwrap();
+    parser.print_dot_graphs(&debug_graph_file);
+    parser.parse("const zero = 0", None).unwrap();
+
+    debug_graph_file.seek(std::io::SeekFrom::Start(0)).unwrap();
+    let log_reader = BufReader::new(debug_graph_file)
+        .lines()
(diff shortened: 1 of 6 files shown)


cli/src/util.rs
```diff
@@ -55,6 +55,9 @@ impl Drop for LogSession {
             {
                 Command::new("open").arg(&self.0).output().unwrap();
             }
+
+            #[cfg(any(debug_assertions, test))]
+            validate_graph_log(&self);
```
maxbrunsfeld: Very cool that you're verifying the graph logs!

My only suggestion is that I think we should perform this check in an explicit unit test of this functionality. Right now, this `LogSession` struct isn't used during a normal test run; it's just a tool that we use to debug failing tests (via `script/test THE_TEST_NAME -D`). And it's kind of an optional feature of the test suite: you can only use it if you have Graphviz installed on your system.

What do you think about adding a specific unit test in `parser_test.rs`, something like this:

```rust
#[test]
#[cfg(unix)]
fn test_parsing_with_debug_graph_enabled() {
    let mut debug_graph_file = tempfile::tempfile().unwrap();
    parser.print_dot_graphs(&debug_graph_file);

    // validate the debug graph 
}	
```

I think that it might be better to perform this validation

cli/src/tests/parser_test.rs
```diff
@@ -56,6 +56,10 @@ fn test_parsing_with_logging() {
         "reduce sym:struct_item, child_count:3".to_string()
     )));
     assert!(messages.contains(&(LogType::Lex, "skip character:' '".to_string())));
+
+    for (_, m) in &messages {
+        assert!(!m.contains("row:0"));
+    }
```
maxbrunsfeld: Thanks for updating all the log messages too!

docs/section-3-creating-parsers.md
```diff
@@ -86,7 +86,7 @@ tree-sitter parse ./the-file
 This should print the following:
 
 ```
-(source_file [0, 0] - [0, 5])
+(source_file [1, 0] - [1, 5])
```
maxbrunsfeld: ✨

cli/src/util.rs
jmitchell: I like this approach much better. Pushed an update and waiting for CI feedback.

maxbrunsfeld:
Awesome! Thanks @jmitchell.

PR #309, opened 2019-03-24
fix example snippet in creating parsers docs

justinwoo:
Maybe I don't quite understand, but afaict the input `hello;` for a
grammar expecting `hello` yields now the result:

```
(source_file [0, 0] - [1, 0]
  (ERROR [0, 5] - [0, 6]
      (ERROR [0, 5] - [0, 6])))
```

Since there is an error from the unexpected semicolon?


maxbrunsfeld:
Thanks @justinwoo!

PR #319, opened 2019-04-09
Allow hex characters in unicode code points

ervinoro:
Previous version of the CLI incorrectly escaped unicode code points such as `\u{1000A}`, causing error among the lines of 
```
Rule <rulename> regex parse error:
    \u\{1000A\}
       ^
error: invalid hexadecimal digit
```

For background, this fix allows unicode character classes to be used through https://github.com/mathiasbynens/regexpu-core `{'unicodePropertyEscape': true, 'useUnicodeFlag': true}` .

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -12,7 +12,7 @@ use std::i32;
 
 lazy_static! {
     static ref CURLY_BRACE_REGEX: Regex =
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Another way to do this would be to *avoid* escaping curly braces if they are prefixed by `\u`, but if this works I'm cool with it.

Can you add one or more tests for this? The existing tests are [at the bottom of the file](https://github.com/tree-sitter/tree-sitter/blob/8c845f29e02763c94b2df0e22ff7536925dda66f/cli/src/generate/prepare_grammar/expand_tokens.rs#L459-L661).

maxbrunsfeld:
Thanks!

PR #320, opened 2019-04-14
Create issues_label_bot.yaml

hamelsmu:
alias from `feature_request` to `enhancement`


maxbrunsfeld:
I think I'm cool with the default configuration actually. Thanks!

Issue #252, opened 2018-12-19
WebAssembly bindings


drom:
It would be useful to provide WebAssembly bindings similar to https://github.com/tree-sitter/node-tree-sitter then tree-sitter can be used directly on the web.

drom:
Was any work done in this direction?

boggle:
Yeah ultimately it would be great if treesitter just moved from C to rust and by doing so would get wasm support for free.

maxbrunsfeld:
I've hacked on a WASM binding a little bit.

Porting Tree-sitter from C to Rust would not achieve WASM support for free. Rust does have [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen), which is nice, but the problem that it solves (wrapping native functions in a JS interface) is easy to solve in C too.

One interesting problem we'll need to solve with the WASM binding is "dynamic linking/loading" - compiling each parser to WASM separately and then composing them together at runtime. WASM (and emscripten specifically) have facilities for dynamic linking, we'd just have to figure out how to leverage them.

Another problem is that there's currently no way to get notified when a JavaScript object is garbage-collected (except for in Node.js), so applications using Tree-sitter through WASM would have to manually call some kind of `.delete` method on `Parser`, `Tree`, and `TreeCursor` objects whenever they were done using them. This is kind of inconvenient.

drom:
> Yeah ultimately it would be great if treesitter just moved from C to rust and by doing so would get wasm support for free.

I don't like an idea of `tree-sitter` emitting Rust code INSTEAD of C.
  1. C is way more portable
  2. It does not solve the issue of creating `wasm-tree-sitter` bindings.

drom:
> Porting Tree-sitter from C to Rust would not achieve WASM support for free. Rust does have [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen), which is nice, but the problem that it solves (wrapping native functions in a JS interface) is easy to solve in C too.

Agree.

> One interesting problem we'll need to solve with the WASM binding is "dynamic linking/loading" - compiling each parser to WASM separately and then composing them together at runtime. WASM (and emscripten specifically) have facilities for dynamic linking, we'd just have to figure out how to leverage them.

My Verilog parser binary is 45MiB and I guess tree-sitter runtime is ~7MiB.
I am OK with static linking in the first version.

[...]

PR #321, opened 2019-04-24
Add a Tree-sitter Wasm binding and a Wasm-based web playground

maxbrunsfeld:
Closes https://github.com/tree-sitter/tree-sitter/issues/252

## Features

This PR adds the following:

#### A Wasm library

Tree-sitter can now be used from JavaScript on the web!

You can run the new script `script/build-wasm` to create two files `tree-sitter.js` and `tree-sitter.wasm` in the `target/release` directory. This JavaScript library defines a global `TreeSitter` object with `Parser`, `Tree`, and `Node` classes similar to the classes in [node-tree-sitter](https://github.com/tree-sitter/node-tree-sitter). Library initialization and parser loading are now asynchronous in the web environment:

```js
await TreeSitter.init();
const JSLanguage = await TreeSitter.Language.load("/path/to/tree-sitter-javascript.wasm");
const parser = new TreeSitter.Parser();
parser.setLanguage(JSLanguage);
```

#### A CLI command for compiling parsers to Wasm

If you have `docker` installed, you can now compile a given Tree-sitter parser (e.g. `tree-sitter-python`) using the new `tree-sitter build-wasm` subcommand:

```sh
tree-sitter build-wasm path/to/tree-sitter-python
```

You can omit the path if you're already in the `tree-sitter-python` directory. This will output a single file called `tree-sitter-python.wasm` in the current directory.

#### A Web Playground

When this is merged, there will be a new `Playground`  page on the docs site [tree-sitter.github.io](http://tree-sitter.github.io) that allows you to try a number of different parsers interactively from the web browser.

### Code Size

The current implementation uses Emscripten's [side modules](https://github.com/emscripten-core/emscripten/wiki/WebAssembly-Standalone#create-a-dynamic-library) feature to allow each compiled and loaded separately.

Unfortunately, in order to successfully load parser that use C++ in their [external scanners](http://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners), the core library needs to contain any standard library functions that these scanners might use. Currently, I'm following [this advice](https://github.com/emscripten-core/emscripten/issues/8308#issuecomment-474467507) and just including all of libc++ in the core module.

The result is that the core library is quite large 😕 

| file | size (minified) | size (gzipped) |
|----|----------------------|----------------|
| `tree-sitter.js` | 793K | 155K |
| `tree-sitter.wasm` | 813K | 342K |

The individual parser binaries aren't too bad, but are still larger than your average javascript library:

| file | size (uncompressed) | size (gzipped) |
|----|----------------------|----------------|
| `tree-sitter-c.wasm` | 694K | 53K |
| `tree-sitter-javascript.wasm` | 1.0M | 83K |
| `tree-sitter-rust.wasm` | 2.2M | 154K |

### Memory Management

Currently, in order to free the memory that a `Parser` or a `Tree` owns on the Wasm heap, you need to call `.delete()` on it.

In seems like is a common problem for Wasm-backed JS libraries, and a solution is already being implemented: [the ECMAScript Weakrefs API](https://github.com/tc39/proposal-weakrefs), which is in stage 2 of the proposal process. This API will allow us to make get rid of the `.delete` methods.

### Tasks

* [x] Get parsers loading and working
* Implement remaining APIs
  * [x] Tree.edit
  * [ ] TreeCursor
* [x] Build the wasm library on CI
* [x] Add unit tests for the wasm library, based on the ones in node-tree-sitter
* [x] Make the playground page nicer

## Next Steps

There are probably ways to avoid some of this bloat, especially in the core library. For comparison, the native-compiled Tree-sitter library is only around 490K uncompressed. Maybe we can include only a subset of libc++ in the core library. Alternatively, maybe we can somehow include the necessary pieces of libc++ into each parser library.

Emscripten and Wasm both seem to be changing pretty quickly, so I didn't put too much effort into optimizing the builds yet. I think I'll wait and do that in a separate pass.


maxbrunsfeld:
Ok, I've addressed the code size problem with the core library, by just whitelisting the functions that emscripten should export. Here are the new sizes:

uncompressed:
```
55K  tree-sitter.js
252K tree-sitter.wasm
```

gzipped:
```
15K  tree-sitter.js.gz
83K  tree-sitter.wasm.gz
```

curran:
I found myself wondering

>Is the WASM build available as an NPM package by any chance?

In case others end up here wondering the same, here it is:

https://www.npmjs.com/package/web-tree-sitter

pckhoi:
Does the Wasm binary include Tree-sitter Highlight?

PR #330, opened 2019-05-08
Update doc example program file locations

andycraig:
Currently, the compilation instructions for the example C program aren't quite consistent with the actual directory structure, so the example as given doesn't compile. This PR fixes that.


maxbrunsfeld:
Thank you!

PR #334, opened 2019-05-17
Reduce parsers' static memory footprint by storing "small" parse states more compactly

maxbrunsfeld:
### Background

Currently, Tree-sitter's parse table is represented as a two-dimensional array of `uint16_t` values (which represent either action ids or successor state ids), indexed by parse state and by lookahead symbol.

```c
static uint16_t ts_parse_table[STATE_COUNT][SYMBOL_COUNT] = {
  [0] = {
    sym_identifier = ACTIONS(250),
    anon_sym_LBRACE = ACTIONS(251),
    sym_expression = STATE(102),

    // ...
  },

  // ...
};
```

### Problem

With many parsers having around 2000 states and 200 symbols, this array occupies a significant amount of statically-allocated memory. I wrote a script to display the sizes of each symbol in a Tree-sitter parser binary:

```sh
$ ./script/show-symbol-sizes ~/.tree-sitter/bin/javascript.so

total                                                            	 1034.8 kb
_ts_parse_table                                                  	 931.6 kb
_ts_parse_actions                                                	 40.1 kb
_ts_lex                                                          	 34.6 kb
_ts_lex_modes                                                    	 8.4 kb
_ts_lex_keywords                                                 	 7.4 kb
_ts_alias_sequences                                              	 4.7 kb

# ...
```

### Solution

A lot of states in this table are *sparse* - they have very few valid lookahead symbols. This means that we can save space by representing them in a different way.

In this PR, I've introduced the notion of *small parse states*. These states are represented as arrays of (`lookahead`, `value`) pairs instead of as arrays of size `SYMBOL_COUNT`, indexed by lookahead. 

```c
static uint16_t ts_small_parse_table[] = {
  [0] = 3,
    sym_identifier, ACTIONS(1306),
    anon_sym_extern, ACTIONS(27),
    anon_sym_static, ACTIONS(27),
  [7] = 2,
    sym_comment, ACTIONS(3),
    sym_parenthesized_expression, STATE(102),

  // ...
};
```

The small parse states are all stored in a single 1-D array. The *starting index* of each small state is stored in a separate array:

```c
static uint32_t ts_small_parse_table_map[] = {
  [SMALL_STATE(385)] = 0,
  [SMALL_STATE(386)] = 97,
  [SMALL_STATE(387)] = 194,

  // ...
};
```

So the procedure for *looking* up a value in the parse table [now](https://github.com/tree-sitter/tree-sitter/blob/8af0afac2eb2cf644e5d7a029647e55c3c48e4a4/lib/src/language.h#L55-L77) has a little bit of conditional logic.

### Results

This reduces the size of language binaries (and their static memory footprint) by 50% to 75%:

```sh
 $ ./script/show-symbol-sizes ~/.tree-sitter/bin/javascript.so
total                                                            	 570.9 kb
_ts_parse_table                                                  	 362.4 kb
_ts_small_parse_table                                            	 99.7 kb
_ts_parse_actions                                                	 40.1 kb
# ...
```

In Python, the *majority* of parse states are actually *small*, so there's more than a 50% savings:

```sh
$ ./script/show-symbol-sizes ~/.tree-sitter/bin/python.so
total                                                            	 339.5 kb
_ts_small_parse_table                                            	 217.3 kb
_ts_parse_actions                                                	 39.3 kb
_ts_parse_table                                                  	 24.9 kb
```

### Notes

* **ABI versioning** - This PR entails another backward-compatible ABI change, so I've bumped `TREE_SITTER_LANGUAGE_VERSION` up to 11. But the library will still be able to *load* parsers that were compiled with ABI version 9 and up, so the transition will be easy to manage.

* **Runtime cost** - This change doesn't make a measurable difference in parsing speed. The symbols within a small parse state are ordered, so we *could* search them using a binary search, but for small arrays, I'm not sure it's worth it. Currently I just use a linear search with an early `break` based on the ordering.

* **WASM** - Unfortunately, the *gzipped* size of the binaries is not really affected by this change. Some binaries gzip *slightly* smaller, and some have actually gotten slightly larger. On average, they remain around 70k gzipped. Still, I think the memory savings are worthwhile in their own right.

🎩 to @marijnh for pointing out how much room for optimization exists due to sparse parse states.

### DIFF
--- a/cli/src/generate/build_tables/build_parse_table.rs
+++ b/cli/src/generate/build_tables/build_parse_table.rs
@@ -108,6 +108,7 @@ impl<'a> ParseTableBuilder<'a> {
                 self.parse_table.states.push(ParseState {
                     id: state_id,
                     lex_state_id: 0,
+                    external_lex_state_id: 0,
                     terminal_entries: HashMap::new(),
                     nonterminal_entries: HashMap::new(),
                     core_id,
@@ -777,6 +778,7 @@ pub(crate) fn build_parse_table(
         parse_table: ParseTable {
             states: Vec::new(),
             symbols: Vec::new(),
+            external_lex_states: Vec::new(),
             production_infos: Vec::new(),
             max_aliased_production_length: 1,
         },
--- a/cli/src/generate/build_tables/build_parse_table.rs
+++ b/cli/src/generate/build_tables/build_parse_table.rs
@@ -25,10 +25,11 @@ struct AuxiliarySymbolInfo {
 type SymbolSequence = Vec<Symbol>;
 type AuxiliarySymbolSequence = Vec<AuxiliarySymbolInfo>;
 
+pub(crate) type ParseStateInfo<'a> = (SymbolSequence, ParseItemSet<'a>);
+
 struct ParseStateQueueEntry {
-    preceding_symbols: SymbolSequence,
-    preceding_auxiliary_symbols: AuxiliarySymbolSequence,
     state_id: ParseStateId,
+    preceding_auxiliary_symbols: AuxiliarySymbolSequence,
 }
 
 struct ParseTableBuilder<'a> {
@@ -38,13 +39,13 @@ struct ParseTableBuilder<'a> {
     variable_info: &'a Vec<VariableInfo>,
     core_ids_by_core: HashMap<ParseItemSetCore<'a>, usize>,
     state_ids_by_item_set: HashMap<ParseItemSet<'a>, ParseStateId>,
-    item_sets_by_state_id: Vec<ParseItemSet<'a>>,
+    parse_state_info_by_id: Vec<ParseStateInfo<'a>>,
     parse_state_queue: VecDeque<ParseStateQueueEntry>,
     parse_table: ParseTable,
 }
 
 impl<'a> ParseTableBuilder<'a> {
-    fn build(mut self) -> Result<ParseTable> {
+    fn build(mut self) -> Result<(ParseTable, Vec<ParseStateInfo<'a>>)> {
         // Ensure that the empty alias sequence has index 0.
         self.parse_table
             .production_infos
@@ -70,9 +71,10 @@ impl<'a> ParseTableBuilder<'a> {
         while let Some(entry) = self.parse_state_queue.pop_front() {
             let item_set = self
                 .item_set_builder
-                .transitive_closure(&self.item_sets_by_state_id[entry.state_id]);
+                .transitive_closure(&self.parse_state_info_by_id[entry.state_id].1);
+
             self.add_actions(
-                entry.preceding_symbols,
+                self.parse_state_info_by_id[entry.state_id].0.clone(),
                 entry.preceding_auxiliary_symbols,
                 entry.state_id,
                 item_set,
@@ -81,7 +83,7 @@ impl<'a> ParseTableBuilder<'a> {
 
         self.remove_precedences();
 
-        Ok(self.parse_table)
+        Ok((self.parse_table, self.parse_state_info_by_id))
     }
 
     fn add_parse_state(
@@ -104,7 +106,9 @@ impl<'a> ParseTableBuilder<'a> {
                 };
 
                 let state_id = self.parse_table.states.len();
-                self.item_sets_by_state_id.push(v.key().clone());
+                self.parse_state_info_by_id
+                    .push((preceding_symbols.clone(), v.key().clone()));
+
                 self.parse_table.states.push(ParseState {
                     id: state_id,
                     lex_state_id: 0,
@@ -115,7 +119,6 @@ impl<'a> ParseTableBuilder<'a> {
                 });
                 self.parse_state_queue.push_back(ParseStateQueueEntry {
                     state_id,
-                    preceding_symbols: preceding_symbols.clone(),
                     preceding_auxiliary_symbols: preceding_auxiliary_symbols.clone(),
                 });
                 v.insert(state_id);
@@ -751,12 +754,12 @@ fn populate_following_tokens(
     }
 }
 
-pub(crate) fn build_parse_table(
-    syntax_grammar: &SyntaxGrammar,
-    lexical_grammar: &LexicalGrammar,
-    inlines: &InlinedProductionMap,
-    variable_info: &Vec<VariableInfo>,
-) -> Result<(ParseTable, Vec<TokenSet>)> {
+pub(crate) fn build_parse_table<'a>(
+    syntax_grammar: &'a SyntaxGrammar,
+    lexical_grammar: &'a LexicalGrammar,
+    inlines: &'a InlinedProductionMap,
+    variable_info: &'a Vec<VariableInfo>,
+) -> Result<(ParseTable, Vec<TokenSet>, Vec<ParseStateInfo<'a>>)> {
     let item_set_builder = ParseItemSetBuilder::new(syntax_grammar, lexical_grammar, inlines);
     let mut following_tokens = vec![TokenSet::new(); lexical_grammar.variables.len()];
     populate_following_tokens(
@@ -766,14 +769,14 @@ pub(crate) fn build_parse_table(
         &item_set_builder,
     );
 
-    let table = ParseTableBuilder {
+    let (table, item_sets) = ParseTableBuilder {
         syntax_grammar,
         lexical_grammar,
         item_set_builder,
         variable_info,
         state_ids_by_item_set: HashMap::new(),
         core_ids_by_core: HashMap::new(),
-        item_sets_by_state_id: Vec::new(),
+        parse_state_info_by_id: Vec::new(),
         parse_state_queue: VecDeque::new(),
         parse_table: ParseTable {
             states: Vec::new(),
@@ -785,5 +788,5 @@ pub(crate) fn build_parse_table(
     }
     .build()?;
 
-    Ok((table, following_tokens))
+    Ok((table, following_tokens, item_sets))
 }
--- a/cli/src/generate/build_tables/item.rs
+++ b/cli/src/generate/build_tables/item.rs
@@ -1,5 +1,8 @@
-use crate::generate::grammars::{LexicalGrammar, Production, ProductionStep, SyntaxGrammar};
-use crate::generate::rules::{Associativity, Symbol, SymbolType, TokenSet};
+use crate::generate::grammars::{
+    LexicalGrammar, Production, ProductionStep, SyntaxGrammar,
+};
+use crate::generate::rules::Associativity;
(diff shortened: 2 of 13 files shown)


maxbrunsfeld:
To help debug where the parse states originate from, I added a `--report-states` flag to `tree-sitter generate` that outputs something like this:

```sh
$ tree-sitter generate --report-states
```

```
binary_expression              	175
member_expression              	161
call_expression                	132
subscript_expression           	128
ternary_expression             	125
update_expression              	124
for_of_statement               	101
object                         	83
for_in_statement               	74
assignment_expression          	68
_expression                    	65
arrow_function                 	54
jsx_self_closing_element       	50
augmented_assignment_expression	48
function                       	41
formal_parameters              	40
array                          	38
class                          	37
jsx_opening_element            	36
method_definition              	32
sequence_expression            	32
_property_name                 	30
assignment_pattern             	29
jsx_closing_element            	28
jsx_fragment                   	27
new_expression                 	27
object_repeat1                 	24
string                         	24
statement_block                	23
generator_function             	21
anonymous_class                	20

...
```

And then then there's a second flag that allows you to view each state's parse item set, and the symbol sequence that would *lead* to that state:

```sh
$ tree-sitter generate --report-states-for-rule for_in_statement
```

which produces a bunch of output like this, showing the item sets with their lookahead context:

```
state index: 525
state id: 3668
symbol sequence: for ( let array in _expression
items:
call_expression → _expression • (12) template_string	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
call_expression → _expression • (12) arguments	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
member_expression → _expression • (14) . identifier@property_identifier	[*, ,, (, ), in, [, <, >, /, ., =, +=, -=, *=, /=, %=, ^=, &=, |=, >>=, >>>=, <<=, **=, ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
subscript_expression → _expression • (14 Right) [ _expressions ]	[*, ,, (, ), in, [, <, >, /, ., =, +=, -=, *=, /=, %=, ^=, &=, |=, >>=, >>>=, <<=, **=, ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
ternary_expression → _expression • (1 Right) ? _expression : _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (2 Left) || _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (2 Left) ^ _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (2 Left) | _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (3 Left) && _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (3 Left) & _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) in _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) < _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) > _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) <= _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) == _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) === _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) != _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) !== _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) >= _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (4 Left) instanceof _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (5 Left) + _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (5 Left) - _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) * _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) / _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) >> _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) >>> _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) << _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (6 Left) % _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
binary_expression → _expression • (7 Left) ** _expression	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
update_expression → _expression • (11 Left) ++	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
update_expression → _expression • (11 Left) --	[*, ,, (, ), in, [, <, >, /, ., ?, &&, ||, >>, >>>, <<, &, ^, |, +, -, %, **, <=, ==, ===, !=, !==, >=, instanceof, ++, --, `]
sequence_expression → _expression • (-1) , _expression	[)]
sequence_expression → _expression • (-1) , sequence_expression	[)]
for_in_statement → for ( let array@array_pattern in _expression • ) _statement	[export, {, import, var, let, const, if, switch, for, (, await, while, do, try, with, break, continue, debugger, return, throw, ;, yield, [, <, /, class, async, function, =>, new, +, -, !, ~, typeof, void, delete, ++, --, ", ', `, number, identifier, this, super, true, false, null, undefined, @, get, set, program]

```

This output helped eventually track down the bug fixed in https://github.com/tree-sitter/tree-sitter/pull/354.

I can see using this command that in a lot of languages, there are a lot of states that currently cannot be merged because of sub-optimal structures in the grammars.

maxbrunsfeld:
Ok, I think this is in good shape. I'm not 100% ready to commit to this ABI though, so I've added a new flag to the `generate` command called `--next-abi`. By default, Tree-sitter will continue to generate the parsers with the same ABI that it did before. It will only generate the new structure if you run `tree-sitter generate --next-abi`.

This way, I can stop having to rebase this PR against all of my changes, and I can experiment with the new ABI more easily. Prior to enabling the new ABI by default, I may still make some structural changes to optimize the parser file size further.

PR #336, opened 2019-05-21
Reorder objects in documentation for clarity

ccharles:
The [documentation lists][1] four types of objects, the C names for them, and a summary for each. Unfortunately, the ordering isn't consistient.

This pull request simply reorders the C names so they match up with the other two orderings for readability and clarity.

[1]: http://tree-sitter.github.io/tree-sitter/using-parsers#the-objects


maxbrunsfeld:
😁 thank you!

PR #342, opened 2019-05-27
Add type definitions

georgewfraser:
This is based on https://github.com/tree-sitter/node-tree-sitter/blob/master/tree-sitter.d.ts with slight modifications because web-tree-sitter has a slightly different interface than node-tree-sitter


georgewfraser:
Huh weird, I actually made this PR in the github UI, not sure why the indent ended up that way.

PR #343, opened 2019-05-28
Add criterion benchmarks

mkrupcale:
### Summary

The previous benchmark system was unreliable since it did not attempt to iterate on the unit under test. The criterion benchmark library, on the other hand, produces useful statistical results for the benchmarks and allows for reliable, statistically significant comparisons. However, the criterion benchmarks take significantly longer to run (~9 hours vs. ~9 seconds) and therefore are not used by default.

I should also add that this was motivated by some other changes I will submit soon. In particular, I wanted a way to reliably measure performance improvements. While this PR is not required for those changes, this was used in my testing of those changes and could be useful for later performance comparisons.

### Usage

To run the new benchmarks, use

`cargo bench --bench crit-benchmark`

To simply test that they all run, use

`cargo bench --bench crit-benchmark -- --test`

For other options, see the full documentation[1].

### Detailed changes

 * `.gitignore`: ignore CLI target output files produced by critereon
 * `Cargo.lock`: update lock file
 * `cli/Cargo.toml`: add criterion to benchmarks and dev dependencies
 * `cli/benches/crit-benchmark.rs`: this is based on the original `benchmarks.rs` but uses criterion and makes some functions for reading the input
 * `script/benchmark`: specify which benchmark to run and default to the original benchmark
 * `script/benchmark.cmd`: likewise

### References

[1] [Command-Line Options - Criterion.rs Documentation](https://bheisler.github.io/criterion.rs/book/user_guide/command_line_options.html)


maxbrunsfeld:
Very cool, thanks for diving into this.

> the criterion benchmarks take significantly longer to run (~9 hours vs. ~9 seconds)

That is indeed significant. If possible, I'd prefer it if we could find some some kind of middle ground. Can we make the benchmarks *more* reliable than the existing ones, but still have them complete within 10 or 15 minutes? How many repetitions is Criterion doing here, and can we have it do far fewer?

In Tree-sitter's benchmarking suite, there's already a lot of redundancy between the examples. We have hundreds of (language, source file) pairs that all exercise mostly the same code paths. That may mean that we need fewer repetitions than we would need if each `.bench` call was measuring separate functionality.

If you think it's feasible, I am open to changing the existing benchmarking suite to just make it more rigorous. For context, the existing benchmark suite serves the following purposes:
* When run locally, it is accurate enough to detect the effects of *some* bigger optimizations or regressions.
* It does catch regressions on CI. Although there's significant noise and I don't carefully track *small* variations in the output, some bugs (especially in the error-recovery algorithm) can cause very large slowdowns that wouldn't be caught by the normal test suite.
* When a major slowdown occurs, it's easy to spot which language and source file triggered the slowdown most strongly.

> I should also add that this was motivated by some other changes I will submit soon. In particular, I wanted a way to reliably measure performance improvements.

That's awesome! I'm excited to see what changes you have come up with.

mkrupcale:
> Can we make the benchmarks more reliable than the existing ones, but still have them complete within 10 or 15 minutes? How many repetitions is Criterion doing here, and can we have it do far fewer?

Yes, this should be possible. The number of iterations performed is determined based on the `sample_size` parameter, `N`, some multiplier, `d`, and the measurement time[1]. Currently, `N=100`, and for the longer benchmarks (i.e. `> 1 ms / iteration`), `d` is set to the minimum of `d=1`, resulting in `\sum_{i=1}^N i = N*(N+1)/2 = 5050` iterations at a minimum. If this minimum estimated total time is within the configured measurement time, then `d` will be increased to maximize the number of iterations within the measurement time, which by default is `5 s` on top of the `3 s` warmup time.

There are currently 420 benchmarks, so if we set our target at 10 minutes, that leaves `~1.5 s` per benchmark on average. Thus, we should set our warmup time and measurement time to e.g. `500 ms` and `1 s`, respectively. However, this is insufficient, because the total time is currently dominated by the longer benchmarks, since there are many benchmarks `> 1 ms / iteration` (some as large as `200 ms / iteration`) and `n_iter >= N*(N+1)/2`. Thus, we also have to reduce the sample size. To go from ~9 hours to ~10 minutes then requires about a factor of 50 reduction in iterations, at least for the longest benchmarks. So we can set `N=10` or so, which should put us in range.

> In Tree-sitter's benchmarking suite, there's already a lot of redundancy between the examples. We have hundreds of (language, source file) pairs that all exercise mostly the same code paths. That may mean that we need fewer repetitions than we would need if each .bench call was measuring separate functionality.

Yeah, essentially the benchmarks are parameterized on both the parser and the input to said parser. If you want to benchmark each combination, you will have to reduce the iterations to get a reasonable total time, yes. But I wouldn't say that because they are exercising similar code paths that we "need" fewer repetitions other than from the pragmatic standpoint just mentioned. To get reliable comparisons, you should still try to do as many iterations as you can for each parameter.

On the other hand, it's possible that only some of these parameter combinations are worth optimizing or the parameter set is linearly dependent. In which case case, it might be better to just eliminate these unnecessary parameter combinations from the benchmark set. This minimal set might be somewhat tricky to properly determine, however. Although, with the above changes to the criterion configuration, I think we can have both reliable benchmark comparisons and "full" parameter space coverage.

> If you think it's feasible, I am open to changing the existing benchmarking suite to just make it more rigorous. For context, the existing benchmark suite serves the following purposes:

The alternative to making these changes to reduce criterion run time is to leave the PR as is, with the intention of only using criterion benchmarks when necessary for detecting small changes while continuing to use the existing benchmark with CI to detect large optimizations or regressions. That was kind of my thought process when submitting this PR. However, if we can achieve reasonable 10-15 minute runtimes and still achieve reliable benchmark comparisons, it might be worthwhile to use the criterion benchmarks in the CI.

One potential problem though is that since criterion benchmarks can have such high sensitivity, using it on the CI systems might falsely indicate either improvements or regressions[2]. Do you know if the CI systems are sufficiently stable that they could be used for benchmark comparisons? If not, it might be better to rely on developers running benchmarks on more stable systems and reporting the results.

I've tested the changes on my machine, and I can get the runtime down to ~15 minutes with criterion benchmarks by setting `N=8` and the reducing the number of resamples during analysis from `10^5` to `10^3`, so let me know if you think I should push these changes for this PR.

> That's awesome! I'm excited to see what changes you have come up with.

Don't get too excited lol. They are fairly minor, I would say, and also had an alternative motivation besides potential performance improvements.

[1] https://bheisler.github.io/criterion.rs/book/analysis.html#measurement
[2] https://bheisler.github.io/criterion.rs/book/analysis.html#comparison

maxbrunsfeld:
> The alternative to making these changes to reduce criterion run time is to leave the PR as is, with the intention of only using criterion benchmarks when necessary for detecting small changes

I just think that with a 9 hour run time, I don't think I'm likely to ever use it. I'd also like to add as little code as possible to the repository that *doesn't* get exercised by CI, because I think it's likely to bit-rot over time.

> I've tested the changes on my machine, and I can get the runtime down to ~15 minutes

That sounds promising. Let's shoot for *replacing* the current benchmark suite rather than adding a second one. What's the output look like? Does it report the average runtimes of individual files, like the current benchmark suite does? If not, is it possible to make it do that?

maxbrunsfeld:
> One potential problem though is that since criterion benchmarks can have such high sensitivity, using it on the CI systems might falsely indicate either improvements or regressions[2]. Do you know if the CI systems are sufficiently stable that they could be used for benchmark comparisons?

I think we should just set things up so that a benchmark regression does not fail the build (unless of course it causes the benchmarks to take longer than the CI provider's time limit).

mkrupcale:
> I just think that with a 9 hour run time, I don't think I'm likely to ever use it. I'd also like to add as little code as possible to the repository that doesn't get exercised by CI, because I think it's likely to bit-rot over time.

Yeah, makes sense.

> That sounds promising. Let's shoot for replacing the current benchmark suite rather than adding a second one.

Okay, I will update the PR with these changes.

> What's the output look like? Does it report the average runtimes of individual files, like the current benchmark suite does? If not, is it possible to make it do that?

If `gnuplot` is installed, criterion can generate HTML and graphical reports[1,2]. If not, the printed output includes the time and throughput confidence intervals and outlier information. If additionally the `--verbose` option is passed, slope, R^2, mean, standard deviation, median and median absolute deviation are also printed.

The criterion benchmarks are parameterized in the same way as the current benchmarks and thus output the above information for each (parser, input) combination. However, it does not currently report aggregate data for each language and overall the {mean,worst}*{normal,error} cases. It's probably possible to generate this aggregate data by post-processing the output, however.

> I think we should just set things up so that a benchmark regression does not fail the build (unless of course it causes the benchmarks to take longer than the CI provider's time limit).

Yeah, the benchmarks should not cause the build to fail, except for failure to build the benchmarks themselves or time constraints as you said.

[1] https://bheisler.github.io/criterion.rs/book/user_guide/html_report.html
[2] https://bheisler.github.io/criterion.rs/book/user_guide/plots_and_graphs.html

maxbrunsfeld:
Yesterday I had to change the benchmark file on master to get CI passing after we [restructured the typescript grammar](https://github.com/tree-sitter/tree-sitter-typescript/pull/68), so that caused some merge conflicts.

mkrupcale:
Yeah, no problem I was able to fix the merge conflicts, so we'll see how this goes in the CI.

mkrupcale:
Hey, have you been able to review the latest changes? I know the CI says that the Win x64 build failed, but it failed during the tests which I believe is unrelated to this PR and is some sort of spurious failure.

mkrupcale:
Is there still interest in this PR? Is there something which should be changed?

maxbrunsfeld:
A few things about this give me pause:

1. Right now, it still seems like the output is very verbose

2. This warning also worries me:

    ```
    Warning: Sample sizes < 10 will be disallowed in Criterion.rs 0.3.0
    ```

    In reality, our true sample size is already much much greater than 10, it's just that the samples are distributed over a large set of input files. There's a bit of a mismatch between Criterion's expectation and the desire to performance-test our function with a large set of inputs.

3. The CI build time is increased by over 10 minutes. The current benchmark script takes about 6 seconds to run. If we manually ran ever parse 10 times in a loop, it still probably wouldn't take more than 1 minute. Why does the Criterion version take so long? Is it because it performs a lengthy warm-up step for each individual sample?

Overall, I'm not yet sold that this new script is more practically useful than what we have currently. Do think it's possible (and worthwhile) to address these issues? Ultimately, is it more practical to just increase the repetition count in our existing benchmark approach? For the performance improvement that you were interested in making, would the improvement be impossible detect using that kind of benchmarking?

maxbrunsfeld:
In the meantime, I pushed a tiny change that adds a `-r` (for repetition count) flag to the benchmark script, so that you can parse each sample a configurable number of times. The default is 5. Which causes the script to take around 30 seconds to run on my laptop.

Not sure if this is helpful to you when detecting performance changes due to optimization work. If not, let me know what you think can be done to address some of the drawbacks of this use of Criterion.

mkrupcale:
>     1. Right now, it still seems like the output is very verbose

This can be decreased by removing `--verbose` option, although you will get less statistics shown.

>     2. This warning also worries me:
>        ```
>        Warning: Sample sizes < 10 will be disallowed in Criterion.rs 0.3.0
>        ```

Yeah, I never really intended to reduce it below 10, but unfortunately the benchmark runtime was closer to ~25-30 minutes with `N=10` than the current ~15-20 minutes.

>        In reality, our true sample size is already much much greater than 10

Recall from above that `N=10` implies that a minimum of `N*(N+1)/2=55` iterations are run.

>        it's just that the samples are distributed over a large set of input files. There's a bit of a mismatch between Criterion's expectation and the desire to performance-test our function with a large set of inputs.

This is not technically correct. If the benchmark time were independent of the variates (parser, input), then you could argue this, but the parse time is heavily dependent on both. See for example for fixed input [1] vs. [2] and for fixed parser [2] vs. [3].

In any case, even if you only wanted to perform one benchmark for each parser (i.e. the benchmark is only parameterized on (parser)), with the benchmark itself iterating over all possible inputs, this would have the essentially the same effect as running each (parser, input) benchmark separately. The effects of this would be

1. Less total warmup time due to fewer benchmarks, which currently account for ~20% of the benchmark runtime (this can be reduced further as desired).
2. Most (all?) benchmarks would be limited by the minimum iterations rather than measurement time due to longer benchmarks

In the case of Criterion benchmarking, point 2 would probably actually result in a higher total benchmark time if we were to use this definition of "benchmark".

>     3. The CI build time is increased by over 10 minutes. The current benchmark script takes about 6 seconds to run. If we manually ran ever parse 10 times in a loop, it still probably wouldn't take more than 1 minute. Why does the Criterion version take so long? Is it because it performs a lengthy warm-up step for each individual sample?

The actual time taken is `1223.43 s = 20.39 min`, with `2m 09s` of compiling, leaving `1094.43 s = 18.24 min` of actual benchmarking.

As indicated above, with `N=8`, there will be a minimum of `N*(N+1)/2=36` iterations for the longer benchmarks, while measurement time will be limited to `1 s` for the shorter benchmarks. The warmup time accounts for `0.5 s` (as currently specified) out of a minimum of `1.5 s` total time per benchmark. With 468 benchmarks currently, the minimum time currently expected then is 11.7 minutes. Note that these are minimums because longer benchmarks will be dominated not by the maximum measurement time but by the minimum number of iterations (see e.g. [4] for an extreme case). Thus, Criterion takes so long because it is simply doing more iterations and more work.

> Overall, I'm not yet sold that this new script is more practically useful than what we have currently.

Criterion is almost certainly the more technically correct way to do benchmarks and comparisons between benchmarks. Whether it is more practically useful depends on how you actually use the results, of course.

> Do think it's possible (and worthwhile) to address these issues?

Reducing the runtime further would involve reducing the warmup time and the measurement time. Due to the warning above about `N<10`, we really should increase the minimum number of iterations as well, though. Even so, this would probably still reduce the runtime since there are quite a few benchmarks limited by the measurement time rather than minimum iterations.

110/468 benchmarks are currently limited by minimum iterations, with total and average est. time of `351.158 s = 5.85 min` and `3.19234 s`, respectively. Increasing the minimum number of iterations from 36 to 55 in these cases should result in a total and average est. time of `536.491 s = 8.94 min` and `4.87719 s`, respectively. If we decrease the warmup time to 0.1 s and the measurement time to 0.5 s, we expect the total time then to be approximately

`468 * 0.1 s + (468-110) * 0.5 s + 536.491 s = 762.291 s = 12.70 min`

which is down 30% from 18.24 minutes.

Yes, this is still about an order of magnitude more time benchmarking than the simple iterated benchmark, but this is expected because Criterion is doing more than an order of magnitude more benchmarking work.

> Ultimately, is it more practical to just increase the repetition count in our existing benchmark approach?

I don't believe so. Criterion benchmarks are simply the more correct way to do benchmarking, and I believe that runtimes can be reduced sufficiently while still providing more useful results.

> For the performance improvement that you were interested in making, would the improvement be impossible detect using that kind of benchmarking?

I don't know--I would need to test that, but it would not technically be possible without some sort of confidence interval or uncertainty measurement to detect statistically significant differences.

> In the meantime, I pushed a tiny change that adds a `-r` (for repetition count) flag to the benchmark script, so that you can parse each sample a configurable number of times. The default is 5. Which causes the script to take around 30 seconds to run on my laptop.

I will try to reduce the runtime of these Criterion benchmarks further. The current PR as is targets the 10-15 minutes you suggested previously, although it falls on the higher end of that (or above, depending on hardware).

> Not sure if this is helpful to you when detecting performance changes due to optimization work. If not, let me know what you think can be done to address some of the drawbacks of this use of Criterion.

As stated above, I'm not sure whether or not the changes would be perceptible or even technically possible to compare without confidence intervals or uncertainties.

[1] https://travis-ci.org/tree-sitter/tree-sitter/jobs/556286288#L1678
[2] https://travis-ci.org/tree-sitter/tree-sitter/jobs/556286288#L3615
[3] https://travis-ci.org/tree-sitter/tree-sitter/jobs/556286288#L3637
[4] https://travis-ci.org/tree-sitter/tree-sitter/jobs/556286288#L3725

maxbrunsfeld:
> Criterion is almost certainly the more technically correct way to do benchmarks and comparisons between benchmarks. Whether it is more practically useful depends on how you actually use the results, of course.

To be clear, practical utility is **the** relevant concern. Period.

I outlined my current practical uses for this script [above](https://github.com/tree-sitter/tree-sitter/pull/343#issuecomment-496653850), and I am open to expanding its uses further as well, if we can do so while still serving the existing purposes.

The phrase "technically correct" is completely vague and just does not apply here. Correctness is how closely a program's actual behavior matches its intended/documented/expected behavior. The script we're talking about is a small dev tool that's mainly run by me. It is correct. The small sample size is not a *bug*. A lot of performance improvements and regressions are very clearly detectable with the current script, which runs in less than one minute and produces couple of screens' worth of very readable output.

What you're doing is changing the script's intended behavior to something entirely different. Now it uses a large warm up time and much larger sample size, its runtime is on a different time scale, and its output is structured very differently. Yes, this has *some* advantage - in theory, you can detect the effects of smaller optimizations this way. It also has some major disadvantages.

If the new behavior makes it possible for you to contribute optimizations to Tree-sitter that you absolutely *couldn't* have contributed before, then that *is* compelling, and I may want to incorporate that into my workflow. But right now I don't have a concrete example of that to consider, so I'm more concerned with keeping the script *useful*. To me.

mkrupcale:
> To be clear, practical utility is **the** relevant concern. Period.

I was not making a comment on whether or not practical utility is of the highest importance. Rather, I was stating that how one _defines_ practically useful places constraints on how one should perform benchmarking analysis. Similarly, how one is able to utilize the data can determine its practical utility. In particular:

 * If you are only interested in measuring and recognizing "large" performance changes and are confident in your ability to do so using the mean of a few iterations of said benchmarks, then the current benchmarking method can be of practical utility.
 * If, however, you are interested in measuring and recognizing statistically significant changes, then a procedure such as that used by Criterion is necessary to be of practical utility.

Thus, your definition of practical utility is not the same as mine, and furthermore, your definition relies on heuristics and intuition only known to you, whereas the latter definition has a mathematical basis for usefulness.

> I outlined my current practical uses for this script [above](https://github.com/tree-sitter/tree-sitter/pull/343#issuecomment-496653850), and I am open to expanding its uses further as well, if we can do so while still serving the existing purposes.

All of these objectives can be accomplished with Criterion benchmarks.

> The phrase "technically correct" is completely vague and just does not apply here.

When I say "technically correct" here, I refer to the mathematical basis for comparison as outlined by the Criterion manual[1]. This is a relatively well-understood and accepted method for comparing statistical estimators[2,3]. Referring to the above distinction of utility, I can see how it would not apply "here" if _you_ do not see the need for statistically significant comparisons.

> Correctness is how closely a program's actual behavior matches its intended/documented/expected behavior. It is correct. The small sample size is not a _bug_.

Whether or not it is correct again depends on your definition of correct. If you have defined the intended behavior to be simply exactly as the code behaves, then obviously it is trivially correct and therefore not a bug. However, to do so is to make the assertion that your definition of correct and its corresponding desired behavior is the only relevant or important one. This is quite presumptuous and short-sighted, to say the least.

> The script we're talking about is a small dev tool that's mainly run by me.

I have no idea how you could know this, but this claim is something entirely different from the claim that the desired benchmark behavior is the one which you define.

> A lot of performance improvements and regressions are very clearly detectable with the current script, which runs in less than one minute and produces couple of screens' worth of very readable output.

It's nice that you are able to read the current benchmark results and deduce when there is a performance improvement using whatever heuristics and intuition you have, but other people might be interested in relying on statistically significant comparisons. This is why I initially created a separate benchmark executable which could be useful to other people in addition to you.

> What you're doing is changing the script's intended behavior to something entirely different.

This is a mischaracterization of the current PR. Every single benchmark run by the current benchmark is run by the current PR. The difference is how many iterations are run and how the results of these iterations are analyzed to produce a meaningful estimator.

> Now it uses a large warm up time and much larger sample size, its runtime is on a different time scale, and its output is structured very differently.

Frankly, it's disappointing that your tune seems to have changed since the original PR regarding the practical utility of this benchmarking method. The current PR meets all requirements you outlined originally, but you seem to have changed your mind with respect to what is an acceptable amount of runtime as well as added some requirements which have to do with your own definition of utility.

I'm going to continue working on this, but it seems to me that if this PR is going to be merged, the current benchmarking system must stay, and this must be a parallel benchmarking system, despite your [previous](https://github.com/tree-sitter/tree-sitter/pull/343#issuecomment-497780764) desire to replace the existing one.

[1] https://bheisler.github.io/criterion.rs/book/analysis.html#comparison
[2] https://en.wikipedia.org/wiki/Bootstrapping_(statistics)#Bootstrap_hypothesis_testing
[3] https://en.wikipedia.org/wiki/Welch%27s_t-test

mkrupcale:
Please see the updated PR which reduces the benchmark runtime by ~28% from `1094.43 s = 18.24 min` to `786.42 s = 13.107 min`. This results in an increase of Travis CI effective runtime relative to the current benchmarking system by ~34% from ~26 minutes to ~35 minutes and "total" time by ~90% from ~32 minutes to ~61 minutes.

I don't know whether these are acceptable times for you, but personally I think the best solution here is closer to my original PR where the Criterion benchmarks are separate from your benchmarks. Then you can:

 * Use either Criterion or your benchmarks for CI
 * Use the Criterion benchmarks for rigorous benchmark comparisons
 * Use your benchmarks for (your and other) quick benchmark comparison purposes

In order to avoid the duplication of benchmarking code, I can extract the portions of the benchmarks which are common to both (i.e. reading the parser inputs) and share them between the two benchmark executables. This should ensure that regardless of whichever benchmarks are used in the CI, both benchmarks should continue to work.

maxbrunsfeld:
Sorry, I'm having too hard a time working with you on this for it to be worth it to me.

I don't want to adopt the code as it stands, and based on the way this discussion has gone thus far, I don't think it's worth going through the back-and-forth that would be required to get it into an acceptable state.

Good, general things to keep in mind when contributing to open source projects in general (sorry if some if this incredibly obvious):

* Writing code is fun. *Maintaining* code is a job. Everything added to a project makes the project more time-consuming to maintain.
* When you submit an additive pull request to an open-source project (that is, any change that is not a bug-fix), **you are requesting that the maintainers of that project please maintain that code for you, indefinitely**. It's a big ask, regardless of how great you think your code is. 
* When beginning to contribute to a new open-source project, a common successful approach is to start by making non-controversial changes such as bug-fixes or optimizations, which benefit users of the project. This allows you to gradually gain familiarity with the project and earn the credibility to make larger decisions, like what tools *other developers* of the project should be using.

I appreciate the time that you put into this, and I do think Criterion is a nice idea in theory, but I don't want to waste any more of your time than I already have. 😕

Issue #341, opened 2019-05-25
[Language Request] Lua


Stanzilla:
I have no real experience with this, but would it be possible to craft Lua support from an existing project like oxyc/luaparse?

Arcanemagus:
You could potentially use it as a reference, but generally the language itself will have a better reference to use while building the grammar. If you mean could that be used _directly_, the answer is no, as even a grammar system that has the same style of parser would need some translation to be used as a Tree-sitter grammar.

If you were interested in creating a Lua grammar the documentation on getting started can be found here: [Creating Parsers](http://tree-sitter.github.io/tree-sitter/creating-parsers)

There also seem to be at least two other projects that started a Lua grammar, I would look there if you were simply looking for one to use:
* [Azganoth/tree-sitter-lua](https://github.com/Azganoth/tree-sitter-lua)
* [Reisz/tree-sitter-lua](https://github.com/Reisz/tree-sitter-lua)

maxbrunsfeld:
Yeah, maybe we should just add Azganoth/tree-sitter-lua to the list; that one looks quite complete.

maxbrunsfeld:
I added a link to [Azganoth's Lua grammar](https://github.com/Azganoth/tree-sitter-lua) in the docs.

PR #346, opened 2019-05-31
Add elm tree sitter

Razzeee:
Hey, 

I'm not sure if you allow third party tree-sitters in that list, but we might as well just talk about that :)

Especially as https://github.com/tree-sitter/tree-sitter/issues/341 seems to aim in the same direction.


maxbrunsfeld:
Just took a look at Tree-sitter-elm. I don't know Elm, but it looks really good. It's great to see that you're successfully parsing all of those elm packages. Nice work @Razzeee!

Razzeee:
Thank you, really appreciate it.

Issue #344, opened 2019-05-29
Release and publish to npm?


paulyoung:
There are changes since the last release that I'd like to depend on but it would be much easier if I could use the release binaries and npm. Thanks in advance 🙇

paulyoung:
It turns out `web-tree-sitter` was already published to npm as `0.15.3` a day ago. A corresponding 0.15.3 GitHub release would be much appreciated 🙏

paulyoung:
Actually, that version doesn't include the TypeScript definitions, so a 0.15.4 release would be great 👌

paulyoung:
FWIW, I'm seeing `Error: "abort({}). Build with -s ASSERTIONS=1 for more info."` when using the latest release of `tree-sitter.js`, so that might be worth including too.

maxbrunsfeld:
Sorry for the delay there. [Tree-sitter-cli 0.15.2](https://www.npmjs.com/package/tree-sitter-cli/v/0.15.2) is out.

paulyoung:
Thank you!

paulyoung:
Could we also get a version published to npm that includes the TypeScript definitions? Thanks again.

maxbrunsfeld:
Published `web-tree-sitter@0.15.4`.

paulyoung:
It appears that the TypeScript definitions are missing from that release too. I think perhaps the `.npmignore` needs to be updated.

maxbrunsfeld:
Oh, ha. Doing that now.

PR #347, opened 2019-05-31
Ensure TypeScript bindings are published to npm

paulyoung:
As mentioned in https://github.com/tree-sitter/tree-sitter/issues/344#issuecomment-497809222, TypeScript definitions were added in #342 but aren't being published to npm.

See https://unpkg.com/web-tree-sitter@0.15.4/tree-sitter-web.d.ts

I confirmed this worked locally via `npm pack`


paulyoung:
Beaten to it 😄

PR #351, opened 2019-06-05
Fix ALLOW_MEMORY_GROWTH setting

paulyoung:
Fixes `ERROR:root:ALLOW_MEMORY_GROWTH: No such file or directory ("ALLOW_MEMORY_GROWTH" was expected to be an input file, based on the commandline arguments provided)` according to https://emscripten.org/docs/optimizing/Optimizing-Code.html#memory-growth

I'm on macOS so that could be why I experienced this and CI didn't.

### DIFF
--- a/script/build-wasm
+++ b/script/build-wasm
@@ -84,7 +84,7 @@ mkdir -p target/scratch
 $emcc                                \
   -s WASM=1                          \
   -s TOTAL_MEMORY=33554432           \
-  -s ALLOW_MEMORY_GROWTH             \
+  -s ALLOW_MEMORY_GROWTH=1           \
   -s MAIN_MODULE=2                   \
   -s NO_FILESYSTEM=1                 \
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Huh, CI does run this on both linux and macOS, but maybe things changed in a later emscripten version? This change seems like an improvement anyway, since it better matches the docs. Thanks!

paulyoung:
It could be due to a different version of emscripten and an artifact of not using the docker image.

After fixing this I still wasn't able to build from source because of exactly [this issue](https://github.com/bakape/meguca/issues/767) but now I should be able to rely on docker instead.

If that works I can confirm if it works without this patch but agree that it's still a worthwhile change.

paulyoung:
Yep, appears to have worked when using the docker image so perhaps something changed in emscripten at some point.

PR #355, opened 2019-06-08
Add .firstChildForIndex and .firstNamedChildForIndex to wasm bindings

Razzeee:
This needs testing, but I'm not sure how to do that.

### DIFF
--- a/lib/binding_web/binding.c
+++ b/lib/binding_web/binding.c
@@ -315,6 +315,16 @@ uint32_t ts_node_named_child_count_wasm(const TSTree *tree) {
   return ts_node_named_child_count(node);
 }
 
+void ts_node_first_child_for_byte_wasm(const TSTree *tree, uint32_t index) {
+  TSNode node = unmarshal_node(tree);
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
Just curious - are you using these methods for anything? I don't think they're used anymore in Atom and I wasn't sure if it was still worth having them.

Razzeee:
No, I'm not using them. The only one that's missing, that I'm using is `.descendantsOfType`

Just thought somebody else might profit. 

I'm also not using the `.closest` method that's also missing from the wasm bindings, but maybe due to the fact, that these are not documented and it's not clear what it does?

maxbrunsfeld:
I actually don't think these methods need to exist any more. Thanks for diving into this code anyways.

Razzeee:
Would be cool, if you can drop the `closest` tests then, just in case somebody else stumbles over that?

maxbrunsfeld:
I'll implement `closest` at some point, so I think I'll keep those tests around.

PR #360, opened 2019-06-12
Create ts_tree_cursor_copy function

zajac:
It is impossible to implement this function in user-space using existing functions without breaking encapsulation of TSTreeCursor. 
Sometimes it is useful to save current position in tree and `ts_tree_cursor_new` will not do the trick because cursor created for some node will not be able to ascend to it's parent with `ts_tree_cursor_goto_parent`

### DIFF
--- a/lib/src/zipper.c
+++ b/lib/src/zipper.c
@@ -1,0 +1,146 @@
+//
+//  zipper.c
+//  jsitter
+//
+//  Created by Andrey Zaytsev on 12.06.19.
+//
+
+#include "tree_sitter/api.h"
+#include "./alloc.h"
+#include "./language.h"
+#include "./tree.h"
+
+typedef struct _TSZipper {
+    const struct _TSZipper *parent;
+    const TSSymbol *parent_alias_sequence;
+    Subtree subtree;
+    uint32_t byte_offset;
+    uint32_t child_index;
+    uint32_t structural_child_index;
+} _TSZipper;
+
+bool __is_visible(_TSZipper *zip) {
+    if (ts_subtree_visible(zip->subtree)) {
+        return true;
+    } else {
+        bool extra = ts_subtree_extra(zip->subtree);
+        if (!extra && zip->parent_alias_sequence) {
+            return zip->parent_alias_sequence[zip->structural_child_index];
+        } else {
+            return false;
+        }
+    }
+}
+
+bool __ts_zipper_right(const _TSZipper *zip, _TSZipper *result) {
+    if (!zip->parent) {
+        return false;
+    }
+    Subtree parent_subtree = zip->parent->subtree;
+    if (zip->child_index == parent_subtree.ptr->child_count - 1) {
+        return false;
+    }
+    const Subtree sibling = parent_subtree.ptr->children[zip->child_index + 1];
+    
+    bool extra = ts_subtree_extra(zip->subtree);
+    uint32_t structural_child_index;
+    if (!extra && zip->parent_alias_sequence) {
+        structural_child_index = zip->structural_child_index + 1;
+    } else {
+        structural_child_index = zip->structural_child_index;
+    }
+    
+    *result = (_TSZipper) {
+        .parent = zip->parent,
+        .parent_alias_sequence = zip->parent_alias_sequence,
+        .subtree = sibling,
+        .byte_offset = ts_subtree_size(zip->subtree).bytes + ts_subtree_padding(sibling).bytes,
+        .child_index = zip->child_index + 1,
+        .structural_child_index = structural_child_index
+    };
+    return true;
+}
+
+bool _ts_zipper_right(const _TSZipper *zip, _TSZipper *result) {
+    while (__ts_zipper_right(zip, result)) {
+        if (__is_visible(result)) {
+            return true;
+        }
+        zip = result;
+    }
+    return false;
+}
+
+bool _ts_zipper_down(const _TSZipper *zip, _TSZipper *result, TSLanguage *lang) {
+    if (ts_subtree_child_count(zip->subtree) == 0) {
+        return false;
+    } else {
+        Subtree child = zip->subtree.ptr->children[0];
+        const TSSymbol *alias_sequence = ts_language_alias_sequence(lang, zip->subtree.ptr->production_id);
+        *result = (_TSZipper) {
+            .parent = zip,
+            .parent_alias_sequence = alias_sequence,
+            .subtree = child,
+            .byte_offset = zip->byte_offset,
+            .child_index = 0,
+            .structural_child_index = 0
+        };
+        if (__is_visible(result)) {
+            return true;
+        } else {
+            uint32_t visible_child_count = ts_subtree_visible_child_count(child);
+            if (visible_child_count > 0) {
+                _TSZipper *z = (_TSZipper *)malloc(sizeof(_TSZipper));
+                *z = *result;
+                return _ts_zipper_down(z, result, lang);
+            } else {
+                return _ts_zipper_right(result, result);
+            }
+        }
+    }
+}
+
+void _ts_zipper_new(TSTree *tree, _TSZipper *result) {
+    *result = (_TSZipper) {
+        .parent = NULL,
+        .parent_alias_sequence = NULL,
+        .subtree = tree->root,
+        .byte_offset = 0,
+        .child_index = 0,
+        .structural_child_index = 0
+    };
+}
+
+void ts_zipper_new(TSTree *tree, TSZipper *result) {
+    return _ts_zipper_new(tree, (_TSZipper *)result);
+}
+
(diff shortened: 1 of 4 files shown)


zajac:
@maxbrunsfeld What do you think of making Cursor immutable by using a linked list instead of an array, making it more like Zippers https://en.wikipedia.org/wiki/Zipper_(data_structure) ?

zajac:
Sorry, the zippers thing was not meant to be in this pull-request. 
But anyway, here is the idea. Immutable cursor = zipper, which can be freely shared between threads, and possibly used for "mutations".
The missing piece is the ref-counter which should be stored inside zipper and used for releasing the memory.
Here is the clojure's implementation of zippers:
https://github.com/clojure/clojure/blob/master/src/clj/clojure/zip.clj

lib/src/zipper.c
```diff
@@ -0,0 +1,146 @@
...
+            .byte_offset = zip->byte_offset,
+            .child_index = 0,
+            .structural_child_index = 0
+        };
+        if (__is_visible(result)) {
+            return true;
+        } else {
+            uint32_t visible_child_count = ts_subtree_visible_child_count(child);
+            if (visible_child_count > 0) {
+                _TSZipper *z = (_TSZipper *)malloc(sizeof(_TSZipper));
```
zajac: this malloc should be something else

maxbrunsfeld:
I definitely think `ts_tree_cursor_copy` is a good feature, but I'm not sure about the zipper thing. The benefit of the current approach (representing the stack as a mutable array) is that it's simple and most tree operations require zero allocations. With a zipper, descending operations require memory allocation. You could mitigate that somewhat by using some kind of shared object pool, but then you're back to having mutability.

From the use cases that I've built so far, I don't think the TreeCursor itself needs to be thread safe. You can just create separate cursor instances for separate threads, since cursors are pretty cheap to instantiate.

@zajac What are your thoughts? Maybe if you describe your use case in more detail I could better understand how the limitations of the cursor are affecting you.

zajac:
I'm investigating the possibility to adopt trees-sitter for use in IDE context. In order to do that I need to be able to use if from JVM. 

Here we have options:

- Port it to Java. It requires too much work and probably some loss of performance.
- Wrap tree nodes with Java objects. Leads to significant memory pressure and redundancy.
- Somehow navigate through trees in native memory

The last option looks most promising to me. From this perspective TSTreeCursor looks promising to me. However if it is the only API we have it becomes hard to track where the particular Cursor positioned. You can not simply pass cursor as function arguments because it is sort of a one-shot thing. If you passed the cursor somewhere it is no longer accessible to you. To avoid that problem one have to do protective copy of cursor (which is not necessary cheap, because trees are arbitrary deep) and again it is not clear who's responsibility to do that. This approach composes poorly.

The solution would be to default to immutable cursors which could be freely shared not just between functions but also between threads. Writing functional code implies that functions don't have side-effects and can be run as many times as one wish and on an arbitrary threads. It is hard to say usually  at which point the cursor/tree has crossed a thread boundary (it might be stored in some closure).

So both TSNode and TSCursor APIs don't look suitable to meet these requirements. 
Right now I consider moving the zipper code to JVM, and consuming the ts_subtree_* api directly using JNI. That kind of solves the problem with allocations during descend because it will be performed by JVM which is super cheap. 
I really like the Subtree API and data structure more both than TSCursor and TSTree/TSNode: 
It is immutable and has ref-counted granular ownership policy, which implies that one could retain just a particular subtree without keeping the entire tree in memory. 

Most of the time IDE is busy with all kinds of search. It runs it on almost every keystroke and tries to denote every identifier to some thing (we call this process Resolve). For this process to be performant it has to be incremental (we should exploit the fact that not everything changed when user typed one latter). And that involves aggressive caching and this is where Subtree being immutable value shines. Consider function f(subtree). In order to cache it all we need is just a check if it is the same subtree which is basically just an identity equality.
But Subtree being essentially an immutable tree provides no ability to ascend. And this is where Zippers come in. Basically Zipper is just a same Subtree "hooked" by a particular node.
And that is not the only thing Zippers are useful for. Imagine if you want to modify a deep immutable data structure. To do that you just copy a path from the root the the node you are trying to modify and return a new root. Zippers helps to do exactly that by making this more efficient by delaying the process of the path-copying and also (which is more important) decouples process of search (for some place in a tree) and modification itself. It makes a Zipper a perfect abstraction for the Place in a Tree.

maxbrunsfeld:
> Wrap tree nodes with Java objects. Leads to significant memory pressure and redundancy.

This is what we do in Atom, except with JavaScript instead of Java. Basically, most of the time there are no JavaScript nodes in memory. When you want to access a node (`TSNode`), we lazily create a JavaScript wrapper for it. So accessors like `tree.rootNode`, `node.child(i)`, and `node.parent` all lazily create JavaScript node wrappers. This works well for most analysis, and allows the nodes to be treated as immutable value objects.

Then, in specific code paths that need to traverse a very *large* number of nodes, those functions will create a tree cursor, walk the tree with it, and delete it. This requires us to use a more stateful algorithm, but gives the best performance.

> Right now I consider moving the zipper code to JVM, and consuming the ts_subtree_* api directly using JNI. That kind of solves the problem with allocations during descend because it will be performed by JVM which is super cheap.

The `Subtree` isn't a public API, and is subject to all kinds of changes for optimization purposes. So that binding is probably going to be more work to maintain. The `TSNode` API is basically just a public wrapper around a `Subtree`. It seems like you could use `TSNode` in your zipper too.

zajac:
TSNode retains the entire Tree which makes it not suitable as the caching basis.

> This is what we do in Atom, except with JavaScript instead of Java.
As far as I understand Atom is mostly single-threaded due to JS limitations and doesn't have a problem with ts_tree_copy

Also, keep in mind the JNI costs. It is much bigger than JS interop relative to overall JVM performance.
Taking into account generation hypothesis, the wrappers produced (and cached) may easily outlive and be propagated to old generation which will have impact on performance.

At the moment, I've re-implemeted zippers in JVM and accessing Subtree with Unsafe which basically involves reading from the memory directly, breaking all the encapsulation. If somebody else is interested in JVM API maybe we could put the code into main repo and maintain these accessors as Subtree layout changes. Are you generally interested in effecient JVM API?

maxbrunsfeld:
> TSNode retains the entire Tree which makes it not suitable as the caching basis.

I think that if you're using parts of a Tree, retaining the entire tree is fine. It will probably be retained anyway for other purposes. It's fine to retain many different trees for the same buffer - they share most of the same internal structure. This is pretty much the same use case that the existing API is designed for.

> Taking into account generation hypothesis, the wrappers produced (and cached) may easily outlive and be propagated to old generation which will have impact on performance.

For this concern, it shouldn't matter whether you use `TSNode` or `Subtree`. The `TSNode` is just a pointer to a `Subtree`, plus three other numeric fields that describe the node's current position. It doesn't require any extra heap allocation.

>  Are you generally interested in efficient JVM API?

I am, but I don't think I want to include it in the main repo. I'd rather it be in its own repo like `py-tree-sitter` and `node-tree-sitter`.

maxbrunsfeld:
I don't know if this is relevant, but it's also totally ok to *split* a `TSNode` into its component parts, if it's easier to store them all in separate primitive fields. You could recombine the primitives into a `TSNode` struct just when calling the `ts_node_*` APIs. The node binding currently does that for efficiency, so that nodes can be represented as plain JS objects without having to hold pointers to C structs internally.

zajac:
Here it is
https://github.com/JetBrains/jsitter

maxbrunsfeld:
Nice! The API looks pretty nice and small.

Breaking the encapsulation is a bit scary to me, but ultimately if it works and it's fast, then that's the most important thing.

What do you want to do with this PR? I know you're not using the `TSTreeCursor` from Java anymore, but if you get rid of the zipper commits, I'll merge it in just in case anyone else wants that functionality. Otherwise, should we close it out?

zajac:
Sure thing, I'll make another one with just `ts_tree_cursor_copy` I'm closing this one.

Yes, breaking all of the encapsulation is what scares me too. But this way I eliminate all of the interop altogether making tree-walking blazingly fast (just as if I were using TSTreeCursor from C directly).

zajac:
Also, consider this: 
```
fun <T : NodeType> parser(nodeType: T): Parser<T>
```

I wish it would be possible to implement this. Namely, the ability to parse different node types. It is really useful if I just want to parse just an expression or type, or anything else. IDEA does have this functionality.

maxbrunsfeld:
> the ability to parse different node types

Yeah, I agree that would be a cool feature. It should be doable, but it doesn't fit in quite as naturally in a bottom-up parser as it would in a recursive-descent parser (which I think the JetBrains IDEs mostly use), so it might be a significant amount of work.

zajac:
Should I file an Issue about that?

PR #363, opened 2019-06-14
Ts tree cursor copy

zajac:
It is impossible to implement this function in user-space using existing functions without breaking encapsulation of TSTreeCursor.
Sometimes it is useful to save current position in tree and ts_tree_cursor_new will not do the trick because cursor created for some node will not be able to ascend to it's parent with ts_tree_cursor_goto_parent

### DIFF
--- a/lib/src/tree_cursor.c
+++ b/lib/src/tree_cursor.c
@@ -293,10 +293,10 @@ const char *ts_tree_cursor_current_field_name(const TSTreeCursor *_self) {
 }
 
 TSTreeCursor ts_tree_cursor_copy(const TSTreeCursor *_cursor) {
-    const TreeCursor *cursor = (const TreeCursor *)_cursor;
-    TSTreeCursor res = {NULL, NULL, {0, 0}};
-    TreeCursor *copy = (TreeCursor *)&res;
-    copy->tree = cursor->tree;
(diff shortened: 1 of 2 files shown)


zajac:
@maxbrunsfeld  Here it is

lib/src/tree_cursor.c
```diff
@@ -291,3 +291,12 @@ const char *ts_tree_cursor_current_field_name(const TSTreeCursor *_self) {
...
   }
 }
+
+TSTreeCursor ts_tree_cursor_copy(const TSTreeCursor *_cursor) {
+    const TreeCursor *cursor = (const TreeCursor *)_cursor;
+    TSTreeCursor res = {NULL, NULL, {0, 0}};
+    TreeCursor *copy = (TreeCursor *)&res;
+    copy->tree = cursor->tree;
+    array_push_all(&copy->stack, &cursor->stack);
+    return res;
```
maxbrunsfeld: It looks like this has a different indentation than the other functions. Could you adjust it to match?

zajac: BTW shouldn't we increment ref-count of the tree here?

zajac: The cursor doesn't seem to retain it at all. Probably that's fine

PR #368, opened 2019-06-18
Wasm file generation note in readme

cancerberoSgx:
* adds a section in readme on how to generate .wasm language file since they are mentioned but not explained how to obtain. 
 * adds working example showing how to loading .asm file in node.js or browser.

### DIFF
--- a/lib/binding_web/README.md
+++ b/lib/binding_web/README.md
@@ -106,36 +106,39 @@ const tree = parser.parse((index, position) => {
 });
 ```
 
-### .wasm language files
+### Generate .wasm language files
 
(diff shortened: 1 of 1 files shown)


cancerberoSgx:
Perhaps is better to include this section in the docs (`docs/section-3-creating-parsers.md`) and reference that ? Also I would mention `tree-sitter web-ui` there, what do you think ?

maxbrunsfeld:
Yeah, it would be a good idea to mention `tree-sitter web-ui` in the Creating Parsers section, as an additional way of testing parsers. If you want to add that to this PR, I would be ok with that. I'm also ok with merging this as is. Nice work!

PR #375, opened 2019-06-29
Use Z option for docker-run volume mount

mkrupcale:
### Summary

This is necessary on systems with SELinux to prevent file permission/access errors when building WASM output.

I made these changes so that I could locally debug (on Fedora 30) some Travis-CI build failures[1] during `script/test-wasm`, and I am able to locally reproduce them with these changes.

### Detailed changes

 * cli/src/wasm.rs: Use Z volume mount option for docker-run
 * script/build-wasm: Likewise.

[1] https://travis-ci.org/tree-sitter/tree-sitter/jobs/552175504#L1452

### DIFF
--- a/cli/src/wasm.rs
+++ b/cli/src/wasm.rs
@@ -33,12 +33,12 @@ pub fn compile_language_to_wasm(language_dir: &Path, force_docker: bool) -> Resu
         let mut volume_string;
         if let (Some(parent), Some(filename)) = (language_dir.parent(), language_dir.file_name()) {
             volume_string = OsString::from(parent);
-            volume_string.push(":/src");
+            volume_string.push(":/src:Z");
             command.arg("--workdir");
             command.arg(&Path::new("/src").join(filename));
         } else {
             volume_string = OsString::from(language_dir);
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Can you explain how the behavior with `:Z` differs from the default, and from `:z`? For example, why not use `:z` instead?

The [Docker docs](https://docs.docker.com/storage/bind-mounts/) don't explain these in much detail, but they do say

> Use extreme caution with these options

mkrupcale:
`z` and `Z` options are used for SELinux re-labeling inside the container volume. `z` labels items such that they can be shared between containers, whereas `Z` creates private, unshared labels specific to the current container[1].

The warning is pertaining to the fact that it modifies the host file/directory SELinux label. In general, this could cause problems outside of the container, but since we are only talking about mounting the `tree-sitter` source directory, I don't think much harm should come of it.

Worst case, the user can force re-label it to the default: `restorecon -RF`.

[1] https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

maxbrunsfeld:
Thanks!

PR #381, opened 2019-07-08
Refine "sloppy brackets" handling

### LINKED ISSUE #380: CURLY_BRACE_REGEX mangles regex properties
(Side note: the exact intended supported flavor of regex should be made clearer.)

`/\p{XID_Start}/` is mangled into the regex `\p\{XID_Start\}`, which is an invalid regex ([and gives a very poor error message](https://github.com/rust-lang/regex/issues/594)).

The mangling at fault:

https://github.com/tree-sitter/tree-sitter/blob/d274e81d0df44334ec328958defa8631cefa4f35/cli/src/generate/prepare_grammar/expand_tokens.rs#L13-L16
https://github.com/tree-sitter/tree-sitter/blob/d274e81d0df44334ec328958defa8631cefa4f35/cli/src/generate/prepare_grammar/expand_tokens.rs#L48

Blame history:

- e2717a6ad14c6d1db056b55e89526b70eeb48a83: Preprocess regexes to allow non-standard escape sequences
Also allow unescaped curly braces to match literal curly braces when 
they don't form a valid repetition operator.
- b799b46f790ccfee6c3e77f98b9129c0d13a021e: Handle repetition range operators with commas in regexes
- 8c845f29e02763c94b2df0e22ff7536925dda66f: Allow hex characters in unicode code points

To be opinionated, I think that the regex-syntax library has the better behavior here. Definitely, trying to sneak by their filter will lead to unexpected results at best. If regex-syntax offered a "sloppy" mode that tried to parse correctly even in the face of what "strict" mode considers questionable, then I could see using that. But currently, you're changing the regex the user wrote to try to do guess what they meant based on an obviously imperfect regex heuristic.

CAD97:
Closes #380.

A better approach would be to write a "sloppy" parser to `regex_syntax::Ast` to use instead of the "strict" one provided, but for now, this adjusts the regex mangling to more precisely target brackets which would be rejected by the strict parser but accepted by the JS regex engine.

I've attempted to make the regex somewhat approachable, but this pushes the edge of regex's usability.

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -685,12 +685,15 @@ mod tests {
                     Rule::pattern(r#"\{[ab]{3}\}"#),
                     // Unicode codepoints
                     Rule::pattern(r#"\u{1000A}"#),
+                    // This isn't a repetition
+                    Rule::pattern(r#"{DEADBEEF}"#),
                 ],
                 separators: vec![],
                 examples: vec![
                     ("u{1234} ok", Some((0, "u{1234}"))),
                     ("{aba}}", Some((1, "{aba}"))),
                     ("\u{1000A}", Some((2, "\u{1000A}"))),
+                    ("{DEADBEEF}", Some((3, "{DEADBEEF}"))),
                 ],
             },
         ];
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -11,8 +11,16 @@ use regex_syntax::ast::{
 use std::i32;
 
 lazy_static! {
-    static ref CURLY_BRACE_REGEX: Regex =
-        Regex::new(r#"(^|[^\\])\{([^}]*[^0-9A-F,}][^}]*)\}"#).unwrap();
+    static ref CURLY_BRACE_REGEX: Regex = Regex::new(r#"(?x)
+        # captures: $1: lookbehind, $2, $4: close bracket; $3, $5: content
+        ( ^[^\\\{]?? | \\[^pPuUx] | [^\\][^\\] )               # lookbehind is not \p, \P, \u, \U, \x, \
+        \{                                                     # followed by a openning bracket
+        (?: $ | (\}) |                                         # followed by eoi or close bracket, or
+            ( (?:\\\}|[^}])*?(?:[^0-9,}]|\\\})(?:\\\}|[^}])*?  # content and
+              [^\\] ) (\}) |                                   # a non-escaped close bracket, or
+            ( (?:\\\}|[^}])*?(?:[^0-9,}]|\\\})(?:\\\}|[^}])*?)$# content and lacks a close bracket
+        )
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Can you add one or more unit tests for the regexes that are now handled properly, which were not handled before?

CAD97:
The problem is that while e.g. `\p{XID_Start}` will pass through to the later stages, it's still rejected by `expand_regex`. I suppose I could test e.g. ` {10F}`, which would currently fail.

I'm on Windows 10 Home, so I can't use docker and the shell scripts don't help me, so I'm basically developing off of Travis for testing.

maxbrunsfeld:
Currently, this feels like too big a jump in complexity, given how much of a benefit we get. I think that a better fix to #380 would be a preliminary step that detects the use of unicode property syntax `\p{BLA}` and returns an error saying "Tree-sitter currently does not support unicode property syntax in regexes", or something like that. Thoughts?

CAD97:
The main problem is that this kind of pre-processing [is inherently suspect](https://twitter.com/burntsushi5/status/1148187796609425409). It's super complicated to do this correctly.

If we give up on covering some edge cases that currently aren't covered (e.g. unpaired brackets, more than one pair of brackets, brackets where only some are escaped), we can reduce the regex to just

```regex
(?x)
( ^[^\\\{]??   # lookbehind is the start anchor
| \\[^pPuUx]   # or a non parameterized escape
| [^\\][^\\] ) # or not an escape at all
\{             # open bracket
( [^}]*?       # content of the brackets
  [^0-9,}]     # content not a repetition
  [^}]*?       # content of the brackets
|)             # or empty bracket contents
\}             # close bracket
```

which would be used simply as `CURLY_BRACE_REGEX.replace(content, r"$1\{$2\}")`, the same as today. (Using `replace_all` instead of `replace` is an easy way, even with today's regex, to support `{}{}` which isn't today. But it still doesn't support nested brackets (e.g. `{{}}`) since `replace_all` is a non-overlapping check.)

The loop as it exists currently could also probably be simplified to `replace_all(content, r"$1\{$3$5$2$4")`, if I'm reading the `Regex::replace` docs correctly, and we don't care about nested brackets. (I over-complicated that bit a lot.)

Nested brackets could be solved simply by repeating `replace_all` until it makes no changes. If the replacement complexity is the entire issue, I think I can even fix `{10F}` and `}` to behave as "expected" of a "lossy" parse by tweaking the regex.

The complexity of trying to do this somewhat correctly with a regex should suggest that [this isn't a problem to be solved with regexes](https://stackoverflow.com/a/1732454/3019990), not to ignore the complexities of the problem (or worse, blame the user).

Alternatively, would you support a patch to parse the regex with tree-sitter-regex? That would then allow for more control over the actual parsing to behave in the "sloppy" mode expected. (Personally, I'd still prefer the "strict" mode, but I can see the reasoning behind acting more like JS when the DSL is pretending to be JS.) Having a "sloppy" frontend to `regex-syntax` is something that would be useful throughout the ecosystem as a whole for when the default strict behavior isn't desired.

CAD97:
Closing as superseded.

PR #382, opened 2019-07-08
Use regex Hir instead of regex Ast

CAD97:
This is mainly a proof-of-concept to move to using `regex_syntax::Ast` instead of `regex_syntax::Hir`. This enables all of the "fancy" regex features provided by the regex crate "for free", as they get translated into the simpler Hir language.

This includes #381 because it's most useful with it.

A lot of regexes will suddenly accept a _lot_ more things because of unicode-aware translation, so I suspect this will very unacceptably regress performance until `CharacterSet` uses a strategy similar to [`ClassUnicode`](https://docs.rs/regex-syntax/0.6.8/regex_syntax/hir/struct.ClassUnicode.html) wherein ranges are stored rather than every single character. Especially since, uh, this translates negated sets into positive ones.

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -74,6 +74,12 @@ impl From<regex_syntax::ast::Error> for Error {
     }
 }
 
+impl From<regex_syntax::Error> for Error {
+    fn from(error: regex_syntax::Error) -> Self {
+        Error::new(error.to_string())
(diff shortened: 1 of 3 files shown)


CAD97:
(closed to reopen properly marked as a draft)

PR #383, opened 2019-07-08
Use regex Hir instead of regex Ast

CAD97:
This is mainly a proof-of-concept to move to using regex_syntax::Ast instead of regex_syntax::Hir. This enables all of the "fancy" regex features provided by the regex crate "for free", as they get translated into the simpler Hir language.

This includes #381 because it's most useful with it.

A lot of regexes will suddenly accept a lot more things because of unicode-aware translation, so I suspect this will very unacceptably regress performance until CharacterSet uses a strategy similar to ClassUnicode wherein ranges are stored rather than every single character. Especially since, uh, this translates negated sets into positive ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -727,6 +727,7 @@ dependencies = [
  "tiny_http 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "tree-sitter 0.3.10",
  "tree-sitter-highlight 0.1.6",
+ "unic-char-range 0.9.0 (git+https://github.com/open-i18n/rust-unic/?branch=CAD97-patch-1)",
  "webbrowser 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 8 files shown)


CAD97:
You might want to cancel these builds as I posit they'll clog your CI for a good while 😅

CAD97:
I did a decent amount of work towards using ranges in `CharacterSet` the whole way through. I don't know how the NFA states are constructed, though, so I got stuck trying to track down the reason for [this failure](https://www.diffchecker.com/bUfE1Ksj) in `generate::nfa::tests::test_group_conditions`.

maxbrunsfeld:
@CAD97 Do you think you'll be working on this one further? I don't think I have time to push this forward, so if you feel like you're stuck, I might close this out for now.

CAD97:
I'm slowly experimenting with pushing the char-range-ification from the other direction, so I'll close this for now.

PR #405, opened 2019-07-31
Mention differing alias behavior based on argument types in the docs 

simonrepp:
Also fixes a broken link I discovered by chance.

Thanks for the explanations so I could even write this. :) :+1:


maxbrunsfeld:
Thanks!

PR #406, opened 2019-07-31
Instruct parcel to skip analysis of fs calls

chrismwendt:
Prior to this change, https://github.com/parcel-bundler/parcel would throw errors saying it can't statically analyze some calls to `fs.readFileSync`.

I noticed all of the calls are guarded by environment checks so that the calls are not made in the browser environment. Parcel has a feature to skip analysis by module, but the original `package.json` needs to be modified. Found at https://github.com/parcel-bundler/parcel/issues/496#issuecomment-366993459

After this change, those `fs` calls are ignored by Parcel, which enables bundling to complete successfully.

### DIFF
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -25,6 +25,9 @@
     "url": "https://github.com/tree-sitter/tree-sitter/issues"
   },
   "homepage": "https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_web",
+  "browser": {
+    "fs": false
+  },
   "devDependencies": {
     "chai": "^4.2.0",
     "mocha": "^6.1.4",


maxbrunsfeld:
Does the module then run and work properly when used via parcel? Just checking in case there are any other parcel-related changes we'd want to piggyback onto this.

chrismwendt:
Parcel succeeds (meaning that bundling works), but I haven't actually tested the resulting JS output. I won't be working on this for the foreseeable future, but I doubt there will be more parcel-related changes.

maxbrunsfeld:
I'm curious how the bundling will work with the separate `.js` and `.wasm` files though. Do we need to switch to fetching the `.wasm` file with `await import` instead of `fetch`? Looking at Parcel's WASM docs [here](https://parceljs.org/webAssembly.html).

wingrunr21:
I just finished looking at this as I couldn't get parcel or webpack to correctly bundle the separate `.wasm` file for a node target. Apparently `emscripten`'s default output of "can run anywhere" messes with the static analysis of these bundlers and they can't determine how or when to bundle the wasm.

Potentially relevant: emscripten-core/emscripten#6565 and emscripten-core/emscripten#6542

maxbrunsfeld:
I think that you may need to have parcel treat the .wasm file similarly to how it would treat an image or a font - a file that remains separate, but gets copied into the output folder alongside the JS, while possibly updating the URL that’s used to fetch it in the JS.

wingrunr21:
Yes, manually copying the wasm file solves the problem (although Parcel has another issue as it doesn't support `__dirname` rewriting which means you have to be very specific in how you place the location of the grammar wasm). But, supposedly both Parcel and Webpack can automatically figure this out if you specify either node or web in the emscripten config.

maxbrunsfeld:
Hmm, it seems like Parcel really *should* be able to work with Emscripten's default output. We want the WASM output to work with both Node and the browser. I'd think that any project that uses Emscripten and Parcel would have this issue.

maxbrunsfeld:
I'm going to close this one out for the time being, as it doesn't seem like it's sufficient to get the module working under parcel. Thanks for taking a look @chrismwendt, and feel free to reopen if you do get a chance to work on this.

I too would like to see it work with Parcel; I might take a look myself at some point.

PR #415, opened 2019-08-08
Fix Rust bindings example

rfdonnelly:
`Parser::parse` returns `Option<&Tree>`.  An `unwrap()` or similar is required to get the actual `Tree`.


maxbrunsfeld:
👍  Thanks for catching that.

PR #422, opened 2019-08-12
build: add support for git submodules

mkrupcale:
The `.git` file can contain a reference to the actual git directory as is commonly the case for a submodule[1]. When this is the case, read the `.git` file to discover the actual git directory.

This fixes an issue where tree-sitter would always be rebuilt if it were used as a git submodule in a larger project due to non-existent `.git/*` files in `cargo:rerun-if-changed={}`.

 * `cli/build.rs`: Read the `.git` file to discover the git directory

[1] https://git-scm.com/docs/gitrepository-layout

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -25,7 +25,15 @@ fn read_git_sha() -> Option<String> {
         }
     }
 
-    let git_head_path = git_path.join("HEAD");
+    let git_dir_path;
+    if git_path.is_dir() {
+        git_dir_path = git_path;
+    } else if let Ok(git_path_content) = fs::read_to_string(&git_path) {
+        git_dir_path = repo_path.join(git_path_content.get("gitdir: ".len()..).unwrap().trim_end());
+    } else {
+        return None;
+    }
+    let git_head_path = git_dir_path.join("HEAD");
     if let Some(path) = git_head_path.to_str() {
         println!("cargo:rerun-if-changed={}", path);
     }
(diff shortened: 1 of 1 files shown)


cli/build.rs
```diff
@@ -25,7 +25,14 @@ fn read_git_sha() -> Option<String> {
         }
     }
 
-    let git_head_path = git_path.join("HEAD");
+    let git_dir_path;
+    if git_path.is_dir() {
+        git_dir_path = git_path;
+    } else {
+        git_dir_path = repo_path.join(fs::read_to_string(&git_path)
+                                      .unwrap().get(8..).unwrap().trim_end());
```
maxbrunsfeld: Just for self-documentation purposes, could you use `"gitdir: ".len()` instead of `8` here?

Also, instead of using `unwrap` here, could you make the function return `None` if the `.git` file is unreadable?

This functionality (putting the sha in the version number) is so unimportant that I'd like the build to succeed even if some occurs when getting the sha.

maxbrunsfeld:
Also, disregard the windows-specific CI failures. That was a mistake that I made on master.

cli/build.rs
mkrupcale: I've made both of these changes, although I'm definitely a Rust novice, so let me know if this is not the correct way of doing it.

maxbrunsfeld:
Thanks!

dineshkewat125318-ship-it:
> The `.git` file can contain a reference to the actual git directory as is commonly the case for a submodule[1]. When this is the case, read the `.git` file to discover the actual git directory.
> 
> This fixes an issue where tree-sitter would always be rebuilt if it were used as a git submodule in a larger project due to non-existent `.git/*` files in `cargo:rerun-if-changed={}`.
> 
> * `cli/build.rs`: Read the `.git` file to discover the git directory
> 
> [1] https://git-scm.com/docs/gitrepository-layout



> The `.git` file can contain a reference to the actual git directory as is commonly the case for a submodule[1]. When this is the case, read the `.git` file to discover the actual git directory.
> 
> This fixes an issue where tree-sitter would always be rebuilt if it were used as a git submodule in a larger project due to non-existent `.git/*` files in `cargo:rerun-if-changed={}`.
> 
> * `cli/build.rs`: Read the `.git` file to discover the git directory
> 
> [1] https://git-scm.com/docs/gitrepository-layout

PR #424, opened 2019-08-14
cli: loader: Move language loader from library into its own function

mkrupcale:
Loading the language from an already-compiled library could be useful on its own without having to go through (potential) source compilation.

I admit I don't completely understand how the `Loader` is used, but I thought that users might want to compile the library externally themselves rather than invoking the C/C++ compiler through TS CLI and having to provide the source code alongside it. I don't know if this is an issue in practice, but perhaps there could be ABI issues loading and using the library which is built on the host machine by TS CLI (or maybe the host machine doesn't have a C/C++ installed compiler at all) when TS itself was built on some other build machine with different configuration.

With this PR, instead the user would be able to compiler both TS and the library produced by compiling the TS CLI parser/scanner on the same build machine with the same configuration/compiler and then deploy it to whatever host machine on which they wish to run it. Furthermore, this gives the user some flexibility in the actual naming of the library file (although the symbols inside must still have suffix `_name`).

 * `cli/src/loader.rs`: Move the compiled library language loading in `load_language_from_source` into its own function

### DIFF
--- a/cli/src/loader.rs
+++ b/cli/src/loader.rs
@@ -322,20 +322,7 @@ impl Loader {
             }
         }
 
-        let library = Library::new(&library_path).map_err(Error::wrap(|| {
-            format!("Error opening dynamic library {:?}", &library_path)
-        }))?;
-        let language_fn_name = format!("tree_sitter_{}", replace_dashes_with_underscores(name));
-        let language = unsafe {
-            let language_fn: Symbol<unsafe extern "C" fn() -> Language> = library
-                .get(language_fn_name.as_bytes())
-                .map_err(Error::wrap(|| {
-                    format!("Failed to load symbol {}", language_fn_name)
-                }))?;
-            language_fn()
-        };
-        mem::forget(library);
-        Ok(language)
+        load_language_from_library(&name, &library_path)
     }
 
     fn find_language_configurations_at_path<'a>(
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Right now, the `cli` module isn't really meant to be consumed as a library; only as a command-line tool. So currently, users won't have a way of accessing the extracted functionality. Is there a specific CLI command that you have in mind in which `load_language_from_library` is called outside of the context of `load_language_from_sources`?

mkrupcale:
Yeah, it's possible I misunderstood how the `Loader` is used. On closer inspection, if it is only really used by CLI for testing (i.e. `tree-sitter test`), benchmarks (i.e. `cli/benches/benchmark.rs`), highlighting (i.e. `tree-sitter highlight`), parsing (i.e. `tree-sitter parse`), and dumping languages (i.e. `tree-sitter dump-languages`), then this may not be terribly useful externally from CLI or even from the `load_language_from_sources` function.

On the other hand, I'm not sure it hurts to encapsulate that functionality in its own function for clarity. My concerns outlined above about ABI/lack of compiler on the host might still be relevant as well, depending on how the TS CLI is actually used (e.g. `tree-sitter highlight` or `tree-sitter parse`).

maxbrunsfeld:
Closing this one out for now, since it's not clear what public-facing change this is service of. Thanks anyway for exploring it.

PR #426, opened 2019-08-15
binding: Make directory structure hierarchical

mkrupcale:
The current directory structure uses a prefix "binding_" in the directory name to indicate that its contents are bindings for a given language. A hierarchical directory structure removes this redundancy by placing all language bindings in a shared binding directory.


maxbrunsfeld:
I think I slightly prefer the less-deeply-nested structure. More importantly though, since it doesn't really matter, I don't think the change is worth the churn that it causes. Thanks anyway though.

PR #428, opened 2019-08-16
Parsers: Add C parsers

bryphe:


bryphe:
Apologies for the noise! Meant to include this in a fork.

PR #430, opened 2019-08-19
Ignore hidden files in grammar test directories

rfdonnelly:
This fixes the "stream did not contain valid UTF-8" error due to `tree-sitter test` attempting to parse Vim's hidden binary swap files.

This change avoids the error by ignoring hidden files in grammar test directories.

A separate problem that this PR does not address is the vagueness of the error message.  It would be helpful if the error message included the problematic file path.

## Example Error

```text
$ tree-sitter test

stream did not contain valid UTF-8
```


maxbrunsfeld:
Wonderful, thanks for this fix!

PR #435, opened 2019-08-24
Add some missing APIs in Rust binding

ubolonton:


maxbrunsfeld:
This all looks great. Thank you!

PR #436, opened 2019-08-26
lib: remove utf8proc dependency

mkrupcale:
### Introduction

This removes the only external library dependency, which is on `utf8proc` for UTF-8 decoding. It does so by implementing its own UTF-8 decoder. The new decoder is both faster and has a simpler API. It also re-writes the UTF-16 decoder to follow this API and have a clearer implementation.

### Motivation

The original motivation of this work was to remove the dependency on `utf8proc` since all that was actually needed was the UTF-8 decoder, whereas `utf8proc` is a more comprehensive library for dealing with UTF-8 and Unicode data. Additionally, this would remove the only external runtime library dependency for tree-sitter. This led to me trying to find the best possible implementation of a UTF-8 decoder that I could. Thus, I made a project[1] for testing and benchmarking several UTF-8 decoder implementations, based on some previous work by Bob Steagall[2].

#### UTF-8 decoders

You can read the detailed performance comparison of the different UTF-8 decoders I tested[3]. In summary the ICU `U8_NEXT` implementation performed quite well for the `decode_next` implementation necessary for Tree Sitter. Furthermore, the `utf8proc` implementation is actually one of the slowest, being about 2-5 times slower than the fastest implementations, depending on input and compiler. Thus, I implemented an expanded form of this C macro as a function following the `decode_next` API and broke it into a few smaller functions for TS.

#### tree-sitter

I then benchmarked using Criterion[4] the TS library with the `utf8proc` and the ICU-based UTF-8 decoders. This can be accomplished using a separate repo[5] which includes `tree-sitter` and uses the existing set of benchmarks but uses Criterion as the benchmarking framework.

On the whole, there is not a big performance difference, but there does appear to be a 1-2% improvement on average (across all parsers and inputs) when using the ICU-based decoder[6]. In particular, out of 455 benchmarks, there were 17 regressions, 265 unchanched or within the noise variations, and 173 improvements of statistical significance[7]. Full details with graphical comparisons are given in the complete Criterion report[8].

### Final comments

This PR removes the only runtime library dependency of TS, `utf8proc`, and replaces its UTF-8 decoder with a simpler API and a significantly faster implementation. While the overall performance improvement of TS is modest at 1-2% on average, it is fairly consistent and statistically significant.

#### References

[1] https://gitlab.com/mkrupcale/libutf8decoders
[2] https://github.com/BobSteagall/CppNow2018/tree/master/FastConversionFromUTF-8
[3] https://gitlab.com/mkrupcale/libutf8decoders-bench-results/blob/master/results/results.md
[4] https://bheisler.github.io/criterion.rs/book/criterion_rs.html
[5] https://gitlab.com/mkrupcale/tree-sitter-criterion
[6] [`critcmp.log`](https://mkrupcale.gitlab.io/tree-sitter-criterion-results/reports/master-105b0ba52ce64e85eb2d6f09f5fe3ad508adeafb-2019-08-24T14:56:14-04:00_utf8proc-remove-a8bb087ad75178b397266ba91a8b8fd46491cf30-2019-08-24T16:04:21-04:00/critcmp.log)
[7] [`criterion.log`](https://mkrupcale.gitlab.io/tree-sitter-criterion-results/reports/master-105b0ba52ce64e85eb2d6f09f5fe3ad508adeafb-2019-08-24T14:56:14-04:00_utf8proc-remove-a8bb087ad75178b397266ba91a8b8fd46491cf30-2019-08-24T16:04:21-04:00/criterion.log)
[8] [Criterion report](https://mkrupcale.gitlab.io/tree-sitter-criterion-results/reports/master-105b0ba52ce64e85eb2d6f09f5fe3ad508adeafb-2019-08-24T14:56:14-04:00_utf8proc-remove-a8bb087ad75178b397266ba91a8b8fd46491cf30-2019-08-24T16:04:21-04:00/report)


maxbrunsfeld:
I like it! I have a somewhat significant change I'd like to propose - I decided to do this in the form of a PR on top of your PR: 😮  https://github.com/mkrupcale/tree-sitter/pull/1. Let me know what you think.

maxbrunsfeld:
Thanks for this detailed investigation @mkrupcale. This will make the library simpler and easier for everyone to use.

PR #439, opened 2019-08-31
docs: add YAML/TOML parser

ikatyang:
- https://github.com/ikatyang/tree-sitter-yaml
- https://github.com/ikatyang/tree-sitter-toml

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -44,8 +44,10 @@ Parsers for these languages are fairly complete:
 * [Ruby](https://github.com/tree-sitter/tree-sitter-ruby)
 * [Rust](https://github.com/tree-sitter/tree-sitter-rust)
 * [SystemRDL](https://github.com/SystemRDL/tree-sitter-systemrdl)
+* [TOML](https://github.com/ikatyang/tree-sitter-toml)
 * [TypeScript](https://github.com/tree-sitter/tree-sitter-typescript)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Both of these look great; awesome work!

I noticed that you had to patch Tree-sitter to allow for `---` lines in the YAML tests. I'll try to fix that limitation sometime soon.

maxbrunsfeld:
Since both of these are such popular languages, I added them to the playground on the website.

maxbrunsfeld:
@ikatyang Ok, in https://github.com/tree-sitter/tree-sitter/commit/a8dedbee0d7ed8df0f359298577858703f6765c2, I *think* I've addressed the problem with `tree-sitter test` that you were having to patch.

Now, when parsing unit test files, in order to find the dividing line between input and expected output, Tree-sitter will look for the *longest* line of dashes, not the first line of dashes. So for YAML, you should be able to write the tests like this, if you want to:

```
===========================
Document start on last line
===========================
---
a: b
---

--------

(stream
  (document
    (block_node
      (block_mapping
        (block_mapping_pair
          key: (flow_node
            (plain_scalar))
          value: (flow_node
            (plain_scalar))))))
  (document))
```

ikatyang:
Thanks, it [works](https://github.com/ikatyang/tree-sitter-yaml/commit/3439d6f2edc5487f7c3ce08c6a177bb8de88ff04)!

PR #440, opened 2019-09-01
fix: allow lowercase unicode escape

ikatyang:
Similar to #319, lowercase unicode escapes should be allowed as well.

```
Error processing rule <rule_name>
Details:
  regex parse error:
    [\u{10000}-\u\{10ffff\}]
                 ^
error: invalid hexadecimal digit
```


maxbrunsfeld:
Thanks!

PR #442, opened 2019-09-01
feat(cli): support snapshot testing with `--update` flag

### LINKED ISSUE #828: Support a test-promote cycle
Hi! First of all thanks a million for `tree-sitter`. Super happy it exists and its a ton of fun to work with 🙏🏼  &mdash; I'm currently migrating my Erlang [Menhir parser](https://github.com/AbstractMachinesLab/caramel/blob/main/erlang/src/erl_parser.mly) to tree-sitter here: [tree-sitter-erlang](https://github.com/AbstractMachinesLab/tree-sitter-erlang).

I'd like to contribute a flag to the `tree-sitter test` subcommand, to promote corpuses with new expectations.

The reason is that as my grammars have been changing a lot, the update process has been rather painful.

In short, I've been removing all the expected s-expressions and copy-pasting the new ones, which can be pretty time consuming and because the test outputs are not pretty printed, means that most of my corpus slowly become less readable 🙈 

### The test-promote cycle

Coming from the OCaml world, where working with s-expressions is commonplace for testing too, the `dune` build tool has a similar testing cycle where you essentially diff expected vs generated sexp trees.

If they are off but you know they need to be updated, you _promote_ the new outcome. Your immediate next run of the same test will pass, since the expectations have been updated.

The cycle looks like the picture below. Here I have updated my Erlang menhir parser and the new output is expected, but I haven't updated it in my corpus yet.

Screenshot from 2020-12-03 14-16-47

---

I'd like to implement this but could use some mentoring 🤓, and of course if this isn't something that you'd like to have as a part of `tree-sitter`, I could look into building a tool _around_ it instead.

ikatyang:
This PR adds an `--update` flag to the `tree-sitter test` command, which adds the ability to replace the _expected_ output in the corpus.txt with the _actual_ output produced by the parser, that is, we can now simply use this `--update` flag to write all the corresponding parser output back to the corpus.txt, and we just need to check the output without typing its actual sexp.

- use the same output format as `tree-sitter parse`, except there won't be any position information printed.
- the corpus.txt won't be touched if there's no difference between the _expected_ output and the _actual_ output in that file.
- if there're differences between _expected_ and _actual_, _expected_ will be replaced by _actual_ and the whole file will be reformatted, i.e., all the output sexp will be formatted just like the output from `tree-sitter parse` and all the delimiters `===`/`---` will be normalized as 80-column long.
- this flag also works with `--filter` flag.

Fixes #828


ikatyang:
I've rewrote the whole logic, and this PR now only does the minimal necessary changes. (Descriptions in the OP are updated.)

maxbrunsfeld:
Thanks for implementing this, and sorry that it took over a year to merge.

PR #443, opened 2019-09-02
[WIP] Highlight Improvements

ubolonton:
I'm working on tree-sitter-based syntax highlighting for Emacs.

I added a trait that would enable callback-based [sources](https://github.com/ubolonton/emacs-tree-sitter/blob/3e1b1feb180d82604c6c1a56b9e2428dc60da047/src/highlight.rs#L22) for `Highlighter`, and was able to make it work with Emacs (up to the part where highlight events are generated, the rest would be Emacs-specific details).

It seems to be a bit slow, due to the overhead of a large number of callbacks. On a 23K JavaScript file:
- Non-incremental parsing takes a couple of ms.
- Looping through all the highlight events (using the already parsed tree) takes around 60-70ms.

As far as I understand, the highlighter uses `source` for several things:
1. Supporting regex selectors in CSS mappings.
2. Emitting highlight events with source fragments.
3. Tracking names of local references and definitions.
4. Something to do with embedded language, in `injection_language_string`?

In my quick tests, most of the source accesses come from 1. and 2.

I have some questions:
- Can 1. be refactored to access the parse tree instead of the source?
- Should 2. be moved to specific rendererers, leaving `HighlightEvent` to contain only the byte-offset range? A text editor renderer would apply these to its internal datastructures, while renderers that want to output texts (HTML/terminal) would hold the original source themselves.
- Can `Highlight` be modified to support custom classes (e.g. `Custom(&'static str)`)? That would give editors some freedom to define their own mappings.
- Should I continue with this refactoring, or should it be a different highlight implementation?

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -12,7 +12,7 @@ use std::i32;
 
 lazy_static! {
     static ref CURLY_BRACE_REGEX: Regex =
-        Regex::new(r#"(^|[^\\])\{([^}]*[^0-9A-F,}][^}]*)\}"#).unwrap();
+        Regex::new(r#"(^|[^\\])\{([^}]*[^0-9A-Fa-f,}][^}]*)\}"#).unwrap();
 }
 
 const ALLOWED_REDUNDANT_ESCAPED_CHARS: [char; 4] = ['!', '\'', '"', '/'];
@@ -653,12 +653,15 @@ mod tests {
                     Rule::pattern(r#"\{[ab]{3}\}"#),
                     // Unicode codepoints
                     Rule::pattern(r#"\u{1000A}"#),
+                    // Unicode codepoints (lowercase)
+                    Rule::pattern(r#"\u{1000b}"#),
                 ],
                 separators: vec![],
                 examples: vec![
                     ("u{1234} ok", Some((0, "u{1234}"))),
                     ("{aba}}", Some((1, "{aba}"))),
                     ("\u{1000A}", Some((2, "\u{1000A}"))),
+                    ("\u{1000b}", Some((3, "\u{1000b}"))),
                 ],
             },
         ];
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -293,7 +293,7 @@ pub fn ansi(
     {
         let event = event.map_err(|e| e.to_string())?;
         match event {
-            HighlightEvent::Source(s) => {
+            HighlightEvent::Source(s, _) => {
                 if let Some(style) = highlight_stack.last().and_then(|s| theme.ansi_style(*s)) {
                     write!(&mut stdout, "{}", style.paint(s))?;
                 } else {
--- a/cli/src/tests/highlight_test.rs
+++ b/cli/src/tests/highlight_test.rs
@@ -481,7 +481,7 @@ fn to_token_vector<'a>(
             HighlightEvent::HighlightEnd => {
                 highlights.pop();
             }
-            HighlightEvent::Source(s) => {
+            HighlightEvent::Source(s, _) => {
                 // Safety: This is safe because we know the original source is kept around
                 // immutably for the duration of the test.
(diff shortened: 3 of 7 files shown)


maxbrunsfeld:
> I'm working on tree-sitter-based syntax highlighting for Emacs.

Awesome! As you can probably tell, the current implementation is quite limited - it's currently only designed to support "batch" syntax highlighting of entire files. I wasn't sure if anyone else would want to directly use my rust implementation for highlighting in an editor (vs using it as a reference, but re-implementing it in the editor's scripting language).

> Should I continue with this refactoring, or should it be a different highlight implementation?

Yes, I'm in favor of generalizing the `highlight` crate to support incremental highlighting.

> Can 1. be refactored to access the parse tree instead of the source?

Unfortunately not - the parse tree does not store any source code, so in order to support regex-based selectors, we'll need to go back to the source code.

> Should 2. be moved to specific renderers, leaving HighlightEvent to contain only the byte-offset range? A text editor renderer would apply these to its internal datastructures, while renderers that want to output texts (HTML/terminal) would hold the original source themselves.

👍 Absolutely. I will actually go ahead and make that change on master.

> Can Highlight be modified to support custom classes (e.g. Custom(&'static str))? That would give editors some freedom to define their own mappings.

Yes. I knew the classes would need to be generalized in order to support other use cases, but haven't decided on an API yet. The Neovim team might have opinions on this matter as well - @bfredl [mentioned](https://github.com/neovim/neovim/pull/9219#issuecomment-466631000) [a similar](https://github.com/neovim/neovim/pull/9219#issuecomment-467114891) desire for flexible class names.

Adding a single `Custom` variant would work. I might also be open to redesigning it more drastically, to get rid of the fixed `enum`. This will take some design work.

maxbrunsfeld:
Ok, that was a fairly big change. A number of other big changes are needed in order to make this module work in a text editor. We may want to these all into separate PRs.

1. `Highlighter` should take a callback to provide the source, instead of requiring the source as a single slice. I see that you've started this already. I don't think we need to define our own trait for this though; I'd prefer to just use `Fn(&tree_sitter::Node) -> T` where `T: AsRef<[u8]>`. The only thing missing is a way to provide the length. Maybe `source_len: usize` could just be a separate argument. What do you think of that?

2. `Highlighter` needs to be able to re-use existing trees, rather than parsing from scratch every time. This is going to be a little more complex than just taking one `Tree` as a parameter in the constructor, because of multi-language documents. For example, in an HTML file with some large `script` and `style` tags, we'd really want to reuse *both* the HTML tree and the CSS and JavaScript trees. So I think we might want to introduce a new stateful type that represents a highlighted document, which contains one *or more* `Tree` instances. I'm not sure yet exactly exactly what this should be called. In Atom it's called [`LanguageMode`](https://github.com/atom/atom/blob/master/src/tree-sitter-language-mode.js).

3. We need to be able to iterate over select regions of a document, not always the entire document. For performance, after an edit I think you'll want to use the `Tree::changed_ranges()` to determine which lines need to be re-highlighted, and then walk only those documents. It'd be good to be able to `seek()` the highlight iterator to the changed part of the document without having to use `.next()` to step over every single leaf node *before* the change.

Maybe in this PR, let's just try and complete step 1. Thoughts?

ubolonton:
>  I wasn't sure if anyone else would want to directly use my rust implementation for highlighting in an editor (vs using it as a reference, but re-implementing it in the editor's scripting language).

I originally thought about reimplementing it in Emacs Lisp, but then decided that being able to reuse the exact same implementation would be better for maintainability (and probably performance).

> Maybe in this PR, let's just try and complete step 1

Yeah, I think these should be split up as well, but maybe we can keep this PR as a discussion (with code), and make smaller changes in other PRs, with reference to this?

> In Atom it's called LanguageMode.

I want to read Atom's highlighting implementation that uses tree-sitter, to understand e.g. how it does incremental highlighting. Is it all in this file, or should I look into other places in Atom's source as well?

> Maybe source_len: usize could just be a separate argument.

Do you mean a separate argument to `Highlighter`? I that works, but how about the use of `source` to provide the parser with chunks of the sub-language's text?

ubolonton:
Closing now that we have #444 (tree query).

PR #444, opened 2019-09-12
Introduce the 'Tree query' - an API for pattern-matching on syntax trees

maxbrunsfeld:
### Overview

This pull request adds a new data type to the Tree-sitter C library: `TSQuery`. A query represents one or more *patterns* of nodes in a syntax tree. You can instantiate a query from a series of [S-expressions](https://en.wikipedia.org/wiki/S-expression) (similar to those used in Tree-sitter's [unit testing system](http://tree-sitter.github.io/tree-sitter/creating-parsers#command-test)). You can then execute the query on a syntax tree, which lets you efficiently iterate over all of the occurrences of the patterns in the tree. This works in C, Rust, and JavaScript (via Wasm).

### Background

Many code analysis tasks involve searching for patterns in syntax trees. Some of these analysis tasks are very common, and it'd be nice to avoid implementing them multiple times. Examples of some common tasks you might want to perform with a Tree-sitter syntax tree:

* Computing syntax highlighting
* Computing code-folding regions
* Finding *nested* documents to parse separately (JavaScript within HTML, Ruby within ERB, etc)

The Tree-sitter C library is used from several languages, so in order for these analyses to be reusable, they have to be specified in a way that doesn't depend on any particular high-level language's runtime.

### Prior Solutions

* In https://github.com/tree-sitter/tree-sitter/pull/204, I added a system called *property sheets*, that let you use CSS to select syntax nodes and assigning them properties. It was based on a compile-time step that converted the `.css` files into finite state machines encoded as JSON.
* In https://github.com/tree-sitter/tree-sitter/pull/283, I added a Rust implementation of syntax highlighting based on these property sheets.

But these solutions have some major drawbacks:

1. It's awkward having to compile the CSS files into JSON files ahead-of-time. It means that the JSON files need to be checked into the git repository (or somewhere) and applications can't easily use the feature for other purposes. Also, The generated JSON files are larger than the source CSS, which is bad for front-end use of Tree-sitter.
2. We could switch to processing the CSS at *runtime*, but the CLI currently relies on [a Rust library](https://github.com/kaj/rsass) for parsing the CSS, and a bunch of Rust code for transforming it into a DFA. It would be hard to consolidate this into the core C library, and it would bloat the library.
3. Most immediately-pressing - CSS has limited expressive power. It's awkward to select nodes based on details of their siblings. This turns out to be important for certain syntax highlighting tasks. For example, in this JavaScript code:

    ```js
    var foo = function(a) { /* ... */ }
    ```

    The variable `foo` is normally highlighted as a function. The current `tree-sitter-highlight` crate can't do this, because properties of nodes can't depend on their siblings like that.

### The Query Language

Instead of using CSS, compiled into a DFA ahead-of-time, the new `TSQuery` API uses S-expressions, compiled into an NFA at runtime. Here are some examples of what the query language currently looks like.

To select all of the methods, together with the name of the class:

```clojure
(class_declaration
  name: (identifier) @the-class-name
  body: (class_body
    (method_definition
      name: (property_identifier) @the-method-name)))
```

To select variables to which functions *or* arrow functions are assigned:

```clojure
(assignment_expression
  (identifier) @function-name
  (function))

(assignment_expression
  (identifier) @function-name
  (arrow_function))
```

To select all null checks:

```clojure
(binary_expression
  left: (*) @null-checked-object
  operator: "=="
  right: (null))
```

The annotations that start with `@`, like `@the-class-name` are called *captures*. These identify which nodes should be *returned* from the query, and what *name* should be associated with them when they are returned.

With the exception of captures, the query language is identical to the format in which Tree-sitter's unit tests are written: S-expressions.

### Static Verification of Queries

When a query is instantiated, all of the node names and field names are transformed into integer ids, and an error is raised if any of the names are not actually defined in the grammar. 

Because the queries are so easy to parse, it would be easy to add an even more thorough check that uses the `node-types.json` to check that all of the parent-child relationships are valid, and *could* actually occur. I'll implement this in a follow-up PR. My plan is that reusable queries could be stored in a top-level `queries` folder in the grammar repositories, and the `tree-sitter test` command could be augmented to check the validity of the queries using the existing node type information.

### Trying it Out

You can write and execute queries interactively in the web UI, both on the docs site and via the `tree-sitter web-ui` command in your own grammar repos.

playground-queries


bfredl:
@maxbrunsfeld Very nice, this seems it will simplify the implementation of tree-sitter highlighting in neovim a lot (which will be a priority again after the imminent 0.4 release).

I have one question though. The property sheet cursor in `lib.rs`, which I partially reimplemented in our lua bindings, implemented the DFA execution on top of the random-access cursor, so it supported the full set of movement operations of a "raw" TSTreeCursor, and updated the DFA state to it. I e we only needed to wrap one cursor API, which optionally could have a property sheet assigned to it.

 This PR looks like it implements a separate cursor type with much more limited cursor movement (only forward-advancement, though still skipping irrelevant sub-trees). Would it possible to support updating this new NFA on top of the unrestricted TSTreeCursor as with the property sheets? It would simplify the API if we only supported one fully-featured cursor type. Or is the new query semantics (perhaps conditioning on future siblings and such) making it impossible/hard to support random access cursor?

bfredl:
@turbo It is work in progress https://github.com/neovim/neovim/pull/10124/files#diff-5ee91216d7ccd53f980bbf60ffcc29a6 . Eventually we might separate out the parts that are not specific to to neovim as a separate library. Though we make some simplifying assumptions (like all trees accessible to lua are treated as immutable), that might not be appropriate for completely general bindings.

turbo:
@maxbrunsfeld This is great. I'm maintaining an internal parsing toolkit similar to tree-sitter and was just about to start implementing a query language for static analysis purposes (security, code duplication etc). So now I can aim for compatibility with this query format, too. Though in my case, the query would be run non-realtime against a previously serialized, immutable tree-sitter syntax tree.

@bfredl Thanks! for the most part, that fits my use case. I'll try to play around with that soon. Our whole system is a mix of Rust and Lua, so I should be able to make it work either way.

maxbrunsfeld:
> Or is the new query semantics (perhaps conditioning on future siblings and such) making it impossible/hard to support random access cursor?

Yeah, queries can capture nodes based on patterns that match against their later siblings, so the process of identifying captures isn't as 'local' as it was with the property sheet system.

The result is that `TSQueryCursor` has a more limited API than `TSTreeCursor`, but I actually think it's an easier API to use. It's more like a classic *iterator* conceptually: a simple sequence of values that you can lazily consume. In Rust, we expose it using the `Iterator` trait.

For Lua, I might expose the `TSQueryCursor` as a simple iterator. 

```lu
start = {row = 30, column = 0}
end = {row = 55, column = 0}

for i, match in query.exec(tree, start, end) do
  print(match.pattern_index)
  for capture_name, node in ipairs(match.captures) do
    print(capture_name, node)
  end
end
```

In that design, `TSQueryCursor` itself isn't even exposed directly to Lua. The downside of that is there *is* a small cost to instantiating and deleting a `TSQueryCursor`. In the JavaScript WASM bindings, since there's only one thread, I just maintain a global `TSQueryCursor` instance, and reuse it any time `Query.exec` is called. I'm not sure if that would be good for NeoVim.

Another binding idea - In JavaScript/WASM, FFI is more costly than allocating arrays, so I don't even expose the iterator interface. Instead, `exec` just returns an array of matches. I'm not sure if the same applies to Lua though.

maxbrunsfeld:
> in my case, the query would be run non-realtime against a previously serialized, immutable tree-sitter syntax tree.

@turbo That's interesting. Is that project open source by chance? I'd be curious to check it out.

bfredl:
@maxbrunsfeld Thanks, that makes sense. Our plan is to first implement fine-grained lua API calls and then profile that solution. If it is too slow there would be multiple ways forward, including to do baseline highlighting (everything that can be expressed as queries only, which now is most things I guess) in C only, and use lua API for specific additional highlighting only.

maxbrunsfeld:
@ubolonton You mentioned that you're working on syntax highlighting in Emacs. My goal with this feature is to *replace* the current Property-sheet system for syntax highlighting. I think that because this feature is totally implemented in the C library, it will make it easier to share more of the implementation between different applications.

@bfredl @ubolonton I would like to get the design of this right for my use case at GitHub as well as both of your use cases, since NeoVim and Emacs are important ecosystems. Some notes about how this system differs from property sheets:

1. There is no [fixed `enum`](https://github.com/tree-sitter/tree-sitter/blob/4151a428ec9059b70f3a7beecba8384b649eb621/highlight/src/lib.rs#L60-L89) containing all of the valid highlight names. The idea is that the `@capture` names can be arbitrary strings that you can use as highlighting identifiers.
2. Obviously, this is part of the core C library, not a separate Rust library
3. This already works with existing trees, instead of operating on the source code.
4. You can already use this to search only the *visible* lines.
5. The queries can be automatically checked for validity against the grammar.

However, there are still some issues to figure out:

1. Different applications have different naming conventions for highlights. I see two ways of handling this.
    * My preference is that we would still *share* the queries, but each dependent app would apply a mapping from Tree-sitter's highlight names into the application's native highlighting names. In this design, there would be a `queries/highlightquery` file checked into each Tree-sitter grammar repo. These files would all use consistent capture names like `@function.builtin`, `@variable.parameter`, `@number.float`, etc. Emacs and Vim could use these queries directly, but would have to maintain mappings for these highlight names.
    * Alternatively, Vim and Emacs could just use their own highlight queries that use the applications' native highlight names directly. This is at least easier now than with Property Sheets, because queries are compiled from source at runtime, so you wouldn't have to store the generated JSON file.

2. The query API isn't highly tailored toward syntax highlighting. Currently, it lets you iterate over the *matches*, where each match can have multiple captures. Matches can overlap, so individual captures are not guaranteed to appear in order. For example, in this JavaScript query:

    ```clojure
    (object
      (pair
        key: (identifier) @method.def
        value: (function)))

    ":" @delimiter.pair
    ```

    With this code:

    ```js
    x = {foo: function() {}}
    ```

    The `delimiter.pair` capture for `:` would be returned *before* the `method.def` capture for `foo`, even though `foo` appears first in the code.

    For syntax highlighting purposes, we'd want to iterate over the individual *captures* (instead of the matches), in the order that the captures appear. So a little bit of post-processing logic is required. Right now, I'm not sure if this should be implemented as part of the C `TSQuery` API, or if it should be left to applications to implement this part.

bfredl:
> My preference is that we would still share the queries, but each dependent app would apply a mapping from Tree-sitter's highlight names into the application's native highlighting names. In this design, there would be a queries/highlightquery file checked into each Tree-sitter grammar repo. These files would all use consistent capture names like `@function.builtin, @variable.parameter, @number.float`, etc. Emacs and Vim could use these queries directly, but would have to maintain mappings for these highlight names.

I think this is the correct baseline solution. We wouldn't want to create highlighting rules from scratch for all languages that tree-sitter supports. Though I think we (neovim) should allow both actually. As queries can be parsed at runtime, some users will want to add their own matches/captures on top of the baseline rules, which can use vim highlight group names directly. As canonical tree-sitter capture names use `lower.case` and vim highlight groups are `UpperCase`, the intention of any rule should be unambiguous.

> For syntax highlighting purposes, we'd want to iterate over the individual captures (instead of the matches), in the order that the captures appear. So a little bit of post-processing logic is required. Right now, I'm not sure if this should be implemented as part of the C TSQuery API, or if it should be left to applications to implement this part.

I will need to experiment a bit with this in practice, but I think we can handle this ourselves, as long as the _matches_ are ordered by their start position. I e if we see a match on line 197, we know that line 196 is "done", and we can flush that line to the terminal/UI. But if a match on line 197 has a capture on line 200 it should be no problem, we can just stash it in some temporary storage until the renderer has reached line 200.

maxbrunsfeld:
> I will need to experiment a bit with this in practice, but I think we can handle this ourselves, as long as the matches are ordered by their start position.

Yeah, this is a good point. I'm going to need to add or tweak APIs to make this possible. I will do that before merging.

maxbrunsfeld:
Ok, I've made a change to the `TSQueryCursor` API so that it's even simpler. You can either iterate over the matches using `ts_query_cursor_next_match`, or iterate over the captures using `ts_query_cursor_next_capture`. In Rust, these are currently exposed via `matches()` and `captures()`, which both return iterators. In JavaScript, it's the same except that they both return arrays.

This should make syntax highlighting pretty simple to implement.

maxbrunsfeld:
> everything that can be expressed as queries only, which now is most things I guess

That reminds me. One thing that currently *can't* be expressed in Queries is matching based on a node's text. For example, in JavaScript, we [highlight capitalized identifiers](https://github.com/tree-sitter/tree-sitter-javascript/blob/a730b5c210904e2e3c1f601125a059fde1b35850/properties/highlights.css#L96-L102) as "constants", and [highlight certain function names](https://github.com/tree-sitter/tree-sitter-javascript/blob/a730b5c210904e2e3c1f601125a059fde1b35850/properties/highlights.css#L11-L16) as "builtins".

Right now, I'm thinking that maybe these things could be done via post-processing, but I'm not 100% sure how it should work.

turbo:
@maxbrunsfeld 

> For example, in JavaScript, we highlight capitalized identifiers as "constants", and highlight certain function names as "builtins".

Does that also apply to languages where exporting depends on capitalization rules? A good example would be Go, but there are languages where this is configured at runtime, e.g. MoonScript:

```moon
local * -- forward declare every top-level identifier
local ^ -- same, but only (first letter) capitalized identifiers

export * -- make all top-level identifiers part of the module returns
export ^ -- same, but for caps, equivalent to Golang

-- example:

local *
export ^

-- local function, usable everywhere because fwd-decl
foobar = -> 42

-- exported table
FizzBuzz = { test: 42 }

-- private class
class helper
  -- ...

-- public / exported class
class Producer
 -- ...
```

as you can see, GitHub highlights them based on just the first letter (implying they are exported), but without the `export ^`, both `Producer` and `FizzBuzz` would be private forward-declared module-local identifiers.

So I guess the question is, could (if there was a grammar for MoonScript), a query identify *actually* exported identifiers, not based on their casing, but based on whether they are explicitly returned and or implicitly `export`ed?

> @turbo That's interesting. Is that project open source by chance? I'd be curious to check it out.

No, sadly not, it's internal tooling. But we could have a chat about that later this year if you're interested. I'm certainly interested in your take on some of the things we're doing.

ubolonton:
This is awesome! I'm going to try integrating it with Emacs this week.

My main question is: How would this support incremental highlighting? Is it by limiting `exec` to the changed ranges?

> Different applications have different naming conventions for highlights. I see two ways of handling this.

I also prefer having a baseline set of queries, but still allowing adding new queries and mappings at runtime.

> Right now, I'm thinking that maybe these things could be done via post-processing, but I'm not 100% sure how it should work.

I think we can start by having applications (specific language modes) post-process these, until it turns out there are patterns common enough across languages.

maxbrunsfeld:
Ok, in order to deal with the Text-matching post-processing, I have added one new construct to the query language that I'm calling *predicates*. They look like this:

```clojure
; treat ALL_CAPS identifiers as constants
((identifier) @constant
 (match? @constant "^[A-Z][A-Z_]+"))

; recognize certain built-in functions
((identifier) @function.builtin
 (eq? @identifier "require"))
```

To add *predicates* to a pattern, you wrap the pattern in a list. All of the remaining elements of the list after the pattern are interpreted as predicates. Predicates are lists of symbols, strings and `@`-prefixed capture names, which reference nodes in the pattern.

The core C library itself does not process predicates. It just *exposes* them in a structured form, associated with each pattern id, so that higher-level code can *implement* particular predicate functions.

For now, I have added built-in implementations of these predicate functions in both the Rust and Wasm bindings:

* `match?` - Check that a captured node's text matches a given regular expression.
* `eq?` - Check that a captured node's text equals a certain string.

    You can also compare the text contents of *two* captured nodes. This is pretty interesting - it lets you query for more sophisticated patterns. For example, this pattern would match assignments that update a variable based on its own previous value, like `x = x + 1`:

    ```clojure
    ((assignment_expression
       left: (identifier) @id1
       right: (binary_expression
         left: (identifier) @id2))
     (eq? @id1 @id2))
    ```

This does mean that in the Rust binding, you have to pass a `text_callback`, which provides access to a given node's text. @ubolonton I know this was a source of problems with the old highlighting system, but I think this is an important feature, and the callback shouldn't be called unnecessarily - only when a node's text is needed due to a predicate.

maxbrunsfeld:
> My main question is: How would this support incremental highlighting? Is it by limiting exec to the changed ranges?

Yes, exactly. You can limit the range in which the `TSQueryCursor` will search. In Rust, you can use `QueryCursor::set_point_range` or `set_byte_range`:

```rust
let captures = cursor
    .set_point_range(Point::new(30, 0), Point::new(70, 0))
    .captures(&query, tree.root_node(), text_callback);
```

So in a text editor, you can update the UI by getting the changed ranges, searching for captures within those ranges, and re-rendering those lines.

lib/src/bits.h
```diff
@@ -0,0 +1,29 @@
...
+  _BitScanReverse(&result, x);
+  return 31 - result;
+}
+
+#else
+
+static inline uint32_t count_leading_zeros(uint32_t x) {
+  if (x == 0) return 32;
+  return __builtin_clz(x);
+}
```
maxbrunsfeld: When matching queries, we need to keep track of multiple *lists* of captured nodes. I use an object pool data structure called `CaptureListPool` to manage these lists. The `CaptureListPool` uses a bitmap to keep track of which lists are in use. I find the first free list in the bitmap using this `count_leading_zeros`, which maps to different intrinsic functions on POSIX and windows.

In the WASM world, I wasn't sure if emscripten was going to support `__builtin_clz`, but it does!

lib/src/tree_cursor.c
```diff
@@ -264,25 +269,61 @@ TSFieldId ts_tree_cursor_current_field_id(const TSTreeCursor *_self) {
...
+  if (field_id) {
+    for (const TSFieldMapEntry *i = field_map; i < field_map_end; i++) {
+      if (i->field_id == field_id && i->child_index > child_index) {
+        *field_has_additional = true;
+      }
+    }
+  }
+
+  return field_id;
+}
```
maxbrunsfeld: This is kind of a weird new internal function. It returns the field associated with the current node, and it also indicates (via the `*field_has_additional` out-parameter) whether any other, later siblings of this node can possibly have the *same* field.

This allows for an important optimization in the pattern-matching process. Most of the time, field names are single-valued: they are only associated with one child node. For a pattern like this:

```clojure
(method_definition
  name: (identifier) @the-function-name)
```

when we find the `identifier` node with the `name` field, we can know for sure that the *only* way that this pattern can match is for it to capture the current node; there will not be any other `identifier` nodes with the `name` field. This allows us to update our `QueryState` for this pattern *in-place*, instead of splitting it as we would normally need to do when adding a capture, in order to perform a truly exhaustive search.

cli/src/tests/query_test.rs
```diff
@@ -0,0 +1,689 @@
...
+        assert_eq!(
+            Query::new(language, r#"((identifier) ()"#),
+            Err(QueryError::Syntax(16))
+        );
+        assert_eq!(
+            Query::new(language, r#"((identifier) @x (eq? @x a"#),
+            Err(QueryError::Syntax(26))
+        );
+    });
+}
```
maxbrunsfeld: The `TSQuery` API is mainly tested through these Rust unit tests. There are also some additional tests in the JavaScript-based WASM test suite.

cli/src/query.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::fs;
+use std::io::{self, Write};
+use std::path::Path;
+use tree_sitter::{Language, Parser, Query, QueryCursor};
+
+pub fn query_files_at_paths(
+    language: Language,
+    paths: Vec<&Path>,
+    query_path: &Path,
+) -> Result<()> {
```
maxbrunsfeld: This file implements a new `tree-sitter query` command. This command isn't really very useful right now, but it was useful at a certain stage for testing queries from the command line.

Right now, it just prints a bunch of messages to stdout. I might change it in the future to just print the contents of the file, but with query matches colored in some way. Kind of like the `tree-sitter highlight` command, but where you'd specify an arbitrary query filename, instead of always looking for the *syntax-highlighting* query.

lib/binding_rust/lib.rs
```diff
@@ -921,6 +956,340 @@ impl<'a, P> TreePropertyCursor<'a, P> {
...
+                    ffi::ts_query_string_value_for_id(ptr, i as u32, &mut length as *mut u32)
+                        as *const u8;
+                let value = slice::from_raw_parts(value, length as usize);
+                let value = str::from_utf8_unchecked(value);
+                value.to_string()
+            })
+            .collect::<Vec<_>>();
+
+        // Build a vector of predicates for each pattern.
+        for i in 0..pattern_count {
```
maxbrunsfeld: This loop reads the structured information about the query's *predicates* and converts it into a Rust `enum` which is used to implement the `eq?` and `match?` predicates that I mentioned in the comments.

There's very similar code in `binding.js` for the WASM binding. In both cases, I use the higher-level language's regex-matching facilities to implement `match?`.

So the plan is, every Tree-sitter binding is going to have a little bit of its own duplicated logic for actually implementing the predicate stuff. All the C library does there is expose the parsed predicates to the caller.

bfredl: @maxbrunsfeld I think this is reasonable, but I'm not sure I have groked the C API yet. How much does a `TSQueryPredicateStep` represent? In  `(eq? @ab hi)`, is each of `eq?` and `@ab` and `hi` a separate "step"? So there can never be a predicate like `(func? @aa (xx yy))` (no nesting), right? Can a singe capture have multiple predicates, and how is that represented then?

maxbrunsfeld: >  In `(eq? @ab hi)`, is each of `eq?` and `@ab` and `hi` a separate "step"?

Yeah, exactly.

> So there can never be a predicate like `(func? @aa (xx yy))` (no nesting), right?

Yeah, with the current API, nested lists like that are not supported in predicates.

> Can a singe capture have multiple predicates, and how is that represented then?

A single pattern can indeed have multiple predicates. So, for example, you can do:

```clojure
; Match bounds checks like `i < items.length`
((binary_expression
  left: (identifier) @variable
  operator: "||"
  right: (member_expression
    property: (property_identifier) @property))

 (match? @variable "^(i|j|k)")
 (eq? @property "length"))
```

The `ts_query_predicates_for_pattern` function returns the *total* number of steps, and a pointer to the first step of the first patttern. Within this buffer, the two individual patterns `(match? @variable "^(i|j|k")` and `(eq? @property "length")` are separated by a *sentinel* step whose `type` is `TSQueryPredicateStepTypeNone`.

It's a little weird, but I wasn't sure of a simpler way to represent an array of arrays with a C API, especially since *internally* I want to store all of the steps in a single buffer.

lib/include/tree_sitter/api.h
```diff
@@ -602,6 +635,147 @@ int64_t ts_tree_cursor_goto_first_child_for_byte(TSTreeCursor *, uint32_t);
...
+ * Get the name and length of one of the query's captures, or one of the
+ * query's string literals. Each capture and string is associated with a
+ * numeric id based on the order that it appeared in the query's source.
+ */
+const char *ts_query_capture_name_for_id(
+  const TSQuery *,
+  uint32_t id,
+  uint32_t *length
+);
+const char *ts_query_string_value_for_id(
```
bfredl: Why not have capture names and this kind of general "strings" (only used by predicates, and not the query engine itself?) share a single "namespace" of interned string id:s, and a single function to look them up? It could still have all the capture names in the beginning, if there is a need to index a dense array of captures or something. It would simplify FFI by only managing one table of interned strings, and not two.

lib/binding_rust/lib.rs
bfredl: Thanks, it makes sense now.

lib/include/tree_sitter/api.h
maxbrunsfeld: The main reason I lead toward keeping them separate is that callers will need to use them slightly differently. When creating bindings to this API, the `capture_names` need to stored in an array that will be used every time you run a query, whereas the `strings` only need to be loaded temporarily, while constructing a query.

With this approach, we can preallocate and populate the `captureNames` array using `ts_query_capture_count / ts_query_capture_name_for_id`, and preallocate and populate a separate `stringValues` array using `ts_query_string_count / ts_query_string_value_for_id`. Then, at the end of the query constructor function, we can simply drop the `stringValues` array.

> It could still have all the capture names in the beginning, if there is a need to index a dense array of captures or something.

That's true. With that approach, we could at least *shrink* the array once we are done using the `strings`. But in order to do this, callers still need to know how many captures there are, so there still needs to be a separate `ts_query_capture_count` function.

In the end, I think the APIs are slightly more obvious in their meaning if `captures` and `strings` are just presented as two separate arrays, where each one has identical APIs for accessing its length and its contents.

lib/binding_rust/lib.rs
maxbrunsfeld: I made this a bit clearer in the docs in my latest commit.

lib/include/tree_sitter/api.h
bfredl: I see  what you mean, though this seems to assume very specific details how you would use the API. In the first iteration I would do as much as possible lua-side, at least the non-hot preprocessing of the queries. So this would just imply needing two code paths (or helper functions) for converting an internal id to a lua string instead of just one. Though it is not much duplicated code, probably only like 3-5 lines.

maxbrunsfeld: Yeah, good point. I'm writing the API with the assumption that FFI has some noticeable cost, based on experience with JavaScript via Node and WebAssembly. If you aren't keeping a managed copy of the captures array, then none of the things that I said above are applicable.

I think I'll keep it separate for now, though I see your point about the potential reduction in API functions.

On a related note, in 1af85dc3f758e5b4e2179fec81f95b89dadc26e3 I did delete two APIs related to captures and string values, which I realized were unused by either of my bindings. So now there are slightly fewer 😁.

bryphe:
Thanks for the heads up @maxbrunsfeld ! This looks promising - our current implementation of [collecting tokens for syntax highlighting](https://github.com/onivim/reason-tree-sitter/blob/4947eadefea0ac40a230eb158cc0d8f16e9ae010/src/Syntax.re#L139) and applying highlights is slow - we weren't even using the `TSCursor` API yet. 

Cool to see all the progress happening here!

> My preference is that we would still share the queries, but each dependent app would apply a mapping from Tree-sitter's highlight names into the application's native highlighting names. In this design, there would be a queries/highlightquery file checked into each Tree-sitter grammar repo. These files would all use consistent capture names like @function.builtin, @variable.parameter, @number.float, etc. Emacs and Vim could use these queries directly, but would have to maintain mappings for these highlight names.

This sounds good to me. We're supporting VSCode themes, so we'd just need to map these capture names to ones that the VSCode themes expect (aka, textmate scopes). Seems like with this solution, we could write a single mapping function, applicable across languages, to go from the capture names like `@function.builtin` to the textmate scopes we require for theming. 

This gets us out of the business of needing per-language mappings, which would be great 👍

Razzeee:
I guess your aware, but VSCode is supposed to move away from textmate scopes pretty soon.

bryphe:
> I guess your aware, but VSCode is supposed to move away from textmate scopes pretty soon.

@Razzeee - Indeed, they are exploring alternate tokenization strategies (outlined here, I believe: https://github.com/microsoft/vscode/issues/77140), but I believe they will still output tm-compatible scopes - otherwise they would break all the existing theme extensions.

Razzeee:
Sounds correct this issue might also be involved in theming. 
https://github.com/microsoft/vscode/issues/77133
Both are planned for this iteration.

jeff-hykin:
> * You can also compare the text contents of _two_ captured nodes. This is pretty interesting - it lets you query for more sophisticated patterns. For example, this pattern would match assignments that update a variable based on its own previous value, like `x = x + 1`:
>   ```clojure
>   ((assignment_expression
>      left: (identifier) @id1
>      right: (binary_expression
>        left: (identifier) @id2))
>    (eq? @id1 @id2))
>   ```

This is absolutely fantastic, thank you for building awesome stuff like this. This has been my largest external obstacle in terms of using the tree sitter practically. Later this year I'd love to work on contributing some tools so that theme makers can easily take advantage of these kinds of selections. I'd also be interested in connecting with people to agree upon a mostly-universal naming convention for capture names.

The only piece of concern I have is for the capture names though, like `@function.builtin`.  TextMate has the terrible legacy of making users to pick between multiple incorrect answers. Sometimes tokens are both parameters, and constants, and sometimes they're equally a function, class, variable, and object all at the same time. So long as there's a way to tag/label a token with multiple capture names, I'll be 100% on board.

PR #448, opened 2019-09-19
Reimplement syntax highlighting with tree queries instead of property sheets

maxbrunsfeld:
### Summary

This PR reimplements the [`tree-sitter-highlight` crate](https://github.com/tree-sitter/tree-sitter/tree/master/highlight) (including [its C API](https://github.com/tree-sitter/tree-sitter/blob/master/highlight/include/tree_sitter/highlight.h)) and the `tree-sitter highlight` subcommand. Syntax highlighting is no longer based on [property sheets](https://github.com/tree-sitter/tree-sitter/pull/204), and instead use the new 'tree query' API introduced in #444.

### The Query Structure

The new syntax highlighting implementation is based on three query files:

1. `queries/highlights.scm`

    This file contains patterns with capture names that correspond to syntax highlighting styles. Examples are `@type`, `@function.builtin`, `@punctuation.bracket`, etc. The names are dot-separated, and the idea is that users may want to style in a course grained way (e.g. `@function`) or in a more fine-grained way (e.g. `@function.method.builtin`), and both will work.

    **Tie-breaking convention** - In the event that two captures in this query both capture a given node with two different capture names, the **first pattern is preferred**.

2. `queries/injections.scm`

    This file contains patterns that, when matched, cause portions of the syntax tree to have their text *re-parsed* with a new "injected" grammar. These patterns can specify three fixed captures:

    * `@injection.site` - This is the parent node that contains all of the information regarding the injection.
    * `@injection.content` - This node will have its text re-parsed using some other grammar. If there are multiple matches with the same `site` but *different* `content` nodes, then the all of the `content` nodes will be parsed together as one nested document. This allows you to parse multiple *disjoint* ranges of text as one injection, which is important for templating languages like [ERB](https://en.wikipedia.org/wiki/ERuby), in which code can be interspersed with other text.
    * `@injection.language` (optional) - This node's text will be used to determine which language to use for the injection. For example, in a Ruby HEREDOC, where the heredoc delimiter often indicates the language, you would capture the delimiter as the `language`.

    The following *predicates* are recognized for patterns in this file:

    * `(set! injection.language "the-language")` - This allows you to *hard-code* a specific language that should be injected, instead of inferring one from the text of a captured node.
    * `(set! injection.include-children)` - By default, when an `injection.content` node is captured, only its *own* text is included in the injection, not the text of its children. This directive allows you to override that, including *all* of the text contained by a given node instead.

3. `queries.locals.scm`

    This file allows you to keep track of local variables, ensuring that they are styled the same way in every place where they occur. Patterns in this file can have these captures:

    * `@local.scope` - This indicates that the captured node introduces a new scope.
    * `@local.definition` - This indicates that the captured node represents a newly-introduced local variable.
    * `@local.reference` - This indicates that the captured node may be a reference to a local variable introduced earlier in the current scope, or some surrounding scope.

### Tasks

* [x] Get the tests passing
* [x] Update the `tree-sitter-highlight` readme
* [x] Add queries for all of the languages which previously had property sheets
* [x] Add the ability for queries match `ERROR` nodes (🎩 @bfredl for raising this)
* [ ] Add some docs about this on the documentation site

### Example Highlighting Queries

* [x] HTML https://github.com/tree-sitter/tree-sitter-html/pull/12
* [x] JS https://github.com/tree-sitter/tree-sitter-javascript/pull/123
* [x] Rust https://github.com/tree-sitter/tree-sitter-rust/pull/55
* [x] ERB/EJS https://github.com/tree-sitter/tree-sitter-embedded-template/pull/2
* [x] Go https://github.com/tree-sitter/tree-sitter-go/pull/33/files
* [x] Python https://github.com/tree-sitter/tree-sitter-python/pull/54
* [x] Ruby https://github.com/tree-sitter/tree-sitter-ruby/pull/110


bfredl:
@maxbrunsfeld A question about using `ts_query_cursor_next_capture` together with predicates. It seems with the current API we would either need to (1) recompute the match predicate for every capture in the match or (2) cache the predicate status using the `TSQueryMatch.id` as a key (assuming it is unique until the cursor has been restarted?). It seems the current rust implementation does (1). 

Do you think it would be reasonable to add a method to the query-cursor API to disregard the entire match from the previous `next_capture` call ( IIUC remove it early from `self->finished_states` ) ? Or perhaps it can be assumed that highlight matches are going to have like only 1-2 captures each typically, so that it doesn't really matter.

maxbrunsfeld:
Yeah, good question. That is why I added the `id` field; I was planning on caching the last few predicate results as an optimization, but I never got around to implementing that. The match ids should be unique until the cursor is restarted, as you said.

> Do you think it would be reasonable to add a method to the query-cursor API to disregard the entire match from the previous next_capture call (IIUC remove it early from self->finished_states)? 

Yeah, that's a good idea. It would avoid the problem of callers not knowing how many predicate results to cache. Maybe something like

```c
void ts_query_cursor_remove_match(TSQueryCursor *self, uint32_t match_id);
```

bfredl:
> The match ids should be unique until the cursor is restarted, as you said. ...  Maybe something like

Nice. I was thinking of allowing arbitrary lua code to define new predicate kinds, and it would be good to ensure these are called just once per match/query.

Another question/comment. Now captures in a match could be used to both return information from the query (i e the nodes to highlight) but also as terms in predicates, even if these captures won't be used later on. Perhaps it would be useful if there was a naming convention or something for captures that are _only_ used as part of predicates, so that we can filter them out before returning them as part of a list/iterator etc.

maxbrunsfeld:
Yes! Maybe capture names beginning with an underscore should be hidden from client code. I was thinking about that same idea, but wasn’t sure if it would be useful, so I hadn’t pursued it yet.

bfredl:
I've started to prototype highlighting in neovim https://github.com/neovim/neovim/pull/11113 . The logic is much simpler now with queries :+1:. 

However I get a segfault inside the query cursor sometimes:
```
=================================================================
==14002==ERROR: AddressSanitizer: SEGV on unknown address 0x614000169a3e (pc 0x55ae7b48b209 bp 0x7fff6182cae0 sp 0x7fff6182bd00 T0)
==14002==The signal is caused by a READ memory access.
    #0 0x55ae7b48b208 in ts_query_cursor__advance /home/bjorn/dev/neovim/build/../src/tree_sitter/query.c
    #1 0x55ae7b493807 in ts_query_cursor_next_capture /home/bjorn/dev/neovim/build/../src/tree_sitter/query.c:1328:10
    #2 0x55ae7a70a701 in query_next_capture /home/bjorn/dev/neovim/build/../src/nvim/lua/treesitter.c:669:7
```
This seems to consistently happen for certain kinds of errors, like if a C source contains an unfinished `/*` comment. (Though it is possible that it is our code is missing some error checking somewhere, I will investigate more on our end).

bfredl:
BTW, I guess it would be a nice feature also to match error/mising nodes specifically, in order to highlight syntax errors.

maxbrunsfeld:
> This seems to consistently happen for certain kinds of errors, like if a C source contains an unfinished /* comment.

Oh, thanks for reporting that. I will try to investigate this on my end also, but if you can you find a C file and query that causes the error, that would be super helpful.

bfredl:
@maxbrunsfeld This is the basic query I started prototyping with :
```
"const" @keyword
"else" @keyword
"for" @keyword
"if" @keyword
"return" @keyword
"static" @keyword
"while" @keyword
(string_literal) @string
(primitive_type) @type
(comment) @comment
```

This C source file seems to be enough:
```
int main() {
  /*
}
```

maxbrunsfeld:
@bfredl I haven't been able to reproduce the problem. I'm running this command:

```sh
tree-sitter query --scope source.c --captures tree-sitter-c/queries/highlights.scm test.c
```

Where `tree-sitter-c/queries/highlights.scm` has your query, and `test.c` is your C source code. I've also tried running this on various other files which would have parse errors (such as non-C source files) and haven't found a problem so far.

Can you try to reproduce the problem through the `tree-sitter` CLI? You'll have to build master in order to use the `query` command. Alternatively, you could try to reproduce the issue via the [web playground](http://tree-sitter.github.io/tree-sitter/playground). You can now use the `Query` checkbox to enable testing out queries.

maxbrunsfeld:
I added a few more patterns just to increase the number of matches, but I'm still not seeing problems.

```clj
"break" @keyword
"case" @keyword
"const" @keyword
"continue" @keyword
"do" @keyword
"else" @keyword
"enum" @keyword
"extern" @keyword
"for" @keyword
"if" @keyword
"inline" @keyword
"return" @keyword
"sizeof" @keyword
"static" @keyword
"struct" @keyword
"switch" @keyword
"typedef" @keyword
"union" @keyword
"while" @keyword

"#define" @keyword
"#else" @keyword
"#endif" @keyword
"#if" @keyword
"#ifdef" @keyword
"#ifndef" @keyword
"#include" @keyword
(preproc_directive) @keyword

"--" @operator
"-" @operator
"-=" @operator
"->" @operator
"!=" @operator
"*" @operator
"&" @operator
"&&" @operator
"+" @operator
"++" @operator
"+=" @operator
"<" @operator
"==" @operator
">" @operator
"||" @operator

"." @delimiter
";" @delimiter

(string_literal) @string
(system_lib_string) @string

(null) @constant
(number_literal) @constant
(char_literal) @constant

(field_identifier) @property

(type_identifier) @type
(primitive_type) @type
(sized_type_specifier) @type

(comment) @comment

(call_expression
  function: (identifier) @function)
(function_declarator
  declarator: (identifier) @function)
(preproc_function_def
  name: (identifier) @function.special)
```

bfredl:
@maxbrunsfeld I tested a bit more in nvim and the error is only visible when compiling with AddressSanitizer (even though the actual crash looks like a SEGV and not something directly detected by ASAN). I'll try and see if I can build the tree-sitter cli with ASAN (could perhaps be tricky with mixed rust/c executable).

maxbrunsfeld:
Ok, thanks for the update. We used to run the Tree-sitter test suite with ASAN, but unfortunately I gave up that benefit when I ported the CLI (and the test suite) from C++ to Rust. It would be great if it were possible to instrument a Rust executable with ASAN.

maxbrunsfeld:
Supposedly the feature is available when using nightly build of Rust: https://github.com/rust-lang/rust/pull/38699.

bfredl:
@maxbrunsfeld I'll create a small test program in C instead.

bfredl:
```c
#include <string.h>
#include <stdint.h>
#include <stdio.h>
#include "tree_sitter/api.h"

int main() {
  TSLanguage *tree_sitter_c(void), *lang = tree_sitter_c();
  ts_parser_set_language(parser, lang);

  const char *src = "int main() {\n/* \n}";

  TSTree *tree = ts_parser_parse_string(parser, NULL, src, strlen(src));

  const char *query_src = "(string_literal) @string (primitive_type) @type (comment) @comment";

  uint32_t error_offset;
  TSQueryError err;
  TSQuery *query = ts_query_new(lang, query_src, strlen(query_src), &error_offset, &err);
  if (!query) {
    printf("%u\n", error_offset);
    exit(1);
  }

  TSQueryCursor *cursor = ts_query_cursor_new();
  ts_query_cursor_exec(cursor, query, ts_tree_root_node(tree));

  TSQueryMatch match;
  uint32_t index;
  while (ts_query_cursor_next_capture(cursor, &match, &index)) {
    printf("match\n");
  }
}
```

This seem enough to reproduce (when compiling with ` clang -g -fsanitize=address,undefined crashtest.c lib/src/lib.c ../tree-sitter-c/src/parser.c -Ilib/include -I../utf8proc/ `

maxbrunsfeld:
@bfredl I got the Tree-sitter test suite running with `-Z sanitizer=address` and found a crash (which I believe is the crash you're seeing). Fixed on master.

maxbrunsfeld:
Confirmed that it fixes the crash in your sample program. Beautiful reproduction case by the way. Thank you very much for finding this issue!

bfredl:
Thank you for fixing it! BTW, what do you think about my second question? Is it possible to have a capture `@error` which matches any error?

maxbrunsfeld:
@bfredl That is not possible right now, but it should be. I think we should allow `ERROR` to be used just like any other node name, like this:

```clj
; match any syntax error
(ERROR) @error

; match only errors directly inside of statement blocks
(compound_statement (ERROR) @error-inside-of-block)
```

I will add that feature.

kkoomen:
I'm trying to get these queries to work with the node binary https://npmjs.com/package/tree-sitter but I do get errors:

```node
const Parser = require('tree-sitter');
const { Query } = require('tree-sitter');
const Python = require('tree-sitter-python');

const parser = new Parser();
parser.setLanguage(Python);

const sourceCode = `
def call(p1, p2):
    pass
`;



const tree = parser.parse(sourceCode);
console.log("tree.rootNode.toString() >>>>", tree.rootNode.toString());

const results = Query.prototype.captures(tree, 1)
console.log("results >>>>", results);
```

I know this is not the place to ask, but all the modules seem very inactive. If there's a place to ask question to get some help then I'd like to know.

PR #450, opened 2019-09-22
Fix some compiler warnings regarding function prototypes

bfredl:
I get these warnings when building latest tree-sitter (as part of neovim build system, which has enabled these warnings):

```
query.c:978:28: warning: implicit declaration of function 'ts_tree_cursor_current_status' is invalid in C99 [-Wimplicit-function-declaration]

api.h:741:35: warning: this function declaration is not a prototype [-Wstrict-prototypes]
TSQueryCursor *ts_query_cursor_new();
```


maxbrunsfeld:
Thanks!

PR #454, opened 2019-09-30
fix compiler warning for comparing `char` with `TSSymbolType'

bfredl:


maxbrunsfeld:
Wow, I don't know what I was thinking there. Thanks for the fix!

PR #462, opened 2019-10-14
docs: fix `lookahead` type

mkrupcale:
The `TSLexer` struct uses signed integer `int32_t` for the `lookahead` Unicode code point[1].

 * docs/section-3-creating-parsers.md: fix `lookahead` type

[1] https://github.com/tree-sitter/tree-sitter/blob/ee9a3c0ebb218990cf391ed987be7f2448c54a73/lib/include/tree_sitter/parser.h#L43


maxbrunsfeld:
Thanks!

PR #467, opened 2019-10-20
[WIP] lib: Add meson build system

mkrupcale:
### Introduction

This adds support for building using the meson meta-build system with ninja build backend. This makes building, installing, and using the tree-sitter library much simpler and faster. The tree-sitter library versions for the web and Rust bindings are also built using meson and ninja.

### Additional build dependencies

This requires some additional build dependencies:

 - `python3`
 - `meson`
 - `ninja`

### Build performance

There's a fairly significant (~70%) reduction in compile times (on my 4C/8T server) as a result of this change:

#### `master`

```shell
$ time (for i in $(seq 100); do
./script/build-lib
rm libtree-sitter.a
done)

real    8m51.558s
user    8m40.819s
sys     0m4.115s
```

#### `meson-build`

```shell
$ time (for i in $(seq 100); do
CC=gcc ./script/build-lib &> /dev/null
rm -rf build
done)

real    2m56.688s
user    9m21.832s
sys     0m30.126s
```

Furthermore, incremental rebuilds should be much faster during development.

### Usage

Meson can be configured to build a static library, shared library, or both libraries during the build using `-Ddefault_library={shared,static,both}`[1]. It is easy to modify the build type (e.g. `release`, `debug`, `debugoptimized`), `CFLAGS` (in addition to `-Dc_args=`), as well as install the libraries and public headers to the appropriate platform-specific location[2]. See `script/build-lib` for example usage as well as the quickstart guide[3] and the Meson reference manual[4] for details.

### Future work

This is not yet completed and requires:

 - [ ] Discussion and approval of the idea
 - [ ] Updated documentation
 - [ ] Shared library: discussion of exported symbols and soname/soversion

### References

1. [Meson - built-in options](https://mesonbuild.com/Builtin-options.html)
2. [Meson - installing](https://mesonbuild.com/Installing.html)
3. [Meson - quickstart guide](https://mesonbuild.com/Quick-guide.html)
4. [Meson - reference manual](https://mesonbuild.com/Reference-manual.html)


maxbrunsfeld:
👋 I don't think I'm interested in switching how Tree-sitter is built. The library is currently build-system-agnostic and is already trivial to integrate into any of the countless C build systems that exists (e.g. Node's [gyp](https://github.com/tree-sitter/node-tree-sitter/blob/355bb66699eb90f6391a9ad6a1027af8bb90f428/binding.gyp#L35-L49), Python's [`setup.py`](https://github.com/tree-sitter/py-tree-sitter/blob/7e3eb09b50d0a364076707b4cbb91c3028a81de4/setup.py#L29-L41), Rust's [`build.rs`](https://github.com/tree-sitter/tree-sitter/blob/49c632ae90e53dfa3f297031733b0cb733ead0b7/lib/binding_rust/build.rs#L37-L44), haskell's [Cabal](https://github.com/tree-sitter/haskell-tree-sitter/blob/7e9995c919467019801debee0e1cc45249357312/tree-sitter/tree-sitter.cabal#L23-L27), [CMake](https://github.com/JetBrains/jsitter/blob/9d651539ece649d676123b5a24b7eb240786c97f/native/CMakeLists.txt#L11-L15), etc). Compile time is already so fast that it's a non-issue (2.5 second clean build on my laptop). I don't see any advantage in adding build dependencies and tying it to a particular build system.

Thanks anyway for exploring the idea.

maxbrunsfeld:
Do you want to explain, in a general sense, how you're using this library, and which parts of the build you're finding difficult or restrictive?

mkrupcale:
As I already indicated, using tree-sitter as a shared library is essentially impossible right now. There is no ABI tracking through SONAME or similar platform-specific symbol versioning. This is a basic requirement for C libraries on Linux systems.

Then one must manually install all artifacts including public headers, which requires the user to delve into the internals of the library source tree. Essentially, anyone that wants to use this library will have to build their own build system from scratch every single time rather than using standard build systems. For Rust this may not matter, but for C libraries, this matters very much, especially for distributors/packagers.

maxbrunsfeld:
> As I already indicated, using tree-sitter as a shared library is essentially impossible right now.

Nope.

```sh
$CC $CFLAGS \
  -shared \
  -I lib/src \
  -I lib/include \
  lib/src/lib.c \
  -o libtree-sitter.so
```

dcreager:
> As I already indicated, using tree-sitter as a shared library is essentially impossible right now. There is no ABI tracking through SONAME or similar platform-specific symbol versioning.

I think a confusion here is that there aren't (to our knowledge) any users who are trying to install `libtree-sitter.so` as a single shared library into a system-managed location like `/usr/lib`, nor any distributions that have been considering putting together pre-built packages.  Instead, applications that have been using tree-sitter are using their existing plugin mechanisms to load it, with a copy of tree-sitter built specifically for that plugin.

mkrupcale:
> Nope.

You have just ignored everything I said about SONAME and ABI versioning. I find your dismissive tendencies to be extremely insulting.

> I think a confusion here is that there aren't (to our knowledge) any users who are trying to install `libtree-sitter.so` as a single shared library into a system-managed location like `/usr/lib`, nor any distributions that have been considering putting together pre-built packages.

Yes, that may be the case currently, but this may change. In any case, there are many reasons to prefer using shared libraries to static ones, and as adoption of tree-sitter increases, this will only become more important and not less. Shared libraries really must be considered first-class artifacts for tree-sitter library usage on Linux.

anthraxx:
@dcreager  hmm, looks like the day has come where someone considers packaging this as a shared system library for linux as its a dependency, and as a distro we don't want to have static libs compiled in random binaries but have control and proper deduplication x_X

maxbrunsfeld:
@anthraxx Are you finding it difficult to compile a Tree-sitter shared library for arch? Feel free to open new issues about specific problems you encounter.

anthraxx:
Well that not really an issue itself (while it would be optimal to have it properly maintained inside the repo itself). The main issue is the lack of ABI versioning, which is also maintained inside a project, infinite .0 sonames are quite a pita.

mkrupcale:
I'd also like to add that my comments on distribution packaging were not entirely hypothetical because I am in fact a packager for Fedora. As @anthraxx hinted at, this is basically un-packagable as is according to the Fedora packaging guidelines not only due to the lack of support for shared library ABI versioning but also lack of fine-grained control over the build flags (and no, `script/build-lib` does not suffice as a proper C library build system).

maxbrunsfeld:
Yes, so feel free to open separate issues about ABI versioning, as that doesn’t have anything to do with the Meson build system.

anthraxx:
@maxbrunsfeld  Sure, will that also consistently distribute header files accordingly or or that a second issue?

maxbrunsfeld:
I’d say just open one issue that describes all the troubles you’ve hit with packaging. Thanks!

SantiagoTorres:
I wanted to add that at this point it's also a chicken-and-egg type of situation, as downstream tools can't assume there's a reasonable library/header pair to build and link against. If I can't build a SO using reasonable tooling I can't ask downstream tools (e.g., radare2) to use them. It's also a little silly to ask each and every downstream packager to write their own toolchain to make a SO out of this codebase.

I personally don't understand the resistance to having tree-sitter provide one of the most common units of exchange when it comes to software.

maxbrunsfeld:
> provide one of the most common units of exchange when it comes to software.

@SantiagoTorres I can't tell what you're specifically requesting that we maintain. Can you clarify?

anthraxx:
I think the general idea of everyone requesting this: Something that properly builds this project automatically and distributes a final shared library plus required header files where they belong.
Ultimately its about you folks choosing one system you want to have, be it meson, cmake, autotools, whatever but please choose something 😸

FFY00:
Btw, this is causing the radare2 package on Arch Linux to be 6 releases behind.

We all are busy, figuring out how to properly build this and integrating that with radare2's build system is time-consuming.

Personally, I have already spent quite a bit of time trying to figure this out but I end up just giving up because I have other responsibilities and can't justify the time and effort invested here.

@anthraxx and @SantiagoTorres probably had the same experience

mkrupcale:
I think the problem here is that @maxbrunsfeld seems to feel that the so-called "build-system-agnostic" shell script or the "do-it-yourself" approach suffice as a means of building and distributing TS. Why burden him with maintaining a proper build system when he could offload that responsibility to all the downstream users?

The most obvious reason to me why you should not bother him with such requests--and the reason why I will no longer work on TS--is that he does not care about such things as shared libraries, ABI versioning, downstream consumption, and long-term maintenance. These are basic responsibilities of a C library maintainer, but I guess as long as it works for him and his envisioned use cases then all is fine.

FFY00:
CCing @radare @XVika @ret2libc

maxbrunsfeld:
Hey everyone, I would really *really* like to help address any problems that you're experiencing when  building or distributing Tree-sitter. With the comments on this thread, I'm having a hard time understanding the separate problems that each person is facing.

I have [asked](https://github.com/tree-sitter/tree-sitter/pull/467#issuecomment-568578930) repeatedly for people to open separate issues and to explain in more detail what roadblocks they are hitting, but nobody has done that yet. Just in case, I will ask again - **If you are having any problem building or distributing Tree-sitter, please open a new issue, and describe what you're trying to do, and why it's not working!**

The reason that this particular PR is a non-starter is that Tree-sitter is used in a very large number of downstream projects which all use different C build systems, and which all rely on the fact that they can trivially incorporate Tree-sitter into their build statically by adding [one source file and two include paths](https://github.com/tree-sitter/tree-sitter/blob/master/script/build-lib#L15-L17). This is by far the most popular way of using this library, and it will most likely remain as the *main* way, which is recommended in the docs and exercised by the project's CI build.

That doesn't mean I don't intend to support dynamic linking though! I would like to support the dynamic linking use case well too.

FFY00:
I don't have much time to reply right now but providing an alternative build system, like meson, does not break anything. You can have meson and still have people using the project the same way they did before.

From the top of my mind the issues are:
  - There's no way to build a dynamic library
  - There's no ABI version (this requires you to keep track of breaking changes and update the ABI when required, optimally you could adhere to a versioning scheme like semver)
  - Public headers should be distributed

FFY00:
On another note, I don't understand why you are trying to reinvent the wheel. Your problem is not new and already has a standard solution: pkgconfig. That's how basically any C build can make use of project. You just need to ship a .pc file (which meson can automatically generate) and them the various downstreams just query pkgconfig to get the required build flags.

Could you please provide me with a list of projects that use tree-sitter? I will have a look and tell if any is not compatible with this release method.

maxbrunsfeld:
> is that he does not care about such things as shared libraries, ABI versioning, downstream consumption, and long-term maintenance.

I do, in fact, care about downstream consumption and long term maintenance. That's why I have spent so much time helping people who want to use Tree-sitter in projects.

ABI versioning is actually something that I've put a ton of work into, in terms of ABI compatibility between the core library and the generated parsers, which can be generated by earlier versions of the CLI. There already is a versioning scheme that can be used at runtime to detect if libraries are compatible with parsers. 

At the same time, I'm aware that dynamic linking poses yet *another* set of compatibility issues. I *may* be up for taking on more responsibility in this area, but it does depend on what the benefits are. You haven't proposed anything that would help with ABI versioning in any way, so this topic is completely orthogonal to your PR.

> These are basic responsibilities of a C library maintainer

My basic responsibilities are to help develop and maintain a number of working applications that a lot of end users care about. This library helps my team to do that. It also also used to great effect by a lot of other developers in the open-source world! I don't think I agree that you are somehow the decided of what my responsibilities are.

maxbrunsfeld:
@FFY00 I'm not sure how I can make this more clear. Can you please open a separate issue and describe what actual problem you are trying to solve?

PR #473, opened 2019-10-25
fix invalid docs for ts_tree_get_changed_ranges

bfredl:
`ts_tree_get_changed_ranges` works very well but only if `old_tree` is passed as first parameter.

Also added a comment for a confusing initialization I stumbled upon while debugging this : cursor2 is used with "other" tree but is initialized with "self" tree, though this works as it is reinitialized later.


maxbrunsfeld:
Thanks!

PR #479, opened 2019-11-06
feat(cli): typescript definition

intellild:
This PR adds typescript definition file to improve edit experience (auto complete and bug checks). For example:

```js
// @ts-check

module.exports = grammar({
  name: "",
  extras: $ => [],
  externals: $ => [],
  conflicts: $ => [$.bar],
  inline: $ => [$.aaa],
  rules: {
    bar: $ => seq("foo"),
    foo: $ => $.foo,
    baz: $ => $.aaa
  }
});
```

```
Property 'aaa' does not exist on type 'Record<"bar" | "foo" | "baz", Rule>'. ts(2339) [8,19]
Property 'aaa' does not exist on type 'Record<"bar" | "foo" | "baz", Rule>'. ts(2339) [12,17]
```

### DIFF
--- a/cli/npm/global.d.ts
+++ b/cli/npm/global.d.ts
@@ -1,0 +1,44 @@
+type Rule =
+  | string
+  | RegExp
+  | {
+      /** @internal */
+      _ruleBrand: never;
+    };
+
+interface BaseGrammar<Keys extends string> {
+  name: string;
+
+  extras($: Record<Keys, Rule>): Rule[];
+
+  externals($: Record<Keys, Rule>): Rule[];
+
+  conflicts($: Record<Keys, Rule>): Rule[];
+
+  inline($: Record<Keys, Rule>): Rule[];
+
+  rules: Record<Keys, ($: Record<Keys, Rule>) => Rule>;
+}
+
+declare function blank(): Rule;
+declare function field(name: string, rule: Rule): Rule;
+declare function choice(...elements: Rule[]): Rule;
+declare function optional(rule: Rule): Rule;
+declare function prec(precedence: number, rule: Rule): Rule;
+declare namespace prec {
+  function left(precedence: number, rule: Rule): Rule;
+  function right(precedence: number, rule: Rule): Rule;
+  function dynamic(precedence: number, rule: Rule): Rule;
+}
+declare function repeat(rule: Rule): Rule;
+declare function repeat1(rule: Rule): Rule;
+declare function seq(...elements: Rule[]): Rule;
+declare function sym(name: string): Rule;
+declare function token(value: string | RegExp): Rule;
+declare namespace token {
+  function immediate(value: string | RegExp): Rule;
+}
+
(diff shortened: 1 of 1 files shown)


cli/npm/global.d.ts
```diff
@@ -0,0 +1,44 @@
...
+      /** @internal */
+      _ruleBrand: never;
+    };
+
+interface BaseGrammar<Keys extends string> {
+  name: string;
+
+  extras($: Record<Keys, Rule>): Rule[];
+
+  externals($: Record<Keys, Rule>): Rule[];
```
SKalt: Checking https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/grammar-schema.json, only `rules` and `name` are required.

cli/npm/global.d.ts
```diff
@@ -0,0 +1,44 @@
+type Rule =
+  | string
+  | RegExp
+  | {
+      /** @internal */
+      _ruleBrand: never;
```
SKalt: What's does the internal `_ruleBrand` do?

cli/npm/global.d.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+interface BaseGrammar<Keys extends string> {
+  name: string;
+
+  extras($: Record<Keys, Rule>): Rule[];
+
+  externals($: Record<Keys, Rule>): Rule[];
+
+  conflicts($: Record<Keys, Rule>): Rule[];
```
SKalt: According to https://tree-sitter.github.io/tree-sitter/creating-parsers#the-grammar-dsl, `conflicts` should return `Array<Array<Rule>>`.

maxbrunsfeld:
👋 @SKalt @intellild I'm open to including this in the npm module, but I don't use TypeScript myself. I think we'd just need to make sure that all of the grammars in [the Tree-sitter org](https://github.com/tree-sitter) type check without error, before merging this.

From what I can tell, @SKalt's suggestions are correct.

cli/npm/global.d.ts
intellild: I  am not implementing details here, for  it seems to be unnecessary. So I make use of nominal typing to prevent mistakes. Typescript does not provide nominal typing and this is a trick to get something similar with nominal typing. You can refer https://spin.atomicobject.com/2018/01/15/typescript-flexible-nominal-typing/

SKalt:
The currently-failing test (`tests::highlight_test::test_highlighting_with_content_children_included`) doesn't seem related to these changes.

jacobwgillespie:
Hello all, I spent some time today creating a `d.ts` file for the DSL for my own use, and afterwards discovered this PR.

Since the declaration file I'm using covers a few more features of the API and of TypeScript, I have opened a new PR to contribute (#658). It could probably replace this PR if so desired, or its ideas could be incorporated here. Let me know if I can help either way!

PR #489, opened 2019-11-18
Properly handle productions that contain named aliases

tclem:
This fixes the last failing test over on https://github.com/tree-sitter/haskell-tree-sitter/pull/241 which had to do with a python example like this (call expression with parenthesized list splat as an argument):

```python
i((*a))
```

which has this tree:

```
(module [0, 0] - [1, 0]
  (expression_statement [0, 0] - [0, 7]
    (call [0, 0] - [0, 7]
      function: (identifier [0, 0] - [0, 1])
      arguments: (argument_list [0, 1] - [0, 7]
        (parenthesized_expression [0, 2] - [0, 6]
          (list_splat [0, 3] - [0, 5]
            (identifier [0, 4] - [0, 5])))))))
```

The trouble comes from node-types.json which doesn't get properly generated for [an alias](https://github.com/tree-sitter/tree-sitter-python/blob/225e757b4580ecc2b3da36d13679188b7c8abe4e/grammar.js#L421) within the `argument_list` production:
```
alias($.parenthesized_list_splat, $.parenthesized_expression)
```

[`parenthesized_expression` only has `_expression` and `yield`](https://github.com/tree-sitter/tree-sitter-python/blob/225e757b4580ecc2b3da36d13679188b7c8abe4e/src/node-types.json#L1753-L1770), but should also include all the children of `parenthesized_list_splat`.

``` json
{
    "type": "parenthesized_expression",
    "named": true,
    "fields": {},
    "children": {
      "multiple": false,
      "required": true,
      "types": [
        {
          "type": "_expression",
          "named": true
        },
        {
          "type": "yield",
          "named": true
        }
      ]
    }
  }
```

I wrote a test that demonstrates the problem (could probably be a bit more concise, but it was tricky to figure out what exactly was going wrong) and then did some debugging to figure out where this information was getting lost.

cc @maxbrunsfeld

### DIFF
--- a/cli/src/generate/node_types.rs
+++ b/cli/src/generate/node_types.rs
@@ -567,6 +567,8 @@ pub(crate) fn generate_node_types_json(
         } else if variable.kind.is_visible()
             && !syntax_grammar.variables_to_inline.contains(&symbol)
         {
+            // If a rule is aliased under multiple names, then its information
+            // contributes to multiple entries in the final JSON.
             for alias in aliases_by_symbol
                 .get(&Symbol::non_terminal(i))
                 .unwrap_or(&HashSet::new())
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
The Travis failure on https://github.com/tree-sitter/tree-sitter/pull/489/commits/e2325102d14e28f982208919f21d3f779459167c was just a spurious failure in Travis's caching system; that commit should be green.

PR #490, opened 2019-11-20
Fix a few compiler warnings

damieng:
Two instances of

`warning: comparison between signed and unsigned integer expressions`

And one

`warning: value computed is not used`


maxbrunsfeld:
Awesome!

PR #498, opened 2019-12-04
SyntaxNode#isNamed is a function

jrieken:
The d.ts-file declared `isNamed` as boolean but it is a function returning a boolean.


maxbrunsfeld:
Sorry for the delay. This was fixed in #622.

PR #521, opened 2020-01-10
removed duplicate key in grammar-schema.json

SKalt:
The key "required" was duplicated on "symbol-rule". I removed the more permissive copy.


maxbrunsfeld:
Thanks!

Issue #523, opened 2020-01-12
grammar-schema.json does not allow `type: "FIELD"`


SKalt:
[tree-sitter-bash's grammar.json](https://github.com/tree-sitter/tree-sitter-bash/blob/master/src/grammar.json) has many nodes like 
```
{
  "type": "FIELD",
  "name": "body",
  "content": {
    "type": "SYMBOL",
    "name": "_statement"
  }
}
```
Nothing in to [grammar-schema.json](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/grammar-schema.json) allows for `type: "FIELD"`. Should I add extend the definition of `rule` to include
```JSON
{
  "field-rule": {
    "required": ["name", "type", "content"],
    "properties": {
      "name": { "type": "string" },
      "type": {
        "type": "string",
        "pattern": "^FIELD$"
      },
      "content": {
        "$ref": "#/definitions/rule"
      }
    }
  }
}
```
?

maxbrunsfeld:
Yes!

PR #525, opened 2020-01-13
added field-rule to grammar-schema.json

SKalt:
Addresses #523.
Requesting review: is there any order the rule definitions should be in?  Are there any tests involving the schema file that I should pay attention to?


SKalt:
Apologies: looks like my IDE `prettier`'d the JSON.

PR #537, opened 2020-02-12
Syntax highlighting primer from PRs

kramred:
A first shot to piece together the current status of the CLI highlight subcommand (but also the API) from PRs mentioning highlight that I could find - feel free to rewrite or drop whatever you see fit; there might be errors.
I'd really appreciate some more/centralized documentation on the syntax highlighting, hope this can serve as a starting point

PR #538, opened 2020-02-12
Syntax highlighting docs

maxbrunsfeld:
This PR adds a separate docs page that's all about syntax highlighting. I think this is needed now that GitHub is using Tree-sitter for highlighting. Once this is done, I think we should link to it from some GH or Linguist docs.

* 'creating parsers' page
  * [x] section about `highlight` command
  * [x] in `test` command section, mention new highlight tests
* 'syntax highlighting' page
  * [x] section: overview
  * [x] section: highlights query
  * [x] section: locals query
  * [x] section: injections query
  * [x] section: testing

See also https://github.com/tree-sitter/tree-sitter/pull/537


maxbrunsfeld:
/cc @bfredl @ubolonton - Just a heads up that there are finally docs for the highlighting system.

http://tree-sitter.github.io/tree-sitter/syntax-highlighting

Issue #524, opened 2020-01-13
Document supertypes in the parser creation page and in grammar-schema.json


SKalt:
Looks like supertypes defined in [the JS DSL](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/dsl.js#L344) are missing from the [grammar creation docs](https://github.com/tree-sitter/tree-sitter/blob/master/docs/section-3-creating-parsers.md) and [grammar-schema.json](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/grammar-schema.json) as well.

Creating a separate ticket from the inspiration, #393, since that was a question and these are bug/doc fixes.

maxbrunsfeld:
Fixed via #542.

PR #542, opened 2020-02-15
feat(cli): documented optional supertypes string[] in grammar schema

SKalt:
Addressed #524 .

drive-by fixes: I also added a `title` key and a `$schema` key to pin the JSON schema version.

Before merge, reviews should ascertain
- I've entered the correct type definition for `supertypes` and 
-  the `description` of `supertypes` correctly describes what `supertypes` does.

### DIFF
--- a/cli/src/generate/grammar-schema.json
+++ b/cli/src/generate/grammar-schema.json
@@ -1,4 +1,6 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "tree-sitter grammar specification",
   "type": "object",
 
   "required": ["name", "rules"],
@@ -7,6 +9,7 @@
 
   "properties": {
     "name": {
+      "description": "the name of the grammar",
       "type": "string",
       "pattern": "^[a-zA-Z_]\\w*"
     },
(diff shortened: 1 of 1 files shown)


cli/src/generate/grammar-schema.json
```diff
@@ -57,6 +60,15 @@
     "word": {
       "type": "string",
       "pattern": "^[a-zA-Z_]\\w*"
+    },
+
+    "supertypes": {
+      "description": "each supertype instructs the parser to replace a set of subtypes in the `types` list with a single supertype.",
```
maxbrunsfeld:
```suggestion
      "description": "A list of hidden rule names that should be considered supertypes in the generated node types file. See http://tree-sitter.github.io/tree-sitter/using-parsers#static-node-types.",
```

cli/src/generate/grammar-schema.json
```diff
@@ -57,6 +60,15 @@
...
+      "type": "array",
+      "items": {
+        "description": "the name of a rule in `rules` or `extras`",
```
maxbrunsfeld: I would just remove this description. I don't think it makes sense to put substantial documentation directly into this grammar schema file, since most users won't look at this. I think that if anything, there should be a few links to the main documentation site. I suggested putting one link in the description of the parent object.

SKalt: I practically agreed; I've merged your changes.
I feel that the schema should be the source of truth about the JSON grammar structure. However, I don't feel strongly enough to automate making legible docs out of the schema.  
Thanks for the PR review!

Issue #543, opened 2020-02-16
Compiling a tree-sitter parser that defines cpp source will break on linking a rust object file due to name mangling


alyssaverkade:
Hello!

I should start off by saying that I have a potential fix for this, but I wanted to make sure my approach is correct before I submitted a PR!

#### Repro
I was trying to use the `tree-sitter-ruby` parser in a rust program on MacOS `10.14.6` with the following contents using `cc` version `1.0.50` and `tree-sitter` version `0.6.1` (and clang `1100.0.33.12`):

`git submodule add --depth 1 https://gitub.com/tree-sitter/tree-sitter-ruby`
`cd tree-sitter-ruby && npm i && cd -`

```rust
// build.rs
use std::path::PathBuf;

fn main() {
    let dir: PathBuf = ["tree-sitter-ruby", "src"].iter().collect();

    cc::Build::new()
        .include(&dir)
        .file(dir.join("parser.c"))
        .file(dir.join("scanner.cc"))
        .compile("treesitter-ruby");
}

```

```rust
// main.rs
use tree_sitter::{Language, Parser};

extern "C" {
    fn tree_sitter_ruby() -> Language;
}

fn main() {
    let src = "def foo; end";
    let mut parser = Parser::new();
    let ruby = unsafe { tree_sitter_ruby() };
    parser.set_language(ruby).unwrap();
    let tree = parser.parse(src, None).unwrap();
    println!("{:?}", tree);
}
```

Compiling this spawns the cthulhu of c-based ffi:
```
 Undefined symbols for architecture x86_64:
            "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__grow_by(unsigned long, unsigned long, unsigned long, unsigned long, unsigned long, unsigned long)", referenced from:
                std::__1::enable_if<(__is_forward_iterator<char const*>::value) && (__libcpp_string_gets_noexcept_iterator<char const*>::value), std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&>::type std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::assign<char const*>(char const*, char const*) in libtreesitter-ruby.a(scanner.o)
            "___cxa_allocate_exception", referenced from:
                std::__1::__throw_length_error(char const*) in libtreesitter-ruby.a(scanner.o)
# ...
 NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
[...]

PR #548, opened 2020-02-19
Add a language linkage declaration to parsers

alyssaverkade:
Submitting after discussion in #543 

I looked at the `parser.c` files of the `go` and `c` parsers after regenerating to see if this PR would handle parsers that did not contain C++ sources, and the output looked correct to me. Coupling that with all the tests passing (on my machine) I have enough confidence in this working to submit this, but I've been wrong before!

Previously, in order to compile a `tree-sitter` grammar that contained
c++ source in the parser (ie the `scanner.cc` file), you would have to
compile the `parser.c` file separately from the c++ source. For example,
in rust projects this would result in a `build.rs` close to the following:
```rust
extern crate cc;

fn main() {
  let dir: PathBuf = ["tree-sitter-ruby", "src"].iter().collect();

  cc::Build::new()
    .include(&dir)
    .cpp(true)
    .file(dir.join("scanner.cc"))
    // NOTE: must have a name that differs from the c static lib
    .compile("tree-sitter-ruby-scanner");

  cc::Build::new()
    .include(&dir)
    .file(dir.join("parser.c"))
    // NOTE: must have a name that differs from the c++ static lib
    .compile("tree-sitter-ruby-parser");
}
```

This was necessary at the time for the following grammars: `ruby`,
`php`, `python`, `embedded-template`, `html`, `cpp`, `ocaml`,
`bash`, `agda`, and `haskell`.

To solve this, we specify an `extern "C"` language linkage declaration
to the functions that must be linked against to compile a parser with the
scanner, making parsers linkable against c++ translation units.
 On all major compilers (gcc, clang, and msvc) this should be the only change
 needed due to the combination of clang and gcc both supporting designated
 initialization for years and msvc 2019 adopting designated initializers as a part of the C++20
conformance push.

Subsequently, for rust projects, the necessary `build.rs` would become
(which also brings these parsers into sync with the current docs):
```rust
extern crate cc;

fn main() {
  let dir: PathBuf = ["tree-sitter-ruby", "src"].iter().collect();

  cc::Build::new()
    .include(&dir)
    .cpp(true)
    .file(dir.join("scanner.cc"))
    .file(dir.join("parser.c"))
    .compile("tree-sitter-ruby");
}
```

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -1058,6 +1058,10 @@ impl Generator {
         let language_function_name = format!("tree_sitter_{}", self.language_name);
         let external_scanner_name = format!("{}_external_scanner", language_function_name);
 
+        add_line!(self, "#ifdef __cplusplus");
+        add_line!(self, r#"extern "C" {{"#);
+        add_line!(self, "#endif");
+
         if !self.syntax_grammar.external_tokens.is_empty() {
             add_line!(self, "void *{}_create(void);", external_scanner_name);
             add_line!(self, "void {}_destroy(void *);", external_scanner_name);
@@ -1183,6 +1187,9 @@ impl Generator {
         add_line!(self, "return &language;");
         dedent!(self);
         add_line!(self, "}}");
+        add_line!(self, "#ifdef __cplusplus");
+        add_line!(self, "}}");
+        add_line!(self, "#endif");
     }
 
     fn get_parse_action_list_id(


maxbrunsfeld:
Thanks!

maxbrunsfeld:
I want to thank you again for this work @alyssaverkade. It has turned out to be beneficial to me already, because recently Emscripten has lost the ability to be able to compile C and C++ together in one invocation. So it is very useful to be able to compile `parser.c` in C++ mode.

PR #550, opened 2020-02-20
Output also HTML_FOOTER with highlight --html

kramred:
Currently `HTML_FOOTER` is being defined but not output. This is a suggestion to add the output of `HTML_FOOTER` right after calling `highlight::html`


cli/src/main.rs
```diff
@@ -279,6 +279,7 @@ fn run() -> error::Result<()> {
             if let Some(highlight_config) = language_config.highlight_config(language)? {
                 if html_mode {
                     highlight::html(&loader, &config.theme, &source, highlight_config, time)?;
+                    println!("{}", highlight::HTML_FOOTER);
```
maxbrunsfeld: This line gets executed for each path that's passed in, so we could end up with multiple `</body>` lines. I think we just need to move it after this `for` loop.

kramred: Yes, good catch. I'll edit the PR.

maxbrunsfeld:
Thanks!

PR #551, opened 2020-02-20
Flag to output only a html snippet

kramred:
Currently a full standalone HTML page is created, this is a suggestion to add a flag where you can leave out the head and footer part.

`-s` might not be the best option as e.g. [pandoc](https://pandoc.org/) outputs HTML snippets by default and uses `-s` or [`--standalone`](https://pandoc.org/MANUAL.html#option--standalone) – thus, `-s` would be creating the opposite effect here vs. pandoc


amaanq:
Superseded by #3879 I believe

PR #561, opened 2020-02-27
Fix out-of-bounds array access in `Language` APIs

ubolonton:
I found these while working on the Emacs Lisp's binding.


maxbrunsfeld:
Thanks!

PR #565, opened 2020-03-04
add eslint for web bindings

benlongo:
Linting `prefix.js` and `binding.js` is somewhat awkward, but I don't think there's a better way to do it. 

I chose to create a local plugin to provide eslint the emscripten runtime variables as we could potentially rip that out in the future for other people to use. It would certainly be simpler to list the globals in a comment, so can do that instead of the complexity isn't worth it.


lib/binding_web/prefix.js
```diff
@@ -1,6 +1,6 @@
 (function (root, factory) {
-  if (typeof define === 'function' && define.amd) {
-    define([], factory);
+  if (typeof define === 'function' && define.amd) { 
+    define([], factory);
```
Razzeee: I don't think you meant to include those whitespaces

benlongo: Good catch 👍

benlongo:
This would be superseded by #569

PR #570, opened 2020-03-07
Output an ES6 module with emscripten and bundle it with rollup.

benlongo:
This is the first stage from #569. 

There are a couple additional changes I'd like to point out:

- The readme for the web bindings suggests that `tree-sitter.js` and `tree-sitter.wasm` should be included in the releases, however `.travis.yml` appears to be pointing to the wrong place. Looking at the current releases I don't see these files so I've adjusted the paths. If we want those files to end up in `target/release` it would be easy to copy them there in `script/build-wasm`.
- `script/build-wasm` is now also running `rollup -c`. I think this makes sense because it preserves the same invariant of having the final artifacts exist after running it. 

Given that `tree-sitter web-ui` works, I believe this works in the browser exactly as before, but I'm not positive.


.travis.yml
```diff
@@ -21,6 +21,7 @@ before_install:
 
 script:
   # Build the WASM binding
+  - npm install --prefix lib/binding_web
```
benlongo: `script/build-wasm` requires rollup to be present so we need to run this first, I suspect some docs also need to be updated.

benlongo: we could also have `script/build-wasm` run `npm install` but that seems somewhat annoying

maxbrunsfeld:
Thanks for extracting this out. It's great to have syntactically valid source files   instead of the partial files that we had with `--pre-js` and `--post-js`.

I'm actually not understanding why Rollup is needed though. Is it just to convert the ES6 module to a UMD module? Because it looks as if Emscripten will actually output a UMD module directly, if you *don't* select `EXPORT_ES6`:

https://github.com/emscripten-core/emscripten/blob/eee2723cf2c62ac9ffae3629944ef92be2faa91a/emcc.py#L3426-L3434

benlongo:
Rollup is required to combine the output of emscripten with the binding code that was extracted from the `--post-js` file. The UMD module from emscripten only contains the `Module` definition if I understand correctly. Without rollup we would have two files instead of just one.

benlongo:
Rollup just hit [2.0 today ](https://github.com/rollup/rollup/releases/tag/v2.0.0), I'll look into updating that on this branch.

maxbrunsfeld:
> Rollup is required to combine the output of emscripten with the binding code that was extracted from the --post-js file.

No, a major part of Emscripten's functionality is to optimize your own hand-written JavaScript code *together* with its own generated code. Emscripten needs to be in charge of combining the binding code with the generated code; it shouldn't be handled after-the-fact via Rollup.

It's true that `--post-js` may be removed, but `--pre-js` is not. I think that `--pre-js` is the normal way of passing JavaScript input to Emscripten.

benlongo:
I don't believe emscripten does anything to the current contents of `--post-js`. To verify this, I compared the output of emscripten on master using `--pre-js` and `--post-js` with the output of rollup on `modularize-web`. I don't see any real differences in the output. 

Here are the steps I followed:
 
1. `git checkout master`
1. Disable terser in `script/build-wasm`
2. `script/build-wasm && cp lib/binding_web/tree-sitter.js ~/onmaster.js`
3. `git checkout modularize-web`
4. `script/build-wasm`
4. `cd lib/binding_web && npm run build -- --environment BUILD:development` to disable terser
5. `cp lib/tree-sitter.js ~/onrollup.js`

Now in `~` we have `onmaster.js` and `onrollup.js`. 

1. `prettier onrollup.js | tail -n +2009 | awk 'NF' | sed 's/Module\.//g' > rollup-clean.js`
1. `prettier onmaster.js| tail -n +2013  | awk 'NF' | sed 's/C\.//g' > master-clean.js`
1. `git diff -wb master-clean.js rollup-clean.js` (ignore whitespace)

2009 and 2013 are the lines where `const INTERNAL = {};` is. 

```
diff --git a/master-clean.js b/rollup-clean.js
index fe5c5a0..fcd37a2 100644
--- a/master-clean.js
+++ b/rollup-clean.js
@@ -11,14 +11,15 @@
   var VERSION;
   var MIN_COMPATIBLE_VERSION;
   var TRANSFER_BUFFER;
-  var currentParseCallback;
-  var currentLogCallback;
   var initPromise = new Promise(resolve => {
-    Module.onRuntimeInitialized = resolve;
+    onRuntimeInitialized = resolve;
   }).then(() => {
     TRANSFER_BUFFER = _ts_init();
     VERSION = getValue(TRANSFER_BUFFER, "i32");
-    MIN_COMPATIBLE_VERSION = getValue(TRANSFER_BUFFER + SIZE_OF_INT, "i32");
+    MIN_COMPATIBLE_VERSION = getValue(
+      TRANSFER_BUFFER + SIZE_OF_INT,
+      "i32"
+    );
   });
   class Parser {
     static init() {
@@ -306,6 +307,7 @@
       if (!Array.isArray(types)) types = [types];
       if (!startPosition) startPosition = ZERO_POINT;
       if (!endPosition) endPosition = ZERO_POINT;
+      // Convert the type strings to numeric type symbols.
       const symbols = [];
       const typesBySymbol = this.tree.language.types;
       for (let i = 0, n = typesBySymbol.length; i < n; i++) {
@@ -313,10 +315,12 @@
           symbols.push(i);
         }
       }
+      // Copy the array of symbols to the WASM heap.
       const symbolsAddress = _malloc(SIZE_OF_INT * symbols.count);
       for (let i = 0, n = symbols.length; i < n; i++) {
         setValue(symbolsAddress + i * SIZE_OF_INT, symbols[i], "i32");
       }
+      // Call the C API to compute the descendants.
       marshalNode(this);
       _ts_node_descendants_of_type_wasm(
         this.tree[0],
@@ -327,8 +331,12 @@
         endPosition.row,
         endPosition.column
       );
+      // Instantiate the nodes based on the data returned.
       const descendantCount = getValue(TRANSFER_BUFFER, "i32");
-      const descendantAddress = getValue(TRANSFER_BUFFER + SIZE_OF_INT, "i32");
+      const descendantAddress = getValue(
+        TRANSFER_BUFFER + SIZE_OF_INT,
+        "i32"
+      );
       const result = new Array(descendantCount);
       if (descendantCount > 0) {
         let address = descendantAddress;
@@ -337,6 +345,7 @@
           address += SIZE_OF_NODE;
         }
       }
+      // Free the intermediate buffers
       _free(descendantAddress);
       _free(symbolsAddress);
       return result;
@@ -452,11 +461,15 @@
     }
     get nodeIsNamed() {
       marshalTreeCursor(this);
-      return _ts_tree_cursor_current_node_is_named_wasm(this.tree[0]) === 1;
+      return (
+        _ts_tree_cursor_current_node_is_named_wasm(this.tree[0]) === 1
+      );
     }
     get nodeIsMissing() {
       marshalTreeCursor(this);
-      return _ts_tree_cursor_current_node_is_missing_wasm(this.tree[0]) === 1;
+      return (
+        _ts_tree_cursor_current_node_is_missing_wasm(this.tree[0]) === 1
+      );
     }
     get nodeText() {
       marshalTreeCursor(this);
@@ -502,7 +515,9 @@
     }
     gotoNextSibling() {
       marshalTreeCursor(this);
-      const result = _ts_tree_cursor_goto_next_sibling_wasm(this.tree[0]);
+      const result = _ts_tree_cursor_goto_next_sibling_wasm(
+        this.tree[0]
+      );
       unmarshalTreeCursor(this);
       return result === 1;
     }
@@ -520,7 +535,9 @@
       this.types = new Array(_ts_language_symbol_count(this[0]));
       for (let i = 0, n = this.types.length; i < n; i++) {
         if (_ts_language_symbol_type(this[0], i) < 2) {
-          this.types[i] = UTF8ToString(_ts_language_symbol_name(this[0], i));
+          this.types[i] = UTF8ToString(
+            _ts_language_symbol_name(this[0], i)
+          );
         }
       }
       this.fields = new Array(_ts_language_field_count(this[0]) + 1);
@@ -827,7 +844,10 @@
         endPosition.column
       );
       const count = getValue(TRANSFER_BUFFER, "i32");
-      const startAddress = getValue(TRANSFER_BUFFER + SIZE_OF_INT, "i32");
+      const startAddress = getValue(
+        TRANSFER_BUFFER + SIZE_OF_INT,
+        "i32"
+      );
       const result = new Array(count);
       let address = startAddress;
       for (let i = 0; i < count; i++) {
@@ -838,7 +858,7 @@
         const captures = new Array(captureCount);
         address = unmarshalCaptures(this, node.tree, address, captures);
         if (this.predicates[pattern].every(p => p(captures))) {
-          result[i] = { pattern: pattern, captures: captures };
+          result[i] = { pattern, captures };
           const setProperties = this.setProperties[pattern];
           if (setProperties) result[i].setProperties = setProperties;
           const assertedProperties = this.assertedProperties[pattern];
@@ -865,7 +885,10 @@
         endPosition.column
       );
       const count = getValue(TRANSFER_BUFFER, "i32");
-      const startAddress = getValue(TRANSFER_BUFFER + SIZE_OF_INT, "i32");
+      const startAddress = getValue(
+        TRANSFER_BUFFER + SIZE_OF_INT,
+        "i32"
+      );
       const result = [];
       const captures = [];
       let address = startAddress;
@@ -918,7 +941,7 @@
       address += SIZE_OF_INT;
       const node = unmarshalNode(tree, address);
       address += SIZE_OF_NODE;
-      result[i] = { name: query.captureNames[captureIndex], node: node };
+      result[i] = { name: query.captureNames[captureIndex], node };
     }
     return address;
   }
@@ -956,7 +979,7 @@
     const result = new Node(INTERNAL, tree);
     result.id = id;
     result.startIndex = index;
-    result.startPosition = { row: row, column: column };
+    result.startPosition = { row, column };
     result[0] = other;
     return result;
   }
```

Based on this diff, the two files are pretty much identical barring some minor syntactical differences.

benlongo:
I did a similar experiment with `--pre-js` and got the same result - I'm fairly confused about why it's not doing anything.

benlongo:
@maxbrunsfeld given that emscripten isn't touching the code do we still want to pass it with `--pre/post-js`?

Menci:
How is it going?

lerencao:
It would be nice to see this happen. :)

maxbrunsfeld:
Interesting; it's good to know that Emscripten isn't currently doing any optimization there. Thanks for investigating that. Maybe it depends on what flags you pass to `emcc`.

Still, I don't see any advantage in *avoiding* Emscripten's standard feature for this (`--post-js`) and adding a separate JS dependency instead. Depending on how the code evolves, Emscripten might improve its C/JS integrated optimization. And in general, I'd prefer fewer dependencies. Are there advantages of using Rollup that I'm missing?

Menci:
So how is it going now?

maxbrunsfeld:
Ok, we never quite got on the same page about this because I was reluctant to add a dependency on Rollup when it seemed like Emscripten was capable of doing the bundling by itself.

In https://github.com/tree-sitter/tree-sitter/pull/822, I did adjust the build process so that the main JavaScript file, `binding.js`, is now a syntactically valid module, instead of needing to be concatenated with other code in order to parse. I didn't massively restructure the build, but just added a separate *suffix* file that adds the final return statement and the closing delimiters.

I'm going to close this out, because I think the main goal - allowing JavaScript-based tools to operate on `binding.js` has been achieved.

PR #583, opened 2020-03-20
Add a 'tags' crate, for computing ctags-style code navigation tags

maxbrunsfeld:
### Motivation

We're exploring ways to make GitHub's [code navigation](https://help.github.com/en/github/managing-files-in-a-repository/navigating-code-on-github#about-navigating-code-on-github) features more easily extensible by external contributors. Now that we've shipped our new [syntax highlighting system](https://tree-sitter.github.io/tree-sitter/syntax-highlighting) based on [tree queries](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries), we think that we may be able to use a similar system to drive our "fuzzy" code navigation data as well.

This PR adds a new sub-crate to the project called `tree-sitter-tags`, which provides functionality for computing [ctags](https://ctags.io/)-style code navigation data based on tree queries.

### API

Like the `tree-sitter-highlight` crate, `tree-sitter-tags` has two build targets: a rust library, and a static C library. The Rust library is consumed by a new subcommand in the CLI: `tree-sitter tags`. Right now, this subcommand is similar to the `query` subcommand, in that it's mainly useful for testing debugging.

The C library is designed to easily be linked into a project written in another programming language. Its API is declared in [`tags.h`](https://github.com/tree-sitter/tree-sitter/blob/9665f1ba70765719b2a5871e0feacbff2db2d1f2/tags/include/tree_sitter/tags.h).

### Query Format

The idea is that, to enable tagging, you would add one new query to a tree-sitter grammar repository: `queries/tags.scm`. This query has a pretty simple structure. The patterns should have the following captures:

* `@name` (required) - a node containing a symbol name
* `@doc` (optional) - a node whose text contains the documentation for a symbol. Multiple nodes can be captured in cases where the documentation is spread across multiple nodes.
* Any one of the following:
  * `@call` - a node that represents a function or method call
  * `@method` - a node that represents a method definition
  * `@function` - a node that represents a function definition
  * `@class` a node that represents a class definition
  * `@module` a node that represents a module definition

Patterns can also have the following predicates:

* `(strip! @docs "some-regex")` - This allows certain parts of the `@docs` node's text to be removed from the captured documentation.
* `(set-adjacent! @docs @some-node)` - Specifies that docs should only be taken from nodes that are on adjacent lines as `@some-node`.

### Fundamental Query API Improvements

As part of this work, we needed to slightly increase the power of the tree query language. We have added a `+` operator that can be used to capture an arbitrary number of repeating nodes. Examples:

You can match a class declaration and capture one or more preceding comments. This is the use case we found in the `tree-sitter-tags` crate.

```
(*
  (comment)+ @doc  .
  (class_declaration
    name: (identifier) @name))
```

But repetition also works for nodes with children. For example, you can match a method definition and and capture one or more decorators:

```
(method_definition
  (decorator (identifier) @deco)+
  name: (property_identifier) @name)
```

### Future Work

Now that there is a `+` operator in queries, which is analogous to `+` in regular expressions, it might make sense to make the `*` operator work like it does in regexes too. But the `*` is currently used as a "wildcard". It might be worth revisiting the query syntax as a whole, to try and make it as familiar and consistent as possible. I think that we can defer this to a later PR.

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -248,6 +248,9 @@ static CaptureListPool capture_list_pool_new() {
 
 static void capture_list_pool_reset(CaptureListPool *self) {
   self->usage_map = UINT32_MAX;
+  for (unsigned i = 0; i < 32; i++) {
+    array_clear(&self->list[i]);
+  }
 }
 
 static void capture_list_pool_delete(CaptureListPool *self) {
--- a/tags/src/c_lib.rs
+++ b/tags/src/c_lib.rs
@@ -143,6 +143,8 @@ pub extern "C" fn ts_tagger_tag(
             let tag = if let Ok(tag) = tag {
                 tag
             } else {
+                buffer.tags.clear();
+                buffer.docs.clear();
                 return TSTagsError::Timeout;
             };
 
--- a/tags/src/lib.rs
+++ b/tags/src/lib.rs
@@ -212,6 +212,7 @@ impl TagsContext {
         self.parser
             .set_language(config.language)
             .map_err(|_| Error::InvalidLanguage)?;
+        self.parser.reset();
         unsafe { self.parser.set_cancellation_flag(cancellation_flag) };
         let tree = self.parser.parse(source, None).ok_or(Error::Cancelled)?;
 


Razzeee:
Will there be bindings for JS similar to how tree sitter works at the moment and will there be docs on this?

maxbrunsfeld:
We'll write docs for this soon, but currently we're still changing the interface as we integrate this into github.com.

maxbrunsfeld:
We probably won't include this as part of the wasm library, but the actual code that implements this is pretty small: https://github.com/tree-sitter/tree-sitter/blob/master/tags/src/lib.rs. So once we stabilize and document the interface, I think it would be easy to implement an equivalent system in JavaScript, or another language.

PR #592, opened 2020-04-11
Fix exporting get/set timeout in web binding

andy0130tw:
Add two symbols "_ts_parser_set_timeout_micros", "_ts_parser_timeout_micros" due to usage in `tree-sitter.js`. Also the `getTimeoutMicros` function was not working, also fixed it.

The fix was trivial enough so I did not add tests.


maxbrunsfeld:
Thank you!

PR #602, opened 2020-04-21
Add a simple Makefile-based build system.

### LINKED ISSUE #581: RFC: provide a build system for Linux distribution integration
Linux distributions would like to provide a system libtree-sitter.so and headers in /usr/include, sufficient to build and link against the tree-sitter software. This would be available as a mutually inclusive alternative to the currently documented way to use tree-sitter, which is to copy the project wholesale as a vendored subdirectory inside another project's git sources, and manually integrate it into that other project's build system.

Requirements for distribution packaging:
- respect standard conventions such as `$CFLAGS` and `$LDFLAGS`
- Compile the sources using one authoritative command
- Link the sources into a shared library, preferably with a [soname](https://en.wikipedia.org/wiki/Soname) indicating its current ABI level.
- (optionally) allow the user to enable a static library instead of, or in addition to, the shared library
- Install the library(/ies) to /usr/lib (or some build system convention like --libdir)
- Install the public headers needed to link to libtree-sitter.(so|a) to /usr/include (or some build system convention like --includedir)
- Install a pkg-config file to /usr/lib/pkgconfig/, e.g. tree-sitter.pc, which will cause the command `pkg-config --libs --cflags tree-sitter` to emit whichever required C compiler flags needed to link to tree-sitter. Typically, that would be `-I/usr/include -L/usr/lib -ltree-sitter`.

Not in scope:
Preventing people from continuing to copy source files into their project, if they really want to.

Suggestion:
The meson build system is a fairly popular build system that can automatically handle most integration details for building and installing software. You simply define a `library()`, tag it with `install: true`, and add all source files and headers, everything else is handled automatically. #467 implements meson for this project, and could be used as inspiration for a mutually satisfactory solution. One bonus of meson is that it can automatically generate pkg-config files via its [Pkgconfig module](https://mesonbuild.com/Pkgconfig-module.html).

eli-schwartz:
This produces static/shared libraries as well as a pkg-config file, and installs them to the standard Unix hierarchy. GNU Make is assumed as features like $(wildcard ...) or $(shell uname) may not work elsewhere, but it should otherwise build on most/all Unix platforms.

Note that we assume the following POSIX default rules/macros exist, so we don't bother redefining them:
- pattern rules for compiling .c -> .o
- $(CC)
- $(AR)

Special note on the .pc file generation: we do most variable replacements in one go, but delay `@PREFIX@` so that we can first find existing substring instances of the prefix value (if libdir/includedir reside within the prefix), and update them to use a literal pkg-config variable `${prefix}`. This is fairly compact (one single sed) while still letting us produce pkg-config files that support runtime redefinition à la cross-compilation.

Resolves #581

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,71 @@
+VERSION := 0.6.3
+
+# install directory layout
+PREFIX ?= /usr/local
+INCLUDEDIR ?= $(PREFIX)/include
+LIBDIR ?= $(PREFIX)/lib
+PCLIBDIR ?= $(LIBDIR)/pkgconfig
+
+# collect sources
+ifneq ($(AMALGAMATED),1)
+	SRC := $(wildcard lib/src/*.c)
+	# do not double-include amalgamation
+	SRC := $(filter-out lib/src/lib.c,$(SRC))
+else
+	# use amalgamated build
+	SRC := lib/src/lib.c
+endif
+OBJ := $(SRC:.c=.o)
+
+# define default flags, and override to append mandatory flags
+CFLAGS ?= -O3
+override CFLAGS += -std=gnu99 -fPIC -Ilib/src -Ilib/include
+
+# ABI versioning
+SONAME_MAJOR := 0
+SONAME_MINOR := 0
+
+# OS-specific bits
+ifeq ($(shell uname),Darwin)
+	SOEXT = dylib
+	SOEXTVER_MAJOR = $(SONAME_MAJOR).dylib
+	SOEXTVER = $(SONAME_MAJOR).$(SONAME_MINOR).dylib
+	LINKSHARED += -dynamiclib -Wl,-install_name,$(LIBDIR)/libtree-sitter.$(SONAME_MAJOR).dylib
+else
+	SOEXT = so
+	SOEXTVER_MAJOR = so.$(SONAME_MAJOR)
+	SOEXTVER = so.$(SONAME_MAJOR).$(SONAME_MINOR)
+	LINKSHARED += -shared -Wl,-soname,libtree-sitter.so.$(SONAME_MAJOR)
+endif
+ifneq (,$(filter $(shell uname),FreeBSD NetBSD DragonFly))
+	PCLIBDIR := $(PREFIX)/libdata/pkgconfig
+endif
+
+all: libtree-sitter.a libtree-sitter.$(SOEXTVER)
+
+libtree-sitter.a: $(OBJ)
+	$(AR) rcs $@ $^
+
+libtree-sitter.$(SOEXTVER): $(OBJ)
+	$(CC) $(LDFLAGS) $(LINKSHARED) $^ $(LDLIBS) -o $@
+	ln -sf $@ libtree-sitter.$(SOEXT)
+	ln -sf $@ libtree-sitter.$(SOEXTVER_MAJOR)
+
+install: all
+	install -d '$(DESTDIR)$(LIBDIR)'
+	install -m755 libtree-sitter.a '$(DESTDIR)$(LIBDIR)'/libtree-sitter.a
+	install -m755 libtree-sitter.$(SOEXTVER) '$(DESTDIR)$(LIBDIR)'/libtree-sitter.$(SOEXTVER)
+	ln -sf libtree-sitter.$(SOEXTVER) '$(DESTDIR)$(LIBDIR)'/libtree-sitter.$(SOEXTVER_MAJOR)
+	ln -sf libtree-sitter.$(SOEXTVER) '$(DESTDIR)$(LIBDIR)'/libtree-sitter.$(SOEXT)
+	install -d '$(DESTDIR)$(INCLUDEDIR)'/tree_sitter
+	install -m644 lib/include/tree_sitter/*.h '$(DESTDIR)$(INCLUDEDIR)'/tree_sitter/
+	install -d '$(DESTDIR)$(PCLIBDIR)'
+	sed -e 's|@LIBDIR@|$(LIBDIR)|;s|@INCLUDEDIR@|$(INCLUDEDIR)|;s|@VERSION@|$(VERSION)|' \
+	    -e 's|=$(PREFIX)|=$${prefix}|' \
+	    -e 's|@PREFIX@|$(PREFIX)|' \
+	    tree-sitter.pc.in > '$(DESTDIR)$(PCLIBDIR)'/tree-sitter.pc
+
+clean:
+	rm -f lib/src/*.o libtree-sitter.a libtree-sitter.$(SOEXT) libtree-sitter.$(SOEXTVER_MAJOR) libtree-sitter.$(SOEXTVER)
+
+.PHONY: all install clean
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,9 @@ script:
   # Build the WASM binding
   - (eval "$WASM_ENV" && script/build-wasm)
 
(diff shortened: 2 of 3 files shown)


eli-schwartz:
I've assumed here that soname versioning of the form X.Y will be used, starting with 1.0 because this is stable and the first soname-versioned release.

The general release process would essentially boil down to bumping the VERSION variable in sync with the repository tag, and deciding whether to bump SONAME_MAJOR on ABI-breaking versions, or SONAME_MINOR on ABI-compatible versions.

Due to the use of wildcards, new source files in lib/src/ will be automatically picked up, and you can optionally do an amalgamated build with `make AMALGAMATED=1`, because I figured hey, why not -- the amalgamated source already exists, lets make it usable.

Makefile
```diff
@@ -0,0 +1,68 @@
...
+OBJ := $(SRC:.c=.o)
+
+# define default flags, and override to append mandatory flags
+CFLAGS ?= -O3
+override CFLAGS += -std=gnu99 -fPIC -Ilib/src -Ilib/include
+LDFLAGS ?=
+override LDFLAGS +=
+
+# ABI versioning
+SONAME_MAJOR := 1
```
dcreager, line 27: Minor nit: Can we start at SOVERSION 0?  That seems to be more standard, and will hopefully make it more obvious that the shared library version string has ~nothing to do with the project semver version number.

Makefile
```diff
@@ -0,0 +1,68 @@
...
+	SOEXT = so
+	SOEXTVER_MAJOR = so.$(SONAME_MAJOR)
+	SOEXTVER = so.$(SONAME_MAJOR).$(SONAME_MINOR)
+	LINKSHARED += -shared -Wl,-soname,libtree-sitter.so.$(SONAME_MAJOR)
+endif
+ifneq (,$(filter $(shell uname),FreeBSD NetBSD DragonFly))
+	PCLIBDIR := $(PREFIX)/libdata/pkgconfig
+endif
+
+all: libtree-sitter.a libtree-sitter.$(SOEXTVER)
```
dcreager, line 46: Can you add a `.PHONY: all install` clause just after this?

Makefile
```diff
@@ -0,0 +1,68 @@
+VERSION := 0.16.5
```
dcreager, line 1: Somewhat confusingly, the library and CLI have different semvers.  This should track the library's semver, which is currently [0.6.3](https://github.com/tree-sitter/tree-sitter/blob/1635aab801152b8b36975dbc6cd02ea01ddbf778/lib/Cargo.toml#L4).

Makefile
dcreager: (Unless the SOVERSION 1 is intended to bump past any SOVERSION-0 packages that might have already been built?)

Makefile
```diff
@@ -0,0 +1,68 @@
...
+	# use amalgamated build
+	SRC := lib/src/lib.c
+endif
+OBJ := $(SRC:.c=.o)
+
+# define default flags, and override to append mandatory flags
+CFLAGS ?= -O3
+override CFLAGS += -std=gnu99 -fPIC -Ilib/src -Ilib/include
+LDFLAGS ?=
+override LDFLAGS +=
```
maxbrunsfeld, line 24: Just curious - what is the effect of these two `LDFLAGS` lines (compared to not explicitly defining LDFLAGS at all)?

Makefile
eli-schwartz: Good point, I should really do that.

Makefile
eli-schwartz: That is a perfectly reasonable point to make. This version number is currently stored in a rust binding configuration file, though, which is why I didn't find it. :D

I expected a bit of fine-tuning -- I guessed at what I hoped were the right values, but I guess I was off.

Makefile
eli-schwartz: I don't know how standard this is, so I'll go with whatever @maxbrunsfeld wants.

Makefile
eli-schwartz: Nothing at all, but they serve as documentation on how to do it properly if you ever do add something here.

As demonstrated by the CFLAGS handling, `FOO ?=` will set the default if you haven't specified it in the environment, and `override FOO +=` will append to FOO (the override keyword is needed to ensure this happens even the defaults were passed as `make FOO=...` rather than `FOO=... make`).

It could be removed and left to be inferred from the previous two lines, or if there are any LDFLAGS you'd like to use, they could be added here.

Makefile
```diff
@@ -0,0 +1,68 @@
...
+	# do not double-include amalgamation
+	SRC := $(filter-out lib/src/lib.c,$(SRC))
+else
+	# use amalgamated build
+	SRC := lib/src/lib.c
+endif
+OBJ := $(SRC:.c=.o)
+
+# define default flags, and override to append mandatory flags
+CFLAGS ?= -O3
```
FFY00, line 21: Just curious, why did you choose `-O3`? Did you do any tests against `-O2`?

eli-schwartz: I faithfully copied this from https://github.com/tree-sitter/tree-sitter/blob/1635aab801152b8b36975dbc6cd02ea01ddbf778/script/build-lib#L10-L12

This is an existing project policy for default cflags, and I'm focused here on only creating a Makefile build with parity to the existing script.

Makefile
eli-schwartz: Fixed.

Makefile
eli-schwartz: Implemented.

eli-schwartz:
I pushed changes for two of the review comments, tell me what you want for the other two (if anything).

Makefile
maxbrunsfeld: I would say let’s start at 0 like @dcreager said.

Makefile
maxbrunsfeld: I think I’d slightly prefer to leave these out then. Thanks for the explanation.

eli-schwartz:
http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

```
DEPFLAGS = -MT $@ -MMD -MP -MF $*.d
CFLAGS += $(DEPFLAGS)
-include lib/src/*.d
```

This works on at least gcc and clang to generate Makefile fragments and include them, which cause the object files to depend on all the header files that went into them. As those are the only two compilers I use, I'm not familiar with which compilers might *not* support it, but they're pretty useful and fairly common, I think... it would be able to be suppressed with `make DEPFLAGS=` if so. Thoughts?

Makefile
eli-schwartz: Okay, libtree-sitter.so.0.0 it is.

maxbrunsfeld:
When developing the library, we mostly are using the rust build system to compile it, since the test suite and the CLI both use the rust binding. we do a clean build of the C code every time, because the cost is negligible compared to the rust build time.

So I don’t think we’ll be doing a lot of incremental builds via the Makefile. Maybe it’d even be better to default to the amalgamated build, because it has less possibility of stale object files.

Overall, I favor making the makefile as simple and compatible as possible for users of the library, versus adding features to help in developing the library.

Makefile
eli-schwartz: Done.

Makefile
eli-schwartz: Done.

eli-schwartz:
All right then, I won't bother. I've just resolved the other two review comments and thrown in a make clean target.

maxbrunsfeld:
Thanks for your work on this, and the RFC.

PR #607, opened 2020-04-28
Do not use multiple unnamed structs inside of unions

ret2libc:
Related to #604 .

I could not completely avoid naming some of the structures, as it seems GCC 4 (at least as provided on CentOS 6) has problems with multiple unnamed structs inside a union.

What do you think about this @maxbrunsfeld  ?

### DIFF
--- a/lib/include/tree_sitter/parser.h
+++ b/lib/include/tree_sitter/parser.h
@@ -62,13 +62,13 @@ typedef struct {
       TSStateId state;
       bool extra : 1;
       bool repetition : 1;
-    };
+    } shift;
     struct {
       TSSymbol symbol;
       int16_t dynamic_precedence;
       uint8_t child_count;
       uint8_t production_id;
-    };
+    } reduce;
   } params;
   TSParseActionType type : 4;
 } TSParseAction;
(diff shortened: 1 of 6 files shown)


lib/include/tree_sitter/parser.h
```diff
@@ -62,13 +62,13 @@ typedef struct {
       TSStateId state;
       bool extra : 1;
       bool repetition : 1;
-    };
+    } status;
```
maxbrunsfeld, line 65: I think this field should be called `shift`. "Shift" and "reduce" are the two main types of actions.

lib/include/tree_sitter/parser.h
```diff
@@ -167,22 +167,22 @@ struct TSLanguage {
...
 
-#define SHIFT(state_value)              \
-  {                                     \
-    {                                   \
-      .type = TSParseActionTypeShift,   \
-      .params = {.state = state_value}, \
-    }                                   \
+#define SHIFT(state_value)                \
+  {                                       \
+    {                                     \
```
maxbrunsfeld, line 172: I think we should use the field name here, for consistency with the `REDUCE` macro.

lib/include/tree_sitter/parser.h
ret2libc: yeah sure. I didn't know how to name it as I'm not that familiar with tree-sitter internals

maxbrunsfeld:
Thanks!

PR #615, opened 2020-05-11
Improve the tree query syntax

maxbrunsfeld:
This PR changes the tree query syntax in the ways described in #612, plus some additional ways.

### 1. Wildcards with `_`

The "wildcard" node is now written `_` (to match wildcard patterns in Rust and Haskell) instead of `*`.

### 2. Repetition with `*`

There is now a `*` postfix operator to go along with the `+` and `?` operators, and they all now match their counterpart in regular expression syntax (`*` is zero-or-more, `+` is one-or-more, `?` is zero-or-one).

### 3. Grouping everywhere with `(`  `)`

Parentheses can now be used everywhere for grouping sibling nodes.

### 4. Predicates with `#`

To disambiguate predicate expressions for named nodes, predicate names now must begin with `#`. I think this makes patterns with predicates much more readable, and it simplifies the parser.

## Example:

```clj
(
  (comment)* @doc
  (function_declaration
    name: (identifier) @name) @function
  (#strip! @doc "^//\\s+")
)
```

## Notes

* The docs are updated to use and explain the new syntax.
* I have tried to make the query parser maintain backward compatibility with existing queries. In most cases, it is possible to detect where the old syntax is being used, and keep the old behavior. AFAIK, the only case where this breaks down is if you wrote a predicate using the old syntax without the `#`, and the predicate name did *not* end with `?` or `!`.


SkyBlack1225:
> This PR changes the tree query syntax in the ways described in #612, plus some additional ways.
> 
> ## Changes
> ### 1. Wildcards with `_`
> The "wildcard" node is now written `_` (to match wildcard patterns in Rust and Haskell) instead of `*`.
> 
> ### 2. Repetition with `*`
> There is now a `*` postfix operator to go along with the `+` and `?` operators, and they all now match their counterpart in regular expression syntax (`*` is zero-or-more, `+` is one-or-more, `?` is zero-or-one).
> 
> ### 3. Grouping everywhere with `(` `)`
> Parentheses can now be used everywhere for grouping sibling nodes.
> 
> ### 4. Predicates with `#`
> To disambiguate predicate expressions for named nodes, predicate names now must begin with `#`. I think this makes patterns with predicates much more readable, and it simplifies the parser.
> 
> ## Example:
> ```clojure
> (
>   (comment)* @doc
>   (function_declaration
>     name: (identifier) @name) @function
>   (#strip! @doc "^//\\s+")
> )
> ```
> 
> ## Notes
> * The docs are updated to use and explain the new syntax.
> * I have tried to make the query parser maintain backward compatibility with existing queries. In most cases, it is possible to detect where the old syntax is being used, and keep the old behavior. AFAIK, the only case where this breaks down is if you wrote a predicate using the old syntax without the `#`, and the predicate name did _not_ end with `?` or `!`.

PR #616, opened 2020-05-13
Fix wasm tests on CI

maxbrunsfeld:
I've had to make a series of tweaks to the way the library and parsers are compiled to Wasm due to a change to emscripten. I'm seeing different results on CI than I do locally, so I'm opening this PR to push a series of changes to try and fix the issue.

### DIFF
--- a/cli/src/wasm.rs
+++ b/cli/src/wasm.rs
@@ -89,11 +89,11 @@ pub fn compile_language_to_wasm(language_dir: &Path, force_docker: bool) -> Resu
     let scanner_cc_path = src.join("scanner.cc");
     let scanner_cpp_path = src.join("scanner.cpp");
 
-    if scanner_cc_path.exists() {
+    if language_dir.join(&scanner_cc_path).exists() {
         command.arg("-xc++").arg(&scanner_cc_path);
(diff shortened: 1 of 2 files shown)


Razzeee:
This seems to introduce problems for me.

If I bump to 0.16.6 or 0.16.7 I run into this error when loading the wasm bundle.

`e[Object.keys(...).find(...)] is not a function`

It should be testable via (this assumes your using vscode):

https://github.com/elm-tooling/elm-language-client-vscode#contributing--debugging

```
git clone --recursive git@github.com:elm-tooling/elm-language-client-vscode.git
cd elm-language-client-vscode
npm install
cd server
npm install --dev tree-sitter-cli
npm install
npm run build-tree-sitter
npm run compile 
```
then hit F5 wait for a new vscode to launch and open an elm project

maxbrunsfeld:
Is your parser generated with the latest tree-sitter CLI? What version of emscripten are you using?

Razzeee:
Sorry, haven't updated either. Will check back after I do.

Razzeee:
Updating the parser to use the latest cli solved the problem. I did not update emscripten. Sorry for the noise.

PR #620, opened 2020-05-16
[cli]: add an option to no open browser in web-ui command

### LINKED ISSUE #506: Don't force local browser to open with web-ui
When using `tree-sitter` over a remote SSH terminal, I'd like to be able to use `tree-sitter web-ui` in order to test my grammar in the browser of my *host* operating system, that is, the system that am using `ssh` from.

Currently, `web-ui` uses the `webbrowser` crate in order to attempt to open the browser on the host tree-sitter is executing on, and it forces this behavior. Because my server is headless and does not run X11, this fails, because `xdg-open`, etc do not work.

I believe this is a relatively simple fix: `webbrowser::open` is simply called with `.unwrap()` here in the [web_ui.rs](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/web_ui.rs#L40) module, which causes the panic. I believe simply handling this error case rather than `.unwrap`'ing will fix it.

lerencao:
I need to test my grammar in browser frequently.
`tree-sitter web-ui` will open a new browser each time, which may be unnecessary.
This option disable it.
I think this can also close #506 .

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -149,7 +149,8 @@ fn run() -> error::Result<()> {
                 .arg(Arg::with_name("path").index(1).multiple(true)),
         )
         .subcommand(
-            SubCommand::with_name("web-ui").about("Test a parser interactively in the browser"),
+            SubCommand::with_name("web-ui").about("Test a parser interactively in the browser")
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks!

PR #622, opened 2020-05-17
TypeScript: Correct SyntaxNode isNamed type

skovhus:
`isNamed` is a function not a boolean, see https://github.com/tree-sitter/tree-sitter/blob/40993195b8da0d042161a7b7bb0f6831dcfb4da1/lib/binding_web/binding.js#L237


maxbrunsfeld:
Sorry for the delay on this one.

PR #624, opened 2020-05-18
Improve tree queries' ability to handle large numbers of nested matches

maxbrunsfeld:
### Motivation

I noticed that Tree-sitter's syntax highlighting was failing for JavaScript code like this, in cases where there were more than about 8 chained calls:

```js
foo
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
```

With the currently-deployed syntax highligthing code, you can see ☝️ that the first two methods aren't highlighted correctly.

### Problem

To guarantee predictably good performance, tree queries impose a strict limit on the amount of memory that can be allocated in pursuing different possible matches. This limit means that most of the time, query matching entails no memory allocation at all.

For the vast majority of real use case, the system never comes close to hitting this limit. The reason that the example above *did* hit the limit is that unlike most real-world coding patterns, long "chains" of method calls produce arbitrarily *deep* syntax trees.

```clj
(call_expression
  function: (member_expression
    object: (call_expression
      function: (member_expression
        object: (call_expression
          function: (member_expression
            object: (call_expression
              function: (member_expression
                object: (call_expression
                  function: (member_expression
                    object: (call_expression
                      function: (member_expression
                        object: (call_expression
                          function: (member_expression
                            object: (call_expression
                              function: (member_expression
                                object: (call_expression
                                  function: (member_expression
                                    object: (call_expression
                                      function: (member_expression
                                        object: (identifier)
                                        property: (property_identifier)
                                      arguments: (arguments
                                        (identifier))))
                                    property: (property_identifier)
                                  arguments: (arguments
                                    (identifier))))
                                property: (property_identifier)
                              arguments: (arguments
                                (identifier))))
                            property: (property_identifier)
                          arguments: (arguments
                            (identifier))))
                        property: (property_identifier)
                      arguments: (arguments
                        (identifier))))
                    property: (property_identifier)
                  arguments: (arguments
                    (identifier))))
                property: (property_identifier)
              arguments: (arguments
                (identifier))))
            property: (property_identifier)
          arguments: (arguments
            (identifier))))
        property: (property_identifier)
      arguments: (arguments
        (identifier))))
    property: (property_identifier)
  arguments: (arguments
    (identifier))))

```

A common query pattern for method calls entails some overlap between the different matches:

```clj
(call_expression
  function: (member_expression
    property: (property_identifier) @the-method-name))
```

Even though there's only one capture (`the-method-name`), there can be a lot of concurrent *matches*, because within a match, before we see the method name, we see another nested `call_expression`.

### Solution

I've maintained a hard limit on the number of concurrent "capture lists" that can be allocated, but I've restructured the matching logic so that a given `QueryState` only *acquires* a `CaptureList` when it processes its first capture, as opposed to acquiring a capture list as soon as we create a new `QueryState`.

This means that it will be much, much more difficult to hit the capture list limit. Before, you just needed to create many matches that all spanned overlapping nodes. Now, the matches would need to *capture* lists of nodes that all overlapped. Notice that in the example above, the matches overlap (because the overall `call_expression` nodes are nested), but none of the *captured* nodes (the method names) overlap at all.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -339,7 +339,7 @@ fn test_query_matches_with_nesting_and_no_fields() {
 }
 
 #[test]
-fn test_query_matches_with_many() {
+fn test_query_matches_with_many_results() {
     allocations::record(|| {
         let language = get_language("javascript");
         let query = Query::new(language, "(array (identifier) @element)").unwrap();
@@ -349,6 +349,47 @@ fn test_query_matches_with_many() {
             &query,
             &"[hello];\n".repeat(50),
             &vec![(0, vec![("element", "hello")]); 50],
+        );
+    });
+}
+
+#[test]
+fn test_query_matches_with_many_overlapping_results() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+        let query = Query::new(
+            language,
+            r#"
+            (call_expression
+                function: (member_expression
+                    property: (property_identifier) @method))
+            (call_expression
+                function: (identifier) @function)
+            ((identifier) @constant
+             (#match? @constant "[A-Z\\d_]+"))
+            "#
+        ).unwrap();
+
+        let count = 80;
+
+        // Deeply nested chained function calls:
+        // a
+        //    .foo(bar(BAZ))
+        //    .foo(bar(BAZ))
+        //    .foo(bar(BAZ))
+        //    ...
+        let mut source = "a".to_string();
+        source += &"\n  .foo(bar(BAZ))".repeat(count);
+
+        assert_query_matches(
+            language,
+            &query,
+            &source,
+            &[
+                (0, vec![("method", "foo")]),
+                (1, vec![("function", "bar")]),
+                (2, vec![("constant", "BAZ")])
+            ].iter().cloned().cycle().take(3 * count).collect::<Vec<_>>(),
         );
     });
 }
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -1268,6 +1268,9 @@ static bool ts_query_cursor__first_in_progress_capture(
   uint32_t *pattern_index
 ) {
   bool result = false;
+  *state_index = UINT32_MAX;
+  *byte_offset = UINT32_MAX;
+  *pattern_index = UINT32_MAX;
   for (unsigned i = 0; i < self->states.size; i++) {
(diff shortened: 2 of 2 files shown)


maxbrunsfeld:
Here's an example snippet with even more chaining. After we redeploy Treelights with this fix, the following code will highlight correctly 👇 

```js
foo
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
```

maxbrunsfeld:
Here's an example snippet with even more chaining. After we ship this fix, the following code will highlight correctly 👇 

```js
foo
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
    .one(a)
    .two(b)
    .three(c)
    .four(d)
    .five(e)
    .six(f)
    .seven(g)
    .eight(h)
    .nine(i)
    .ten(j)
```

PR #630, opened 2020-05-30
Add alternative syntax in queries

maxbrunsfeld:
### Summary

Currently, there is a fair amount of duplication in some of our highlighting and tagging queries. Some of this could be avoided if Tree queries had a notion of *alternation* - writing a sub-pattern that matches any *one of* a list of possible patterns. This PR adds a general alternation syntax to Tree queries.

### Syntax

An alternation is written as a pair of *square brackets* containing a list of alternative patterns. Using square brackets seems practical because:

* It fits neatly into the existing S-expression syntax, which is based on the `(` `)` delimiters
* It is reminiscent of the regular expression *character class* syntax `[abc]`, which matches either `a`, `b`, or `c`.

Alternations are (I believe) fully composable with the other syntax constructs in queries (captures, `+`, `*`, and `?`, etc).

### Examples

All of these examples are for the JavaScript language:

1. Match a function **or an arrow function** being assigned to a variable **or an object property**. If the left hand side is a variable, capture it as `@function`. If it is a property, capture it as `@method`.

```clj
(assignment_expression
  left: [
    (identifier) @function
    (member_expression
      property: (property_identifier) @method)
  ]
  right: [(function) (arrow_function)]
)
```

2. Match a call to either a variable **or an object property**. In the case of a variable, capture it as `@function`. In the case of a property, capture it as `@method`.

```clj
(call_expression
  function: [
    (identifier) @function
    (member_expression
      property: (property_identifier) @method)
  ])
```

3. Match a set of possible keyword tokens, capturing them as `@keyword`.

```clj
[
  "break"
  "catch"
  "delete"
  "do"
  "else"
  "finally"
  "for"
  "function"
  "if"
  "new"
  "return"
  "throw"
  "try"
  "var"
  "while"
] @keyword
```

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -367,8 +367,9 @@ fn test_query_matches_with_many_overlapping_results() {
                 function: (identifier) @function)
             ((identifier) @constant
              (#match? @constant "[A-Z\\d_]+"))
-            "#
-        ).unwrap();
+            "#,
+        )
+        .unwrap();
 
         let count = 80;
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Still need to write docs for this feature.

Issue #575, opened 2020-03-10
Oxidize core lib


dodomorandi:
It could be interesting to port the core lib to Rust, in order to increase the maintainability of the project and to gain more safety.

Instead of creating the lib from ground up, it should be possible to use [C2Rust](https://github.com/immunant/c2rust) in order to convert the existing C code into unidiomatic and unsafe Rust, then gradually convert the functions to safe and idiomatic versions, one by one.

It is also reasonable to integrate [cbindgen](https://github.com/eqrion/cbindgen) in order to automatically create the C header files and keep a 100% API/ABI compatibility respect to the current state of the project.

If I am not wrong, there are some tests available, but they only relate to the CLI tool. Maybe it is interesting to create unit tests for the ported Rust code, in order to easily detect the source of regressions during the oxidation process.

maxbrunsfeld:
The core library is tested pretty thoroughly with unit tests as well as randomized tests. All the tests use the Rust binding (and the CLI's functionality for dynamically loading generated parsers). They are located [here](https://github.com/tree-sitter/tree-sitter/tree/master/cli/src/tests).

I think this may be worth pursuing, but currently it's very convenient to be able to build Tree-sitter with nothing but a C compiler. It may make more sense to do this work once Rust has become a little bit more ubiquitous, so that that tools like Node.js, Python, Ruby and Haskell are all able to compile Rust code automatically.

dodomorandi:
Thank you for your feedback!

Pardon me if I could be misunderstood -- I noticed that the library is well tested and there is a very nice work behind it! What I just wanted to say is that with Rust it is straightforward to create tests for each function. This obviously does not give less importance to the actual tests, it just give another layer of reliability over the single functions, hopefully easing the detection of specific regressions.

About the 'nothing but a C compiler' point: my personal objection is that the project already uses a full Rust ecosystem in order to build the C library. I knew that, in theory, one could just perform a `cc lib.c -o lib.so --shared` (or similar) on the source directory, but in practice I think that people interested in the project will just use cargo to build everything.

In any case, these are obviously just my opinions :smiley:. Just let me know if you will accept an eventual contribution focused on the oxidation of the C lib, or if you consider it premature for the current state of the project. Nevertheless, thank you for your work and your time :wink:!

maxbrunsfeld:
> in practice I think that people interested in the project will just use cargo to build everything

In practice, most users of the project are not using Rust; they're relying on build systems which have built-in support for C code:

[...]

PR #638, opened 2020-06-06
Convert the C library to Rust

Luni-4:
Hello! :smiley: 

As [Rust Torino](https://github.com/rust-torino), we tried to convert the `C` library of `tree-sitter` to transpiled `Rust` code. The transpiled code is then used in the `binding_rust` workspace. 

## Web Bindings

Unfortunately, we failed at creating `js` and `wasm` files for the `binding_web` workspace, that's why the old `C` library has been maintained.

## Tests

Among the 159 tests, only 2 are currently failing on `Travis`, but we don't understand why, perhaps someone could give us an hint to fix them.

We haven't test it on `AppVeyor` yet.

## Create C-libraries and headers from Rust

It is also possible to create the C-libraries and the `api.h` header from the `Rust` code in this way.

```
cd core
cargo install cargo-c
cargo cinstall --destdir=.
```
After that, you will find the output of this process inside the `/usr` directory contained in the `core` directory.

## Acknowledgements

Even if I'm opening this PR, I didn't work at this project alone, so I would also like to thank the entire team: @lu-zero, @dodomorandi, @xrmx

I hope this is appreciated and thanks in advance for your review! :smiley:


maxbrunsfeld:
This is very cool. Was there very much manual effort involved, or can the entire thing be automated now, using c2rust?

Overall, I don't think I am ready to take over maintenance of this. There are still some advantages for me in using C for the core library, as mentioned in #575. If there is an automated way to perform the conversion, I would be open to including some kind of script in the repo for now. It might be useful for projects wanting to use Tree-sitter with wasm-bindgen, due to the [current incompatibility between clang and Rustc](https://github.com/WebAssembly/tool-conventions/issues/130).

Luni-4:
Thanks for your reply @maxbrunsfeld! 

> This is very cool. Was there very much manual effort involved, or can the entire thing be automated now, using c2rust?

The main effort was splitting up the big `c2rust` file into different Rust files and replacing `libc`'s functions in order to make the workspace cross-platform.

> 
> Overall, I don't think I am ready to take over maintenance of this. There are still some advantages for me in using C for the core library, as mentioned in #575. If there is an automated way to perform the conversion, I would be open to including some kind of script in the repo for now. It might be useful for projects wanting to use Tree-sitter with wasm-bindgen, due to the [current incompatibility between clang and Rustc](https://github.com/WebAssembly/tool-conventions/issues/130).

I've read #575 and it seems the main problem is building some of the software related to `tree-sitter`. Does this problem will be fixed if we automatically release the output of `cargo-c`? I mean:
- Header files (api.h and parser.h)
- Static libs (`.lib` for Windows and `.a` for Linux and macOS)

I'm willing to create a GitHub Actions script to do so, in this way the bindings and the other software can automatically use these headers and binaries.

mcepl:
>     * Static libs (`.lib` for Windows and `.a` for Linux and macOS)

Also dynamic libs (`.dll` and `.so` respectively).

Issue #314, opened 2019-04-01
`get_column` causing side effects


banacorn:
I'm not sure if this is intended, however, triggering `get_column` would cause the lexer to advance: 

https://github.com/tree-sitter/tree-sitter/blob/49392c8502091eff4340e9318a5e84f16c5bb81c/lib/src/lexer.c#L141-L145

If this side effect is intended (or inevitable somehow), then I think it should be documented.

Note: I found this when I was working on a workaround for https://github.com/tree-sitter/tree-sitter/issues/312.

banacorn:
After some tinkering, I think that this is a bug, intended or not.

Is there any plan to fix this?

maxbrunsfeld:
Calling `get_column()` mutates the lexer internally, but it should leave the lexer in the same exact state as it was before: it moves to the beginning of the line and then advances back to the original position in order to count the number of unicode characters. I'm not sure how it could be causing side effects.

banacorn:
I've managed to setup a minimal experiment for reproducing the issue.
Please checkout [this branch](https://github.com/tree-sitter/tree-sitter-agda/tree/tree-sitter%23314). 

```bast
git clone git@github.com:tree-sitter/tree-sitter-agda.git
cd tree-sitter-agda
git checkout tree-sitter#314
```

I've put the description in README.
(calling `lexer->get_column(lexer)` more `n` times would set the line numbers off by `n`)

banacorn:
I want to fix this. 

I'm looking at directories like [this one](https://github.com/tree-sitter/tree-sitter/tree/master/test/fixtures/test_grammars/external_extra_tokens), trying to figure out how to setup a test that utilizes `get_column()`. 
That `grammar.json` should be generated from some `grammar.js, but I can't seem to find it. @maxbrunsfeld can you help me?

maxbrunsfeld:
Yeah, for those tests you just write the `grammar.json` manually, and add a `scanner.c`.

banacorn:
@maxbrunsfeld is there anyway of testing against the source location of the output syntax tree?

The corpus test only make sure that the generated abstract syntax tree looks right (without source locations), but the problem I'm trying to fix lies in those source locations.

bglgwyng:
#589 I was trying to fix this issue and I found that when the token includes line break(In my case, this was for the line fold implementation), _rollback-and-re-advance_ in `get_column` increases `self->current_position.extent.row`.

banacorn:
[...]

Issue #589, opened 2020-04-04
Each call of get_column increase the line number of tokens by 1


bglgwyng:
https://github.com/bglgwyng/tree-sitter-sandbox/tree/get-column-increases-line-no
Here is my simple example.
You can find `lexer->get_column(lexer);` on the line 35 of `scanner.c`.
Each time I add the function call, the result span's line number increases by 1.

maxbrunsfeld:
Thanks for taking the time to create these reproduction steps. This is very very helpful. I will try to take a look at this bug later this week.

bglgwyng:
@maxbrunsfeld As I mentioned in #314, I think it happens when the token includes newline character.

PR #640, opened 2020-06-07
[WIP] fix get_column side effect

bglgwyng:
It's a fix for #314. It's work in progress and I have some questions.

I've asked that why tree-sitter doesn't just return `extent.column` for get_column in that issue.
Now I found that `extent.column` is not the character count. It's the byte offset from the start of the row. If there's some unicode character in the text, then `extent.column` can be different from expected.

So I added `column_` to `TSPointer`.
It passed every test. But I don't think we can rely on it. Since a very rough fix(https://github.com/bglgwyng/tree-sitter/tree/7dd503cac220e5ffaac0db8ae07fdd35a42dc2ea) also passed 100%.
I guess subtree editing is not covered by tests. How can I test it then?

I didn't figure out how to fix `ts_subtree_size` in lib/src/subtree.h so I left it.
I think we need to modify `ts_subtree_new_leaf` and `ts_subtree_edit`.
Am I doing it right?

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -66,7 +66,7 @@ module.exports = grammar({
 });
 ```
 
-Then run the the following command:
+Then run the following command:
 
 ```sh
 tree-sitter generate
@@ -152,7 +152,7 @@ func x() int {
 
 These tests are important. They serve as the parser's API documentation, and they can be run every time you change the grammar to verify that everything still parses correctly.
 
-By default, the `tree-sitter test` command runs all of the tests in your `corpus` or `test/corpus/` folder. To run a particular test, you can use the the `-f` flag:
+By default, the `tree-sitter test` command runs all of the tests in your `corpus` or `test/corpus/` folder. To run a particular test, you can use the `-f` flag:
 
 ```sh
 tree-sitter test -f 'Return statements'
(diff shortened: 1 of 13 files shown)


maxbrunsfeld:
I don't think I want to pursue this solution, because it means a very large increase in the memory consumption of a syntax tree. Every node in the tree would need to store two additional 32-bit integers, which can add up to many megabytes in a large tree.

This is why I took the approach of computing the unicode character column count only when `get_column` is called. I think we need to understand why that code has the possibility of introducing side effects, and fix that.

bglgwyng:
@maxbrunsfeld I remove additional unicode character count field `column_` and reimplement it following your guide.

bglgwyng:
@maxbrunsfeld Have you checked the new commit?
The new one has very small changes from the master branch.
It just counts unicode characters from the beginning of the row using `ts_decode_utf8` or `ts_decode_utf16`.
It doesn't call `ts_lexer__advance` not to make side effects.

maxbrunsfeld:
I can't tell why this fixes the issue; it seems like the existing implementation *should* have the same effect. The side effects are temporary, and the state should be restored by the end of the `get_column` call.

Can you add a test that reproduces the problem? I think you would need to create a minimal grammar that demonstrates the problem: something similar to [this directory](https://github.com/tree-sitter/tree-sitter/tree/master/test/fixtures/test_grammars/external_tokens).

bglgwyng:
#589 I had posted the reproduction step here before. I should have mentioned that.
You can try `tree-sitter parse example-file-1`.
- with this fix
```
(source_file [0, 0] - [2, 1]
  (parethesized [0, 0] - [2, 1]
    (apply [0, 1] - [0, 4]
      (identifier [0, 1] - [0, 2])
      (whitespace [0, 2] - [0, 3])
      (identifier [0, 3] - [0, 4]))
    (whitespace [0, 4] - [0, 5])
    (whitespace [0, 5] - [2, 0])))
```
- with the current version
```
(source_file [0, 0] - [4, 1]
  (parethesized [0, 0] - [4, 1]
    (apply [0, 1] - [0, 4]
      (identifier [0, 1] - [0, 2])
      (whitespace [0, 2] - [0, 3])
      (identifier [0, 3] - [0, 4]))
    (whitespace [0, 4] - [0, 5])
    (whitespace [2, 3] - [4, 0])))
```
The last row number is different.

I just checked the current test suits and I found that `corpus.txt` does not contain row number and columnnumber for each token. Without them, I think I can't inspect the side effect of `get_column`. Is there any way to make test suits to tell if row number and column number are correct?

PR #644, opened 2020-06-10
Analyze queries on construction to identify impossible patterns, and patterns that will definitely match

### LINKED ISSUE #634: .captures method returns fewer results than .matches when query cursor has range restriction
I'm using `.captures` with range restriction for syntax highlighting. Sometimes it doesn't return nodes that `.matches` would otherwise return. It works correctly if the query cursor is not restricted to a range.

matches-vs-captures

This is the JSON file: https://github.com/ubolonton/emacs-tree-sitter/blob/matches-vs-captures/lisp/test-files/matches-vs-captures.json#L57-L60.

This is the highlight query: https://github.com/ubolonton/emacs-tree-sitter/blob/matches-vs-captures/langs/queries/json/highlights.scm.

Here's the output of [the code that reproduces the issue](https://github.com/ubolonton/emacs-tree-sitter/blob/matches-vs-captures/src/matches_vs_captures.rs):
```
.matches
  501 . 509 : "fields"
  519 . 528 : "keyword"
  540 . 546 : "type"
  567 . 581 : "ignore_above"
  608 . 624 : "abc_def_ghk_04"
  632 . 638 : "type"
  652 . 660 : "fields"
  670 . 679 : "keyword"
  691 . 697 : "type"
  718 . 732 : "ignore_above"
  759 . 775 : "abc_def_ghk_05"
  783 . 789 : "type"
  803 . 811 : "fields"
  821 . 830 : "keyword"
  842 . 848 : "type"
  869 . 883 : "ignore_above"
  910 . 926 : "abc_def_ghk_06"
  934 . 940 : "type"
  954 . 962 : "fields"
  972 . 981 : "keyword"
  993 . 999 : "type"
.captures
  501 . 509 : "fields"
  519 . 528 : "keyword"
  540 . 546 : "type"
  567 . 581 : "ignore_above"
  608 . 624 : "abc_def_ghk_04"
  632 . 638 : "type"
  652 . 660 : "fields"
  670 . 679 : "keyword"
  691 . 697 : "type"
  718 . 732 : "ignore_above"
  759 . 775 : "abc_def_ghk_05"
  783 . 789 : "type"
  803 . 811 : "fields"
  821 . 830 : "keyword"
  842 . 848 : "type"
  869 . 883 : "ignore_above"
  910 . 926 : "abc_def_ghk_06"
```

maxbrunsfeld:
Fixes #634

## Problem 1 - Static validation of Queries

Currently, when constructing a query, the engine checks that all of the node types and field names exist in the grammar. This prevents simple typos. But that's it. The engine doesn't check that all of the patterns make sense. It is possible to query for invalid child or sibling relationships. It's also possible to query for field names which cannot possibly occur in the given parent node.

## Problem 2 - Dropping matches when iterating through captures

The bug described in #634 is caused by a combination of three aspects of the query engine:

**a) Intersecting matches** - When using Tree queries, it's often possible for one match to span a large chunk of source code that contains many other matches. For example, consider this JavaScript query:

```clj
(object
  "{" @bracket
  "}" @bracket)

(pair
  key: (string) @property-name)
```

When matching against a large `object` literal, a match of the first pattern (which captures the curly braces) may *enclose*  many matches of the second pattern (which captures each key).

**b) Maintaining Capture Order** - This situation is fine when consuming matches one by one (via the `ts_query_cursor_next_match` API). But for syntax highlighting, we consume query results in a different way, via `ts_query_cursor_next_capture`. This API allows us to iterate over the individual *captures* in the order that they appear, regardless of which matches they belong to. 

So in this example, we would receive the `{` node first (captured by the first pattern), then a series of `string` nodes (captured by the second pattern), and finally the `}` node (part of the same match as the earlier `{`).

**c) Limited buffering** - When using `ts_query_cursor_next_capture`, the `{` node cannot be returned until the engine determines that its pattern will successfully match. So until the query engine sees the `}` node, it must store the `{` node. And in order to preserve ordering, it must store any subsequent matches as well.

In order to maintain acceptable performance, there is a limit to how many matches will be stored. If this limit is exceeded, then the in-progress match with the earliest capture (the `{` node in this case) will be dropped in order to unblock the buffered captures from being returned.

## Solution - Abstract Analysis of Patterns

This PR adds a new step to the internal query-construction procedure. After parsing a query, we *analyze* it to determine two things:

1. **Possibility** - For each pattern in the query, is it possible to construct a syntax node that will match the pattern? This can be used to catch mistakes in queries. The engine will return an error if any pattern in the query is impossible to match.

2. **Definiteness** - For each *step* of each pattern in the query, once that step is reached, is it *inevitable* that the pattern will successfully match?

    Definiteness can be used to address Problem 2 described above. In the example above, the query engine shouldn't need to store the `{` token until it sees the `}` token, because we know in advance that **an `object` node always ends with a `}`**, so the pattern is guaranteed to match. This PR changes the `ts_query_cursor_next_capture` function so that it does not necessarily wait to return a capture until its pattern has successfully matched. Instead, it will return captures as soon as it can be verified that their patterns *will* definitely match.

## Steps

* [x] Implement a system for symbolically executing each non-terminal sub-pattern in a query, and keeping track of the possible steps where matching could terminate. Using this set of steps, determine
  * whether or not each step in the pattern is *definite*
  * whether or not the sub-pattern is *possible* to match
* [x] Return an error when constructing a query with impossible patterns, indicating which pattern is impossible
* [x] When iterating through *captures*, use steps' *definite* flags to "early return" certain captures before their parent match has completed, when possible
  * [x] Don't "early return" captures for patterns with predicates, unless all of the captures mentioned in the predicates have already been matched.
* [x] Ensure that analysis works correctly for repeated nodes, optional nodes, and alternation nodes
* [x] Ensure that analysis works correctly for parent symbols that may be aliases
* [x] Ensure that analysis works correctly (or degrades gracefully) for sub-patterns that require recursion through arbitrarily many hidden nodes
* [x] Optimization - prevent analysis from recurring through hidden nodes when it is clearly unnecessary
* [x] Optimization - iterate through possible child symbols more efficiently, taking advantage of the structure of "small parse states", in which valid lookaheads are already listed explicitly.


maxbrunsfeld:
Here's a GIF of the web playground with this feature enabled. You get live feedback on the structural soundness of your queries:

query-checking

maxbrunsfeld:
@ubolonton If you get a chance, could you try this branch in `emacs-tree-sitter`? To enable this analysis, you do have to rebuild the `tree-sitter` CLI and use it to regenerate the parsers, because some extra metadata is needed. With existing versions of the parsers, this PR will be a no-op.

I believe that the original bug is fixed. I also am hoping that you won't get any false positive *structure errors* when constructing queries. I believe the analysis logic is correct and it works for all of the queries in the `tree-sitter-*` repositories, but I know you have many of your own queries, and you may use some different patterns than I do.

If you don't have time to test right now, that's ok too. I might merge this tomorrow or the next day, and we can always fix up any bugs that we find after the fact.

vigoux:
@maxbrunsfeld we will try this too in neovim, and see if it fixes some errors on our side too.
Thanks for all the great work!

ubolonton:
@maxbrunsfeld I'm using this now. The original bug is fixed, and JSON syntax highlighting now works. I also tried it with all the languages for which `emacs-tree-sitter` brings its own highlight queries, and got no *structure errors*.

Thanks for working on this!

lib/src/array.h
```diff
@@ -66,6 +66,47 @@ extern "C" {
...
+      comparison = compare(&((self)->contents[mid_index] suffix), (needle)); \
+      if (comparison <= 0) *(index) = mid_index; \
+      size -= half_size; \
+    } \
+    comparison = compare(&((self)->contents[*(index)] suffix), (needle)); \
+    if (comparison == 0) *(exists) = true; \
+    else if (comparison < 0) *(index) += 1; \
+  } while (0)
+
+#define _compare_int(a, b) ((int)*(a) - (int)(b))
```
dcreager, line 88: Can you add a comment to clarify that only `a` is a pointer? Looking at the definition of `array__search_sorted` I think that's right, but it's definitely not typical for a binary search callback.

script/test
```diff
@@ -83,10 +83,14 @@ done
...
-else
-  top_level_filter=$1
+top_level_filter=$1
+
+if [[ \
+  -n $TREE_SITTER_TEST_LANGUAGE_FILTER || \
+  -n $TREE_SITTER_TEST_EXAMPLE_FILTER || \
+  -n $TREE_SITTER_TEST_TRIAL_FILTER \
+]]; then
+  echo ${top_level_filter:=corpus}
```
dcreager, line 93: Do you want to echo `corpus` here?  If not you can replace the `echo` with `:` to let the substitution assignment happen in an otherwise no-op statement.

maxbrunsfeld: Thanks, TIL about `:`!

I knew there was something like that, but forgot to follow up after hacking this in 😁 .

PR #653, opened 2020-06-20
Added working commands for Windows users

### LINKED ISSUE #652: Parser creation doc instructions fail on Windows 10
I was following [these docs](https://tree-sitter.github.io/tree-sitter/creating-parsers) on Windows 10. Came to this step:

> You can test this parser by creating a source file with the contents “hello” and parsing it:
> 
> ```
> echo 'hello' > example-file
> tree-sitter parse example-file
> ```
> This should print the following:
> 
> (source_file [1, 0] - [1, 5])

But if I run these exact commands on Windows 10 in PowerShell, I get the following error:

```
(ERROR [0, 0] - [1, 1]
  (ERROR [0, 0] - [1, 1]))
.\example-file  0 ms    (ERROR [0, 0] - [1, 1])
```

This is tricky and frustrating because the file *looks* as expected. Creating a new blank file and manually typing in "hello" results in the file parsing just fine. The `echo` command in PowerShell just inserts some weird invisible garbage characters at the start of the file. Even WinMerge can't tell the two files apart, I had to use `fc`:

```
>fc /b example-file-fail example-file-succeed
Comparing files example-file-fail and EXAMPLE-FILE-SUCCEED
00000000: FF 68
00000001: FE 65
00000002: 68 6C
00000003: 00 6C
00000004: 65 6F
00000005: 00 0D
00000006: 6C 0A
```

I think it has to do with windows-1252 vs iso-10646 character encodings.

Also: if you instead run the echo command in cmd instead of PowerShell you'll get a file containing the string "'hello'" (single quotes included).

I don't really know of a good cross-platform way to write this step, but figure it's worth reporting because it frustrated me for a good bit before figuring it out, and hopefully by reporting it I can give someone else a great experience with the tutorial!

ahelwer:
Windows users are most likely to be using the VS Code terminal, Windows Terminal, or PowerShell.

Fixes #652

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -77,10 +77,10 @@ This will generate the C code required to parse this trivial language, as well a
 You can test this parser by creating a source file with the contents "hello" and parsing it:
 
 ```sh
-echo 'hello' > example-file
+echo hello > example-file
 tree-sitter parse example-file
 ```
-Alternatively, in Windows Terminal:
+Alternatively, in Windows PowerShell:
 ```pwsh
 "hello" | Out-File example-file -Encoding utf8
 tree-sitter parse example-file
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -52,11 +52,6 @@ The last command will install the CLI into the `node_modules` folder in your wor
 # In your shell profile script
 export PATH=$PATH:./node_modules/.bin
 ```
-Alternatively, in Windows Terminal:
-```pwsh
-# In an init.ps1 file in the directory
-$env:Path += ";.\node_modules\.bin"
-```
 
 Once you have the CLI installed, create a file called `grammar.js` with the following contents:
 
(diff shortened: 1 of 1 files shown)


Neppord:
I think it should say "powershell" instead of "windows terminal". The normal instructions should work in "windows command line" (cmd.exe)

ahelwer:
@Neppord unfortunately the normal instruction does not quite work in cmd, it will put the text 'hello' in the file with single-quotes included. The single-quotes could be removed from this command and it would then work in both cmd and bash (I think), although it would still fail mysteriously in PowerShell/Windows Terminal/VS Code terminal.

Neppord:
I though VSCode terminal and windows terminal used CMD as standard and only used powers shell if you have selected it to do so.

I always configure the to use git bash. So I have no machine to test this hypothesis on.

I do really like your changes, and the details above doesn't matter.

ahelwer:
@Neppord who do I need to ping to get this merged?

ahelwer:
Closing since I doubt this will ever be merged.

ahlinc:
@ahelwer I'm sorry that this good docs additions didn't have attention. When I saw it before I wasn't sure that I can make such decisions to merge many PR on my own in this repo. I got a green light to make improvements in this repo that don't related to a core parser logic. You can freely ping me on every such case, I see all notices.

ahelwer:
@ahlinc oh okay neat! I do have this PR open that would handle a lot of questions about lexer precedence I think, I will look at polishing it a bit more maybe: https://github.com/tree-sitter/tree-sitter/pull/1612

Issue #541, opened 2020-02-15
Feature request: typescript definition for the DSL


SKalt:
**Why**: to give earlier, more informative feedback when writing a `grammar.js` in an IDE that understands typescript. Feedback might come in the form of error messages on incorrect DSL usage or hover-over documentation of specific DSL functions.

**How**: Either provide a typescript stub file as in PR #479 or rewrite the DSL in typescript as I've done at [github.com/skalt/ts-dsl-tree-sitter](https://github.com/skalt/ts-dsl-tree-sitter).  Note that the port of the DSL exactly reproduces the current behavior of the DSL (see the [test setup](https://github.com/SKalt/ts-dsl-tree-sitter/blob/91294bcd55fd2b99296d17a2ce280065d1656666/test/test.js) and [test results](https://travis-ci.com/SKalt/ts-dsl-tree-sitter/jobs/287620918)).

Adding stubs would require:
- including the .d.ts file in the NPM deploy of `tree-sitter-cli`
- trusting that the stub file correctly represents the activity of `dsl.js`.

Adopting the TypeScipt port of the DSL would involve:
- adding a build step to the NPM deploy of `tree-sitter-cli`
- adopting the tests of that the ported DSL creates `grammar.json`s as expected by other `tree-sitter-$LANG` packages.

I'm hoping the port and its test suite will make tweaking the JS DSL easier in the future.

jacobwgillespie:
I've opened #658 to implement stubs, as a place to start. 🙂

SKalt:
closed by PR #658

PR #658, opened 2020-06-22
Add TypeScript definition file for the DSL

jacobwgillespie:
This PR adds a TypeScript definition file for the DSL API, complete with documentation comments for IntelliSense. It models the behavior of the DSL functions, including all grammar options and extending from other grammars. The types will also infer and suggest rule names for code completion, although they still allow for referencing any symbol rule via `$.name` even if `name` is not a rule, since it may refer to one of the `externals`.

This is related to #541. This PR would likely replace #479.

## Usage

To use these types, the user would reference them via a TypeScript [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) at the top of their `grammar.js` to load the definitions into the ambient scope:

```typescript
```

Then if the user had `"checkJs": true` in their `tsconfig.json` or added `// @ts-check` to the top of their grammar file, TypeScript would check the types! ✨ 

## Existing Grammars

These types should be compatible with existing `tree-sitter-*` grammars. To demonstrate this, I have prepared a CodeSandbox example, which should allow you to hover over values to inspect their types, view function IntelliSense documentation, or experiment with adding a new rule to experience code completion:

➡️ [**CodeSandbox Example**](https://codesandbox.io/s/tree-sitter-typescript-types-demo-dmiem?file=/grammar-python.js)

Notes about this example:

* This example sets TypeScript's `strict` mode to `true`, and as such, there were a few places where JSDoc type annotations were added to the original to avoid TypeScript inferring the `any` type, for instance the `commaSep1` function at the bottom of `grammar-python.js`:

	```typescript
	/**
	 * @param {Rule} rule
	 */
	function commaSep1(rule) { ... }
	```    
* The grammar files use the triple-slash directive `<reference path="./dsl.d.ts" />` since they refer to a local copy of the `dsl.d.ts` file.
* CodeSandbox automatically runs Prettier on all files, but other than the addition of the declaration reference and `@type` JSDoc comments, the grammar code itself should be the same.

## Screenshots

Example screenshot of IntelliSense:

Screenshot 2020-06-23 at 00 18 56

Example screenshot of code completion:

Screenshot 2020-06-23 at 00 14 41

### DIFF
--- a/cli/npm/dsl.d.ts
+++ b/cli/npm/dsl.d.ts
@@ -1,0 +1,356 @@
+type AliasRule = {type: 'ALIAS'; named: boolean; content: Rule; value: string};
+type BlankRule = {type: 'BLANK'};
+type ChoiceRule = {type: 'CHOICE'; members: Rule[]};
+type FieldRule = {type: 'FIELD'; name: string; content: Rule};
+type ImmediateTokenRule = {type: 'IMMEDIATE_TOKEN'; content: Rule};
+type PatternRule = {type: 'PATTERN'; value: string};
+type PrecDynamicRule = {type: 'PREC_DYNAMIC'; content: Rule; value: number};
+type PrecLeftRule = {type: 'PREC_LEFT'; content: Rule; value: number};
+type PrecRightRule = {type: 'PREC_RIGHT'; content: Rule; value: number};
+type PrecRule = {type: 'PREC'; content: Rule; value: number};
+type Repeat1Rule = {type: 'REPEAT1'; content: Rule};
+type RepeatRule = {type: 'REPEAT'; content: Rule};
+type SeqRule = {type: 'SEQ'; members: Rule[]};
+type StringRule = {type: 'STRING'; value: string};
+type SymbolRule<Name extends string> = {type: 'SYMBOL'; name: Name};
+type TokenRule = {type: 'TOKEN'; content: Rule};
+
+type Rule =
+  | AliasRule
+  | BlankRule
+  | ChoiceRule
+  | FieldRule
+  | ImmediateTokenRule
+  | PatternRule
+  | PrecDynamicRule
+  | PrecLeftRule
+  | PrecRightRule
+  | PrecRule
+  | Repeat1Rule
+  | RepeatRule
+  | SeqRule
+  | StringRule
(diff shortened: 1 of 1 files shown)


cli/npm/dsl.d.ts
```diff
@@ -0,0 +1,356 @@
...
+  | SeqRule
+  | StringRule
+  | SymbolRule<string>
+  | TokenRule;
+
+type RuleOrLiteral = Rule | RegExp | string;
+
+type GrammarSymbols<RuleName extends string> = {
+  [name in RuleName]: SymbolRule<name>;
+} &
```
SKalt, line 40: Saying any string key of `$` will be a `SymbolRule` guarantees that prior grammars remain correct and is accurate to the how the `RuleBuilder` works.  However, allowing access to any key means typescript won't be able to catch the invalid reference `$.foobar`.  Does type inference work without the `& Record<string, SymbolRule<string>>`?

jacobwgillespie: Correct, the `Record<string, SymbolRule<string>>` is modelling the fact that `$` is a Proxy object, and is required specifically for `externals`. Consider this case:

```typescript
grammar({
  externals: ($) => [$._external],
  rules: {
    example: ($) => choice($._external),
  },
})
```

Without the `& Record<...>`, the only key that TypeScript would be able to infer would be `"example"`, and so the `$._external` would be a type error. 

There is unfortunately not a way for TypeScript to infer the usage of `$.name` inside the `externals` builder function, as there isn't currently a way to model an object in TypeScript where its keys return the string literal type of themselves (see https://github.com/microsoft/TypeScript/issues/22509 for specifically that request). 

If that _did_ exist, then we could write the type such that it inferred the names of all rules + all base grammar rules + the externals, but unfortunately it's just not possible. So today, the `& Record<...>` workaround is probably the best we can do.

If the DSL itself was changed in the future, you could do this:

```typescript
grammar({
  externals: () => ['_external'],
})

// or even just
grammar({
  externals: ['_external'],
})
```

And then it would be possible to infer the names exactly. But that's not compatible with current usage.

maxbrunsfeld:
This is really nicely done. Thanks!

PR #666, opened 2020-07-04
Fix wrong file name

intrigus-lgtm:
"build_fuzzers" -> "build-fuzzers".
It should be a hypen and not an underscore.


maxbrunsfeld:
Thanks!

PR #692, opened 2020-07-19
Add '.' as a valid start of a query predicate, in addition to '#'

ubolonton:
From the discussion at https://github.com/ubolonton/emacs-tree-sitter/issues/38#issuecomment-638964308.

This will make it more convenient to embed queries in Lisp dialects, for things like ad-hoc custom highlighting:

```emacs-lisp
;; Before:
(tree-sitter-hl-add-patterns 'python
  "((string) @constant
    (#match? @constant \"^'\"))")

;; After:
(tree-sitter-hl-add-patterns 'python
  ((string) @constant
   (.match? @constant "^'")))
```

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -2087,6 +2087,35 @@ fn test_query_disable_pattern() {
     });
 }
 
+#[test]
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks!

PR #696, opened 2020-07-23
size_t variables need %zu, not %lu

ret2libc:
I got these warnings while compiling a project that uses tree-sitter on windows:
```
c:\projects\radare2\shlr\tree-sitter\lib\src\alloc.h(48): warning C4477: 'fprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t' [C:\projects\radare2\build\shlr\cafa177@@tree_sitter@sta.vcxproj]
c:\projects\radare2\shlr\tree-sitter\lib\src\alloc.h(57): warning C4477: 'fprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t' [C:\projects\radare2\build\shlr\cafa177@@tree_sitter@sta.vcxproj]
c:\projects\radare2\shlr\tree-sitter\lib\src\alloc.h(66): warning C4477: 'fprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t' [C:\projects\radare2\build\shlr\cafa177@@tree_sitter@sta.vcxproj]
```

### DIFF
--- a/lib/src/alloc.h
+++ b/lib/src/alloc.h
@@ -45,7 +45,7 @@ static inline bool ts_toggle_allocation_recording(bool value) {
 static inline void *ts_malloc(size_t size) {
   void *result = malloc(size);
   if (size > 0 && !result) {
-    fprintf(stderr, "tree-sitter failed to allocate %lu bytes", size);
+    fprintf(stderr, "tree-sitter failed to allocate %zu bytes", size);
     exit(1);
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thank you!

PR #715, opened 2020-08-25
Fix a few cases of Clang 10 with UBSAN detecting undefined behavior

bfredl:
Clang 10 considers adding any offset, including 0, to the null pointer to be undefined behavior. `((void *)NULL) + 0 = kaboom`.

Noticed this when trying to compile neovim+tree-sitter after upgrading to clang 10.0.0. There might be more cases, but these are the cases detected by our test suite.


maxbrunsfeld:
> adding any offset, including 0, to the null pointer to be undefined behavior

Good to know!

Thanks for the fix @bfredl. Good to see you on here again.

Issue #707, opened 2020-08-05
TypeScript .d.ts file generator


asgerf:
Hi 👋 

I wrote a mapping from `node-types.json` to TypeScript `.d.ts` files here https://github.com/asgerf/dts-tree-sitter.

I was wondering if you'd want to have a tool like that in the tree-sitter org, or as part of the `node-tree-sitter` bindings, maybe?

jmbockhorst:
This tool is really helpful and I agree that it should be a part of tree-sitter, as it makes working with TypeScript much better.

DanielFlaum:
I would like this badly enough to investigate building and contributing it. But most of my experience is with writing my own code, not getting familiar with other people's.

If people are willing to answer some dumb questions from me I might be able to pull it off.

I guess my first questions are these:

* Would we want declaration files to be generated by default as part of the CLI's `generate` command, generated only when a flag is added to the `generate` command (i.e. `tree-sitter generate --ts-declarations`), or generated with a separate command (i.e. `tree-sitter build-ts`?
* If this generation takes places as part of the `generate` command, then I'd be doing the meat of my work in `tree-sitter/cli/src/generate`, right? (I ask because, again, I have little experience reading codebases I didn't write, and I don't have confidence yet in my ability to find the right place to start.)

Thanks!

dundargoc:
We accept contributions, PRs welcome.

No maintainer has indicated they actually want this (or not want this), so can't really comment on that. The best way to get a reaction is to usually threaten them with a PR.

PR #729, opened 2020-09-10
Web bindings: Auto-generate Node subclasses with getters for node fields

jmbockhorst:
Subclasses will only generate if the node-types object is passed as the second arg to `Parser.Language.Load`. This makes the web bindings more like the node bindings and was modeled after [this commit](https://github.com/tree-sitter/node-tree-sitter/commit/c69d03a9f5d5da44ef70fecd95a0fb10bf7bd6b9#diff-168726dbe96b3ce427e7fedce31bb0bc).

This makes tree-sitter for  very powerful using TypeScript when combined with a tool like described here: https://github.com/tree-sitter/tree-sitter/issues/707. As many users are using the web bindings in a node environment for cross-platform compatibility (like a LSP), I think it is beneficial to allow optional `node-types` to passed when loading the language so they can benefit from subclasses and extra type information.

The motivation for this is I'm implementing type inference for the Elm Language Server, and this will make things much cleaner. 
/cc @Razzeee

### DIFF
--- a/lib/binding_web/binding.c
+++ b/lib/binding_web/binding.c
@@ -30,12 +30,18 @@ static uint32_t byte_to_code_unit(uint32_t byte) {
   return byte >> 1;
 }
 
-static inline void marshal_node(const void **buffer, TSNode node) {
+static inline TSSymbol marshal_node(const void **buffer, TSNode node) {
   buffer[0] = (const void *)node.id;
   buffer[1] = (const void *)byte_to_code_unit(node.context[0]);
   buffer[2] = (const void *)node.context[1];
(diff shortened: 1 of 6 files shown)


Razzeee:
@maxbrunsfeld can you let us know if you are considering this?

Razzeee:
Travis error seems unrelated, probably just needs to be rerun

Razzeee:
@maxbrunsfeld any word on if this is wanted?

Alexizx00789999 (approved):

amaanq:
This will be re-designed and implemented when we improve typed node support for bindings across the board.

PR #733, opened 2020-09-14
Fix "Comparison between x of type TSSymbol and y of wider type unsigned int"

ret2libc:
Detected in:
https://github.com/radareorg/radare2/security/code-scanning/27?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/radareorg/radare2/security/code-scanning/26?query=ref%3Arefs%2Fheads%2Fmaster
https://github.com/radareorg/radare2/security/code-scanning/25?query=ref%3Arefs%2Fheads%2Fmaster

Probably the best thing would be to make TSSymbol uint32_t, but I'm not sure you want to do that. Maybe it breaks compatibility and such? Advices are welcome!

### DIFF
--- a/lib/src/language.c
+++ b/lib/src/language.c
@@ -87,7 +87,7 @@ TSSymbol ts_language_symbol_for_name(
 ) {
   if (!strncmp(string, "ERROR", length)) return ts_builtin_sym_error;
   uint32_t count = ts_language_symbol_count(self);
-  for (TSSymbol i = 0; i < count; i++) {
+  for (TSSymbol i = 0; i < count && i < UINT16_MAX; i++) {
     TSSymbolMetadata metadata = ts_language_symbol_metadata(self, i);
     if (!metadata.visible || metadata.named != is_named) continue;
     const char *symbol_name = self->symbol_names[i];
@@ -134,7 +134,7 @@ TSFieldId ts_language_field_id_for_name(
   uint32_t name_length
 ) {
   uint32_t count = ts_language_field_count(self);
-  for (TSSymbol i = 1; i < count + 1; i++) {
+  for (TSSymbol i = 1; i < count + 1 && i < UINT16_MAX; i++) {
     switch (strncmp(name, self->field_names[i], name_length)) {
       case 0:
         if (self->field_names[i][name_length] == 0) return i;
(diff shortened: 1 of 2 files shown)


ret2libc:
ping.

lib/src/language.c
```diff
@@ -87,7 +87,7 @@ TSSymbol ts_language_symbol_for_name(
 ) {
   if (!strncmp(string, "ERROR", length)) return ts_builtin_sym_error;
   uint32_t count = ts_language_symbol_count(self);
-  for (TSSymbol i = 0; i < count; i++) {
+  for (TSSymbol i = 0; i < count && i < UINT16_MAX; i++) {
```
maxbrunsfeld, line 90: Will this change silence the warning?

```c
for (TSSymbol i = 0; (uint32_t)i < count; i++) {
```

I think I'd rather add a cast than an additional runtime check.

ret2libc: I don't think that would work well. It would just hide the warning but it wouldn't remove the problem, which is that if count is bigger than uint16max this becomes an infinite loop because i  will always be less than count.

maxbrunsfeld: Actually, I think a better fix would be to simply change line 90 to this:

```
uint16_t count = ts_language_symbol_count(self);
```

In practice there will only ever be a few hundred "symbols" in a grammar. This is why we can use the type `uint16_t` to represent `TSSymbol`, which is good because it allows parser binaries to be much smaller. It follows that the variable `count`, which represents the number of distinct `TSSymbol` values in a parser, must itself be representable as `uint16_t`.

ret2libc: Then why does `ts_language_symbol_count` returns `uint32_t` (and `field_count` is `uint32_t` as well)? If you think this is the proper fix, then we should update that API and the type of `field_count` as well. If, however, you think that would break some users, probably just changing the type of the `count` variable in this function won't really fix the issue but just hide it again.

maxbrunsfeld: I think it would be valid for those functions to return `uint16_t`, but it's also fine to keep them as `uint32_t`. For right now, I'd rather leave them as they are.

> just changing the type of the `count` variable in this function won't really fix the issue but just hide it again.

There isn't *really* a bug here; all we need to do is help the static analyzer understand that `count` is, by the way that the parsers are constructed always going to be much less than `UINT16_MAX`.

ret2libc: @maxbrunsfeld is `count` the number of symbols defined by the user? Or is it something else (maybe some auto-generated stuff)?

maxbrunsfeld: There are some auto-generated ones as well as those defined by the user.

ret2libc: So are you sure auto-generated + a big grammar can't reach the limit?

PR #738, opened 2020-09-21
Allow tree queries to match on nodes' supertypes

maxbrunsfeld:
In a tree query, you can now match on a node's [*supertypes*](http://tree-sitter.github.io/tree-sitter/using-parsers#static-node-types). For example, in Python, `_expression` is a supertype. You could match any type of `_expression` within an `argument_list` with this pattern:

```clj
(argument_list
  (_expression) @argument)
```

You can also write a pattern that matches on both a node's *visible* type *and* one of its supertypes. For example, this pattern would only match `identifier` nodes that occur as *expressions* (as opposed to other identifier nodes which represent definitions, attributes, or keyword argument names).

```clj
(_primary_expression/identifier) @id
```

### DIFF
--- a/cli/src/generate/build_tables/minimize_parse_table.rs
+++ b/cli/src/generate/build_tables/minimize_parse_table.rs
@@ -68,6 +68,7 @@ impl<'a> Minimizer<'a> {
                             ..
                         } => {
                             if !self.simple_aliases.contains_key(&symbol)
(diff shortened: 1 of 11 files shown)


maxbrunsfeld:
Another change - supertype nodes no longer need to be prefixed with `_`; they will be automatically hidden. I think this makes more sense, because they are not fully "private" anymore. Soon I will update some of the grammars to remove the leading underscores on some of the supertypes.

PR #743, opened 2020-09-27
Fix compilation with TinyCC

XVilka:
image

As part of the [effort to make radare2 buildable with TinyCC](https://github.com/radareorg/radare2/pull/17295), which is the only option on some systems since easier to port rather than GCC or Clang.

Current [TinyCC compiler](https://bellard.org/tcc/) version is at https://repo.or.cz/tinycc.git/ - `mob` branch.

### DIFF
--- a/lib/src/atomic.h
+++ b/lib/src/atomic.h
@@ -3,7 +3,23 @@
 
 #include <stdint.h>
 
-#ifdef _WIN32
+#ifdef __TINYC__
+
+static inline size_t atomic_load(const volatile size_t *p) {
+  return *p;
+}
+
+static inline uint32_t atomic_inc(volatile uint32_t *p) {
+  *p += 1;
+  return *p;
+}
+
+static inline uint32_t atomic_dec(volatile uint32_t *p) {
+  *p-= 1;
+  return *p;
+}
+
+#elif defined(_WIN32)
 
 #include <windows.h>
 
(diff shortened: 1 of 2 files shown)


XVilka:
Anything I should do to get this merged?

lib/src/atomic.h
```diff
@@ -3,7 +3,23 @@
...
+
+static inline uint32_t atomic_inc(volatile uint32_t *p) {
+	*p += 1;
+	return *p;
+}
+
+static inline uint32_t atomic_dec(volatile uint32_t *p) {
+	*p-= 1;
+	return *p;
+}
```
maxbrunsfeld, line 20: Could you indent these lines similarly to the rest of the file?

lib/src/bits.h
```diff
@@ -7,7 +7,18 @@ static inline uint32_t bitmask_for_index(uint16_t id) {
...
-#if defined _WIN32 && !defined __GNUC__
+#ifdef __TINYC__
+
+static inline uint32_t count_leading_zeros(uint32_t x) {
+	int count = 0;
+	if (x == 0) return 32;
+	x = x - ((x >> 1) & 0x55555555);
+	x = (x & 0x33333333) + ((x >> 2) & 0x33333333);
+	count = (((x + (x >> 4)) & 0x0f0f0f0f) * 0x01010101) >> 24;
+	return count;
```
maxbrunsfeld, line 18: Is this algorithm based on something that's published? Maybe we should include a comment with a link.

maxbrunsfeld:
Thanks for the PR. I left a couple of bits of feedback.

XVilka:
Sorry for the delay. I fixed indentation and added a comment about the algorithm origin.

maxbrunsfeld:
Thanks!

PR #753, opened 2020-10-05
Allow an arbitrary curly-brace-delimited block after each query pattern

maxbrunsfeld:
This expands the query syntax to allow an arbitrary curly brace-delimited "metadata block" after each pattern in a query. For example:

```clj
(identifier)
{
  MY CUSTOM DATA HERE
}

(string_literal)
{
  {nested} (braces) [are] {{allowed}}
}
```

These blocks don't **do** anything. They just let you put your own data into the query which can be retrieved later. I don't know if this is generally useful, but it is helpful for an internal GitHub initiative, and it may be useful more generally, so I thought it would make sense to include in the core library.

### API

This adds a new C API for accessing the byte range associated with each pattern's "metadata block":

```c
bool ts_query_metadata_range_for_pattern(
  const TSQuery *self,
  uint32_t pattern_index,
  uint32_t *start,
  uint32_t *end
);
```

And a Rust binding for it, `Query::metadata_range_for_pattern()`.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -2412,6 +2412,51 @@ fn test_query_start_byte_for_pattern() {
 }
 
 #[test]
+fn test_query_pattern_metadata() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+        let source = r#"
+        (identifier)
+        {ok}
+
+        (string)
+
+        "{"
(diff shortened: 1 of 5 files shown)


amaanq:
It's been 4 years with no movement or activity from GH regarding this, I'll close this out

PR #766, opened 2020-10-16
[feat] add custom allocation override

ikrima:
# Add ability to override allocator

PR meant to be a starting point; happy to implement in a different way

## Problem
Integrating tree-sitter into our main app requires it to use our allocators
Windows does not easily provide a means to override weakly linked symbols (can be done; usually by inserting a trampoline)
We'd like to also leverage all of our existing memory tracking/profiling

## Fix: Macro override
- Simple,  minimal affect on rest of code base. User can define ts_malloc* 
```
#define ts_malloc(_sz)       es2capi_alloc(   OSAlctrHndlUid(), _sz,  0,      __FILE__, __LINE__)
#define ts_calloc(_cnt,_sz)  es2capi_calloc(  OSAlctrHndlUid(), _cnt, _sz, 0, __FILE__, __LINE__)
#define ts_realloc(_ptr,_sz) es2capi_realloc( OSAlctrHndlUid(), _ptr, _sz, 0, __FILE__, __LINE__)
#define ts_free(_ptr)        es2capi_free(    OSAlctrHndlUid(), _ptr, 0,      __FILE__, __LINE__)
```
- Retains default fallback 
- Benefit of macro is allows inserting tracking like __FILE__, __LINE__, etc

## Alternatives
* Add a config struct with callback handlers 
  * - ruffles my perf ocd sensibilities :P Users can decorate their functions with `__pure__ __restrict__` but esoteric knowledge
  * - not easy to handle tracking
* Other variants of the macro override

### DIFF
--- a/lib/src/alloc.h
+++ b/lib/src/alloc.h
@@ -25,6 +25,7 @@ bool ts_toggle_allocation_recording(bool);
 #else
 
 // Allow clients to override allocation functions
+
 #ifndef ts_malloc
 #define ts_malloc  ts_malloc_default
 #endif
--- a/lib/src/alloc.h
+++ b/lib/src/alloc.h
@@ -9,21 +9,6 @@ extern "C" {
 #include <stdbool.h>
 #include <stdio.h>
 
-// Allow clients to override allocation functions
-
-#ifndef ts_malloc
-#define ts_malloc(size)         ts_malloc_default(size)
-#endif
-#ifndef ts_calloc
-#define ts_calloc(count,size)   ts_calloc_default(count,size)
-#endif
-#ifndef ts_realloc
-#define ts_realloc(buffer,size) ts_realloc_default(buffer,size)
-#endif
-#ifndef ts_free
-#define ts_free(buffer)         ts_free_default(buffer)
-#endif
-
 #if defined(TREE_SITTER_TEST)
 
 void *ts_record_malloc(size_t);
@@ -32,23 +17,26 @@ void *ts_record_realloc(void *, size_t);
 void ts_record_free(void *);
 bool ts_toggle_allocation_recording(bool);
 
-static inline void *ts_malloc_default(size_t size) {
-  return ts_record_malloc(size);
-}
-
-static inline void *ts_calloc_default(size_t count, size_t size) {
-  return ts_record_calloc(count, size);
-}
-
-static inline void *ts_realloc_default(void *buffer, size_t size) {
-  return ts_record_realloc(buffer, size);
-}
-
-static inline void ts_free_default(void *buffer) {
-  ts_record_free(buffer);
-}
+#define ts_malloc  ts_record_malloc
+#define ts_calloc  ts_record_calloc
+#define ts_realloc ts_record_realloc
+#define ts_free    ts_record_free
 
 #else
+
+// Allow clients to override allocation functions
+#ifndef ts_malloc
+#define ts_malloc  ts_malloc_default
+#endif
+#ifndef ts_calloc
+#define ts_calloc  ts_calloc_default
+#endif
+#ifndef ts_realloc
+#define ts_realloc ts_realloc_default
+#endif
+#ifndef ts_free
+#define ts_free    ts_free_default
+#endif
 
 #include <stdlib.h>
 
--- a/lib/src/alloc.h
+++ b/lib/src/alloc.h
@@ -9,6 +9,21 @@ extern "C" {
 #include <stdbool.h>
 #include <stdio.h>
 
+// Allow clients to override allocation functions
+
+#ifndef ts_malloc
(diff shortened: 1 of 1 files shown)


lib/src/alloc.h
```diff
@@ -42,7 +42,20 @@ static inline bool ts_toggle_allocation_recording(bool value) {
...
+#define ts_calloc(_cnt,_sz)  ts_calloc_dflt(_cnt,_sz)
+#endif
+#ifndef ts_realloc
+#define ts_realloc(_ptr,_sz) ts_realloc_dflt(_ptr,_sz)
+#endif
+#ifndef ts_free
+#define ts_free(_ptr)        ts_free_dflt(_ptr)
+#endif
+
+static inline void *ts_malloc_dflt(size_t size) {
```
maxbrunsfeld, line 58: I'd prefer to just spell out "default" fully. Can you change these functions to `*_default`?

maxbrunsfeld (commented):
I think this approach is good. I suggested one naming change.

Also, ow that `ts_malloc` etc are macros, I think that for consistency we should also define these symbols as macros in the case where `TREE_SITTER_TEST` is defined (starting on line 20 of this file), instead of defining them as `static inline` functions. Could you make that adjustment too?

lib/src/alloc.h
ikrima: side note: i assume github actions runs through all the necessary tests? lmk if that's incorrect/docs on running any integration/build tests

lib/src/alloc.h
```diff
@@ -9,6 +9,21 @@ extern "C" {
...
+#endif
+#ifndef ts_calloc
+#define ts_calloc(count,size)   ts_calloc_default(count,size)
+#endif
+#ifndef ts_realloc
+#define ts_realloc(buffer,size) ts_realloc_default(buffer,size)
+#endif
+#ifndef ts_free
+#define ts_free(buffer)         ts_free_default(buffer)
+#endif
```
maxbrunsfeld, line 25: Sorry, my earlier feedback was a little bit vague. In that comment I made about the `TREE_SITTER_TEST` case, I actually meant something slightly different. I don't think we should let people override `ts_malloc` etc when running the tests; I just thought you could make the header a little bit simpler.

I was thinking something like this:

```c
#if defined(TREE_SITTER_TEST)

void *ts_record_malloc(size_t);
void *ts_record_calloc(size_t, size_t);
void *ts_record_realloc(void *, size_t);
void ts_record_free(void *);
bool ts_toggle_allocation_recording(bool);

#define ts_malloc  ts_record_malloc
#define ts_calloc  ts_record_calloc
#define ts_realloc ts_record_realloc
#define ts_free    ts_record_free

// ...

#endif
```

Then, below, (and outside of the previous `#if` condition) we could provide the default definitions:

```c
#ifndef ts_malloc
#define ts_malloc(size)         ts_malloc_default(size)
#endif
#ifndef ts_calloc
#define ts_calloc(count,size)   ts_calloc_default(count,size)
#endif
#ifndef ts_realloc
#define ts_realloc(buffer,size) ts_realloc_default(buffer,size)
#endif
#ifndef ts_free
#define ts_free(buffer)         ts_free_default(buffer)
#endif
```

Does that seem ok to you?

maxbrunsfeld: Also, a very minor question: For brevity, can we just define the default macros as plain name-substitutions? For example:

```c
#ifndef ts_malloc
#define ts_malloc ts_malloc_default
#endif
```

I think that since `malloc` and friends are standard functions, it's still readable and clear what the intended arguments are.

maxbrunsfeld (commented):
Thanks for the update! I left two more minor suggestions; don't mean to give you busy work.

lib/src/alloc.h
ikrima: sure, np! that makes more sense since I also thought it odd to allow user overrides in tests

re: #2 - _[tiny nitpick]_  imho, it's better to keep the args in the macros for clarity since the order of parameters changes based on `{platform} x {C,C++} x {langversion}`
granted, it's only for aligned allocations so not applicable for ones above hence the _[tiny nitpick]_

thoughts?

maxbrunsfeld: I guess on that issue, I'm ok with either way. But I think for these 4 standard allocation APIs, `malloc`, `calloc`, `realloc` and `free`, the arguments are pretty clear, and it reads a little more clearly without the argument lists.

ikrima: I'll keep it without the args then since i think consistent style >> nitpick

Swamped with work stuff so will fix it up in a couple of day

maxbrunsfeld:
This looks great. Just waiting for CI.

netjune:
Can you define these functions as callback functions?

PR #772, opened 2020-10-21
Expect text_callback to return a byte iterator rather than a byte slice

nathansobo:
I'm using Tree-sitter in a situation where I can't return byte slices from the `text_callback` passed to `QueryCursor::matches` and `QueryCursor::captures`, because my document is composed of many fragments in a rope-style data structure. To support this use case, this PR changes the expected return value of that callback to be an exact-sized byte *iterator* rather than a byte slice.

The big downside is that I have to copy the bytes into a scratch buffer in order to match regular expressions. However, the scratch buffer is allocated only once when the iterator is created. Other comparisons are performed with `Iterator::eq`, which I assume has performance similar to comparing byte slices, though I could be wrong about that.

One other idea I explored was to take a `TextProvider` object that can return references to itself and buffers internally. The problem with that approach was when I needed to compare the text of two different nodes. The internal buffer of the `TextProvider` can't be used for two different values at the same time.

Happy to explore other approaches with different trade-offs if you have any ideas. Also, no offense taken if you simply don't want to take on complexity to serve this use case. I can maintain this in a fork if necessary.

You know I have always thought this project is a masterpiece, and it keeps getting better.

### DIFF
--- a/cli/src/query.rs
+++ b/cli/src/query.rs
@@ -34,7 +34,7 @@ pub fn query_files_at_paths(
         let source_code = fs::read(&path).map_err(Error::wrap(|| {
             format!("Error reading source file {:?}", path)
         }))?;
-        let text_callback = |n: Node| &source_code[n.byte_range()];
+        let text_callback = |n: Node| source_code[n.byte_range()].iter().cloned();
         let tree = parser.parse(&source_code, None).unwrap();
 
         if ordered_captures {
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -3,6 +3,8 @@ use super::helpers::fixtures::get_language;
 use lazy_static::lazy_static;
 use std::env;
 use std::fmt::Write;
+use std::iter;
+use std::slice;
 use tree_sitter::{
     Language, Node, Parser, Query, QueryCapture, QueryCursor, QueryError, QueryErrorKind,
     QueryMatch, QueryPredicate, QueryPredicateArg, QueryProperty,
@@ -2968,6 +2970,6 @@ fn format_captures<'a>(
         .collect()
 }
 
-fn to_callback<'a>(source: &'a str) -> impl Fn(Node) -> &'a [u8] {
-    move |n| &source.as_bytes()[n.byte_range()]
+fn to_callback<'a>(source: &'a str) -> impl Fn(Node) -> iter::Cloned<slice::Iter<'a, u8>> {
+    move |n| source.as_bytes()[n.byte_range()].iter().cloned()
 }
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -101,7 +101,7 @@ where
 struct HighlightIterLayer<'a> {
     _tree: Tree,
     cursor: QueryCursor,
-    captures: iter::Peekable<QueryCaptures<'a, &'a [u8]>>,
+    captures: iter::Peekable<QueryCaptures<'a, std::iter::Cloned<std::slice::Iter<'a, u8>>>>,
     config: &'a HighlightConfiguration,
     highlight_end_stack: Vec<usize>,
     scope_stack: Vec<LocalScope<'a>>,
@@ -357,7 +357,7 @@ impl<'a> HighlightIterLayer<'a> {
                         vec![(None, Vec::new(), false); combined_injections_query.pattern_count()];
                     let matches =
                         cursor.matches(combined_injections_query, tree.root_node(), |n: Node| {
-                            &source[n.byte_range()]
+                            source[n.byte_range()].iter().cloned()
                         });
                     for mat in matches {
                         let entry = &mut injections_by_pattern_index[mat.pattern_index];
@@ -396,7 +396,7 @@ impl<'a> HighlightIterLayer<'a> {
                     unsafe { mem::transmute::<_, &'static mut QueryCursor>(&mut cursor) };
                 let captures = cursor_ref
                     .captures(&config.query, tree_ref.root_node(), move |n: Node| {
-                        &source[n.byte_range()]
+                        source[n.byte_range()].iter().cloned()
                     })
                     .peekable();
 
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -131,7 +131,7 @@ pub struct QueryMatch<'a> {
 }
 
 /// A sequence of `QueryCapture`s within a `QueryMatch`.
-pub struct QueryCaptures<'a, T: AsRef<[u8]>> {
+pub struct QueryCaptures<'a, T: ExactSizeIterator<Item = u8>> {
     ptr: *mut ffi::TSQueryCursor,
     query: &'a Query,
     text_callback: Box<dyn FnMut(Node<'a>) -> T + 'a>,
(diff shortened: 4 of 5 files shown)


maxbrunsfeld:
Thanks! I do definitely want to handle this use case well.

I was actually mistaken about what `FnMut` allowed you to do, so I thought this case was supported well already, via a pattern like this (which does not actually compile):

```rust
let buffer: &mut Vec<u8> = &mut self.buffer;
let captures = cursor.captures(&query, tree.root_node(), |node| {

    // (Assuming the rope has similar functionality to `xi_rope::Rope`)
    let mut chunks = rope.iter_chunks(node.byte_range());
    let first_chunk = chunks.next();

    match chunks.next() {
        // Often, the node only intersects one chunk.
        None => first_chunk.as_bytes(),

        // When it intersects multiple chunks, we incur some copying,
        // but still no allocation, since we're reusing this buffer.
        Some(next_chunk) => {
            buffer.clear();
            buffer.extend_from_slice(first_chunk.as_bytes());
            buffer.extend_from_slice(next_chunk.as_bytes());
            chunks.for_each(|c| buffer.extend_from_slice(c));
            buffer.as_slice()
        }
    }
});
```

Now I realize that this ☝️  isn't allowed, because the closure's return type would need to have the same lifetime as the closure itself, which can't be expressed until they add GATs.

> One other idea I explored was to take a `TextProvider` object that can return references to itself and buffers internally.

I think that I would prefer this `TextProvider` approach. The iterator solution is sensible, but I'm pretty sure the `TextProvider` solution would be more efficient in the most common code path, which is probably:
* The node is an `identifier` or something with a small range, and it only intersects one chunk of the rope
* We're doing a regex match

In this case, it should be possible to avoid any copying, because the regex can operate directly on the chunk of the rope. And in the case where copying is needed, the `TextProvider` implementation could reuse a buffer to avoid allocations.

> The problem with that approach was when I needed to compare the text of two different nodes.

Yeah, this is a downside. But I think this happens pretty rarely in the queries we're using on GitHub, so it is probably ok if there is one additional slice copy involved. We could still reuse a buffer for this too, that we could store on the `QueryCaptures` iterator.

Overall, it's hard to say what would have better performance, but I also think that it'd be nice to avoid abstracting a byte slice as a `Iterator<Item = u8>`. The compiler can probably optimize away all of that iteration abstraction, but it seems a little simpler in some sense to just deal with slices explicitly.

Maybe some interface like this?

```rust
pub trait TextProvider {
    fn call<'a>(&'a mut self, node: Node<'a>) -> &'a [u8];
}
```

And the cursor methods could take it via an `Into`?

```rust
impl QueryCursor {
    pub fn matches<'a>(
        &'a mut self,
        query: &'a Query,
        node: Node<'a>,
        text_callback: impl Into<Box<dyn TextProvider + 'a>>,
    ) -> impl Iterator<Item = QueryMatch<'a>> + 'a {
        // ...
    }

    pub fn captures<'a>(
        &'a mut self,
        query: &'a Query,
        node: Node<'a>,
        text_callback: impl Into<Box<dyn TextProvider + 'a>>,
    ) -> QueryCaptures<'a> {
        // ...
    }
}
```

Maybe the trait object isn't even needed. I'm not sure. I think this generalization might make the API nicer to use in some of the existing server-side applications, since we could just pass a byte slice directly, instead of having to make a closure.

dcreager:
The [bytes](https://docs.rs/bytes/0.6.0/bytes/) crate might be a good fit here.  If I'm following what y'all are describing is the [`Buf`](https://docs.rs/bytes/0.6.0/bytes/) trait.

maxbrunsfeld:
Yeah, it would be cool if there were some standard trait we could use to describe this.

The `bytes::Buf` trait is a bit different, because its API is focused on processing a byte sequence from beginning to end without backtracking. When using the `::captures` API in particular, the `Query` will return nodes in the order that they appear in the source, but the order in which textual predicates are called is not going to obey a strict ordering.

Another difference with this 'text provider' (and the reason it takes a syntax `Node` instead of just a byte range), is that you can also choose to use the node's `start_position()` and `end_position()` instead of its byte range, in case your text data structure is 'line-oriented' as some editors are. I'm not sure if that functionality is actually used by anyone right now. Maybe `emacs-tree-sitter`? But still I think it's a good feature to provide.

dcreager:
Can you use `Buf + Clone` for the backtracking part?  The `Bytes` type, which is the typical implementation of `Buf`, is cheaply cloneable via ref-counting (_not_ by copying the buffer content), but doesn't provide the rope-like behavior that @nathansobo needs.

maxbrunsfeld:
Yeah, `Buf + Clone` would make it at possible, but I don't think that would be an improvement, compared to just communicating our API's needs explicitly, via a one-method trait. It would be more complex on our end to deal with `Buf`, because it has a stateful API that doesn't necessarily return all of the bytes that you need in one call.

And I don't think it would aid in interoperability. For example, even if someone had their source code stored in a `byte::Bytes`, it would be inefficient for us to interact with that struct through `Buf + Clone`. Cloning implies manipulating atomic reference counts, in order to support multiple ownership, but we don't need that. All we're trying to do is access a particular offset in the buffer.

A custom trait is a nice because we can define `From` implementations for whatever is convenient:
* `&[u8]`
* `FnMut(Node) -> AsRef<u8>` (for compatibility with today)
* And maybe even `FnMut(Node, &'a mut Vec<u8>) -> &'a [u8]`, so that even when you want to make use of a reused buffer, for copying disjoint chunks, you can do by just writing a closure, and we'll automatically provide the buffer.

nathansobo:
Thanks for the thoughtful reply.

I had one additional idea. What if the trait defined predicate methods rather than returning a byte slice? It could define the following methods:

* `capture_eq_capture(&mut self, a: Node, b: Node) -> bool`
* `capture_eq_str(&mut self, node: Node, s: &str) -> bool`
* `capture_matches_regex(&mut self, node: Node, regexp: &Regex) -> bool`

This wouldn't require any copying in the default static case. It would also allow the regex to be run directly on the contents of the rope in cases where the node's range doesn't cross a fragment boundary. The downsides are that it potentially leaks implementation details such as the use of the `Regex` crate. The trait would also need to change if you introduced more types of text predicates. But assuming those trade-offs are acceptable, it seems like it offers the best possible performance while covering the rope use case.

Depending on your feedback, I'll try to make some time to iterate on a new approach this week.

maxbrunsfeld:
I'm open to that, but I feel like before we accept that increase in API complexity, it would be good to confirm that the performance improvement is potentially meaningful, compared to a simpler API that incurs more `memcpy`.

What do you think about profiling some scenario like this 👇 , if you haven't already:

* Set up your editor to just use the *existing* text callback API, with a callback that returns a `Cow<[u8]>`. So, `Cow::Borrowed([u8])` when the node only spans once chunk, and `Cow::Owned(Vec<u8>)` when it spans multiple chunks. This is worse than the alternatives we're talking about, because of the allocation in the `Owned` case, but might still be a useful base for comparison.
* Pick a language that has a fair amount of text predicates in the highlight query: probably [javascript](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/queries/highlights.scm) or [python](https://github.com/tree-sitter/tree-sitter-python/blob/master/queries/highlights.scm#L11-L15)? 
* Set up a text document with really really long variable names, to emphasize the the text copying / matching costs

And then with whatever workload, see if you can confirm that the text copying does indeed show up in the profile.

I know some performance improvements are worth doing a priori (since e.g. some users' machines are slower than any laptop you might have access to, etc), without needing to painstakingly prove their worth, but I'm just wondering if it's worth going all the way to that more specific API you outlined above, or if we should be satisfied with the single-method version, for the sake of avoiding the drawbacks that you outlined.

maxbrunsfeld:
BTW I think it's very cool that you're using Tree-sitter in your editor, and wouldn't want to you to have to maintain a fork to get the performance you want, so let me know if you're not in agreement about the direction.

ahlinc:
@maxbrunsfeld It seems this PR can be closed in favor of #1130 where the `TextProvider` approach was implemented.

maxbrunsfeld:
Yeah, you’re right; thanks!

PR #783, opened 2020-10-29
Rust bindings: add a TreeVisitor

aibaars:
This pull request adds a function for convenient Tree traversal. It's implemented with a TreeCursor and provides a higher level interface than directly using a TreeCursor.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -655,6 +655,12 @@ impl Tree {
         self.root_node().walk()
     }
 
+    /// Traverse the [Tree] starting from its root [Node] applying a visitor at
+    /// all nodes.
+    pub fn traverse(&self, visitor: &impl TreeVisitor) {
+        self.root_node().traverse(visitor)
+    }
+
     /// Compare this old edited syntax tree to a new syntax tree representing the same
     /// document, returning a sequence of ranges whose syntactic structure has changed.
     ///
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
In general, I've been trying to keep these libraries pretty minimal, not providing much functionality that could be implemented on top of the core primitives. In particular, I value having a minimal number of distinct nouns/concepts in the API.

On the other hand, I think you're right that the depth-first traversal loop can be a *little* tricky to write. It's kind of borderline, for me, as to whether it's worth abstracting. In the past, I've thought about adding some kind of iterator-based abstraction for it. Maybe something like:

```rust
for step in tree_cursor.traverse() {
  match step {
    TraversalStep::In(node) => {
      println!("entering node {:?}", node);

      // avoid visiting certain subtrees
      if some_condition(node) {
        step.skip();
      }
    }

    TraversalStep::Out(node) => {
      println!("leaving node {:?}", node);
    }
  }
}
```

The iterator API seems a bit more idiomatic to me than the visitor trait, based on my limited experience with the rust ecosystem, but I might be missing some benefits to it. What do you think?

aibaars:
> On the other hand, I think you're right that the depth-first traversal loop can be a _little_ tricky to write. It's kind of borderline, for me, as to whether it's worth abstracting.

Indeed. It took me a little while to correctly implement depth-first traversal with the TreeCursor interface. That's why I thought it would be good to make a PR so others don't have to do the same thing.

> The iterator API seems a bit more idiomatic to me than the visitor trait, based on my limited experience with the rust ecosystem. And it's a bit more flexible in terms of mutating your environment and use `break`,`continue`,`return`, etc. But I might be missing some benefits to the visitor. What do you think?

That's an interesting idea. I'll try to implement that.

ObserverOfTime:
Closing since an iterator would be preferred over a visitor.

PR #784, opened 2020-10-29
Ensure "extras" symbols are included in the node-types.json file

aibaars:
The symbols marked as "extras" are the start symbols of secondary
languages. These should be included in the aliases map just as done
for start symbol of the main language to ensure their node type and
field information is included in the node-types.json file.

### DIFF
--- a/cli/src/generate/node_types.rs
+++ b/cli/src/generate/node_types.rs
@@ -424,6 +424,14 @@ pub(crate) fn generate_node_types_json(
             aliases
         });
     }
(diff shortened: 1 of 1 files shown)


maxbrunsfeld (commented):
This code looks great to me. Could you add unit test coverage for this?

I think that you could just add a little bit to the first unit test in this file, `test_node_types_simple`.

aibaars:
@maxbrunsfeld I added a test case and checked that this change behaves correctly on the tree-sitter-ruby grammar.

maxbrunsfeld:
Thanks!

PR #786, opened 2020-10-30
Add the link to COBOL parser

XVilka:
See also discussion in https://github.com/github/semantic/issues/452

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -57,6 +57,7 @@ Parsers for these languages are fairly complete:
 Parsers for these languages are in development:
 
 * [Agda](https://github.com/tree-sitter/tree-sitter-agda)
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Based on the README and the small number of tests in the corpus folder, I think I'd like to hold off on recommending that anyone use it until it's a bit further along. Do you think I'm missing anything there?

Neppord:
@maxbrunsfeld no you are correct, this is not even working for small examples. Development have also slowed down due to low interest from the community and my focus needed elsewhere.

You can close this one and we can comment on reopen if the project gets closer to a finnish, thanks!

XVilka:
Yes, you are right. Lets wait then.

PR #787, opened 2020-10-30
Update TS definitions to support the Query API

wingrunr21:
Updates the `web-tree-sitter` TypeScript definitions to support the Query API

### DIFF
--- a/lib/binding_web/tree-sitter-web.d.ts
+++ b/lib/binding_web/tree-sitter-web.d.ts
@@ -48,6 +48,7 @@ declare module 'web-tree-sitter' {
     ) => string | null;
 
     export interface SyntaxNode {
+      id: number;
       tree: Tree;
       type: string;
       isNamed: boolean;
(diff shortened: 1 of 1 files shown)


Razzeee:
You might want to merge that with ours https://github.com/elm-tooling/elm-language-server/blob/master/src/typings/tree-sitter-web.d.ts

wingrunr21:
@Razzeee good call. done

lib/binding_web/tree-sitter-web.d.ts
```diff
@@ -131,8 +132,28 @@ declare module 'web-tree-sitter' {
...
+      operator: string;
+      operands: { name: string; type: string }[];
+    }
+
+    class Query {
+      captureNames: string[];
+
+      delete(): void;
+      matches(node: SyntaxNode, startPosition?: Point, endPosition?: Point): QueryResult[];
+      captures(node: SyntaxNode, startPosition?: Point, endPosition?: Point): QueryResult[];
```
maxbrunsfeld, line 155: The `captures` and `matches` methods return objects of different types. I would call one of them `QueryCapture`, and one of them `QueryMatch`.

A `QueryCapture` [has a `node` and a `name`](https://github.com/tree-sitter/tree-sitter/blob/99cd283e39d8dfb766fb365262fd08a419dd20a2/lib/binding_web/binding.js#L1044). A `QueryMatch` (as you wrote), has a `pattern` index, and a `captures` array, whose elements are `QueryCapture` objects.

wingrunr21: @maxbrunsfeld updated

wingrunr21:
@Razzeee note that the changes @maxbrunsfeld pointed out will be incompatible with your local typedefs

maxbrunsfeld:
The CI failure is because of a change to `tree-sitter-ruby`, which is used in some of the tests. I'll fix it up.

Razzeee:
@wingrunr21 thx,we'll be able to drop that, as soon as this lands, so that's fine

PR #816, opened 2020-11-24
highlight: Make README snippets compile

owlseye:
Some code examples in the highlight README require slight changes to compile, which is fixed by this patch.


maxbrunsfeld:
Thanks!

PR #819, opened 2020-11-26
web binding: fix equals()

joelspadin:
Node.equals() used to always return true. Now it works.
Also added unit tests for it.


maxbrunsfeld:
Thank you! Sorry for the confusion when building the bindings. Not yet sure what was going on there.

PR #827, opened 2020-12-02
cli test: use diff format for expected vs actual

milahu:
fix #655 and #823 

#823 - use better diff algorithm  
the old library sometimes fails to find the longest common sequence (LCS)

#655 - prettify output

old format

<tt>1 failure:</tt>

<tt><span color="green">expected</span> / <span color="red">actual</span></tt>

<tt>1. Name of test</tt>

<tt>(component (element (start_tag (tag_name)) (text) (element (start_tag (tag_name)) (text) <span color="red">(text_expression</span> <span color="green">(expression</span> (raw_text)) (end_tag (tag_name))) (text) (element (start_tag (tag_name) (directive_attribute (directive_name))) (text) <span color="red">(text_expression</span> <span color="green">(expression</span> (raw_text)) (end_tag (tag_name))) (text) (end_tag (tag_name))))</span></tt>

new format

```diff
1 failure:

1. Name of test

+++ expected
--- actual
  (component (element (start_tag (tag_name)) (text) (element (start_tag (tag_name)) (text)
+ (expression
- (text_expression
  (raw_text)) (end_tag (tag_name))) (text) (element (start_tag (tag_name) (directive_attribute (directive_name))) (text)
+ (expression
- (text_expression
  (raw_text)) (end_tag (tag_name))) (text) (end_tag (tag_name))))
```

benefit: compatibility with GNU diff, simply copy-paste to markdown <tt>```diff</tt>

cost: test output cannot be indented

alternative: produce html code to encode colors

todo
* allow to disable colors via CLI, similar to `diff --color=never`
* auto-indent the ascii tree
* find a better algorithm than myers algorithm, to avoid 'sliding errors' (patience diff or histogram diff), compare trees not strings

tree diff:
expected
```diff
  (fragment (element (self_closing_tag (tag_name)))
+ (text)
  )
```
actual
```diff
  (fragment (element (self_closing_tag
+ (tag_name))) (text))
- (tag_name))))
```

PR #832, opened 2020-12-04
Small improvements to test & update-corpus cycle

leostera:
Following #442 and #828, here's a few small changes that make the flow a little smoother, and one rather larger one that gives us S-expressions that are easier on the eyes, and should make for easier to read diffs with #827 as well.

#### Changes:

* Do not reprint the list of changed tests, just the number of them

* Use new character for updated tests (arrow pointing up)

* Depend on the `tree-sitter-sexpr` crate that implements a thin
  idiomatic Rust layer to parse and pretty print S-expressions on top of
  a tree-sitter generated parser fo them.

---

cc/ @maxbrunsfeld @ikatyang

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


amaanq:
Echoing what Max said, I don't think we want to bring in an external grammar, to avoid having a circular dependency, so closing

PR #836, opened 2020-12-07
Add link to R tree sitter grammar

jimhester:


maxbrunsfeld:
Oh wow, R! Great work! How complete is the parser so far? Have you used it on some popular R projects?

jimhester:
It is reasonably complete. It parses all of the code in the base and recommended packages that come with R distributions without failing. I also just checked it on all of the packages in the [tidyverse](https://github.com/tidyverse) and it parsed them as well without failure.

jimhester:
That being said, I am not an expert on writing tree sitter grammars, if you or someone else more familiar with tree sitter could review the code in the parser I am sure there are things that could be improved.

maxbrunsfeld:
It looks pretty good to me, based on a cursory read.

PR #856, opened 2020-12-20
Update CC for additional target support

0xADD1E:
This just explicitly sets a new minimum for `cc`, and allows tree-sitter to build on additional platforms, namely Apple M1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,9 +123,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.25"
+version = "1.0.66"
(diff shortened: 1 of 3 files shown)


0xADD1E:
I'm going to be honest I have no idea how updating cc affects test results at all, let alone changing the invalid ruby heredoc from having a method_call to a call. 

If someone more familiar with the project has any idea where I should be looking, that would be awesome.

maxbrunsfeld:
Hey, sorry for the confusion with the CI build. The Tree-sitter unit tests use a number of grammars that live in separate repositories. Occasionally, we change a grammar in a way that affects these unit tests, and then we have to make a corresponding change in this repository.

Issue #853, opened 2020-12-19
Build fails with gcc 11


agriffis:
I have a nightly build of tree-sitter and neovim on Fedora COPR, see https://copr.fedorainfracloud.org/coprs/agriffis/neovim-nightly/builds/

Lately the builds on Rawhide (which will become Fedora 34 in the spring) have started failing, with this error:

<details><summary>Click to see build error</summary>

```
Mock Version: 2.7
ENTER ['do_with_status'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target x86_64 --nodeps /builddir/build/SPECS/tree-sitter.spec'], chrootPath='/var/lib/mock/fedora-rawhide-x86_64-1608022300.079413/root'env={'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'}shell=Falselogger=<mockbuild.trace_decorator.getLog object at 0x7fc90051e910>timeout=0uid=1001gid=135user='mockbuild'nspawn_args=[]unshare_net=TrueprintOutput=True)
Executing command: ['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target x86_64 --nodeps /builddir/build/SPECS/tree-sitter.spec'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'} and shell False
Building target platforms: x86_64
Building for target x86_64
setting SOURCE_DATE_EPOCH=1607990400
Wrote: /builddir/build/SRPMS/tree-sitter-0.17.3.0.git.2787.d6a3e4c9-1.fc34.src.rpm
Child return code was: 0
[...]

PR #858, opened 2020-12-22
Fix build with GCC 11

agriffis:
It seems redundant to array_init() when the structure is zeroed already
in the static assignment. But probably the fact that this code juggles
between TSTreeCursor and TreeCursor, without using a union, leaves the
compiler unable to tell that it's been initialized.

Fixes https://github.com/tree-sitter/tree-sitter/issues/853


maxbrunsfeld:
Thanks!

PR #863, opened 2020-12-26
Add --xml option to save the parsed code into XML

yijunyu:
I work on Huawei's Trusted Programming Language project, which was recently asked to analyse the source code files of multiple programming languages.  We believe `tree-sitter` is a suitable tool for this purpose. However, `tree-sitter` saves the parsed code into a format that is not interoperable with other tools we use yet. Hence, I've completed a working prototype for saving parsed source code into XML, and add `--xml` or `-x` option to the CLI: 
```bash
tree-sitter parse -q -x <path_to_code>
```
Here the `<path_to_code>` is the path to a source code file.  As a result, the output code will be printed to the standard output in XML format. The XML elements are generated from the `node.kind()`, and the textual value of these elements are extracted from the `source_code` string in-between the `start` and the `end` positions.  If these text strings contain special characters, they will be escaped into XML-compatible entities using the `html-escape` crate, hence the extra dependency. 

If someone with more experience developing in `tree-sitter` can give this an early look and send feedback, that would be really helpful. Apologies in advance for the current lack of tests and limited documentation. I'll be working on this full time for a while, so it will improve rapidly.

Thanks!

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -40,6 +40,7 @@ pub fn parse_file_at_path(
     timeout: u64,
     debug: bool,
     debug_graph: bool,
+    debug_xml: bool,
     cancellation_flag: Option<&AtomicUsize>,
 ) -> Result<bool> {
     let mut _log_session = None;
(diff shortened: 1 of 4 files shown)


cli/src/parse.rs
```diff
@@ -151,6 +152,65 @@ pub fn parse_file_at_path(
...
+                        needs_newline = true;
+                    }
+                    if cursor.goto_first_child() {
+                        did_visit_children = false;
+                        indent_level += 1;
+                    } else {
+                        did_visit_children = true;
+                        let start = node.start_position();
+                        let end = node.end_position();
+                        let value = std::str::from_utf8(
```
maxbrunsfeld, line 201: It looks like you could just use `node.start_byte()` and `node.end_byte()` here, instead of needing to convert the row/column positions back into byte offsets.

cli/src/tests/corpus_test.rs
```diff
@@ -390,7 +390,9 @@ fn flatten_tests(test: TestEntry) -> Vec<(String, Vec<u8>, String, bool)> {
                 }
                 result.push((name, input, output, has_fields));
             }
-            TestEntry::Group { mut name, children, .. } => {
+            TestEntry::Group {
+                mut name, children, ..
+            } => {
```
maxbrunsfeld, line 395: For some of these files, it looks like the only changes are formatting. Could you remove these changes, so that the only files affected by the PR are the ones with code changes?

cli/src/parse.rs
```diff
@@ -277,6 +337,21 @@ fn parse_edit_flag(source_code: &Vec<u8>, flag: &str) -> Result<Edit> {
...
+fn offset_for_lines(input: &Vec<u8>) -> Vec<usize> {
+    let mut offsets = Vec::<usize>::new();
+    offsets.push(0);
+    for (i, c) in input.iter().enumerate() {
+        if *c as char == '\n' {
+            offsets.push(i + 1);
+        }
+    }
+    offsets
+}
```
maxbrunsfeld, line 349: If I understand correctly, you wouldn't need these functions if you updated your code in `parse_file_at_path` to just use the `Node::start_byte` and `Node::end_byte` methods.

maxbrunsfeld (changes requested):
I think this is a good feature. I suggested a few changes.

cli/src/parse.rs
yijunyu: These are really great suggestions, Max, making my changes much simplified. I have just fixed those and reverted the changes to the original code introduced by `cargo-fm`.

maxbrunsfeld (approved):

maxbrunsfeld (approved):
This looks great now. Thanks! I'll just wait for CI to complete.

PR #866, opened 2020-12-31
Fix typo in document

rhysd:
I found missing backtick in document. This PR fixes it.

https://docs.rs/tree-sitter/0.17.1/tree_sitter/struct.Query.html#method.general_predicates


maxbrunsfeld:
Thanks!

Issue #752, opened 2020-10-05
Comments in test files


alemuller:
Allow (or document) how to add comments in test file (`corpus/*`).

Comments can be useful to annotate the expected s-expression, and to fold the text in vim with the default folding expression (if someone already have a folding expression for the corpus format fell free to post here).

maxbrunsfeld:
I think this would be a good feature. I think we should allow `;` as a comment character, like in lisp.

PR #871, opened 2021-01-06
Add ";" as comment in s-expressions

marcel0ll:
Remove everything after a ";" 

So
```
; Comment
(a 
  ; Comment
  (b)
)
```

Becomes
```
(a (b))
```

Related issue: https://github.com/tree-sitter/tree-sitter/issues/752

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -22,6 +22,7 @@ lazy_static! {
         .multi_line(true)
         .build()
         .unwrap();
+    static ref COMMENT_REGEX: Regex = Regex::new(r";.*").unwrap();
     static ref WHITESPACE_REGEX: Regex = Regex::new(r"\s+").unwrap();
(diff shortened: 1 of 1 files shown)


marcel0ll:
Have to take into account ";" in languages...

marcel0ll:
Apparently, error during tests is due to a change in the ruby grammar that was already incorporated inside tree-sitter tests but not actually released by tree-sitter-ruby.

I could not think of a reliable way to write a regex that would understand the difference between s-expression containing a ";", ex: '(MISSING ";"), 'and a final "; comment" so I this PR would only accept comments at line start.

maxbrunsfeld:
This is great work. Thanks for writing a good test for it as well, and sorry for the long delay in merging.

marcel0ll:
Glad to contribute

PR #882, opened 2021-01-17
Add index.js wrapper for wasm-bindgen compatibility

hvithrafn:
This PR adds an `index.js` wrapper module that simply imports and re-exports from the underlying bindings module.

For background, I'm trying to use `web-tree-sitter` with a wasm-bindgen project [`web-tree-sitter-sys`](https://github.com/hvithrafn/web-tree-sitter-sys).

However, when I try to call `Parser::init()` from Rust, based on these wasm-bindgen bindings, I get an error that `init` is undefined:

```rust
#[wasm_bindgen(module = "web-tree-sitter")]
extern {
    #[derive(Clone, Debug)]
    pub type Parser;

    #[wasm_bindgen(static_method_of = Parser)]
    pub fn init() -> Promise;

    // ...
}
```

I found [this issue](https://github.com/rustwasm/wasm-bindgen/issues/1202), which seems to be related.

Based on what was discussed there, I figured that adding an `index.js` module like I've done in this PR might solve the problem, and indeed it does. With this change, I can use the wasm-bindgen bindings without errors.

If adding this wrapper `index.js` module doesn't cause any other problems, could you please merge this and make a new release of `web-tree-sitter`?

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -1145,3 +1145,4 @@ function marshalEdit(edit) {
 }
 
 Parser.Language = Language;
+Parser.Parser = Parser;


hvithrafn:
Actually it looks like this `index.js` wrapper breaks the test suite so maybe this isn't the best approach. I don't really understand why it stops working with JS but now seems to work okay with wasm-bindgen.

In any case, do you have some thoughts about how to resolve this in a better so that the library can be used with wasm-bindgen?

hvithrafn:
So after I first made this PR and the tests broke on CI, I did a little investigating in order to determine why the tests were now failing but things were working okay with wasm-bindgen.

I guess the crux of the issue, which I didn't realize before, is that with wasm-bindgen there isn't really a way to even refer to the `Parser` class, since it is exported directly (i.e., like `exports = Parser` rather than `exports.Parser = Parser`).

This PR changes that, but it also means that you now have to import like this:

```js
const { Parser } = require('web-tree-sitter');
```

But this change will break existing code so I don't know what you think about that. If you'd rather not merge this I could just publish a compatibility module, although it would be nice not to have to do that.

maxbrunsfeld:
Hmm, I am open to this kind of change, if it's really true that "default exports" (`export = A`) are less easy to use. This seems like a missing feature in `wasm-bindgen` though, since default exports are a valid JavaScript pattern.

Alternatively, could we just set `Parser.Parser = Parser`? That way, I think both styles of imports would work.

hvithrafn:
> This seems like a missing feature in `wasm-bindgen` though, since default exports are a valid JavaScript pattern.

I agree. This is actually something I've wanted before with `wasm-bindgen`.

> Alternatively, could we just set `Parser.Parser = Parser`? That way, I think both styles of imports would work.

I think that should work. I'll try to update the PR in a bit.

hvithrafn:
@maxbrunsfeld okay, I changed it similar to what you suggested. Seems to work.

maxbrunsfeld:
Ok, glad to hear that it works.

I think that, instead of adding an `index.js`, you could just add this additional line in `binding.js`. `Parser.Parser = Parser;`.

hvithrafn:
> I think that, instead of adding an `index.js`, you could just add this additional line in `binding.js`: `Parser.Parser = Parser;`, or something like that.

Oh, I see what you mean. I'll give that a try.

hvithrafn:
@maxbrunsfeld okay, I changed it to where the named `Parser` export is made in `binding.js` and tested that it still works. Should be good to go I think.

PR #883, opened 2021-01-17
Refine web bindings typings

silvanshade:
This PR refines the `web-tree-sitter` typings a bit.

I noticed in the tests that it's possible to set the language and logger to null, and of course sometimes getting the language and logger will also return null. So it seems a little more accurate to specify this in the typings as well.

### DIFF
--- a/lib/binding_web/tree-sitter-web.d.ts
+++ b/lib/binding_web/tree-sitter-web.d.ts
@@ -1,8 +1,8 @@
 declare module 'web-tree-sitter' {
   class Parser {
     static init(): Promise<void>;
-    delete(): void;
-    parse(input: string | Parser.Input, previousTree?: Parser.Tree, options?: Parser.Options): Parser.Tree;
+    delete();
+    parse(input: string | Parser.Input, previousTree?: Parser.Tree, options?: Parser.Options): Parser.Tree | null;
     getLanguage(): Parser.Language | null;
     setLanguage(language?: Parser.Language): void;
     getLogger(): Parser.Logger | null;
@@ -61,8 +61,8 @@ declare module 'web-tree-sitter' {
       startIndex: number;
       endIndex: number;
       parent: SyntaxNode | null;
-      children: Array<SyntaxNode>;
-      namedChildren: Array<SyntaxNode>;
+      children: SyntaxNode[];
+      namedChildren: SyntaxNode[];
       childCount: number;
       namedChildCount: number;
       firstChild: SyntaxNode | null;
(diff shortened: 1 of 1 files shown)


hvithrafn:
I've added some additional refinements, specifying that several other methods can return null (based on reading the Rust API). I also added some bindings that seemed to be missing from the TS typings.

I also wanted to ask about some of the differences between the web bindings and the Rust API. I noticed that some methods from the Rust API are not included in the web bindings and several things have different names.

Did they just evolve separately or is there some other reason for wanting it to be that way?

The reason I'm asking is because I'm planning to use both the Rust API and the web API in my language server implementation (the WebAssembly language server), where the latter will be used for running the server in a browser context.

In order to do that, I'll probably need to implement a sort of compatibility layer over both APIs using traits. It might be easier/cleaner to accomplish if the two APIs were a bit more uniform.

I can try to submit some more PRs if there is interest in going in that direction.

hvithrafn:
Not sure why this is failing on that one test.

lib/binding_web/tree-sitter-web.d.ts
```diff
@@ -1,12 +1,15 @@
...
-    setLanguage(language: any): void;
-    getLogger(): Parser.Logger;
-    setLogger(logFunc: Parser.Logger): void;
+    delete();
+    parse(input: string | Parser.Input, previousTree?: Parser.Tree, options?: Parser.Options): Parser.Tree | null;
+    getLanguage(): Parser.Language | null;
+    setLanguage(language?: Parser.Language): void;
+    getLogger(): Parser.Logger | null;
+    setLogger(logFunc?: Parser.Logger): void;
+    reset(): void;
```
maxbrunsfeld, line 10: For functions with no return value, is the `void` useful or not? I noticed you removed some of the explicit `void` annotations, but not all.

maxbrunsfeld (commented):
I think it would be good if the rust and web APIs were as uniform as possible, though there are some differences in idioms from JavaScript to Rust. Feel free to propose some changes, there are likely improvements we can make.

lib/binding_web/tree-sitter-web.d.ts
hvithrafn: It doesn't make any semantic difference as far as I know. I think the IDE tooling or something might warn you about the explicit ones which is why I might have removed them. It's been awhile since I looked at this though, so I don't remember for sure now.

hvithrafn:
> I think it would be good if the rust and web APIs were as uniform as possible, though there are some differences in idioms from JavaScript to Rust. Feel free to propose some changes, there are likely improvements we can make.

Okay. It may be awhile before I can get to this but it would be nice to eventually make things a bit more uniform where possible.

On a related note, I've created two crates for working with the JavaScript bindings in Rust:

- [web-tree-sitter-sys](https://github.com/hvithrafn/web-tree-sitter-sys)
- [tree-sitter-facade](https://github.com/hvithrafn/tree-sitter-facade/tree/main/src)

The `web-tree-sitter-sys` crate just provides the raw interface to the JavaScript bindings for wasm-bindgen projects.

The `tree-sitter-facade` crate provides a uniform interface to both the native Rust `tree-sitter` crate or the `web-tree-sitter-sys` crate depending on whether the compilation target is native or `wasm32-unknown-unknown`.

For the most part, the `tree-sitter-facade` interface is pretty close to what you get directly with `tree-sitter`: [docs.rs/tree-sitter-facade](https://docs.rs/tree-sitter-facade/0.0.5/tree_sitter_facade/)

It might be interesting to consider moving some of this stuff into `tree-sitter` itself rather than maintaining these separate crates. Maybe you have some thoughts about that.

PR #896, opened 2021-01-25
Modify Language.load to accept bytes directly

hvithrafn:
This PR modifies `Language.load` to also accept a `Uint8Array` so that the language grammar wasm bytes can be passed in directly rather than loaded from a URL.

This pattern seems to be more suitable for using `web-tree-sitter` with my wasm-bindgen bindings. It's easier to just include the grammar bytes with `include_bytes!` since cargo doesn't make it easy to control how external assets are installed and later referenced.

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -856,26 +856,31 @@ class Language {
     );
   }
 
-  static load(url) {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks!

Issue #778, opened 2020-10-23
Multiple regex ranges slow parser generation down.


BonaBeavis:
The following grammar takes relative to its size a long time to generate.

```javascript
module.exports = grammar({
  name: 'sparql',

  rules: {
    pn_chars_base: $ => choice(
      /[A-Z]/,
      /[a-z]/,
      /[\u00C0-\u00D6]/,
      /[\u00D8-\u00F6]/,
      /[\u00F8-\u02FF]/,
      /[\u0370-\u037D]/,
      /[\u037F-\u1FFF]/,
      /[\u200C-\u200D]/,
      /[\u2070-\u218F]/,
      /[\u2C00-\u2FEF]/,
      /[\u3001-\uD7FF]/,
      /[\uF900-\uFDCF]/,
      /[\uFDF0-\uFFFD]/,
      /[\u{10000}-\u{EFFFF}]/u
    ),

    pn_chars_u: $ => choice(
      $.pn_chars_base,
      '_'
    ),
  }
})
```

PR #904, opened 2021-01-28
Represent CharacterSet internally as a vector of ranges

maxbrunsfeld:
Fixes https://github.com/tree-sitter/tree-sitter/issues/778

### Background

The Tree-sitter generator internally uses a type called `CharacterSet` to represent character sets associated with lexical state transitions. Currently, a `CharacterSet` is represented as a simple sorted vector of characters. In order to deal with negated character classes like `[^"]`, it is an `enum` with two variants: `Include` and `Exclude`.

During parser generation, we do a lot of `CharacterSet` copies and set-intersection operations. If grammars use regexes with very large character classes, these operations (on very large vectors) can become pretty slow.

### Change

This PR changes the internal representation of `CharacterSet` so that it uses *ranges* of characters. Because most large character classes contain relatively few disjoint *ranges* of characters, they can be represented much more compactly with this scheme.

The new representation also simplifies the code in some ways, because we don't need a special `Exclude` variant. The character set `[^"]` can be represented with just two ranges: `0..'"'` and `('"' + 1)..char::MAX` (in other words, the range of all characters *less than* `"`, and the range of all characters *greater than* `"`.

This is one of the last steps toward being able to efficiently support *unicode property escapes* in regexes.

### DIFF
--- a/cli/src/generate/build_tables/mod.rs
+++ b/cli/src/generate/build_tables/mod.rs
@@ -13,7 +13,7 @@ use self::minimize_parse_table::minimize_parse_table;
 use self::token_conflicts::TokenConflictMap;
 use crate::error::Result;
 use crate::generate::grammars::{InlinedProductionMap, LexicalGrammar, SyntaxGrammar};
-use crate::generate::nfa::{CharacterSet, NfaCursor};
+use crate::generate::nfa::NfaCursor;
 use crate::generate::node_types::VariableInfo;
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
This does change the generated parsers, because the ordering semantics of `CharacterSet` has slightly changed.

In addition to running all the tests (obviously), I've manually inspected the diffs in all the `ts_lex` functions, and it looks like nothing significant has changed.

PR #905, opened 2021-01-29
Fix hiding problems in ./build/Debug/tree_sitter_*_binding.node

ahlinc:
In debug building modules also may happen errors and a current implementation
completely hides them, so errors like 'undefined symbol' can't be
easily identified due to wrong traceback and error message.


maxbrunsfeld:
Thanks! This is a nice improvement.

PR #908, opened 2021-02-01
Add several bindings for Language

hvithrafn:
This PR adds several bindings for `Language` to `web-tree-sitter` to bring it closer to the Rust crate API.


lib/binding_web/test/language-test.js
```diff
@@ -0,0 +1,100 @@
...
+
+  before(async () =>
+    ({JavaScript} = await require('./helper'))
+  );
+
+  describe(".fieldCount", () => {
+    it("returns a number", () => {
+        assert.equal(34, JavaScript.fieldCount);
+    });
+  });
```
maxbrunsfeld, line 14: It looks like the indentation is inconsistent in some parts of this file. Could you adjust it to use 2 spaces throughout?

lib/binding_web/binding.js
```diff
@@ -646,6 +646,34 @@ class Language {
     return this.fields[fieldId] || null;
   }
 
+  idForNodeType(type, named) {
+    const typeLength = lengthBytesUTF8(type);
+    const typeAddress = C._malloc(typeLength + 1);
+    stringToUTF8(type, typeAddress, typeLength + 1);
```
maxbrunsfeld, line 652: I think you need to call `free` on this pointer to avoid a memory leak here.

lib/binding_web/binding.c
```diff
@@ -184,6 +184,20 @@ TSTree *ts_parser_parse_wasm(
...
+/**********************/
+/* Section - Language */
+/**********************/
+
+int ts_node_type_is_named_wasm(const TSLanguage *self, TSSymbol typeId) {
+  const TSSymbolType symbolType = ts_language_symbol_type(self, typeId);
+  return symbolType == TSSymbolTypeRegular;
+}
+
+int ts_node_type_is_visible_wasm(const TSLanguage *self, TSSymbol typeId) {
```
maxbrunsfeld, line 196: For consistency with the other functions, could you rename these new methods so that they begin with the prefix `ts_language_`?

maxbrunsfeld (commented):
This is great; I left a few minor suggestions.

hvithrafn:
@maxbrunsfeld thanks for the feedback. I made the suggested changes.

PR #911, opened 2021-02-03
Implement Clone for TreeCursor

michalmuskala:
Cloning was supported by the C library, but not exposed in Rust bindings


maxbrunsfeld:
Thanks!

By the way sorry for any confusion with the Travis CI build - we are in the process of switching away from Travis.

PR #914, opened 2021-02-05
Add fixture tests for different line endings

### LINKED ISSUE #913: Fixture tests does not test different line endings
Some parsers fail when a file has `\r\n` line endings instead of `\n`. Different line endings are not really tested which led to this issue being overlooked. More details in #717.

lukedan:
Fix #913. We still need to decide whether to include `\r` as a supported line ending. Also tests will fail, and will take *substantially* longer.

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -14,11 +14,13 @@ use std::str;
 use tree_sitter::{Language, LogType, Parser, Query};
 
 lazy_static! {
-    static ref HEADER_REGEX: ByteRegex = ByteRegexBuilder::new(r"^===+\r?\n([^=]*)\r?\n===+\r?\n")
+    // Since parse_test_content() normalizes the line endings, it's not necessary to account for \r\n in these regexs
+    // If you're using these somewhere else, you may need to adding them back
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
I don't think I want to use this approach of randomly mutating files' line endings. I would rather just add an individual test file, for example in the `tree-sitter-bash` repository, that tests a few simple examples with CRLF line endings.

lukedan:
I believe that there are already test cases in the Ruby parser for CRLF line endings - and they failed to find the issues. The reason I added the random tests is to test mixed line endings - although I agree that they're probably not that effective. It may be a better idea to keep the LF and CRLF tests and only remove the random tests. We can decide how to test mixed line endings once we've encountered such issues.

lukedan:
@maxbrunsfeld Any thoughts about my proposal?

lukedan:
@maxbrunsfeld It's been a while - I'll work on reporting the issues for the respective languages. Let me try to explain my opinion again - for any feature of any parser, it should be *required* that it can handle both LF and CRLF line endings. Adding extra tests is just wasted human labor since using existing tests is both easier and more effective.

PR #915, opened 2021-02-05
updated docs & tests for tree-sitter-svelte

Himujjal:
As mentioned I have updated the repo to accomodate Svelte grammar in playground and in the main index.md file.


maxbrunsfeld:
I don't think we want to incorporate the Svelte grammar into the core test suite. Also, for the playground, don't necessarily want to add languages right now.

I would like to add a link to the docs page though. If you back out the other changes, I'll merge just the change to `docs/index.html`.

PR #917, opened 2021-02-08
parser.c - added LOG_LOOKAHEAD macro to escape \n in logging

ahlinc:
Current debug log output hard to understand due to broken lines for the parser part logging when it prints unnamed symbols where new line symbols may happen. The lexer solves this problem by printing just a symbol code instead of a symbol itself and this works for lexer because every time it prints just one symbol. On the other hand the parser prints one or several symbols at once and approach with symbol codes wouldn't be easy understandable. This patch introduces a new preprocessor macro that escapes \s symbols < 127 in the range of codes 9 - 13 except the code 32 - space symbol  itself.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -25,33 +25,34 @@
     ts_parser__log(self);                                                                   \
   }
 
-#define LOG_LOOKAHEAD(symbol_name, size)                     \
-  if (self->lexer.logger.log || self->dot_graph_file) {      \
-    char *buf = self->lexer.debug_buffer;                    \
-    const char *symbol = symbol_name;                        \
-    int off = sprintf(buf, "lexed_lookahead sym:");          \
-    for (                                                    \
-      int i = 0;                                             \
-      symbol[i] != '\0'                                      \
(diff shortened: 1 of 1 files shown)


lib/src/parser.c
```diff
@@ -477,10 +512,10 @@ static Subtree ts_parser__lex(
       self->language
     );
 
-    LOG(
-      "lexed_lookahead sym:%s, size:%u, character:'%c'",
+    LOG_LOOKAHEAD(
       SYM_NAME(ts_subtree_symbol(result)),
       ts_subtree_total_size(result).bytes,
+      "character:'%c'",
```
maxbrunsfeld, line 518: I don't think this `character` part needs to be logged; I think I left it in by accident. It looks like, if you removed this, you could make the macro simpler (no varargs).

ahlinc: I don't remember that this was useful for me, so I completely agree with you. Let's make it simpler, I've pushed additional commit. Also added a backslash escaping to make it clear what it is in case of \\\n.

maxbrunsfeld:
Thanks!

Issue #919, opened 2021-02-09
CLI compilation error introduced by b661050a61c1c318acca2309003fc4e5f0905b58


ahlinc:
## Error
Linker produces a huge error output that contains many complains about undefined symbols.
Bellow is an excerpt of the end of the output:
```
/tmp/tree-sitter.debug.il3ad/lib/src/./tree.c:14: undefined reference to `ts_record_malloc'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/tree-sitter.debug.il3ad/lib/src/./tree.c:20: undefined reference to `ts_record_calloc'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/tree-sitter.debug.il3ad/target/debug/deps/libtree_sitter-99382a6ca8df91d7.rlib(lib.o): in function `ts_tree_delete':
/tmp/tree-sitter.debug.il3ad/lib/src/./tree.c:37: undefined reference to `ts_record_free'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/tree-sitter.debug.il3ad/lib/src/./tree.c:38: undefined reference to `ts_record_free'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/tree-sitter.debug.il3ad/lib/src/./tree.c:39: undefined reference to `ts_record_free'
collect2: error: ld returned 1 exit status
```
If to revert the https://github.com/tree-sitter/tree-sitter/commit/b661050a61c1c318acca2309003fc4e5f0905b58 commit everything compile and work fine.

### Environment
```
# cc --version
cc (Gentoo 9.3.0-r2 p4) 9.3.0
# cargo -V
cargo 1.49.0 (d00d64df9 2020-12-05)
```

The same issue was with GCC 11 also.

### Steps to reproduce the error
1. Checkout fresh tree-sitter repo.
2. Inside of repo root folder:
```
make -j
cargo b
```

### Extra notes
Bellow is a shell script that was used to identify the root cause, it uses `git bisect` to automatically run several compilation attempts until the bad commit has been identified.
```
#!/bin/bash

set -eo pipefail

[...]

PR #934, opened 2021-02-23
Move allocation tracking into lib crate

dcreager:
We have several test cases defined in the `cli` crate that depend on the `lib` crate's `allocation-tracking` feature.  The implementation of the actual allocation tracker used to live in the `cli` crate, close to the test cases that use it.  The `allocation-tracking` feature in the `lib` crate was just used to tell the tree-sitter implementation to expect that the allocation tracker exists, and to use it.

That pattern meant that we had a circular dependency: `cli` depends on `lib`, but `lib` required some code that was implemented in `cli`. That, in turn, caused linker errors — but only [when compiling in certain configurations][1]!

This patch moves all of the allocation tracking implementation into the `lib` crate, gated on the existing `allocation-tracking` feature, which fixes the circular dependency.

Note that this patch does **not** fix the fact that feature unification causes the `lib` crate to be built with the `allocation-tracking` feature enabled, even though it's not a default.  Fixing that depends on the forthcoming [version 2 feature resolver][2], or using the [`dev_dep` workaround][3] in the meantime.

[1]: https://github.com/tree-sitter/tree-sitter/issues/919
[2]: https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-resolver-version-2
[3]: https://github.com/tree-sitter/tree-sitter/issues/919#issuecomment-777107086


maxbrunsfeld (approved):
Thanks for taking care of this! This solution is ultimately much more straightforward than what we had before.

PR #943, opened 2021-02-26
Clean up parse table representation, use 16 bits for production_id

### LINKED ISSUE #511: Tree-sitter generates invalid code for grammars with large numbers of fields and/or aliases
I'm trying to implement a quite complicated grammar using tree-sitter. 

I'm getting compiler warnings like this:

```
> node-gyp rebuild

  CC(target) Release/obj.target/tree_sitter_coccinelle_binding/src/parser.o
../src/parser.c:14813:90: warning: implicit conversion from 'int' to 'uint8_t' (aka 'unsigned char') changes value from 256 to 0 [-Wconstant-conversion]
  [1841] = {.count = 1, .reusable = true}, REDUCE(sym_metavariables, 9, .production_id = 256),
                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
../src/tree_sitter/parser.h:203:9: note: expanded from macro 'REDUCE'
        __VA_ARGS__                              \
        ^~~~~~~~~~~
```

Everything is then built without errors, but the grammar obviously is not working.

I think it is due to definitions like this one (there are more uses of `uint8_t` in that files): https://github.com/tree-sitter/tree-sitter/blob/9a7327738940d6726c2e01c23b598e1704830ea7/lib/include/tree_sitter/parser.h#L70

It would be good if tree-sitter would support more productions, and if the number of productions is more than maximum supported it would be good to have a compile error.

I'm attaching the grammar I'm using now.
[grammar.zip](https://github.com/tree-sitter/tree-sitter/files/4008805/grammar.zip)

maxbrunsfeld:
Fixes #511

### Background

Currently, because of the way that Tree-sitter's parser ABI incrementally evolved, the `production_id` field on parse actions (which is used to identify the sequence of *fields* and *aliases* that applies to a given node's children) is represented as a `uint8_t`. So far, this has not been a problem for any of the grammars that we maintain, but other users have hit the limit, and when it happens, it's very confusing.

### Change

This PR changes `production_id` to a nice `uint16_t`, which should always be more than large enough.

Unfortunately, because of some shortcomings of the old struct layout, this was not feasible to do without making a *breaking change* to the parser ABI. So, for the first time in a long time, I have bumped tree-sitter's `TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION` constant. This means that **new versions of the library will refuse to load parsers that were generated with previous versions of the CLI**. Specifically `ts_parser_set_language` will return `false`. Users will need to regenerate their parsers.

Normally, this is not a problem: I think that for most use cases, users can easily upgrade the library and their parsers together. There are a few exceptions that I know of:
1. Atom - At some point, when Atom adopts this version of Tree-sitter, this change will affect Atom users who have installed packages using third-party Tree-sitter parsers. The packages will needto be updated to use a regenerated version of the parsers. /cc @darangi  - I don't think Atom plans on upgrading Tree-sitter anytime soon, so this is probably not urgent for you I think?

2. Neovim - I think that some end-users of Neovim have installed their own third-party parsers. When Neovim upgrades to the latest version of Tree-sitter, Neovim will refuse to load these older parsers. The end users will need to install new versions of the parsers. /cc @bfredl @vigoux @theHamsta @tjdevries

3. Emacs-Tree-Sitter - /cc @ubolonton I'm not sure if users can install their own parsers. If so, then this is something to be aware of.

### Details

Since an ABI change was already needed, I took this opportunity to fix some dumb aspects of the current ABI. There are no more bitfields (they were used unnecessarily before). Also, I've added one more field to the `TSLanguage` struct - `production_id_count`, which makes it so that all the arrays on `TSLanguage` can have their length known at runtime.

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -11,11 +11,8 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
 use std::mem::swap;
 
-// Currently, the library supports a new ABI version that has not yet been
-// stabilized, and the parser generation does not use it by default.
-const STABLE_LANGUAGE_VERSION: usize = tree_sitter::LANGUAGE_VERSION - 1;
-
 const LARGE_CHARACTER_RANGE_COUNT: usize = 8;
+const SMALL_STATE_THRESHOLD: usize = 64;
 
 macro_rules! add {
     ($this: tt, $($arg: tt)*) => {{
@@ -52,8 +49,6 @@ macro_rules! dedent {
     };
 }
 
-const SMALL_STATE_THRESHOLD: usize = 64;
-
 struct Generator {
     buffer: String,
     indent_level: usize,
@@ -72,6 +67,8 @@ struct Generator {
     unique_aliases: Vec<Alias>,
     symbol_map: HashMap<Symbol, Symbol>,
     field_names: Vec<String>,
+
+    #[allow(unused)]
     next_abi: bool,
 }
 
@@ -109,9 +106,7 @@ impl Generator {
             self.add_alias_sequences();
         }
 
-        if self.next_abi {
-            self.add_non_terminal_alias_map();
-        }
+        self.add_non_terminal_alias_map();
 
         let mut main_lex_table = LexTable::default();
         swap(&mut main_lex_table, &mut self.main_lex_table);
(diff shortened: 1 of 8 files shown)


ubolonton:
Most Emacs users of `tree-sitter` download pre-compiled parsers distributed by the Emacs package [tree-sitter-langs](https://github.com/ubolonton/tree-sitter-langs/), so this shouldn't be too much of an issue.

qjoner (commented):
cli/src/generate/render.rs

Issue #930, opened 2021-02-20
Tree-sitter 1.0 Checklist


maxbrunsfeld:
In the not-too-distant future, I'd like to bump Tree-sitter's version to 1.0, indicating a greater degree of stability and completeness. After that I'd like to regenerate all of the parsers in the `tree-sitter` org, and bump them to 1.0 as well. Before doing this, there are several important problems with the framework that I think should be fixed.

## Tasks

* [ ] **Unicode character properties** - Support ECMAScript unicode property escapes in regexes.
  * [x] Implement basic support for this construct (https://github.com/tree-sitter/tree-sitter/pull/906)
  * [ ] Regenerate all parsers to use unicode property escapes, fix any bugs that surface

* [ ] **Partial Precedence Orderings** - The integer precedence system makes some grammars shockingly difficult to maintain. Enhance the precedence system to allow precedences to be expressed in a pairwise *partial ordering* instead of requiring a total ordering based on integers.
  * [ ] Update `tree-sitter-javascript` and `tree-sitter-typescript` to use this more flexible precedence scheme. Right now, the integer precedence system is making it very difficult to continue development of `tree-sitter-typescript` in particular, because of the mix of different conflicts between types and expressions.
  * *Dynamic* precedence should probably stay integer-only, for simplicity

* [ ] **Mergeable Git Repos** - Make it easier to collaborate on grammars by removing generated files from version control.
  * [ ] **CLI commands** - Add new `pack` and `publish` subcommands to the Tree-sitter CLI, for uploading tarballs and compiled `.wasm` files to the GitHub releases API.
  * [ ] **Downstream** - Update GitHub's internal services to stop consuming Tree-sitter grammars using Git, and to start downloading them as tarballs from the GitHub releases API.
  * [ ] **Cleanup** - Remove generated files from all the grammar repos in the `tree-sitter` org

* [ ] **Grammars with many fields, aliases** - By historical accident, generated parsers use too small an integer type (`uint8_t`) for storing nodes' field and alias information. Parsers with large numbers of fields can cause integer overflows (https://github.com/tree-sitter/tree-sitter/issues/511)
  * [ ] Start representing nodes' `production_id` as a `uint16_t`
  * [ ] **Strategy** - Decide whether we're going to bother to maintain backward compatibility with old generated parsers, if so, the library code will need to become a bit more complicated in order to consume both binary formats.
  * [ ] **Grammars** - Regenerate all the parsers with the new representation.

[...]

PR #946, opened 2021-02-26
Error reporting

### LINKED ISSUE #255: Helpful parser error messages
I can't seem to figure out a way to surface parser errors to the user of my text editor. In particular, I'm working on a VSCode extension using tree-sitter to parse typescript files and surface parsing errors. Tree-sitter is much faster, so far, than the typescript language service, but I can't seem to figure out how to surface a useful error message to the user.

In particular, I'd like something like the `Parsing error: ';' expected.` that eslint gives me. I figure that tree-sitter has some sort of context on the active branches in the parser, so should be able to surface that context int he error object.

If this feature already exists, and I'm just missing something, perhaps it should be included in the documentation site, as I couldn't find anything there.

hvithrafn:
This PR adds error diagnostics based on @alanz's implementation mentioned [here](https://github.com/tree-sitter/tree-sitter/issues/255#issuecomment-786268102) and is intended to fix #255.

I haven't had a chance to actually experiment with the results yet but I just wanted to get this PR in place so that there is some visibility for this. It would really be nice to have some sort of functionality like this ready for 1.0 (see https://github.com/tree-sitter/tree-sitter/issues/930).

Ultimately, I think it may be more useful to have JSON output of these error diagnostics (at least as an option) rather than plain strings. I will probably try to add that to this PR before marking it as ready.

/cc @maxbrunsfeld @alanz


maxbrunsfeld:
:wave: I think this is not quite the right way to report errors. To generate a useful error message, I think we need to actually indicate the location where the error was detected. Right now, this information is not stored: when recovering from an error, we pick out an arbitrary *sequence* of nodes surrounding error-detection-point and wrap those nodes in an `ERROR`.

What we need to do is tweak the error recovery procedure (during parsing) so that it records *where* the error initially occurred, and store this information on the `ERROR` node. I think it could be stored in terms of a child index (e.g. the first node *after* the error-detection-point is the `n`th child of the `ERROR`).

Then, we would be able to compute what parse state we were in at the time the error was detected, and look up the set of tokens that would be valid in that parse state.

hvithrafn:
@maxbrunsfeld After having experimented with this some, I agree with you. I was thinking the output would perhaps be in a tree form or something, but it's basically what you'd get from walking the tree (with a cursor) and collecting errors, which is something I already do in my language servers.

I'd prefer to have something more like what you are describing where the actual parser state is described when an error occurs.

I'll try to take a look at the parser to see if I can figure out how to do what you described. Any further suggestions you might have along those lines are appreciated.

PR #952, opened 2021-02-28
Support highlighting in truecolor, falling back to the closest xterm …

### LINKED ISSUE #758: RGB color on tree-sitter highlight
```json
{
  "parser-directories": [
    "/home/alemuller/git"
  ],
  "theme": {
      "number": "#FF0000"
  }
}
```

When calling `tree-sitter highlight file.ext` the following error is produced:
```
thread 'main' panicked at 'Unsupported color type', cli/src/highlight.rs:271:14
```
When color is a number or an ANSI name it works.

lazytype:
…color if the terminal does not support it

Fixes #758

### DIFF
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -7,6 +7,7 @@ use serde::ser::SerializeMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::{json, Value};
 use std::collections::HashMap;
+use std::fmt::Write;
 use std::sync::atomic::AtomicUsize;
 use std::time::Instant;
 use std::{fs, io, path, str, usize};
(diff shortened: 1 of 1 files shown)


cli/src/highlight.rs
```diff
@@ -252,27 +266,68 @@ fn style_to_css(style: ansi_term::Style) -> String {
...
-        Color::Blue => "blue",
-        Color::Red => "red",
-        Color::Green => "green",
-        Color::Yellow => "yellow",
-        Color::Cyan => "cyan",
-        Color::Purple => "purple",
-        Color::White => "white",
-        Color::Fixed(n) => CSS_STYLES_BY_COLOR_ID[n as usize].as_str(),
-        _ => panic!("Unsupported color type"),
+fn write_color(buffer: &mut String, color: Color) {
```
lazytype, line 275: folded `color_to_css` into this since we can't return the computed RGB string as `&'static str`

maxbrunsfeld:
Nice! Thanks for fixing this.

PR #956, opened 2021-03-01
Slightly improve highlight test hygiene

lazytype:
Realized I forgot to handle a case when restoring the COLORTERM env variable to its pre-test state


maxbrunsfeld:
Thanks for following up.

PR #961, opened 2021-03-02
Fix precedence typings

jonhue:
I think grammars without the `precedences` field are still valid. Also, this renames the first argument of `prec` from `string` to `value`.


maxbrunsfeld:
🤦 Thanks for the quick fix. I'll get CI passing today and do another patch release.

maxbrunsfeld:
This is in v0.19.1.

PR #962, opened 2021-03-02
Add empty dsl.js file to cli

jonhue:
I ran into some problems using

```
import 'tree-sitter-cli/dsl'
```

in my grammars because TypeScript would transpile this to

```
require("tree-sitter-cli/dsl");
```

which would fail because `dsl.js` is not a file. I don't know if there is a better way to fix this, but with this small change things work well 🙂


Razzeee:
Why are you importing it? 
I'm doing it like this: https://github.com/elm-tooling/tree-sitter-elm/blob/master/grammar.js#L1

jonhue:
Oh yeah, that's way better. Thanks!

PR #975, opened 2021-03-08
fix: don't zeroing provided start position by the end zero position

ahlinc:
Hello :wave: 
I've stumbled over strange behavior that when I just pass a start position to a query and want that the end position to be at the end of the input then tree-sitter doesn't takes into account passed start position and captures all from the beginning of the input like the starting position also is `{row: 0, column: 0}`.
I suppose that `POINT_ZERO` what means `{row: 0, column: 0}` is a special marker that can be used to mark undefined position and then the end zero position shouldn't affect starting position in any case cause the starting position will be passed in any case because it passed to the `ts_query_cursor_set_byte_range` function by value.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -4,7 +4,7 @@ use std::env;
 use std::fmt::Write;
 use tree_sitter::{
     allocations, Language, Node, Parser, Query, QueryCapture, QueryCursor, QueryError,
-    QueryErrorKind, QueryMatch, QueryPredicate, QueryPredicateArg, QueryProperty,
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I think you're right. Could you add a test that demonstrates the correct behavior when `end_byte == 0`? I think it could go near here: https://github.com/tree-sitter/tree-sitter/blob/9dc88061d5c56b5bb5f770f68a9697f54a851bd2/cli/src/tests/query_test.rs#L1635.

ahlinc:
Sure, I'll do it on this week.

maxbrunsfeld:
Thank you!

PR #977, opened 2021-03-10
dsl.js - Reuse sym() in RuleBuilder

ahlinc:
This PR replaces duplicated piece of code in `dsl.js` with a `sym()` function call that does the same.


maxbrunsfeld:
Makes sense!

PR #978, opened 2021-03-11
Fix the behavior of Lexer.get_column

### LINKED ISSUE #144: Invalidate external tokens that use `get_column` when their start column changes
This is needed for [tree-sitter-haskell](https://github.com/tree-sitter/tree-sitter-haskell).

### LINKED ISSUE #516: get_column causes infinite loop
I tried `get_column` but it didn't work as I expected.
So I just put `lexer->get_column(lexer);` at the beginning of scan function to check the behavior.
```c
bool tree_sitter_cham_external_scanner_scan(
    void *payload,
    TSLexer *lexer,
    const bool *valid_symbols)
{

  lexer->get_column(lexer);
```
But then for every input it never stops and gives any result neither.
My `tree-sitter-cli` version is 0.16.2.

### LINKED ISSUE #589: Each call of get_column increase the line number of tokens by 1

maxbrunsfeld:
Fixes #516
Fixes #589
Closes #640
Fixes #144

This PR finally makes the `Lexer.get_column` (needed for languages like Haskell, Elm) API work properly.
* Adds unit test coverage for `get_column`, using a test fixture language with Haskell-like layout rules.
* Accounts for tokens' position-dependence when editing a tree, so that incremental parsing is fully correct with respect to these "layout" tokens. This requires tracking which subtrees depend on their start column:
  * any external token where the scanner called `get_column()`
  * recursively, any parent node that has a child node on its *first line* that depends on its own start column
* ⚠️ **Changes the meaning of `get_column` so that it returns a *byte count*, not a character count** ⚠️ 

Regarding this last item - We originally returned a character count from `get_column` because it seemed more technically correct, since GHC counted the unicode characters (as opposed to the bytes) in its implementation of layout. But this adds so much complexity (and perf cost) to our code that I really don't think it's worth it. There could be some slightly incorrect layout-parsing if somebody uses a mixture of different non-ascii whitespace characters for their layout, but this seems like a very uncommon situation.

/cc @Razzeee @tek @bglgwyng @banacorn

### DIFF
--- a/cli/src/tests/helpers/fixtures.rs
+++ b/cli/src/tests/helpers/fixtures.rs
@@ -74,3 +74,9 @@ pub fn get_test_language(name: &str, parser_code: &str, path: Option<&Path>) -> 
         .load_language_from_sources(name, &HEADER_DIR, &parser_c_path, &scanner_path)
         .unwrap()
 }
+
+pub fn get_test_grammar(name: &str) -> (String, Option<PathBuf>) {
+    let dir = fixtures_dir().join("test_grammars").join(name);
+    let grammar = fs::read_to_string(&dir.join("grammar.json")).unwrap();
+    (grammar, Some(dir))
+}
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -1,5 +1,5 @@
 use super::helpers::edits::ReadRecorder;
-use super::helpers::fixtures::{get_language, get_test_language};
+use super::helpers::fixtures::{get_language, get_test_grammar, get_test_language};
 use crate::generate::generate_parser_for_grammar;
 use crate::parse::{perform_edit, Edit};
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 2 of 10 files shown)


maxbrunsfeld:
I ran all of the `tree-sitter-elm` tests, so I think we're good. Just let me know if you see any problems in your language server after upgrading.

Razzeee:
Did you remove the `eof` check for that? https://github.com/elm-tooling/tree-sitter-elm/blob/main/src/scanner.cc#L262

Is my understanding right, that that's obsolete now?

maxbrunsfeld:
Yeah, tests pass without that.

maxbrunsfeld:
Actually, without that check, one of the example files in `elm-ui` has a parse error. So maybe you want the `eof()` check anyway?

Razzeee:
Will have to check, I just realized, that I never added the examples (https://github.com/elm-tooling/elm-language-server/issues/527#issuecomment-782831721) to our test suite. I probably thought, it's fixed now and we will never move that code again 😆

Razzeee:
Added in https://github.com/elm-tooling/tree-sitter-elm/commit/569336c989eb67af548d9c846eaf9151f5f13bed

PR #985, opened 2021-03-13
fix type of prec in dsl

jonhue:
This pull request fixes the type of the `prec` functions to allow for string precedences.


maxbrunsfeld:
❤️

PR #989, opened 2021-03-14
Web bindings: fix descendantsOfType memory allocation

jmbockhorst:
We started seeing some memory out of bounds errors in version 0.18, and eventually I was able to track it down to this bug. Not sure why it never caused issues before.


maxbrunsfeld:
Thank you!

PR #993, opened 2021-03-14
feat: Show leaf nodes text in comments

ahlinc:
When working on a grammar I found that it's hard to realize with standard `tree-sitter parse` command that tokens match right pieces of text. I know about the `--xml` flag but it's output is less readable IMO and doesn't have token ranges what is also useful information.

**Now it looks like:**

_Note: Command line option was renamed: `-c, --node-text-comments`_

screenshot1

What corresponds to the following C code excerpt:
```c
 0  #include <tree_sitter/parser.h>
 1
 2  #if defined(__GNUC__) || defined(__clang__)
 3  #pragma GCC diagnostic push
 4  #pragma GCC diagnostic ignored "-Wmissing-field-initializers"
 5  #endif
 6
 7  #define LANGUAGE_VERSION 13
 8  #define STATE_COUNT 1331
 9  #define LARGE_STATE_COUNT 365
10  #define SYMBOL_COUNT 253
11  #define ALIAS_COUNT 3
```

What's interesting even on the above screenshot there are visible issues with `preproc_arg` tokens they all include preceding space that may be not what is expected.

### Plans:
- [ ] Comments color fallback: RGB -> 256 -> ANSI 8 colors
- [ ] No colors if stdout isn't a tty
- [ ] Show text pieces even on non leaf nodes like starter piece and like additional trailers on leaf node comments
- [ ] Improve s-exp syntax to allow a line trailing comments. Following the work done in #871

_Any feedback or suggestion improvements are welcome!_

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -76,7 +76,12 @@ fn run() -> error::Result<()> {
                         .takes_value(true)
                         .multiple(true)
                         .number_of_values(1),
-                ),
+                )
+                .arg(
+                    Arg::with_name("nodes-text")
+                        .help("Show leaf nodes text as comments")
+                        .long("nodes-text").short("n"),
+                )
         )
         .subcommand(
             SubCommand::with_name("query")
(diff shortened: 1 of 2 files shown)


ahlinc:
Attempt to improve rendering by just add comments to nodes was naive :smiley: I'm closing this in favor of work in progress that happen in https://github.com/ahlinc/tree-sitter/pull/4. It'll be contributed by another PR.

PR #1002, opened 2021-03-19
Simplify script/fetch-fixtures, it clones the right reference by the git clone

ahlinc:
This PR simplifies and a bit speed ups the `script/fetch-fixtures`.
Also it fixes that in a `javascript` fixure repo the actual local `master` branch forcibly moves to required reference instead of to clone required reference directly.

### DIFF
--- a/script/fetch-fixtures
+++ b/script/fetch-fixtures
@@ -11,14 +11,8 @@ fetch_grammar() {
   echo "Updating ${grammar} grammar..."
 
   if [ ! -d $grammar_dir ]; then
-    git clone $grammar_url $grammar_dir --depth=1
+    git clone -b $ref $grammar_url $grammar_dir --depth=1
   fi
-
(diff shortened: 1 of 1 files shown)


script/fetch-fixtures
```diff
@@ -11,14 +11,8 @@ fetch_grammar() {
...
   if [ ! -d $grammar_dir ]; then
-    git clone $grammar_url $grammar_dir --depth=1
+    git clone -b $ref $grammar_url $grammar_dir --depth=1
   fi
-
-  (
-    cd $grammar_dir
-    git fetch origin $ref --depth=1
-    git reset --hard FETCH_HEAD
-  )
```
maxbrunsfeld, line 21: If the grammar repo has already been downloaded, this would update it to the given ref. It seems like that will no longer happen. Am I missing something?

PR #1008, opened 2021-03-19
feat: script/test improvements

ahlinc:
This PR:
* Enables maximally strong bash mode to make the script less error prone, `set -Eeuo pipefail` command.
* Fixes conditions check on the [90](https://github.com/tree-sitter/tree-sitter/compare/master...ahlinc:feat/script-test-improvements?expand=1#diff-2ade384ce3c8dda6cc27f888849ab9af75d1870698b5d6e5b7d2e5d1ba86d064L90) line, that turns into `[ -n || -n ... ]` for any missing var, this wasn't triggered cause the `getopts` also emits cases for missing options with empty arguments. But it's better to don't rely on this occasion.
* Fixes coloring by moving it's control to the script level cause the `--nocapture` disables it without any reasons for this script.
* Allows to do `source script/test` and use the `cargo test` directly.

### DIFF
--- a/script/test
+++ b/script/test
@@ -105,5 +105,6 @@ if [[ "${mode}" == "debug" ]]; then
   )
   lldb "${test_binary}" -- $top_level_filter
 else
-  cargo test $test_flags --jobs 1 $top_level_filter -- --nocapture
+  [ -t 1 ] && color=always || color=never
+  cargo test $test_flags --jobs 1 $top_level_filter -- --nocapture --color "$color"
 fi
--- a/script/test
+++ b/script/test
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # vim:ts=2:sw=2:et
 
+shopts=$-
 set -Eeuo pipefail
 
 function usage {
@@ -96,6 +97,23 @@ if [[ \
   -n "$TREE_SITTER_TEST_TRIAL_FILTER" \
 ]]; then
   : ${top_level_filter:=corpus}
+fi
+
+# Allow to source this script to just set
+# shell environment variables and use `cargo test` directly
+if [ "$0" != "$BASH_SOURCE" ]; then
+    eval set +${-//[is]/}
+    eval set -${shopts//[is]/}
+    cargo() {
+      case "$1" in
+        t|test)
+            command cargo test $test_flags "${@:2}"
+            ;;
+        *)  command cargo "$@"
+            ;;
+      esac
+    }
+    return
 fi
 
 if [[ "${mode}" == "debug" ]]; then
--- a/script/test
+++ b/script/test
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
+# vim:ts=2:sw=2:et
 
-set -e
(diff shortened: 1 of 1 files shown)


ahlinc:
@maxbrunsfeld could you review this PR please? It really makes work within `script/test` a bit more comfortable.

script/test
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
+# vim:ts=2:sw=2:et
```
maxbrunsfeld, line 2: Could you 🔥  this vim comment? I just prefer not to have editor-specific stuff in the code.

script/test
```diff
@@ -84,22 +89,40 @@ done
...
+      case "$1" in
+        t|test)
+            command cargo test $test_flags "${@:2}"
+            ;;
+        *)  command cargo "$@"
+            ;;
+      esac
+    }
+    return
+fi
```
maxbrunsfeld, line 117: I don't fully understand this part. You're defining `cargo` as a function, which will then be called in place of the `cargo` executable after sourcing the script?

Could you do without this part? It seems a bit surprising, and specific to your workflow. It seems like you could still source the script just to set all the env vars that are used by the tests.

maxbrunsfeld (commented):
Sorry for the long delay. There's some good fixes in here. Great catch about setting the `--color` flag explicitly. I left a couple of comments.

maxbrunsfeld:
By the way @ahlinc, I really appreciate all of the contributions that you've been making to Tree-sitter. Awesome work.

script/test
ahlinc: Sure.

script/test
ahlinc: I define this wrapper so it helps to use `cargo test` in any sub folder in the repo and don't think about a path to the `script/test` to call it. There other options like an alias for the git which calls all of it sub commands from the root of the repo, but the wrapper look more comfortable IMO.

And the line `if [ "$0" != "$BASH_SOURCE" ]; then` allows to just source the script into current shell environment without actual execution of it, but set all cargo specifics in the current shell environment.
If you disagree I'll remove this.

ahlinc:
Regarding this PR, I'd would suggest a better solution.
I found that many cargo project uses an xtask approach: 
- https://rust-analyzer.github.io/rust-analyzer/xtask/index.html
- https://github.com/rust-analyzer/rust-analyzer/tree/master/xtask/src
Here is [a spec proposal](https://github.com/matklad/cargo-xtask/) and [a blog post](https://matklad.github.io/2018/01/03/make-your-own-make.html) with some preceding context.

@maxbrunsfeld what do you think about to make a migration of the whole `scripts/` folder to the xtask approach?
The benefits that I see:
- No more shell specific.
- All tasks would be easy to perform even in Windows environments.
- This opens possibility to unify CI for all platforms and use one common GH Actions matrix configuration for all platforms.

@maxbrunsfeld If you open for such changes I'd make it in free time as a practice with `xtask` approach cause I've opened it for myself recently.

PR #1035, opened 2021-04-01
Publish automatically the tree-sitter lib on crates.io

Luni-4:
This PR allows to publish automatically the `tree-sitter` lib  on `crates.io` when a new tag is set. It is necessary to define the `CARGO_REGISTRY_TOKEN` as secret. 

I also tried to publish the cli, but without any effect, I don't know why.

Thanks in advance for your review! :)

### DIFF
--- a/.github/workflows/publish_cratesio.yml
+++ b/.github/workflows/publish_cratesio.yml
@@ -1,0 +1,32 @@
+name: Publish on crates.io
+
+on:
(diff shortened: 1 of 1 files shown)


Luni-4:
It seems that the `macos` job can't start properly

Luni-4:
The `macos` test is fixed now

PR #1040, opened 2021-04-02
Add an action to check code formatting and clippy lints

Luni-4:
This PR adds an action to check code formatting problems through `rustfmt` and some common mistakes through `clippy`.  The various defects found by these two tools have also been fixed.

Thanks in advance for your review! :)


Luni-4:
It seems there is an intermittent test, because on my branch the `ubuntu-test` job passed https://github.com/Luni-4/tree-sitter/runs/2254388484

PR #1044, opened 2021-04-05
Fix build-wasm on Windows

### LINKED ISSUE #434: tree-sitter-cli cannot run emcc on windows
I'm trying to set up WebAssembly compilation on Windows using tree-sitter-cli's `build-wasm` subcommand and even though I have emcc in my path (after running emsdk_env.bat), the CLI complains that "You must have either emcc or docker on your PATH to run this command".

After inspecting the code a bit, it appears that this is the issue:
https://github.com/tree-sitter/tree-sitter/blob/105b0ba52ce64e85eb2d6f09f5fe3ad508adeafb/cli/src/wasm.rs#L25

The problem is that when `emcc` is run by itself, it prints "emcc:WARNING: no input files" and exits with exit code 1 (which counts as failure).

I'd be happy to open up a pull request to fix this (by checking the output from emcc) if that's ok.

**My specs**
```
Windows 10
emcc (Emscripten gcc/clang-like replacement) 1.38.42 (commit da4900b83f6bce7511189b9307569d6c9c344e36)
tree-sitter 0.15.7 (5d43b58751ca4135a42e37b3b52f9bbea3055ab1)
```

aminya:
Fixes #434

Tested on tree-sitter-python and it generates the correct wasm file.


cli/src/wasm.rs
```diff
@@ -116,3 +123,23 @@ pub fn compile_language_to_wasm(language_dir: &Path, force_docker: bool) -> Resu
...
 }
+
+fn get_emcc_path() -> Result<PathBuf> {
+    let emcc_bin;
+    if cfg!(windows) {
+        emcc_bin = "emcc.bat";
+    } else {
+        emcc_bin = "emcc";
+    };
+    let emcc_which = which(emcc_bin);
```
maxbrunsfeld, line 134: Could you explain the reason for using the `which` crate here, compared just checking the result of `Command::new`, since it already looks up the command on your `PATH`?

maxbrunsfeld (commented):
The `emcc.bat` thing looks good. I left one question.

cli/src/wasm.rs
aminya: Built-in Rust's `Command` apparently has a buggy behavior.


Running tree-sitter build-wasm with the following code
```rs
fn get_emcc_path() -> Result<String> {
    let emcc_bin = if cfg!(windows) {
                        "emcc.bat"
                    } else {
                        "emcc"
                    };
    if Command::new(emcc_bin).output().is_ok() {
        return Ok(emcc_bin.to_string());
    }
    return Error::err("emcc binary not found".to_string());
}
```

Results in this error:
```
emcc command failed - C:\Users\aminy\AppData\Local\emsdk\python\3.9.2-1_64bit\python.exe: can't open file 'C:\Users\aminy\Documents\GitHub\Dream\tree-sitter-python\emcc.py': [Errno 2] No such file or directory
```

maxbrunsfeld: Interesting. Thanks for explaining.

aminya: You're welcome

PR #1051, opened 2021-04-08
feat: run highlight tests in subdirectories

### LINKED ISSUE #1027: Tree-sitter test don't run highlighting tests on sub directories of test/highlight
The command `tree-sitter test` runs parsing tests on sub directories of `test/corpus`, but don't run highlighting tests on sub directories of `test/highlight`.

```
$ tree ./test
test
├── corpus
│   ├── bar
│   │   ├── file1.txt
│   │   └── file2.txt
│   └── foo
│       ├── file3.txt
│       └── file4.txt
└── highlight
    └── file1.ext
    └── baz
        └── file2.ext

$ tree-sitter test
...
syntax highlighting:
file1.ext (n assertions)
No language found for path "tree-sitter-ext/test/highlight/baz"

$ tree-sitter --version
tree-sitter 0.19.4 (76862e281b3aa5d93b471bfdf903527fec4f9b29)
```

e-hat:
Closes #1027

This PR is in response to #1027, where syntax-highlighting tests were not run on sub-directories of `test/highlight`. The only file edited was `cli/src/test_highlight.rs`. 

## Description of changes
A wrapper function was created around the previously-named `test_highlights`. This function is now named `test_highlights_indented` because it takes an indentation level as a parameter, which it uses to display the directory structure. 

When this function encounters a directory, it either skips it if the directory is empty (since it obscures the test output), or it prints the directory name and runs the tests within it with one more level of indentation. This type of behavior is consistent with the non-syntax-highlighting tests output (in `test.rs`). This is done with recursion. This function is not marked `pub`.

Now, `test_highlights` refers to a small function that prints the line `syntax highlighting` (in accordance with previous behavior), then runs the syntax-highlighting tests with an indentation level of 2. This level was chosen purely to line up with the non-syntax-highlighting tests' output. This function is marked `pub`.

Also, a small helper function `print_indent` that takes an indentation level was created to have a single source of truth for the number of spaces in an indentation (2 spaces). This number of spaces was also chosen to line up with the output of the non-syntax-highlighting tests' output. This function is not marked `pub`.

## Before/After of test output
Some changes to the output of running tests have been made. Here is the file structure for the following examples
```
$ tree test
test
└── highlight
    ├── a_test.c
    ├── inner_dir
    │   ├── functions.c
    │   └── inner_inner_dir
    │       └── variables.c
    └── keywords.c
```
Before: Here is the result of running `tree-sitter test` *before* the changes made in this PR.

```
$ tree-sitter test
syntax highlighting:
  ✗ keywords.c
    Failure - row: 0, column: 3, expected highlight 'function', actual highlights: 'keyword'

No language found for path "path-to-tree-sitter-proj/test/highlight/inner_dir"
```

After: Here is the result of running `tree-sitter test` *after* the changes made in this PR.

```
syntax highlighting:
    ✗ keywords.c
      Failure - row: 0, column: 3, expected highlight 'function', actual highlights: 'keyword'
    inner_dir:
      ✓ functions.c (20 assertions)
      inner_inner_dir:
        ✓ variables.c (20 assertions)
    ✓ a_test.c (20 assertions)
```

Let me know if there are any issues with the formatting/handling of this problem -- I don't want the formatting changes I made to break a system that parses the results of the test. 

Also, there seems to be a large amount of similarity between the code in `test.rs` and `test_highlight.rs`, with `test.rs` seemingly having a more robust process for running tests. I tried to only imitate the behavior of `test.rs` (to stay within the scope of the issue), but perhaps in the future it would be wise to share code between the two modules.

I would appreciate any feedback since this is one of my first PR's.

### DIFF
--- a/cli/src/test_highlight.rs
+++ b/cli/src/test_highlight.rs
@@ -39,40 +39,60 @@ impl std::fmt::Display for Failure {
 }
 
 pub fn test_highlights(loader: &Loader, directory: &Path) -> Result<()> {
+    println!("syntax highlighting:");
+    test_highlights_indented(loader, directory, 2)
+}
+
+fn test_highlights_indented(loader: &Loader, directory: &Path, indent_level: usize) -> Result<()> {
     let mut failed = false;
     let mut highlighter = Highlighter::new();
 
-    println!("syntax highlighting:");
     for highlight_test_file in fs::read_dir(directory)? {
         let highlight_test_file = highlight_test_file?;
         let test_file_path = highlight_test_file.path();
         let test_file_name = highlight_test_file.file_name();
-        let (language, language_config) = loader
-            .language_configuration_for_file_name(&test_file_path)?
-            .ok_or_else(|| anyhow!("No language found for path {:?}", test_file_path))?;
-        let highlight_config = language_config
-            .highlight_config(language)?
-            .ok_or_else(|| anyhow!("No highlighting config found for {:?}", test_file_path))?;
-        match test_highlight(
-            &loader,
-            &mut highlighter,
-            highlight_config,
-            fs::read(&test_file_path)?.as_slice(),
-        ) {
-            Ok(assertion_count) => {
-                println!(
-                    "  ✓ {} ({} assertions)",
-                    Colour::Green.paint(test_file_name.to_string_lossy().as_ref()),
-                    assertion_count
-                );
-            }
-            Err(e) => {
-                println!(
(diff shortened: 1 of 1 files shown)


amaanq (approved):

amaanq:
I know it's been 2 years and I apologize for no activity on this - but I've squashed it down and tweaked it a tiny bit - looks good otherwise! Thanks!

PR #1065, opened 2021-04-20
Split query-related lifetimes into query/cursor lifetime and tree lifetime

xlambein:
I've noticed that the lifetimes of query matches and captures are overly strict, which makes it impossible e.g. to call a function that takes a node as input, performs a query on it, and then returns the query's matched nodes as output:

```rust
fn do_query<'tree>(source: &str, language: Language, node: Node<'tree>) -> Node<'tree> {
    let query = Query::new(language, "...").unwrap();
    let node = cursor
        .matches(&query, node, move |n| &source.as_bytes()[n.byte_range()])
        .next()
        .unwrap()
        .captures[0]
        .node;
    node
}
```

Here, the rust compiler complains that `node`'s lifetime is tied to `cursor`, because it is.  But it needn't be.

This PR splits the lifetime of `QueryCaptures`, `QueryCapture` and `QueryMatch` into `'a` and `'tree`, the latter being the lifetime of the node passed to `cursor.matches()` or `cursor.captures()`.

(by the way, this library is totally amazing :-) thanks for the awesome work!)

### DIFF
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -101,7 +101,7 @@ where
 struct HighlightIterLayer<'a> {
     _tree: Tree,
     cursor: QueryCursor,
-    captures: iter::Peekable<QueryCaptures<'a, &'a [u8]>>,
+    captures: iter::Peekable<QueryCaptures<'a, 'a, &'a [u8]>>,
     config: &'a HighlightConfiguration,
     highlight_end_stack: Vec<usize>,
     scope_stack: Vec<LocalScope<'a>>,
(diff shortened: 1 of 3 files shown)


ahlinc:
@xlambein thank you for the good reproducing snippet. I realized that I also got the same issue.
But I was wondering why this this PR breaks tests and I after experiment I found a way how to apply such lifetime changes less intrusive, you can find a result in #1072.

xlambein:
So the main difference is that you don't tie the lifetime of a `QueryCapture` back to the cursor, and only ties it to the node?  I thought about doing this, but each capture comes with an `index` field that I think refers back to the original cursor, so I thought it made more "conceptual sense" to have its lifetime tied to the cursor too.  I don't know if that's justified tho!

ahlinc:
>and only ties it to the node?

Yes, but also I used less lifitimes in `captures` and `matches` method signatures.

PR #1069, opened 2021-04-22
package.json generation improvements

ahlinc:
This PR adds generation of:
- An actual tree-sitter-cli version on package.json generation in case on missing file.
- Dashes in a language name inside of package.json and Cargo.toml files to uniform naming style in those files.


maxbrunsfeld:
Thanks! This is a great improvement.

PR #1070, opened 2021-04-22
fix(binding_rust): hide an unsafe internals for a cancellation flag management

ahlinc:
It's a good practice in Rust to don't expose unsafe code in API if it's possible, so there wouldn't be the need to use `unsafe { ... }` surrounding blocks on API calls.
Note: This PR isn't a breaking change and possible end users will just receive a compiler warning to just remove redundant unsafe surrounding for `cancellation_flag` / `set_cancellation_flag` API calls.


maxbrunsfeld:
The reason that this function is unsafe is that the reference that you pass to `set_cancellation_flag` has an unclear lifetime.  The parser holds onto that reference. The C library just assumes that it outlives the `TSParser`. So as things are, we can't correctly mark it as safe, because you then you could pass in a reference to an `AtomicUsize` that lives on the stack, and get undefined behavior.

I guess we could declare the `cancellation_flag` as having a `'static` lifetime, but that is also pretty restrictive for users. Callers would probably end up using `unsafe` and transmuting the reference.

If we wanted to avoid unsafe entirely, you might have to add a lifetime parameter to the `Parser` struct, and declare the `cancellation_flag` with a lifetime that outlives the parser's. The reason I didn't do that is that I was worried it would be annoying and complex, and not worth it just for this feature. I could be wrong though.

I'm open to *some* change to this, but as it stands, I don't think this change is correct.

ahlinc:
@maxbrunsfeld thanks for the explanation, it's not easy to understand all circumstances without the author :slightly_smiling_face:

_It was really impossible to implement this without slightly changing the API._

### In the new implementation:
- I saved unsafe methods cause there are clients like `c_lib.rs` files that expose C API for `tags` and `highlight` crates but renamed these in a style of Rust std library by adding `_unchecked` suffixes.
- Introduced non public `ts_set_cancellation_flag` method to share code between safe and unsafe methods and to avoid double mut borrowing issue in `set_cancellation_flag` method.
- Now there are two really safe methods `cancellation_flag` and `set_cancellation_flag` that operate on `Option<Arc<AtomicUsize>>` and this is checked by compiler. I also don't like the idea to add explicit lifetime for the `Parser` struct cause this will enforce adding explicit lifetime into dozen places, so I added additional field into the `Parser` tuple to bind flag's lifetime with the `Parser` lifetime.

@maxbrunsfeld _waiting for your comments!_

cli/src/main.rs
```diff
@@ -271,7 +271,7 @@ fn run() -> error::Result<()> {
                 debug,
                 debug_graph,
                 debug_xml,
-                Some(&cancellation_flag),
+                Some(cancellation_flag.clone()),
```
maxbrunsfeld, line 274: Do we need the `.clone()` here?

lib/binding_rust/lib.rs
```diff
@@ -603,7 +624,18 @@ impl Parser {
...
+    ///
+    /// This can only be called when
+    /// - There is a guaranty that the flag wouldn't be dropped by the end of a scope.
+    ///
+    pub unsafe fn set_cancellation_flag_unchecked(&mut self, flag: Option<&AtomicUsize>) {
+        self.1 = None;
+        self.ts_set_cancellation_flag(flag);
+    }
+
+    unsafe fn ts_set_cancellation_flag(&self, flag: Option<&AtomicUsize>) {
```
maxbrunsfeld, line 638: What do you think about combining this `ts_set_cancellation_flag` method with the public unsafe version, `set_cancellation_flag_unchecked`? I think that the safe method, `set_cancellation_flag`, could just call `set_cancellation_flag_unchecked` and *then* assign the `self.1` field to the arc. It's pretty minor, but I think it would make things slightly more clear to me.

lib/binding_rust/lib.rs
```diff
@@ -603,7 +624,18 @@ impl Parser {
     /// If a pointer is assigned, then the parser will periodically read from
     /// this pointer during parsing. If it reads a non-zero value, it will halt early,
     /// returning `None`. See [parse](Parser::parse) for more information.
-    pub unsafe fn set_cancellation_flag(&self, flag: Option<&AtomicUsize>) {
+    ///
+    /// # Safety
+    ///
+    /// This can only be called when
+    /// - There is a guaranty that the flag wouldn't be dropped by the end of a scope.
```
maxbrunsfeld, line 631:
```suggestion
    /// - The `flag` reference is guaranteed to be valid until either the parser is
    ///   dropped or the reference is replaced by another call to this method.
```

maxbrunsfeld (commented):
Thanks for adapting based on that feedback. Your new solution looks great. I just left some minor questions.

lib/binding_rust/lib.rs
ahlinc: Good point, I did. The main idea is to prohibit to mix safe and unsafe api calls to set the flag, but this it seems wasn't the good implementation.

cli/src/main.rs
ahlinc: Yes, we need it cause there is a loop above on line [258](https://github.com/tree-sitter/tree-sitter/pull/1070/files#diff-37970e7d817d4316a093bb967d74ff7e26d9c6c01d93f5a79d3d195ba9f6155bL258) and so we can't give exclusive ownership for the `cancellation_flag` to the `parse::parse_file_at_path()` method call just for a one loop iteration.

ahlinc:
**Explanation for the bunch of new commits:**
- I've realized that calls for `tree_sitter_highlight::Highlighter::highlight()` and `tree_sitter_tags::TagsContext::generate_tags()` also have a bug for the Rust API on the current master and suffer from the same problem like the `Parser::set_cancellation_flag` but some how these API calls wasn't marked as unsafe.
- But it's impossible easily change these methods API on the Rust convenience to use `Arc<AtomicUsize>` for cancellation flags cause:
  - There are C API in `c_lib.rs` for crates that contain mentioned methods. And I don't know easy feasible way how to transform `Option<&'a AtomicUsize>` to `Option<Arc<AtomicUsize>>` cause the first is actually the pointer to a piece of memory that was allocated by a caller and may be destructed by the caller side at any time and we have no control about this.
  - There are additional checks for the cancellation flag outside of the parser in iterators like [this](https://github.com/tree-sitter/tree-sitter/blob/c389be5eee14083b6721e8298cb59ca1a36f0187/highlight/src/lib.rs#L633) and [this](https://github.com/tree-sitter/tree-sitter/blob/master/tags/src/lib.rs#L302). And it's also wan't easy to avoid Rust's lifetime restrictions and pass to those places cancellation flags that came from Rust native API and compatible with C API, that is the reason why a `CancellationVariant` enum exists.

To solve unsafety for Rust API and to leave an API for `c_lib.rs` files I've changed `cancellation_flag` types for safe API methods to use `Option<Arc<AtomicUsize>>` and introduced unsafe method counterparts with `Option<&'a AtomicUsize>` for `cancellation_flag` parameters.

_@maxbrunsfeld Could you also give attention for #1008 PR, please?_

highlight/src/lib.rs
```diff
@@ -109,7 +112,28 @@ struct HighlightIterLayer<'a, 'tree: 'a> {
...
+            CancellationVariant::Value(f) => f.as_ref().load(Ordering::Relaxed),
+        };
+        if flag != 0 {
+            return Some(Error::Cancelled);
+        }
+        None
+    }
+}
+
+impl<'a> Highlighter {
```
maxbrunsfeld, line 136: I don't understand why we need to change the `Highlighter` API at all. I think it was already safe, which is why it wasn't marked as `unsafe`. The `Highlighter` encapsulates the `Parser`, so it can dynamically make sure that the `Parser` does not hold onto the cancellation flag longer than the reference is valid.

AFAICT, this API can stay the same as before, but just switched to using `set_cancellation_flag_unchecked` internally. Am I missing something?

tags/src/lib.rs
```diff
@@ -239,25 +264,59 @@ impl TagsConfiguration {
...
+    pub fn parser(&mut self) -> &mut Parser {
+        &mut self.parser
+    }
+
+    pub fn generate_tags(
+        &'a mut self,
+        config: &'a TagsConfiguration,
+        source: &'a [u8],
+        cancellation_flag: Option<Arc<AtomicUsize>>,
+    ) -> Result<(impl Iterator<Item = Result<Tag, Error>> + 'a, bool), Error> {
```
maxbrunsfeld, line 284: I have the same question here as I had about the `Highlighter`. Is it necessary to change this API? The existing API was already safe (because it maintains the contract required by the unsafe `set_cancellation_flag` API).

ahlinc: I've decided to write more tests and will come back soon with a bunch on tests that would help to explain things better.
Also I found and tested a better approach how to implement changes suggested in this PR, also planning to try to finish it on the next week.

PR #1071, opened 2021-04-22
feat(binding_rust): implement Display and Error traits for TS errors to be convertible

ahlinc:
This PR provides implementation for standard error traits for TS Rust binding errors to make tree-sitter easily usable with any external crates like `anyhow` that provides additional uniform errors handling facilities.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -1917,20 +1917,20 @@ impl fmt::Display for LanguageError {
 
 impl fmt::Display for QueryError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let msg = match self.kind {
-            QueryErrorKind::Capture => format!("Invalid capture name {}", self.message),
-            QueryErrorKind::Field => format!("Invalid field name {}", self.message),
-            QueryErrorKind::NodeType => format!("Invalid node type {}", self.message),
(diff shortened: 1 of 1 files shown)


lib/binding_rust/lib.rs
```diff
@@ -1908,6 +1899,46 @@ fn predicate_error(row: usize, message: String) -> QueryError {
...
+            "Incompatible language version {}. Expected minimum {}, maximum {}",
+            self.version, MIN_COMPATIBLE_LANGUAGE_VERSION, LANGUAGE_VERSION,
+        )
+    }
+}
+
+impl fmt::Display for QueryError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let msg = match self.kind {
+            QueryErrorKind::Capture => format!("Invalid capture name {}", self.message),
```
maxbrunsfeld, line 1921: Could we avoid the intermediate string allocation here by combining this with the `write!` call somehow?

ahlinc: I've just slightly adapted trait implementation from here: https://github.com/tree-sitter/tree-sitter/blob/347b3eb24373ccd9458709cb94654e7fe64657ce/cli/src/error.rs#L56
But off course we can do better on a following commit :smile_cat:

maxbrunsfeld:
Awesome! Thanks for improving that implementation to where it's better than what the CLI was doing 😁 .

PR #1074, opened 2021-04-27
cli: Improve const-correctness of the generated parsers.

markus-oberhumer:
This moves most of the generated tables from the data segment into
the text segment (read-only memory) so that it can be shared between
different processes.

As a bonus side effect we can also remove all casts in the generated parsers.
```
Before:
  size --totals target/scratch/*.so
      text    data     bss     dec     hex filename
    853623 4684560    2160 5540343  5489f7 (TOTALS)

After:
  size --totals target/scratch/*.so
      text    data     bss     dec     hex filename
   5472086   68616     480 5541182  548d3e (TOTALS)
```

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -380,7 +380,7 @@ impl Generator {
     }
 
     fn add_unique_symbol_map(&mut self) {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
This is great. Thanks for the fix, and for the clear demonstration of the effects.

PR #1075, opened 2021-04-27
binding_web: Add Query.didExceedMatchLimit

dcreager:
This lets wasm clients check whether a query exceeded its maximum number of in-progress matches.

### DIFF
--- a/lib/binding_web/binding.c
+++ b/lib/binding_web/binding.c
@@ -670,3 +670,10 @@ void ts_query_captures_wasm(
   TRANSFER_BUFFER[0] = (const void *)(capture_count);
   TRANSFER_BUFFER[1] = result.contents;
 }
+
+bool ts_query_did_exceed_match_limit_wasm(
+  const TSQuery *self
+) {
+  if (!scratch_query_cursor) return false;
+  return ts_query_cursor_did_exceed_match_limit(scratch_query_cursor);
+}
--- a/lib/binding_web/binding.c
+++ b/lib/binding_web/binding.c
@@ -622,8 +622,11 @@ void ts_query_matches_wasm(
     }
   }
 
+  bool did_exceed_match_limit =
+    ts_query_cursor_did_exceed_match_limit(scratch_query_cursor);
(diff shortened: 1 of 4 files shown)


lib/binding_web/binding.c
```diff
@@ -670,3 +670,10 @@ void ts_query_captures_wasm(
   TRANSFER_BUFFER[0] = (const void *)(capture_count);
   TRANSFER_BUFFER[1] = result.contents;
 }
+
+bool ts_query_did_exceed_match_limit_wasm(
+  const TSQuery *self
+) {
+  if (!scratch_query_cursor) return false;
+  return ts_query_cursor_did_exceed_match_limit(scratch_query_cursor);
```
maxbrunsfeld, line 678: The only problem with this is that, while the `didExceedMatchLimit` method is on the `Query` object, its behavior right now isn't scoped to that particular query. So it could give an unexpected result if you ran two different queries, and then called `didExceedMatchLimit` on the *first* query (admittedly, an unlikely sequence).

What do you think of the following tweak, which would fix that minor issue:

At the ends of the functions `ts_query_matches_wasm` and `ts_query_captures_wasm`, we'd write a `did_exceed_match_limit` boolean as a third value to our `TRANSFER_BUFFER`. We'd have to cast it to a `void *`, which we're already doing with the result count. Then, on the JavaScript side, we'd just have to store that value as a boolean field on the `Query` object, and then return the same field value from the `didExceedMatchLimit` method.

It's slightly more complicated, but I feel like it's not too bad, so it'd be worth it to be able to avoid confusion if people call the methods in an unusual order.

dcreager: I completely agree re the query / cursor mismatch!  At first I had considered keeping track of which query object the scratch cursor had been used with most recently, and verifying that the new method was called from the same query object.  But I like your suggestion about the transfer buffer better.  I'll take a stab at that.

dcreager: Done

maxbrunsfeld (approved):
⚡  This is great.

PR #1076, opened 2021-04-27
cli: Warn about queries that exceed match limit

dcreager:
This makes it easier to debug particular queries and source files to see if they exhibit the match-limit problem.

### DIFF
--- a/cli/src/query.rs
+++ b/cli/src/query.rs
@@ -91,6 +91,12 @@ pub fn query_files_at_paths(
                 }
             }
         }
+        if query_cursor.did_exceed_match_limit() {
+            writeln!(
+                &mut stdout,
+                "  WARNING: Query exceeded maximum number of in-progress captures!"
+            )?;
+        }
         if should_test {
             query_testing::assert_expected_captures(results, path, &mut parser, language)?
         }


maxbrunsfeld (approved):
Great idea. Thanks for adding this!

tjdevries:
Will this make it so we get a warning for this situation? https://github.com/tree-sitter/tree-sitter/issues/1057

Is be helpful to know when that might happen, at least so I can tweak my queries in the meantime. I'm missing matches for doc generation stuff in some of my projects, so I'm trying to figure out how to get more reliable matches.

dcreager:
@tjdevries I think it should, yes — you mention that `did_exceed_match_limits` is returning true in your case, which is what's being tested in this PR, too.  So hopefully this will give you an easier way to see if particular queries / source files are exhibiting the issue.

tjdevries:
OK, cool. It's still unclear to me _why_ it is in my case 😂 I only have 32 functions to match against or something like that, but I'll see if this can make it more clear. Thanks for the response.

maxbrunsfeld:
It seems like there’s a tree-sitter bug that only the Lua grammar is triggering right now...

tjdevries:
Yeah, it's quite a weird bug. Thanks for this PR either way though, might be able to play with some queries to avoid the bug in the meantime with this.

Thanks :)

PR #1084, opened 2021-04-30
Add Make parser

alemuller:
Add make parser and also change Fennel list from `-` to `*` (that created a list inside a list).


maxbrunsfeld:
Thanks!

PR #1089, opened 2021-05-03
binding_web: Call correct function to reset parser

### LINKED ISSUE #1088: wasm bindings don't work with emscripten 2.0.18
emscripten 2.0.18 was released recently, which causes a regression with our wasm bindings:

- before, [succeeds on 2.0.17](https://github.com/dcreager/tree-sitter-test/runs/2441087580):
- after, [fails on 2.0.18](https://github.com/dcreager/tree-sitter-test/runs/2441025964)

```
RuntimeError: function signature mismatch
    at <anonymous>:wasm-function[80]:0x16cb7
    at <anonymous>:wasm-function[131]:0x26333
    at ts_parser_parse_wasm (<anonymous>:wasm-function[243]:0x2d8fd)
    at Object.Module._ts_parser_parse_wasm (/home/runner/work/tree-sitter-test/tree-sitter-test/node_modules/web-tree-sitter/tree-sitter.js:1:18760)
    at Parser.parse (/home/runner/work/tree-sitter-test/tree-sitter-test/node_modules/web-tree-sitter/tree-sitter.js:1:33876)
    at /home/runner/work/tree-sitter-test/tree-sitter-test/index.js:13:10
```

I think the issue is [here](https://github.com/tree-sitter/tree-sitter/blob/fe5a29a1c64e011de570960ced069c7dab648d40/lib/binding_web/binding.js#L126), where we call `ts_parser_parse_wasm` with the wrong number of parameters.  @maxbrunsfeld, should we call a different function there to reset the parser state?

dcreager:
Fixes #1088

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -123,7 +123,7 @@ class Parser {
   }
 
   reset() {
-    C._ts_parser_parse_wasm(this[0]);
+    C._ts_parser_reset(this[0]);
   }
 
   setTimeoutMicros(timeout) {
--- a/lib/binding_web/exports.json
+++ b/lib/binding_web/exports.json
@@ -69,6 +69,7 @@
   "_ts_parser_enable_logger_wasm",
   "_ts_parser_new_wasm",
   "_ts_parser_parse_wasm",
+  "_ts_parser_reset",
   "_ts_parser_set_language",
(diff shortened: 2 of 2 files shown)


maxbrunsfeld:
Thanks for fixing this! I guess `Parser.reset` wasn't used by anyone before this, because it was completely broken 😓 .

Issue #931, opened 2021-02-21
Rule to disable extras matching


Razzeee:
I would like to be able to express, that an extra is not a valid match right now.

```
  extras: ($) => [
    $.block_comment,
    $.line_comment,
    /[\s\uFEFF\u2060\u200B]|\\\r?\n/,
  ],
```

So my extras are comments and spaces right now.

Usually you could use `token.immediate`, but the structure is not shallow enought:
```
negate_expr: ($) =>
      seq(
        alias($.minus_without_trailing_whitespace, $.operator_identifier),
        $._atom
      ), // todo disallow whitespace
```

```
================================================================================
Negate
================================================================================

negate n =
  -n

--------------------------------------------------------------------------------

(file
  (value_declaration
    (function_declaration_left
      (lower_case_identifier)
      (lower_pattern
        (lower_case_identifier)))
    (eq)
    (negate_expr
      (operator_identifier)
      (value_expr
        (value_qid
          (lower_case_identifier))))))
```

I basically want to make sure that `-n` isn't `- n`.

I think I actually do this in my external scanner, but that's not ideal too I think.

```
     // Handle minus without a whitespace for negate
            if (valid_symbols[MINUS_WITHOUT_TRAILING_WHITESPACE])
            {
                if (lexer->lookahead == '-')
                {
                    advance(lexer);
                    auto lookahead = lexer->lookahead;
                    if ((lookahead >= 'a' && lookahead <= 'z') || (lookahead >= 'A' && lookahead <= 'Z') || lookahead == '(')
                    {
                        lexer->result_symbol = MINUS_WITHOUT_TRAILING_WHITESPACE;
                        lexer->mark_end(lexer);

                        return true;
                    }
                    return false;
                }
            }
```
So I basically peek at the next char and decide which kind of a minus it is based on that.

I think it would also be helpful for comments vs strings. Even if prec and sometimes token.immediate seemt to be enough.

ahlinc:
[...]

Issue #1087, opened 2021-05-01
How to prevent extras from matching in a string?


keidax:
I can't figure out how to allow comments in `extras` while keeping them out of strings. Here's a minimal grammar that shows the problem:
``` javascript
module.exports = grammar({
  name: 'crystal',

  extras: $ => [/\s/, $.comment],

  rules: {
    source_file: $ => repeat($._statement),

    _statement: $ => choice(
      $.string,
    ),

    comment: $ => /#.*\n/,

    string: $ => seq(
      '"',
      repeat(
        token.immediate(/./),
      ),
      token.immediate('"'),
    ),
  }
});
```

If I try to parse text such as
``` crystal
"string"
# comment
"#"
```
then the parser produces an error
```
(source_file [0, 0] - [3, 0]
  (string [0, 0] - [0, 8])
  (comment [1, 0] - [2, 0])
  (string [2, 0] - [3, 0]
    (comment [2, 1] - [3, 0])))
example.cr	0 ms	(MISSING """ [3, 0] - [3, 0])
```

This is a similar issue to https://github.com/tree-sitter/tree-sitter/issues/861, but the workaround there seems to require `$.comment` being an external. Is it possible to keep extras out of rules which aren't externals?

alemuller:
You have to set the comment to have lower precedence than [conflicting tokens](https://tree-sitter.github.io/tree-sitter/creating-parsers#conflicting-tokens).

```js
comment: $ => token(prec(-1,/#.*\n/)),
```

Also. You probably want
```
token.immediate(/[^"]/),
```

instead of
```
token.immediate(/./),
```

keidax:
Your first suggestion works! I had tried setting a negative precedence, but I didn't realize it was necessary to use `token` as well. 

Could you explain why these rules don't work? I'm not sure how to reason about the order of composition.
``` javascript
    comment: $ => prec(-1, /#.*\n/),
```
``` javascript
    comment: $ => prec(-1, token(/#.*\n/)),
```

theHamsta:
You probably want to end stings on line end (or can they be Multi-Line?) You could take the rule of tree-sitter-c

keidax:
@alemuller When re-reading the documentation, I guess I missed this sentence:
> 2. Lexical Precedence - When the precedence functions described above _are used within the token function_, the given precedence values serve as instructions to the lexer.

[...]

PR #1102, opened 2021-05-12
Add immediate rule

aminya:
As I discussed in [this comment](https://github.com/tree-sitter/tree-sitter/issues/931#issuecomment-838582264) in #931, having a way to express an immediate rule is highly needed. 

In this PR, I tried to show what I mean. This code doesn't work in its current state but can be used to discuss the issue.

In my RFC, `immediate` function is called inside `seq`:
```js
seq(rule1, immediate(rule2), rule3)

```

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -167,6 +167,14 @@ impl NfaBuilder {
                 }
                 Ok(true)
             }
+            Rule::Immediate(element) => {
+                self.is_sep = false;
+                if self.expand_rule(element, next_state_id)? {
+                    Ok(true)
+                } else {
+                    Ok(false)
+                }
+            }
             Rule::Seq(elements) => {
                 let mut result = false;
                 for element in elements.into_iter().rev() {
--- a/cli/npm/dsl.d.ts
+++ b/cli/npm/dsl.d.ts
@@ -11,6 +11,7 @@ type PrecRule = {type: 'PREC'; content: Rule; value: number};
 type Repeat1Rule = {type: 'REPEAT1'; content: Rule};
 type RepeatRule = {type: 'REPEAT'; content: Rule};
 type SeqRule = {type: 'SEQ'; members: Rule[]};
+type ImmediateRule = {type: 'Immediate'; content: Rule};
 type StringRule = {type: 'STRING'; value: string};
 type SymbolRule<Name extends string> = {type: 'SYMBOL'; name: Name};
 type TokenRule = {type: 'TOKEN'; content: Rule};
@@ -28,6 +29,7 @@ type Rule =
   | PrecRule
   | Repeat1Rule
   | RepeatRule
+  | ImmediateRule
   | SeqRule
   | StringRule
   | SymbolRule<string>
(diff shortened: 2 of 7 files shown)


cli/src/generate/prepare_grammar/expand_tokens.rs
```diff
@@ -167,6 +167,11 @@ impl NfaBuilder {
                 }
                 Ok(true)
             }
+            Rule::Immediate(_next_rule) => {
+                self.is_sep = false;
+                self.push_advance(CharacterSet::empty(), next_state_id);
+                Ok(true)
```
aminya, line 173: Here we need to somehow instruct the parser to not `SKIP` and immediately `ADVANCE` to the next rule.

ahlinc:
Actually there no a strict need to handle the case of immediate rules in tree-sitter internals, it can be modeled in `grammar.js` file by defining several JS functions.
You can find example here https://github.com/ahlinc/tree-sitter-plantuml/blob/demo/grammar.js#L205
Where I've modeled and fully covered a pretty complex processing for [arrows](https://plantuml.com/sequence-diagram#4764f83f72ed032f) in DSL for PlantUML diagrams. Such arrows can't be represented just as terminal tokens cause these may have additional [incorporated information](https://plantuml.com/class-diagram#4ffca7d13182195f).

aminya:
How is this possible? What feature of tree-sitter did you use to achieve this? It is impossible using the current sets of API. Best thing you can do is to parse it as a token.immediate or use external scanners.

ahlinc:
I've also modeled it with `token.immediate` [here](https://github.com/ahlinc/tree-sitter-plantuml/blob/demo/grammar.js#L450).
For my case it was enough to that `rule2` first terminal would be marked by `token.immediate`.

cli/src/generate/prepare_grammar/expand_tokens.rs
```diff
@@ -167,6 +167,14 @@ impl NfaBuilder {
...
                 Ok(true)
             }
+            Rule::Immediate(element) => {
+                self.is_sep = false;
+                if self.expand_rule(element, next_state_id)? {
+                    Ok(true)
+                } else {
+                    Ok(false)
+                }
+            }
```
aminya, line 177: Here we need to somehow instruct the parser to not `SKIP` and immediately `ADVANCE` to the next rule.

aminya: Maybe we should define a new action like `REJECT_NEXT_EXTRA`, so the lexer rejects if the next lookahead is extra like space.

ahlinc:
If to implement such standalone `immediate` function in some approach than it makes sense to make this function generic and allow that it would be applicable to rules: `immediate($.rule)` and to tokens: `immediate(token(...))`. Also this makes sense to deprecate `token.immediate(...)` in favor of `immediate(token(...))`.

aminya:
> If to implement such standalone `immediate` function in some approach than it makes sense to make this function generic and allow that it would be applicable to rules: `immediate($.rule)` and to tokens: `immediate(token(...))`. Also this makes sense to deprecate `token.immediate(...)` in favor of `immediate(token(...))`.

`token.immediate` has different semantics:
- `token.immediate` only matches a whole token without returning the nodes that make up the parent node. 
- `immediate` matches the consecutive rules, but it returns the children nodes in the resulting AST.


For example,
```js
seq(rule1, immediate(rule2)
```
will give such AST:
```lisp
((rule1) (rul2))
```


But
```
seq(rule1, token.immediate(rule2))
```
will give such AST:
```
((rule1))
```

Notice the missing `rule2`

ahlinc:
> `token.immediate` has different semantics:
> 
> * `token.immediate` only matches a whole token without returning the nodes that make up the parent node.
> * `immediate` matches the consecutive rules, but it returns the children nodes in the resulting AST.

Current tree-sitter's DSL consist of functions that can be thought as simple building elements and their combinations, such combinations may have different meaning, recently there was good accent on this in another issue: https://github.com/tree-sitter/tree-sitter/issues/1087#issuecomment-833198651 that `prec(1, ...)` and `token(prec(1, ...))` means different things.
So my previous note just support such idea, that it's better to have minimal DSL with clear combinations than controversial functions, cause if there will be standalone `immediate` function many users would resonably think that `immediate(token(...))` should do something meaningful.

aminya:
To me `immediate(token` should behave like `token.immediate`. We can make that an exception by providing a method overload.

amaanq:
Closing as stale..

Issue #1090, opened 2021-05-04
ts_tree_cursor_current_field_name API does not return field_name


nhasabni:
Hi,

I'm writing a simple recursive function that is printing some additional details to S-expressions for parse trees than those printed by `ts_node_string` API. For instance, for binary expression `x + y` from C language, `ts_node_string` prints `(binary_expression left: (identifier) right: (identifier))`, whereas I want to print `(binary_expression left: (identifier x) operator: "+" right: (identifier y))`.

I've written following function as a starting step for this purpose:
```
std::string GetSExpression(const TSNode& node) {
  std::string sexpr = "";
  sexpr += "(";
  sexpr += ts_node_type(node);
  sexpr += " ";
  for (size_t i = 0; i < ts_node_named_child_count(node); i++) {
    auto child = ts_node_named_child(node, i);
    auto tree_cursor = ts_tree_cursor_new(child);
    auto field_name = ts_tree_cursor_current_field_name(&tree_cursor);
    ts_tree_cursor_delete(&tree_cursor);
    if (field_name != NULL)
      sexpr += field_name;
    sexpr += " ";
    sexpr += GetSExpression(child);
  }
  sexpr += ")";
  return sexpr;
}
```

The problem is: ` ts_tree_cursor_current_field_name(&tree_cursor)` is returning NULL for every node. I checked API documentation, and cannot find any error with my code. I even tried moving TreeCursor APIs outside of `for` loop so that TreeCursor sees parent `node`, but still no luck. But I guess I'm missing something here.

Thanks.

ahlinc:
@maxbrunsfeld this was fixed by the great @nhasabni contribution and seems can be closed.

PR #1109, opened 2021-05-15
Adding API to get field name of a TSNode

nhasabni:
This PR adds an API to get name of the field of TSNode's child.
It uses same set of arguments as that of ts_node_child, but returns
field name if it is found, otherwise it returns NULL.

This API is useful to implement custom printing of S-expressions such
as following:

"(binary_expression
   (binary_expression_left (identifier))
   (binary_expression_operator ("+"))
   (binary_expression_right (identifier)
)"

Currently, ts_node_string does not allow any customization for printing.
This PR addresses question #1090.

### DIFF
--- a/cli/src/tests/node_test.rs
+++ b/cli/src/tests/node_test.rs
@@ -249,6 +249,25 @@ fn test_node_parent_of_child_by_field_name() {
 }
 
 #[test]
+fn test_node_field_name_for_child() {
+    let mut parser = Parser::new();
+    parser.set_language(get_language("c")).unwrap();
+    let tree = parser.parse("x + y;", None).unwrap();
+    let translation_unit_node = tree.root_node();
+    let binary_expression_node = translation_unit_node
+                                 .named_child(0)
+                                 .unwrap()
+                                 .named_child(0)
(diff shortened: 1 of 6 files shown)


nhasabni:
Hello, this PR addresses question #1090.

maxbrunsfeld (commented):
I'm open to this addition. There's two changes I'd suggest:

* Could you rename it to `ts_node_field_name_for_child`?
* Can you add a unit test for the new function in [this file](https://github.com/tree-sitter/tree-sitter/blob/b8ff6447d7041d198e94c2600a65c10300ddadde/cli/src/tests/node_test.rs#L229)? You'll need to add a Rust binding around [here](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_rust/lib.rs#L858), which will require running the [`generate-bindings`](https://github.com/tree-sitter/tree-sitter/blob/master/script/generate-bindings) script.

nhasabni:
@maxbrunsfeld Thanks for your comments. I've addressed your comments. Please take a look.

maxbrunsfeld:
Thanks!

PR #1110, opened 2021-05-15
Filter empty string in extras

aminya:
This prevents never-ending loops in the parser if someone uses `() => [""]` by mistake in their `extras`. 

This happened to me, and the only way I could stop the code was to hard reset my computer!

### DIFF
--- a/cli/src/generate/parse_grammar.rs
+++ b/cli/src/generate/parse_grammar.rs
@@ -120,7 +120,22 @@ pub(crate) fn parse_grammar(input: &str) -> Result<InputGrammar> {
         precedence_orderings.push(ordering);
     }
 
-    let extra_symbols = grammar_json.extras.into_iter().map(parse_rule).collect();
+    let extra_symbols = grammar_json
+        .extras
+        .into_iter()
+        .try_fold(Vec::new(), |mut acc, item| {
+            let rule = parse_rule(item);
+            if let Rule::String(ref value) = rule {
+                if value.is_empty() {
+                    return Err(anyhow!(
+                        "Rules in the `extras` array must not contain empty strings"
+                    ));
+                }
+            }
+            acc.push(rule);
+            Ok(acc)
+        })?;
+
     let external_tokens = grammar_json.externals.into_iter().map(parse_rule).collect();
 
     Ok(InputGrammar {


cli/src/generate/parse_grammar.rs
```diff
@@ -119,7 +119,10 @@ pub(crate) fn parse_grammar(input: &str) -> Result<InputGrammar> {
         precedence_orderings.push(ordering);
     }
 
-    let extra_symbols = grammar_json.extras.into_iter().map(parse_rule).collect();
+    let extra_symbols = grammar_json.extras.into_iter()
+                                           .map(parse_rule)
+                                           .filter(|rule| !Rule::is_empty_string(rule))
```
aminya, line 124: This still allows a `NamedSymbol` which has an empty string inside. Should we filter that as well?

ahlinc:
@aminya Could you provide a complete grammar sample with empty string in _extras_?

On the current _master_ branch the following grammar doesn't run into infinite loop:
```js
module.exports = grammar({
  name: 'tree_sitter_bugs',

  extras: $ => [""],

  rules: {
    source_file: $ => seq($.hello, $.world),
    hello: $ => 'hello',
    world: $ => 'world',
  }
});
```
Called by:
```sh
tree-sitter parse -d /dev/stdin <<< "hello world"
```
Output:
```wasm
(source_file [0, 0] - [1, 0]
  (hello [0, 0] - [0, 5])
  (ERROR [0, 5] - [0, 6]
    (ERROR [0, 5] - [0, 6]))
  (world [0, 6] - [0, 11])
  (ERROR [0, 11] - [1, 0]
    (ERROR [0, 11] - [1, 0])))
```
Output with commented `extras`:
```wasm
(source_file [0, 0] - [1, 0]
  (hello [0, 0] - [0, 5])
  (world [0, 6] - [0, 11]))
```

aminya:
Currently, I do not have access to a sandbox that I can try this on. The last time I faced this issue, I had to hard reset my computer as it was fully frozen. I think empty strings should be handled better in the codebase because they are "meaningless" in the context of the grammars and can cause severe issues that are not covered by the unit tests or the compilers.

maxbrunsfeld:
I actually think we should return an error from `tree-sitter generate` if you do this, instead of filtering it out.

amaanq:
Sorry for taking so long to get this @aminya, I actually didn't notice this PR until recently, been busy with a lot of stuff inside and outside of tree-sitter, but this is a great fix!

PR #1113, opened 2021-05-17
Make HighlightConfiguration::configure use IntoIterator

edg-l:
Make `HighlightConfiguration::configure` use into iterator instead of a `Vec<String>`

I think it's better this way, so you can define a static array of highlight names without having to turn them into a vector of owned values.

Maybe there is some caveat to this that I don't know or this is not useful at all? Let me know please.

With this, you can do the following:

```rust
static HIGHLIGHT_NAMES: &[&str] = &[
    "attribute",
    "label",
    "constant",
    "function.builtin",
    "function.macro",
    ...
];

let mut highlighter = Highlighter::new();

let mut rust_config = HighlightConfiguration::new(
    tree_sitter_rust::language(),
    tree_sitter_rust::HIGHLIGHT_QUERY,
    "",
    "",
)
.unwrap();

rust_config.configure(&HIGHLIGHT_NAMES);
```

Or checkout how I changed the readme.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -661,7 +661,7 @@ impl<'a> LanguageConfiguration<'a> {
                             }
                         }
                     }
-                    result.configure(&all_highlight_names);
+                    result.configure(&all_highlight_names.as_slice());
                     Ok(Some(result))
                 }
             })
--- a/cli/src/tests/helpers/fixtures.rs
+++ b/cli/src/tests/helpers/fixtures.rs
@@ -50,7 +50,7 @@ pub fn get_highlight_config(
         &locals_query,
     )
     .unwrap();
-    result.configure(highlight_names);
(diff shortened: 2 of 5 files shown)


maxbrunsfeld (commented):
Yeah, the ability to use static `strs` instead of `Strings` is a good improvement. I'm wondering if we really need the generality of `IntoIterator`, or if we could just declare the function like so:

```rust
    pub fn configure(&mut self, recognized_names: &[impl AsRef<str>]) {
        // ...
    }
```

There's something nice about the simplicity of the slice. What do you think?

edg-l:
I think it's fixed now

maxbrunsfeld:
We check the code formatting on CI, and it looks like there's a super minor formatting problem. Could you run `rustfmt` on the changed files?

maxbrunsfeld:
Thank you!

PR #1117, opened 2021-05-19
cli: Improve const-correctness of the generated parsers (part 2 of 2).

markus-oberhumer:
This is a follow-up to my previous commit 1badd131f91

I've made this an extra patch as it requires a minor API change in `<tree_sitter/parser.h>`.

This commit moves the remaining generated tables into the read-only segment.

```
Before:
  $ for f in bash c cpp go html java javascript jsdoc json php python ruby rust; do \
       gcc -o $f.o -O2 -Ilib/include -c test/fixtures/grammars/$f/src/parser.c; \
    done
  $ size --totals *.o
      text    data     bss     dec     hex filename
   5353477   24472       0 5377949  520f9d (TOTALS)

After:
  $ for f in bash c cpp go html java javascript jsdoc json php python ruby rust; do \
       gcc -o $f.o -O2 -Ilib/include -c test/fixtures/grammars/$f/src/parser.c; \
    done
  $ size --totals *.o
   5378147       0       0 5378147  521063 (TOTALS)
```

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -355,7 +355,7 @@ impl Generator {
     }
 
     fn add_symbol_names_list(&mut self) {
-        add_line!(self, "static const char *ts_symbol_names[] = {{");
+        add_line!(self, "static const char * const ts_symbol_names[] = {{");
         indent!(self);
         for symbol in self.parse_table.symbols.iter() {
             let name = self.sanitize_string(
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Awesome! Thanks for following up.

markus-oberhumer:
@maxbrunsfeld BTW many thanks for your great tree-sitter project - this really means an impactful change for future code editor improvements!

PR #1119, opened 2021-05-20
Remove interior mutability for parent-node caching in Tree

maxbrunsfeld:
### Background

A long time ago, in #172, after removing parent pointers from syntax trees, I added a system for caching nodes' parents, in order to optimize client code that repeatedly called `ts_node_parent`. This system relied on *interior mutation* of syntax trees - non-mutating `TSNode` APIs would secretly write to a "parent cache" on the `TSTree`.

The downside of this optimization was that it meant that trees' could not be shared between threads at all, *not even in a read-only fashion*.

### Problem

I think we no longer need the optimization I described above. The slow code path in Atom's `bracket-matcher` packages was later overhauled in https://github.com/atom/bracket-matcher/pull/367. None of GitHub's production systems use APIs like `ts_node_parent` - they all use efficient tree traversals with `TSTreeCursor` or `TSQuery`.

### Change

In this PR, I'm basically reverting all of #172. Trees no longer have a "parent node cache" construct. This means that in Rust's terminology, a syntax tree is `Sync` - it can be shared in a read-only way between multiple threads. So I've implemented the `Sync` trait for `Tree`.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -1942,7 +1942,8 @@ impl error::Error for QueryError {}
 unsafe impl Send for Language {}
 unsafe impl Send for Parser {}
 unsafe impl Send for Query {}
-unsafe impl Send for Tree {}
 unsafe impl Send for QueryCursor {}
+unsafe impl Send for Tree {}
 unsafe impl Sync for Language {}
 unsafe impl Sync for Query {}
+unsafe impl Sync for Tree {}
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -150,9 +150,6 @@ static inline TSNode ts_node__child(
     while (ts_node_child_iterator_next(&iterator, &child)) {
       if (ts_node__is_relevant(child, include_anonymous)) {
         if (index == child_index) {
-          if (ts_node__is_relevant(self, true)) {
-            ts_tree_set_cached_parent(self.tree, &child, &self);
-          }
           return child;
         }
         index++;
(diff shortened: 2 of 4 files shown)


maxbrunsfeld:
/cc @vigoux, @theHamsta - A while back, I noticed that you have [some code](https://github.com/nvim-treesitter/nvim-treesitter/blob/ddc0f1b606472b6a1ab85ee9becfd4877507627d/lua/nvim-treesitter/query_predicates.lua#L23-L60) in Neovim that repeatedly calls `ts_node_parent` (`Node:parent` in the lua binding). I just want to give you a heads up that this PR may make that specific predicate slower.

What you might want to do instead is walk *down* the tree from the root, searching for the target node by its byte offsets, instead of repeatedly stepping *up* the tree. Basically, each `parent` call is now a separate tree traversal with a cost proportional to `Log(Tree-size)`.

theHamsta:
Thank you for the warning! I think we can avoid that predicates.  @p00f  this might be important for bracket matcher plugin.

ahlinc:
As I understand this change also affected performance of all the sibling access API:
https://github.com/tree-sitter/tree-sitter/blob/e4e5ffe517ca2c668689b24cb17c51b8c6db0790/lib/src/node.c#L620-L634
https://github.com/tree-sitter/tree-sitter/blob/e4e5ffe517ca2c668689b24cb17c51b8c6db0790/lib/src/node.c#L246-L249
https://github.com/tree-sitter/tree-sitter/blob/e4e5ffe517ca2c668689b24cb17c51b8c6db0790/lib/src/node.c#L187-L192
Both _prev_ and _next_ methods use the `ts_node_parent` function that now always performs the descent from the root node of a tree. So may be this worth to be noted in the documentation that for the tree traversing operation the cursor API is preferable cause it maintains a path from the root node.

PR #1120, opened 2021-05-21
Fix: cast pointers to `void *` when printing

claudi:
In `lib/src/stack.c` and `lib/src/subtree.c`, there were a couple of instances where pointers were being printed, but where not of type `void *`. This may cause undefined behaviour, and is not supported by the `gnu99` standard, which is being used in compilation.

These were all the instances of `%p` I was able to find (with `git grep "%p"`).

### DIFF
--- a/lib/src/stack.c
+++ b/lib/src/stack.c
@@ -737,7 +737,7 @@ bool ts_stack_print_dot_graph(Stack *self, const TSLanguage *language, FILE *f) 
     if (head->status == StackStatusHalted) continue;
 
     fprintf(f, "node_head_%u [shape=none, label=\"\"]\n", i);
-    fprintf(f, "node_head_%u -> node_%p [", i, head->node);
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Just for my understanding (and potentially for others who may not be familiar with this restriction), could you link to some documentation about the rules for passing pointers to `printf`, and why they need to be void pointers?

claudi:
Sure. 

It is mentioned on the C99 standard, in §7.19.6 Formatted input/output functions.

While looking for references I found [this](https://stackoverflow.com/a/9053835) answer in stackoverflow, which explains it in more detail that I could have (and cites the standard).

In fact, adding the `-pedantic` flag to compilation (which issues all the warnings demanded by strict ISO C, see [reference](https://gcc.gnu.org/onlinedocs/gcc-3.4.4/gcc/Warning-Options.html)) will flag these `printf` arguments with warnings.

I should have included this from the begging!

maxbrunsfeld:
Thanks!

PR #1125, opened 2021-05-23
fix(cli): Correct fallback on docker compilation for all platforms, fixes #1124

### LINKED ISSUE #1124: tree-sitter-cli 0.19.5 does not build with docker
I usually use docker to build the wasm, not emmc

```
> @elm-tooling/elm-language-server@2.1.0 build-tree-sitter
> tree-sitter build-wasm ./node_modules/@elm-tooling/tree-sitter-elm


emcc was not found on PATH
```

This is using 
```
"tree-sitter-cli": "^0.19.5",
```

when I downgrade to 
```
"tree-sitter-cli": "^0.19.4",
```
I can build successfully again.

ahlinc:
This PR fixes regression introduced in #1044 PR by mostly restoring logic before it but also takes into account an issue that that PR tried to fix. All error messages that were introduced in #1044 for absent emcc don't make sense cause there is a fallback to building in docker.
Closes #1124

### DIFF
--- a/cli/src/wasm.rs
+++ b/cli/src/wasm.rs
@@ -3,7 +3,6 @@ use super::generate::parse_grammar::GrammarJSON;
 use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::path::Path;
-use std::path::PathBuf;
 use std::process::Command;
 use which::which;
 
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks for the fix!!

PR #1130, opened 2021-05-28
Improve the QueryCursor API

maxbrunsfeld:
This PR makes the `QueryCursor` object usable in a wider variety of situations.

#### A more flexible text callback (Rust binding)

In order to handle query predicates like `#eq?` and `#match?`, the query cursor takes a callback to retrieve the source text for a given node. Previously, this callback has to return something that converted to a slice of bytes:

```rust
fn matches(/* ... */,  text_callback: F)
where
    F: FnMut(Node<'tree>) -> T,
    T: AsRef<[u8]>;
```

The problem with this is that it doesn't work well when the text is stored in a non-contiguous structure like a [rope](https://en.wikipedia.org/wiki/Rope_(data_structure)). This PR generalizes that callback using a trait called `TextProvider`. A text provider must return an *iterator* that yields slices of bytes:

```rust
pub trait TextProvider<'a> {
    type I: Iterator<Item = &'a [u8]> + 'a;
    fn text(&mut self, node: Node) -> Self::I;
}
```

This way, you can return a node's text as a series of disconnected chunks.

We've provided two built-in `TextProvider` implementations: one for `&[u8]` (the simple case, where the text is stored in a single buffer), and one for closures (`impl FnMut(Node) -> impl Iterator<Item = &[u8]>`) so that you can avoid having to implement the `TextProvider` trait explicitly if you'd rather pass a closure inline.

#### More Useful Range Restrictions (core library)

The query cursor allows you to run a query within a specific *range* of a document (expressed either in terms of bytes or in terms of row/column `Points`). Previously, if you specified a range, the cursor would only report matches that were **fully contained** within that range. So, for example, if a match extended across lines 3 through 8, and you asked for the matches on lines 4 through 10, that match would **not** be reported.

In this PR, we adjust the semantics of `ts_query_cursor_set_byte_range` and `ts_query_cursor_set_point_range` so that the cursor will report all of the matches that *intersect* the range. This is  more useful in a context like a text editor, where you're interested in all of the matches that affect the visible area, or contain the current line.

#### Use `Range` objects for range methods on `QueryCursor` (rust)

Previously, the query cursor's `::set_byte_range` and `::set_point_range` methods took separate arguments for the range start and end. Now, these methods just take a `std::ops::Range`. This fits in better with Rust idioms.

#### Miscellaneous additions (Rust binding)

There are also a few other additive changes, making it more convenient to use queries from Rust.


lib/src/query.c
```diff
@@ -2736,27 +2746,49 @@ static inline bool ts_query_cursor__advance(
...
-        if (step->supertype_symbol && !supertype_count) continue;
-        ts_query_cursor__add_state(self, pattern);
+        QueryStep *step = &self->query->steps.contents[pattern->step_index];
+        if (
+          (node_intersects_range || (!pattern->is_rooted && parent_intersects_range)) &&
+          (!step->field || field_id == step->field) &&
+          (!step->supertype_symbol || supertype_count > 0)
+        ) {
+          ts_query_cursor__add_state(self, pattern);
+        }
```
dcreager, line 2776: This is a style nit, please feel free to ignore, but I tend to like the "early return" style you were using previously better. I find it easier to read than a multi-line `if` conditional, and it means that the success path isn't indented another level.

lib/binding_rust/lib.rs
```diff
@@ -126,18 +134,36 @@ pub struct QueryPredicate {
...
 }
 
-/// A sequence of `QueryCapture`s within a `QueryMatch`.
-pub struct QueryCaptures<'a, 'tree: 'a, T: AsRef<[u8]>> {
+/// A sequence of `QueryMatch`es associated with a given `QueryCursor`.
+pub struct QueryMatches<'a, 'tree: 'a, T: TextProvider<'a>> {
     ptr: *mut ffi::TSQueryCursor,
     query: &'a Query,
-    text_callback: Box<dyn FnMut(Node<'tree>) -> T + 'a>,
+    text_provider: T,
```
dcreager, line 148: Nice, always love :fire:-ing a `Box`

lib/binding_rust/lib.rs
```diff
@@ -1772,6 +1915,26 @@ impl<'a> fmt::Debug for QueryMatch<'a> {
...
+    fn text(&mut self, node: Node) -> Self::I {
+        (self)(node)
+    }
+}
+
+impl<'a> TextProvider<'a> for &'a [u8] {
+    type I = std::option::IntoIter<&'a [u8]>;
+
+    fn text(&mut self, node: Node) -> Self::I {
+        Some(&self[node.byte_range()]).into_iter()
```
dcreager, line 1934: I think `std::iter::once` would be more idiomatic for this

dcreager (approved):
These are some great updates

lib/src/query.c
maxbrunsfeld: Yeah, I feel the same way about the early continues being readable.

The reason that we ended up switching the style was that we wanted to impose these guards in the same style as we do in the loop below (where we iterate over all the patterns with a given start node in the `pattern_map`). But in *that* loop, the `continue` guards don't work as well, because there is state that we need to update at the end of the loop body. There may be some way to restructure that later loop to allow for `continues`, but nothing jumped out at me at the time.

PR #1134, opened 2021-05-31
fix: Cut dangling closing bracket in the playground tree

ahlinc:
This PR removes dangling closing bracket in the playground representation to make it cleaner cause it seems it was migrated from CLI representation where there are opening brackets and closing brackets for every opening bracket.

image


maxbrunsfeld:
😆  I can't believe I never noticed that.

PR #1135, opened 2021-05-31
fix(cli): Panic on broken pipe

ahlinc:
This PR:
* Fixes CLI panic if to call tree-sitter-cli in pipe with a tool that closes pipe without reading to the end:
  ```
  tree-sitter parse ... | head
  ```
* Introduces a concept of ignored errors when the `Error` wrapper struct contains `None` instead of `Vec` of `String` messages.

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -5,26 +5,35 @@ use tree_sitter::{QueryError, QueryErrorKind};
 use walkdir;
 
 #[derive(Debug)]
-pub struct Error(pub Vec<String>);
+pub struct Error(Option<Vec<String>>);
 
 pub type Result<T> = std::result::Result<T, Error>;
 
 impl Error {
     pub fn grammar(message: &str) -> Self {
-        Error(vec![format!("Grammar error: {}", message)])
+        Error(Some(vec![format!("Grammar error: {}", message)]))
     }
 
     pub fn regex(mut message: String) -> Self {
         message.insert_str(0, "Regex error: ");
-        Error(vec![message])
+        Error(Some(vec![message]))
     }
 
     pub fn undefined_symbol(name: &str) -> Self {
-        Error(vec![format!("Undefined symbol `{}`", name)])
+        Error(Some(vec![format!("Undefined symbol `{}`", name)]))
     }
 
     pub fn new(message: String) -> Self {
-        Error(vec![message])
+        let e = Error(Some(vec![message]));
+        e
+    }
+
+    pub fn new_ignored() -> Self {
+        Self(None)
+    }
+
+    pub fn is_ignored(&self) -> bool {
+        self.0.is_none()
     }
 
     pub fn err<T>(message: String) -> Result<T> {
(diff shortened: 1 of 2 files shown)


ahlinc:
@maxbrunsfeld Could you review this PR and if okay merge it, please?

cli/src/error.rs
```diff
@@ -89,6 +106,10 @@ impl From<serde_json::Error> for Error {
 
 impl From<io::Error> for Error {
     fn from(error: io::Error) -> Self {
+        match error.raw_os_error() {
+            Some(32) => return Error::new_ignored(), // Broken pipe
```
maxbrunsfeld, line 110: Will it work to compare the error's [`kind`](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind) here, and compare it to `ErrorKind::BrokenPipe`?

maxbrunsfeld (commented):
This is a nice solution. I left one comment.

cli/src/error.rs
ahlinc: It doesn't work directly but it's possible to use the error kind with construction like
```rust
x if x.kind() == ErrorKind::BrokenPipe => return Error::new_ignored(),
```
I've updated the PR.

maxbrunsfeld: I think you don't even need a `match` statement:

```rust
if error.kind() == ErrorKind::BrokenPipe {
    Error::new_ignored()
} else {
    Error::new(error.to_string())
}
```

cli/src/error.rs
```diff
@@ -1,30 +1,40 @@
...
 
     pub fn undefined_symbol(name: &str) -> Self {
-        Error(vec![format!("Undefined symbol `{}`", name)])
+        Error(Some(vec![format!("Undefined symbol `{}`", name)]))
     }
 
     pub fn new(message: String) -> Self {
-        Error(vec![message])
+        let e = Error(Some(vec![message]));
+        e
```
maxbrunsfeld, line 29: Looks like you could eliminate this `let`.

cli/src/main.rs
```diff
@@ -14,9 +14,12 @@ const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 const BUILD_SHA: Option<&'static str> = option_env!("BUILD_SHA");
 
 fn main() {
-    if let Err(e) = run() {
+    if let Err(ref e) = run() {
```
maxbrunsfeld, line 17: Is the `ref` needed here? I think the error can be moved into `e`, as opposed to borrowed.

maxbrunsfeld:
Thanks!

cli/src/error.rs
ahlinc: Done, I've forgot it after debugging while assigning to a var forces rust-analyzer to show derived type.

cli/src/main.rs
ahlinc: No, it isn't, I've removed it.

ahlinc:
Super fast merge :rocket: :slightly_smiling_face:

cli/src/error.rs
ahlinc: I've fixed this in other small PR: https://github.com/tree-sitter/tree-sitter/pull/1150

PR #1150, opened 2021-06-08
Stylistic fix for #1135 - Panic on broken pipe

ahlinc:
I somehow overlooked that it's possible to clearly express error handling only with the match arms.


maxbrunsfeld:
Thanks!

PR #1152, opened 2021-06-09
cli: Use anyhow and thiserror for errors

dcreager:
This patch updates the CLI to use anyhow and thiserror for error management.  The main feature that our custom `Error` type was providing was a _list_ of messages, which would allow us to annotate "lower-level" errors with more contextual information.  This is exactly what's provided by anyhow's `Context` trait.

(This is setup work for a future PR that will pull the `config` and `loader` modules out into separate crates; by using `anyhow` we wouldn't have to deal with a circular dependency between with the new crates.)

This patch also ensures that all of our error types (in the `tags` and `highlight` crates, for instance) implement `std::error::Error`.


cli/src/main.rs
```diff
@@ -1,32 +1,18 @@
...
+    config, generate, highlight, loader, logger, parse, query, tags, test, test_highlight, util,
+    wasm, web_ui,
 };
 
 const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 const BUILD_SHA: Option<&'static str> = option_env!("BUILD_SHA");
 
-fn main() {
-    if let Err(e) = run() {
-        if e.is_ignored() {
```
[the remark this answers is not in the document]
dcreager: Great catch, thanks!  The new commit on this branch handles this over in `main.rs` now.  Before:

```
$ ~/git/tree-sitter/target/release/tree-sitter parse src/flask/app.py | head -n 3
(module [0, 0] - [2088, 0]
  (import_statement [0, 0] - [0, 16]
    name: (dotted_name [0, 7] - [0, 16]
Error: Broken pipe (os error 32)
```

After:

```
$ ~/git/tree-sitter/target/release/tree-sitter parse src/flask/app.py | head -n 3
(module [0, 0] - [2088, 0]
  (import_statement [0, 0] - [0, 16]
    name: (dotted_name [0, 7] - [0, 16]
```

cli/src/generate/node_types.rs
```diff
@@ -328,10 +328,9 @@ pub(crate) fn get_variable_info(
     for supertype_symbol in &syntax_grammar.supertype_symbols {
         if result[supertype_symbol.index].has_multi_step_production {
             let variable = &syntax_grammar.variables[supertype_symbol.index];
-            return Err(Error::grammar(&format!(
-                "Supertype symbols must always have a single visible child, but `{}` can have multiple",
+            return Err(anyhow!( "Grammar error: Supertype symbols must always have a single visible child, but `{}` can have multiple",
```
maxbrunsfeld, line 331: Is there a stray space in there before the string?

maxbrunsfeld (approved):
It's great to replace all of this custom code with something more standard.

cli/src/generate/node_types.rs
dcreager: Yeah I guess the long string literal confused `rustfmt`?  I added in a `concat!` to make it happy

ahlinc (approved):

PR #1153, opened 2021-06-09
rust: Extract runtime language detection into separate crate

dcreager:
This patch adds a new `tree-sitter-loader` crate, which holds the CLI's logic for finding and building local grammar definitions at runtime. This allows other command-line tools to use this logic too!


maxbrunsfeld (approved):
Great idea!

PR #1157, opened 2021-06-10
cli: Extract CLI configuration into separate crate

dcreager:
This patch adds the `tree-sitter-config` crate, which manages tree-sitter's configuration file.  This new setup allows different components to define their own serializable configuration types, instead of having to create a single monolithic configuration type.  But the configuration itself is still stored in a single JSON file.

Before, the default location for the configuration file was `~/.tree-sitter/config.json`.  This patch updates the default location to follow the XDG Base Directory spec (or other relevant platform-specific spec).  So on Linux, for instance, the new default location is `~/.config/tree-sitter/config.json`.  We will look in the new location _first_, and fall back on reading from the legacy location if we can't find anything.


dcreager:
c.f. #628

dcreager:
Compiled parsers are now placed into the user's cache directory, typically `~/.cache/tree-sitter/lib`.  It's still configurable where we look for the grammar _source_ repo, using the `parser-directories` entry in the config file.  No changes to the default set of source repo directories.  There were suggestions on #628 that those should live under `~/.local`, and I'd be open to including that in the default value for the config parameter, but I think it should remain configurable since most folks will want their grammars to be in their typical "bag of git clones" workspace directory.

cli/config/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            .ok_or(anyhow!("Cannot determine home directory"))?
+            .join(".tree-sitter/config.json");
+        if legacy_path.is_file() {
+            return Ok(legacy_path);
+        }
+
+        Err(anyhow!(concat!(
+            "Cannot find a tree-sitter configuration file.\n",
+            "Please run `tree-sitter init-config` to create one!"
+        )))
```
dcreager, line 45: Note that this makes it a hard error to run `tree-sitter` without having a config file in place, which is different than before.  I'm honestly not sure that I like this part of the change.

ahlinc:
@dcreager @maxbrunsfeld I know that `clap` crate allows to extract sub commands to separate binaries: https://users.rust-lang.org/t/using-external-subcommands-with-clap/10753/4. What are you think about to move `tree-sitter highlight` and `tree-sitter tags` sub commands into separate binaries in the corresponding crates?

dcreager:
> sub commands into separate binaries in the corresponding crates

I quite like that idea, though I think it should be done in a separate PR

cli/config/src/lib.rs
dcreager: Ha! And in fact that makes the test cases fail!  Okay that definitely tells me we don't want this to be a hard error.  I'll push up a fix for that.

maxbrunsfeld (approved):
This is great.

Can you do a cursory read of the [syntax highlighting docs path](https://github.com/tree-sitter/tree-sitter/blob/master/docs/section-4-syntax-highlighting.md) and see if anything should be updated or rephrased, due to the changes to the config file's location?

dcreager:
> see if anything should be updated or rephrased

Done :heavy_check_mark:

PR #1163, opened 2021-06-11
fix(cli): Don't print 'Error:' prefix

ahlinc:
Follow up for #1152

Currently `tree-sitter parse` command shows weird empty error prefix cause there is a code:
https://github.com/tree-sitter/tree-sitter/blob/6ed42747a4e0faee9b65edbbacc86ed0caeae05c/cli/src/main.rs#L303-L305
That should set some system error code without printing any error messages, but the `anyhow` library prints the `Error: ` prefix even for empty error messages:
```console
# tree-sitter parse lib/include/tree_sitter/api.corrupted_file.h
lib/include/tree_sitter/api.corrupted_file.h	2 ms	(ERROR [897, 0] - [897, 47])
Error:
```
This PR removes printing of the prefix for empty messages.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -12,17 +12,20 @@ use tree_sitter_loader as loader;
 const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 const BUILD_SHA: Option<&'static str> = option_env!("BUILD_SHA");
 
-fn main() -> Result<()> {
+fn main() {
     let result = run();
-    // Ignore BrokenPipe errors
     if let Err(err) = &result {
+        // Ignore BrokenPipe errors
         if let Some(error) = err.downcast_ref::<std::io::Error>() {
             if error.kind() == std::io::ErrorKind::BrokenPipe {
-                return Ok(());
+                return;
             }
         }
+        if !err.to_string().is_empty() {
+            eprintln!("{:?}", err);
+        }
+        std::process::exit(1);
     }
-    result
 }
 
 fn run() -> Result<()> {


dcreager (approved):
fyi the `Error: ` text doesn't come from anyhow, it comes from how the cli's `main` function just returns the error directly, instead of catching/printing it.  e.g. [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5b52269ddf35433e25e633a9c1d13bff) prints out `Error: "oh no"`

Maybe we want to remove the `Error: ` text in all cases?  If not I'm :+1: to this change.  @maxbrunsfeld, thoughts?

maxbrunsfeld:
Yeah, let’s just remove that error prefix in all cases.

ahlinc:
> it comes from how the cli's `main` function just returns the error directly

@dcreager good note, thanks! So it's Rust internally [hardcoded](https://users.rust-lang.org/t/pretty-print-errors-from-main-result-failure-error/17669/2?u=ahlnn) thing. That explains why simple search in the TS and `anyhow` code bases didn't get anything meaningful.

ahlinc:
@maxbrunsfeld I've updated the PR, to don't print `Error:` prefix and don't print an empty line if the error message is empty.

maxbrunsfeld (commented):
With this change, I think we could change main so that it doesn’t return a Result anymore.

cli/src/main.rs
```diff
@@ -12,17 +12,20 @@ use tree_sitter_loader as loader;
...
     if let Err(err) = &result {
+        // Ignore BrokenPipe errors
         if let Some(error) = err.downcast_ref::<std::io::Error>() {
             if error.kind() == std::io::ErrorKind::BrokenPipe {
-                return Ok(());
+                return;
             }
         }
+        if !err.to_string().is_empty() {
+            eprintln!("{:?}", err);
```
ahlinc, line 25: Debug format needed to print extended errors with `Caused by:` explanation what Rust makes by default.

ahlinc:
@maxbrunsfeld now it's finally fixed.

Before:
```console
$ tree-sitter generate
Error: Error processing rule NumericLit

Caused by:
    Grammar error: Unexpected rule Symbol(Symbol { kind: NonTerminal, index: 4 })
```
After:
```console
$ tree-sitter generate
Error processing rule NumericLit

Caused by:
    Grammar error: Unexpected rule Symbol(Symbol { kind: NonTerminal, index: 4 })
```
And there is no empty ending `Error:` for `tree-sitter parse` in case of parsing errors.

maxbrunsfeld:
Awesome, thanks @ahlinc .

PR #1166, opened 2021-06-12
Add Editors section to the README.

azizk:
Hey, I think it would be great to know where tree-sitter is actually being employed. If not in the README, maybe at least somewhere in "docs/"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,15 @@ Tree-sitter is a parser generator tool and an incremental parsing library. It ca
 - **Robust** enough to provide useful results even in the presence of syntax errors
 - **Dependency-free** so that the runtime library (which is written in pure C) can be embedded in any application
 
+## Editors
+
+The following editors are making use of tree-sitter for syntax highlighting and other features:
+
+- [Neovim](https://github.com/neovim/neovim): a Vim-fork focused on extensibility and usability.
+- [helix](https://github.com/helix-editor/helix): a post-modern modal text editor.
+- [emacs-tree-sitter](https://github.com/emacs-tree-sitter/elisp-tree-sitter): bindings for Emacs.
+- [lapce](https://github.com/lapce/lapce): an IDE written in pure Rust leveraging OpenGL.
+
 ## Links
 
 - [Documentation](https://tree-sitter.github.io)


ahlinc:
I think there may be a section in a docs main page after a `Parsers` section and it worth to add more editors, other known tools and replace Atom to alive forks.

azizk:
I removed Atom from the list considering it's been sunset. I don't know if there is a well-maintained fork...

Also added helix and lapce to the list. :)

README.md
```diff
@@ -13,6 +13,15 @@ Tree-sitter is a parser generator tool and an incremental parsing library. It ca
 - **Robust** enough to provide useful results even in the presence of syntax errors
 - **Dependency-free** so that the runtime library (which is written in pure C) can be embedded in any application
 
+## Editors
+
+The following editors are making use of tree-sitter for syntax highlighting and other features:
+
+- [Neovim](https://github.com/neovim/neovim): a Vim-fork focused on extensibility and usability.
+- [helix](https://github.com/helix-editor/helix): a post-modern modal text editor.
+- [emacs-tree-sitter](https://github.com/ubolonton/emacs-tree-sitter): bindings for Emacs Lisp for Emacs 29+.
```
jcs090218, line 22: The Emacs repo has moved to https://github.com/emacs-tree-sitter/elisp-tree-sitter

README.md
```diff
@@ -13,6 +13,15 @@ Tree-sitter is a parser generator tool and an incremental parsing library. It ca
 - **Robust** enough to provide useful results even in the presence of syntax errors
 - **Dependency-free** so that the runtime library (which is written in pure C) can be embedded in any application
 
+## Editors
+
+The following editors are making use of tree-sitter for syntax highlighting and other features:
+
+- [Neovim](https://github.com/neovim/neovim): a Vim-fork focused on extensibility and usability.
+- [helix](https://github.com/helix-editor/helix): a post-modern modal text editor.
+- [emacs-tree-sitter](https://github.com/emacs-tree-sitter/elisp-tree-sitter): bindings for Emacs 29+.
```
jcs090218, line 22: Should probably remove "29+", see their main page:

> For Emacs 29+, please use [the built-in integration](https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01443.html) instead of this package.

Emacs 29+ has a built-in module into Emacs, but not the same to [emacs-tree-sitter]()/[elisp-tree-sitter]().

dundargoc:
I question the value of these list of links. They add constant churn as they constantly need to be updated, because everyone and their grandma creates a new editor every other week. I suggest adding these to the wiki instead.

PR #1177, opened 2021-06-17
update set_included_ranges to modify extent if the current position is at the very beginning of the range.  

forbisc:


ahlinc:
@maxbrunsfeld I've tested this patch and it fixes a problem in c-library, description is here: https://github.com/tree-sitter/tree-sitter/discussions/1175#discussion-3413931 Could we merge this?
@forbisc thank you for the finding and the fix :+1:

cli/src/tests/parser_test.rs
```diff
@@ -872,6 +872,48 @@ fn test_parsing_with_multiple_included_ranges() {
...
+        start_point: Point {
+            row: 10,
+            column: 12,
+        },
+        end_byte,
+        end_point: Point {
+            row: 10,
+            column: 12 + end_byte,
+        },
+    };
```
maxbrunsfeld, line 895: Is it intentional that this `start_point` does not match the `start_byte`? The test doesn't appear to make assertions about the row/column positions of the nodes. If it's not necessary to have this mismatch, I think it'd be clearer if the points just matched the ranges (`start_point: Point { row: 0, column: 0 }, end_point: Point { row: 0, column: end_byte }`)

cli/src/tests/parser_test.rs
```diff
@@ -872,6 +872,48 @@ fn test_parsing_with_multiple_included_ranges() {
...
+
+    parser.set_included_ranges(&[range_to_parse]).unwrap();
+
+    let mut callback = |offset, _point| {
+        if offset == 0 {
+            source_code.as_bytes()
+        } else {
+            &[]
+        }
+    };
```
maxbrunsfeld, line 905: Is the callback important here? Could you just pass the `source_code.as_bytes()` directly to `parser.parse`?

cli/src/tests/parser_test.rs
```diff
@@ -872,6 +872,48 @@ fn test_parsing_with_multiple_included_ranges() {
...
+
+    let html_tree = parser.parse_with(&mut callback, None).unwrap();
+
+    assert_eq!(html_tree.root_node().range(), range_to_parse,);
```
maxbrunsfeld, line 909: (nitpick) :fire: trailing comma

maxbrunsfeld (changes requested):
@forbisc Thanks for the great bug report and PR. I left some minor comments.

Also, just for my understanding, if we *undid* your fix, how exactly would your test fail?

cli/src/tests/parser_test.rs
forbisc: It is intentional as I'm trying to provide a different start location.  This line asserts the column and lines match what was provided in set_included_ranges: `assert_eq!(html_tree.root_node().range(), range_to_parse,);`

cli/src/tests/parser_test.rs
forbisc: It is still possible to reproduce the issue by passing the bytes directly.  However because the bytes and position are mismatched, it seemed like it was more appropriate to use the callback instead.  Since it's basically saying here's a piece of the file instead of the entire file (but maybe that's not an important distinction).

cli/src/tests/parser_test.rs
forbisc: > @forbisc Thanks for the great bug report and PR. I left some minor comments.
> 
> Also, just for my understanding, if we _undid_ your fix, how exactly would your test fail?

It's this line here that would fail.  Maybe I could make that more clear by asserting the start_point instead of the whole range.  

assert_eq!(html_tree.root_node().range(), range_to_parse,);

maxbrunsfeld:
@forbisc Thanks for the replies. So, just to make sure I understand - Is the bug that you're fixing *only* related to the bytes and the row/column positions being intentionally mismatched? Or does your code change fix any behaviors in the case where the positions match up?

forbisc:
> @forbisc Thanks for the replies. So, just to make sure I understand - Is the bug that you're fixing _only_ related to the bytes and the row/column positions being intentionally mismatched? Or does your code change fix any behaviors in the case where the positions match up?
> 
> I hadn't considered use-cases where you'd intentionally provide mismatched values. But if it works, I'm glad to hear it.

I believe that this only fixes cases where the bytes and row/column are intentionally mismatched.  If they are matched up, then when the lexer is sitting at the start of a new range, it should already have the right state with respect to the source locations.  

The use case is basically if I have a piece of a file that I want to parse, but that chunk of source code needs to be modified first (un-escaped for the particular context that it appears in).  Because it has to be modified, I'd need to copy the buffer, and I don't want to copy the entire source file, I just want to copy the chunk that needs escaping.  But I still want to parse it and have the (approximately) correct source locations and I want the byte ranges to correspond to my escaped chunk.

cli/src/tests/parser_test.rs
maxbrunsfeld: I'd rather have this test just use the bytes. I think that the use of a callback vs a byte slice can be considered orthogonal to whether the points match the byte offsets. The callback is more needed when we have the text stored in some non-contiguous structure.

maxbrunsfeld:
Gotcha. That totally makes sense. Could you update the name of the test to something like `test_parsing_with_included_range_containing_mismatched_positions`, or something like that? I think that would help tell the story that we're doing it intentionally here, since this would be the only test case exercising that.

maxbrunsfeld (approved):

maxbrunsfeld:
Thank you!

PR #1194, opened 2021-06-22
Close #1032 - fix all weirdness in the generated Cargo.toml

### LINKED ISSUE #1032: Weird Cargo.toml generated with `tree-sitter generate`
Version: tree-sitter 0.19.4 (76862e281b3aa5d93b471bfdf903527fec4f9b29)

When I run `tree-sitter generate` I see a `Cargo.toml` generated like this:

```toml
[package]
name = "..."
description = "... grammar for the tree-sitter parsing library"
version = "0.0.1"
keywords = ["incremental", "parsing", "clay"]
categories = ["parsing", "text-editors"]
repository = "https://github.com/tree-sitter/tree-sitter-javascript"
edition = "2018"
license = "MIT"

build = "bindings/rust/build.rs"
include = [
  "bindings/rust/*",
  "grammar.js",
  "queries/*",
  "src/*",
]

[lib]
path = "bindings/rust/lib.rs"

[dependencies]
tree-sitter = "0.17"

[build-dependencies]
cc = "1.0"
```

There are a few problems:

- tree-sitter dependency is 0.17 even though I generated the bindings with 0.19, which depends on lib in the same tree with version 0.19.
- `repository` field points to `tree-sitter-javascript`

ahlinc:
This PR fixes a template repository link to something more meaningful and uses a grammar name for the suffix.
Also it implements substitution for correct Rust binding version actual to used tree-sitter CLI version that was firstly used to generate binding files in a TS language folder. Generated binding version has `~` prefix to allow patch updates (changes in last number).


cli/src/generate/binding_files.rs
```diff
@@ -13,6 +13,8 @@ const PACKAGE_JSON_TEMPLATE: &'static str = include_str!("./templates/package.js
 const PARSER_NAME_PLACEHOLDER: &'static str = "PARSER_NAME";
 const CLI_VERSION_PLACEHOLDER: &'static str = "CLI_VERSION";
 const CLI_VERSION: &'static str = env!("CARGO_PKG_VERSION");
+const RUST_BINDING_VERSION: Option<&'static str> = option_env!("RUST_BINDING_VERSION");
```
maxbrunsfeld, line 16: Since the `RUST_BINDING_VERSION=...` line is unconditionally printed in `build.rs`, should we switch this from `option_env!` to `env!`, and remove the `.unwrap` call when it is used?

maxbrunsfeld (commented):
Thanks for tackling this; it's a great improvement. Just one question.

cli/src/generate/binding_files.rs
ahlinc: Totally agree, fixed! - _I somehow missed to post my comment._

PR #1196, opened 2021-06-23
feat(cli): Independent binding files generation

### LINKED ISSUE #1195: tree-sitter generate - doesn't restore any independently deleted generated files
Currently `tree-sitter generate` doesn't restore any deleted generated file if it was accidentally deleted or if some one didn't want to commit it and removed it by `git clean -fdx` or similar command but want to use specific binding. Such binding may be broken if important for it files wouldn't be dynamically restored. Another case if some one deletes generated file to get it refreshed for ex. dependency versions.

_Affected files:_
```console
Cargo.toml
binding.gyp
bindings/rust/lib.rs
bindings/rust/build.rs
bindings/node/index.js
bindings/node/binding.cc
```

ahlinc:
Closes #1195

This PR introduces approach that binding files and related folders restore independently but in proper order so tree-sitter will always restore a full shape of all components for all bindings.

### DIFF
--- a/cli/src/generate/binding_files.rs
+++ b/cli/src/generate/binding_files.rs
@@ -54,11 +54,15 @@ pub fn generate_binding_files(repo_path: &Path, language_name: &str) -> Result<(
         &binding_gyp_path,
         |path| generate_file(path, BINDING_GYP_TEMPLATE, language_name),
         |path| {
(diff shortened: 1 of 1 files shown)


ahlinc:
@maxbrunsfeld this also would be great to include in the upcoming release.

PR #1198, opened 2021-06-23
feat(cli): colored help and frequent command aliases

ahlinc:
This PR: 
* Enables builtin Clap's coloring for the help output.
* Adds short aliases for frequent sub commands:
  * _generate - gen, g_
  * _parse - p_
  * _query - q_
  * _test - t_
  * _build-wasm - bw_
  * _playground - play, pg_
* Renames `web-ui` sub command to `playground` with a backward compatible `web-ui` alias. This change is controversial but it seems most users know about [Playground](https://tree-sitter.github.io/tree-sitter/playground) in the documentation but don't realize that they have the same but for their local grammar.
* Removes `help` subcommand, it's the same as `--help, -h` flags but among all sub commands doesn't make any real action.
* Change `-h` -> `-H` for the highlight sub command to don't clash with the standard help flag, cause it's unexpected that it's possible to use `-h` to call the help except for `highlight` sub command.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -169,8 +169,11 @@ fn run() -> Result<()> {
                 .arg(Arg::with_name("path").index(1).multiple(true)),
         )
         .subcommand(
-            SubCommand::with_name("web-ui")
-                .about("Test a parser interactively in the browser")
+            SubCommand::with_name("playground")
+                .alias("play")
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Great changes. Totally agree about `playground` being a better name than `web-ui`.

PR #1199, opened 2021-06-23
cli: Allow `$` in `tree-sitter parse --edit`

dcreager:
You can now specify `$` as the position to apply an edit, signifying the end of the file.  (That prevents you from having to calculate the size of the file yourself.)


maxbrunsfeld:
💸

PR #1208, opened 2021-06-26
cli/src/main.rs: Fix panic with empty file list to parse

CyberShadow:
E.g.: `tree-sitter parse --paths /dev/null`


maxbrunsfeld:
Thanks!

PR #1210, opened 2021-06-27
fix(cli): init-config experience, it shouldn't silently overwrite existing config file

ahlinc:
Add a fuse to decline silently overwrite of existing config file cause users may have there own grammar lookup paths and custom highlighting configuration.
Also this PR adds the `DeriveDisplayOrder` Clap's setting to show sub commands in the defined order that looks more accurate and have more meaning for sub commands use order.

### DIFF
--- a/cli/config/src/lib.rs
+++ b/cli/config/src/lib.rs
@@ -20,7 +20,7 @@ pub struct Config {
 }
 
 impl Config {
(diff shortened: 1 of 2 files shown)


ahlinc:
@maxbrunsfeld this also would be great to include in the upcoming release.

maxbrunsfeld:
Thanks!

PR #1220, opened 2021-06-30
allow `~` or `$HOME` in `parser-directories`

toyboot4e:
This PR allows users to write [parser-directories](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#paths) in `config.json` using `~` or `$HOME`:

```json
{
  "parser-directories": [
    "~/code",
    "~/other-code"
  ]
}
```

Currently, if we write the paths like that, they're not correctly recognized (at least on macOS), so I got such errors:

```
$ tree-sitter highlight path/to/your/file.extension
No language found for path "path/to/your/file.extension"
```

That was very confusing! So this PR fixes it.

I'll appreciate code edit in any form or any better solutions. Thank you :)

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -2,6 +2,7 @@ use anyhow::{anyhow, Context, Error, Result};
 use libloading::{Library, Symbol};
 use once_cell::unsync::OnceCell;
 use regex::{Regex, RegexBuilder};
+use serde::{Deserialize, Deserializer};
 use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::io::BufReader;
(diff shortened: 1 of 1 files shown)


toyboot4e:
Related to #732

cli/loader/Cargo.toml
```diff
@@ -14,6 +14,7 @@ repository = "https://github.com/tree-sitter/tree-sitter"
 anyhow = "1.0"
 cc = "^1.0.58"
 dirs = "3.0"
+home = "0.5.3"
```
dcreager, line 17: Small nit: We've already taken a dependency on `dirs`, which provides a [`home_dir`](https://docs.rs/dirs/3.0.2/dirs/fn.home_dir.html) function that we can use. Can we use that instead of taking a dependency on the `home` crate?

toyboot4e: Sure, let me fix and force push.

toyboot4e: Pushed :)

cli/loader/src/lib.rs
```diff
@@ -18,10 +19,42 @@ use tree_sitter_tags::{Error as TagsError, TagsConfiguration};
...
+{
+    let paths = Vec::<PathBuf>::deserialize(deserializer)?;
+    let home = match dirs::home_dir() {
+        Some(home) => home,
+        None => return Ok(paths),
+    };
+    let standarized = paths
+        .into_iter()
+        .map(|path| standarize_path(path, &home))
+        .collect();
```
dcreager, line 44: One last nit, I promise! :smile:  `standardized` and `standardize_path` are misspelled. (missing `d`)

Otherwise this looks great, thanks!

toyboot4e: Thank you :) Which would you think is better, standardize or normalize?

toyboot4e: Thank you for the review! I fixed the misspellings and force pushed.

dcreager: No strong opinion between standardize and normalize, happy to go with what you wrote.  Thanks for the patch!

dcreager (approved):

cli/loader/src/lib.rs
toyboot4e: Then I'd go with standardize. Thank you <3

maxbrunsfeld:
This is a great improvement. Thanks so much @toyboot4e.

PR #1234, opened 2021-07-03
fix(cli): Improve error messages on config.json loading, closes #1227

### LINKED ISSUE #1227: tree-sitter-cli 0.20.0: better error messages when config.json parsing fails
I recently installed tree-sitter-cli 0.20.0 with `npm install -g`. This issue only occurs with the global install; the tree-sitter that lives in the .node_modules directory of a tree-sitter grammar seems to work fine. The issue is that no matter what command I run, the global tree-sitter command only outputs the following:
```
trailing comma at line 5 column 3
```
That's all. It does this when I run init-config, and parse, and highlight. It happens regardless of what console I am using. The only time it doesn't do this is when I run the `-h` or `-V` commands. Anyone know what could be causing this?

I should also add this occurred after I switched back and forth between 0.20.0 and 0.19.5 trying to isolate/reproduce an issue. Perhaps `npm uninstall -g tree-sitter-cli` doesn't remove everything?

ahlinc:
Closes #1227

Now it looks like:
```console
# tree-sitter highlight foobar
Failed to read /home/user1/.config/tree-sitter/config.json

Caused by:
    Permission denied (os error 13)
```
OR
```console
# tree-sitter highlight foobar
Bad JSON config /home/user1/.config/tree-sitter/config.json

Caused by:
    trailing comma at line 55 column 3
```

### DIFF
--- a/cli/config/src/lib.rs
+++ b/cli/config/src/lib.rs
@@ -1,6 +1,6 @@
 //! Manages tree-sitter's configuration file.
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::path::PathBuf;
@@ -64,8 +64,10 @@ impl Config {
             Some(location) => location,
             None => return Config::initial(),
         };
-        let content = fs::read_to_string(&location)?;
-        let config = serde_json::from_str(&content)?;
+        let content = fs::read_to_string(&location)
+            .with_context(|| format!("Failed to read {}", &location.to_string_lossy()))?;
+        let config = serde_json::from_str(&content)
+            .with_context(|| format!("Bad JSON config {}", &location.to_string_lossy()))?;
         Ok(Config { location, config })
     }
 


cli/config/src/lib.rs
```diff
@@ -64,8 +64,20 @@ impl Config {
             Some(location) => location,
             None => return Config::initial(),
         };
-        let content = fs::read_to_string(&location)?;
-        let config = serde_json::from_str(&content)?;
+        let content = fs::read_to_string(&location).or_else(|err| {
+            Err(anyhow!(
```
maxbrunsfeld, line 68: I think you could use anyhow's [`with_context`](https://docs.rs/anyhow/1.0.42/anyhow/trait.Context.html#tymethod.with_context) method to express this slightly more concisely:

```rust
let content = fs::read_to_string(&location)
    .with_context(|| format!("failed to read {}", location.to_string_lossy()))?;
// ...
```

maxbrunsfeld (commented):
This is a good change. Let me know what you think of the one style suggestion I made.

cli/config/src/lib.rs
ahlinc: Thank you for the suggestion! I've fixed it.

maxbrunsfeld:
⚡️

PR #1247, opened 2021-07-08
use libtool for building

DDoSolitary:
I'm trying to build this project on MSYS (a fork of Cygwin) and got troubles because the current Makefile doesn't have rules to build DLL files.

This patch replaces the manually written rules with libtool commands to support building libraries for different platforms. The new Makefile should produce the same output as the old one (libtree-sitter.a and libtree-sitter.so.0 with proper symlinks), plus a libtool library file libtree-sitter.la. (I've tested on Linux and MSYS, but don't have a Mac device to test on.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ docs/assets/js/tree-sitter.js
 *.exp
 *.lib
 *.wasm
+
+.libs/
+*.la
+*.lo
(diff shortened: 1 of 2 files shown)


ahlinc:
@eli-schwartz Do you have any concerns about this proposed change to use `libtool` in the Makefile?

amaanq:
needs a rebase, also -D for install doesn't exist on macOS and I don't think it's installed by default.

PR #1250, opened 2021-07-09
Use emcc modularize option

jrieken:
tl;dr: fixes https://github.com/tree-sitter/tree-sitter/issues/559

This PR uses the [`MODULARIZE `](https://github.com/emscripten-core/emscripten/blob/667d416d84b33b5bdbf47de10a9780b85589117b/src/settings.js#L1161) and `EXPORT_NAME` options when invoking `emcc`. This makes emcc emit a top-level function which exports itself, thereby making prefix.js and suffix.js obsolete. 

The wanted effect of these changes is that the top-level function can be invoked with the emcc `Module` object. This allows to tweak some things, namely how the wasm-binary is located. E.g this is how I can now use tree-sitter:

```ts
import TreeSitter, { Parser } from 'web-tree-sitter';

TreeSitter({
  locateFile: (path) => {
    // massage `path` and return actual path
    return _applyPathMagic(path);
  }
}).then(module => {
  const parser = new module.Parser();
  // happy parsing!
});
```

Also, `Parser.init()` is becoming obsolete with these changes because the `TreeSitter` function returns a promise that resolves when things are ready. Tho, I didn't remove it yet as it also doesn't cause harm. 

The obvious down-side is that this is a breaking change which existing consumers need to adopt.

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -1206,3 +1206,4 @@ function marshalEdit(edit) {
 
 Parser.Language = Language;
 Parser.Parser = Parser;
+C.Parser = Parser;
--- a/lib/binding_web/check-artifacts-fresh.js
+++ b/lib/binding_web/check-artifacts-fresh.js
@@ -8,7 +8,6 @@ const inputFiles = [
   'binding.js',
   'exports.json',
   'imports.js',
-  'prefix.js',
   ...list('../include/tree_sitter'),
   ...list('../src')
 ]
--- a/lib/binding_web/prefix.js
+++ b/lib/binding_web/prefix.js
@@ -1,9 +1,0 @@
-(function (root, factory) {
-  if (typeof define === 'function' && define.amd) {
-    define([], factory);
-  } else if (typeof exports === 'object') {
-    module.exports = factory();
-  } else {
-    window.TreeSitter = factory();
-  }
-}(this, function () {
--- a/lib/binding_web/suffix.js
+++ b/lib/binding_web/suffix.js
@@ -1,2 +1,0 @@
-return Parser;
-}));
--- a/lib/binding_web/test/helper.js
+++ b/lib/binding_web/test/helper.js
@@ -1,11 +1,11 @@
-const Parser = require(`..`);
+const TreeSitter = require(`..`);
 
 function languageURL(name) {
   return require.resolve(`../../../target/release/tree-sitter-${name}.wasm`);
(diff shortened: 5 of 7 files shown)


maxbrunsfeld:
@jrieken Sorry for the delay on this. I think we definitely need to expose this `locateFile` functionality.

I'm also ok with making a breaking API change if we need to in order to achieve better usability. But for this feature (providing an alternative URL for the WASM file), I'm wondering if we could actually do it in a way that *preserves* backwards-compatibility.

I think the new API structure that you've outlined seems a bit more complex than the old one. Am I correct that with this new API, the `Parser` *type* is provided by the initial `web-tree-sitter` import, but the actual class used at runtime is provided by the value that `TreeSitter` resolves with asynchronously?

I thought it was actually simpler to just provide a single class as the `web-tree-sitter` import, and have `init` just be a static method you could call. What do you think about keeping that API structure, but just changing `init` to take an `options` object, where one of the keys is `wasmURL`?

jrieken:
(I am back from vacation and rolling)

> What do you think about keeping that API structure, but just changing init to take an options object, where one of the keys is wasmURL?

That would be nice but I am not sure that's possible with the current `emcc` options nor with the `MODULARIZE` option. Without the modularize-option the `wasmBinaryFile`-variable is set during load (run `./script/build-wasm --debug` and check `tree-sitter.js#L1176`). This happens before declaring the `init`-function and works with the `Module`-global, which is hidden by the custom prefix-script. 

Once the `MODULARIZE`-option is used, the export is a single [function that returns](https://github.com/emscripten-core/emscripten/blob/667d416d84b33b5bdbf47de10a9780b85589117b/src/settings.js#L1102) the actual module. That makes the `d.ts`-file and hence TS-consumption more complex.

jrieken:
> I haven't actually tested the locateFile bit, but I think this shows the idea. When you get a chance, could you test this approach and see if it works for your use case?

It doesn't work quite yet because `locateFile` is called before `Module.locateFile` is being set from `treeSitterInitOptions`. 

However, seeing your changes and seeing the new structure made me realize that now the `Module`-global can be updated from within `init`. That makes this work with API breakage. The usage is as you outlined it above but all options of the emscripten module-object will be supported. 

I have taken (stolen) your changes and added some on top. See this new PR: https://github.com/tree-sitter/tree-sitter/pull/1367

jrieken:
closing in favour of #1367

PR #1273, opened 2021-07-15
Generate header file

FnControlOption:
This allows us to check node types more efficiently by using `ts_node_symbol()` instead of `ts_node_type()`. For example, this:

```c
assert(strcmp(ts_node_type(root_node), "program") == 0);
```

can be replaced with this:

```c
assert(ts_node_symbol(root_node) == sym_program);
```

Currently, the value of `sym_program` can be retrieved by doing this:

```c
TSSymbol sym_program = ts_language_symbol_for_name(language, "program", 7, true);
```

However, this isn't as efficient as doing `#include "enums.h"`

### DIFF
--- a/cli/src/generate/mod.rs
+++ b/cli/src/generate/mod.rs
@@ -101,7 +101,7 @@ pub fn generate_parser_in_directory(
     Ok(())
 }
 
-pub fn generate_parser_for_grammar(grammar_json: &str) -> Result<(String, String)> {
+pub fn generate_parser_for_grammar(grammar_json: &str) -> Result<(String, String, String)> {
     let grammar_json = JSON_COMMENT_REGEX.replace_all(grammar_json, "\n");
     let input_grammar = parse_grammar(&grammar_json)?;
     let (syntax_grammar, lexical_grammar, inlines, simple_aliases) =
@@ -115,7 +115,7 @@ pub fn generate_parser_for_grammar(grammar_json: &str) -> Result<(String, String
         true,
         None,
     )?;
-    Ok((input_grammar.name, parser.c_code))
+    Ok((input_grammar.name, parser.c_code, parser.header_code))
 }
 
 fn generate_parser_for_grammar_with_opts(
--- a/cli/src/generate/mod.rs
+++ b/cli/src/generate/mod.rs
@@ -87,7 +87,7 @@ pub fn generate_parser_in_directory(
     )?;
 
     write_file(&src_path.join("parser.c"), c_code)?;
-    write_file(&src_path.join("enums.h"), header_code)?;
+    write_file(&src_path.join("parser.h"), header_code)?;
     write_file(&src_path.join("node-types.json"), node_types_json)?;
 
     if next_abi {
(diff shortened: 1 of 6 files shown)


ahlinc:
Good idea! :slightly_smiling_face: I also have thoughts about changing the `parser.c` generation and make some of its structures that may be useful for external scanners available in a header file. May be it make sense to name such file `parser.h`? So it would be used in the `parser.c` and available for other parts?

FnControlOption:
Thank you. ❤️ Yes, that makes sense. In fact, I realized that we could also add a function prototype for `tree_sitter_*`

maxbrunsfeld:
The symbol enum really isn't meant to be publicly accessible right now - it doesn't abide by the C naming convention for "public" symbols, in which they are prefixed by the name of the library to avoid collisions with symbols from other libraries. With this approach, you would almost definitely get a 'duplicate symbol' error from your C compiler if you tried to *include* the headers for multiple tree-sitter parsers.

I also prefer the `ts_language_symbol_for_name` approach because most applications that use Tree-sitter aren't written in C or C++; they're written in some other language, and they use Tree-sitter via bindings. The `ts_language_symbol_for_name` approach is portable to other languages, using bindings to that function. But this header file would only be useful in a more niche case.

Also, requiring each `parser.c` to be compiled against a distinct header file would make it a bit more difficult to compile a set of *multiple* tree-sitter parsers, because now, each `parser.c` would need to be compiled with separate `-H` flags.

Overall, I really value keeping Tree-sitter's public interface as tiny as possible, and so I don't feel like this change offers enough benefit to warrant the additional complexity and public surface area. I'm open to other arguments about why this change is beneficial though. Can you talk more about why the `ts_language_symbol_for_name` approach doesn't meet your application's needs?

dcreager:
I'm also skeptical of the efficiency argument against using `ts_language_symbol_for_name`, since you can do that once _per language_ and cache the results, amortizing that across many parser invocations.

ahlinc:
> it doesn't abide by the C naming convention for "public" symbols ... you would almost definitely get a 'duplicate symbol' error from your C compiler ...

> because now, each `parser.c` would need to be compiled with separate `-H` flags.

> you can do that once per language and cache the results

@maxbrunsfeld @dcreager Strong reasons! Thank you for detailed comments!

My use case that I'd like to discuss to have some generated header file is related to the [`ts_external_scanner_states`](https://github.com/tree-sitter/tree-sitter-html/blob/af9339f3deb131ab99acfac906713b81dbcc41c9/src/parser.c#L663-L709) array that I found would be useful for external scanners in some complex situations. But that's I suppose is better to discuss in a separate issue.

maxbrunsfeld:
@FnControlOption This looks really well implemented, but I think we don't want to go in this direction, for the reasons discussed above.

Scorg:
> I'm also skeptical of the efficiency argument against using `ts_language_symbol_for_name`, since you can do that once _per language_ and cache the results, amortizing that across many parser invocations.

Why do I have to guess or search for the string names of the symbols while they are already defined in an enum?

amaanq:
@Scorg, stop necroposting with unhelpful comments, this is your first & last warning

PR #1278, opened 2021-07-19
feat(cli): Make "test" output more readable

CyberShadow:
Before:

image

After:

image

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -114,7 +114,9 @@ pub fn run_tests_at_path(
             print_diff_key();
             for (i, (name, actual, expected)) in failures.iter().enumerate() {
                 println!("\n  {}. {}:", i + 1, name);
-                print_diff(actual, expected);
+                let actual = format_sexp_indented(&actual, 2);
+                let expected = format_sexp_indented(&expected, 2);
+                print_diff(&actual, &expected);
             }
             Err(anyhow!(""))
         }
@@ -153,8 +155,7 @@ pub fn print_diff_key() {
 }
 
 pub fn print_diff(actual: &String, expected: &String) {
-    let changeset = Changeset::new(actual, expected, " ");
-    print!("    ");
+    let changeset = Changeset::new(actual, expected, "\n");
     for diff in &changeset.diffs {
         match diff {
             Difference::Same(part) => {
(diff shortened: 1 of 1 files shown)


CyberShadow:
By the way, am I the only one who finds the color choice here backwards? Intuitively I would expect the colors to match those shown after running `tree-sitter test --update && git diff`. I.e., I made a change in the grammar (or test source code), and I would expect that resulting additions to the parsed tree to be shown in green and deletions in red.

ahlinc:
@CyberShadow No, you're not alone. I suppose there was logic from a test perspective, what's needed to achieve for passing the test. But from a user actions perspective I agree that may be it would be better to be closer to what a VCS shows in its diff output.
But here would be awesome to here other opinions.

ahlinc:
Also I thought that may be it makes sense to add a flag that may disable such indented formatting to be able to have the compact output as before in a case when every details are unimportant and what's needed it's an ability to quickly estimate corruptions around tests that are happened by some change.

CyberShadow:
Maybe something like `test -q` (similar to `parse -q`) would make sense for that use case?

ahlinc:
As I saw across tree-sitter sub commands the `-q` flag is used **to suppress** all additional output.
I'd like to suggest `-c, --compact` for this use case.

CyberShadow:
> As I saw across tree-sitter sub commands the `-q` flag is used **to suppress** all additional output.

That is what I suggested. Maybe I'm not understanding the situation you're describing that would make this additional switch worthy of inclusion. Generally I can't reasonably expect anyone to make informed decision based on outputs formatted as they currently are for any non-trivially-sized tests.

Maybe it would make more sense to close this PR and instead put more focus on `test --update` in the documentation, as the `git diff` output doesn't have any of the problems discussed here, and (depending on the user's choice for a git UI) the test fallout can be more easily perused for either a summary or detailed diffs.

ahlinc:
I'm for sure :+1: for the changes in this PR.

> That is what I suggested.

Hmm. With a good description for the suggested `-q` flag like "Show tests results in the compact form" I support to name such flag as the `-q`.

ahlinc:
And I've a lot of the interest to improve testing experience to make a test first approach in grammars development easy and comfortable as much as possible. I think this PR is a good step toward of such improvements.

CyberShadow:
OK. So, would it make sense that `-q` succeeds silently if all tests pass, and only prints something in case of a problem? I can see use cases for that (CI or integration tests). Or did you have in mind to just make it keep the current behavior?

ahlinc:
I think it may be as such:
- `test -q` for the compact form.
- `test -qq` for the silence in case of all tests are passed. The `clap` crate allows this:

> Flags / Switches (i.e. bool fields)
>> Supports multiple occurrences (i.e. -vvv or -v -v -v)

maxbrunsfeld:
This is great.

If you're interested in switching the colors as part of this PR (so that "actual" would be green and "expected" would be red), I'd be fine with that too.

maxbrunsfeld:
Merging as-is, and we can always flip the colors later.

chharvey:
Is there a colorblind mode? Some people confuse red and green very easily.

PR #1294, opened 2021-07-25
Allow TextProvider's iterators to generate owned text

ubolonton:
The switch to `TextProvider` made it no longer possible to use a callback that returns owned text, which was enabled by #488 (for Emacs dynamic modules).

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -8,6 +8,7 @@ pub mod allocations;
 use std::os::unix::io::AsRawFd;
 
 use std::{
+    borrow::Cow,
     char, error,
     ffi::CStr,
     fmt, hash, iter,
@@ -162,7 +163,8 @@ pub struct QueryCaptures<'a, 'tree: 'a, T: TextProvider<'a>> {
 }
 
 pub trait TextProvider<'a> {
-    type I: Iterator<Item = &'a [u8]> + 'a;
+    type I: Iterator<Item = Cow<'a, [u8]>>;
+
     fn text(&mut self, node: Node) -> Self::I;
 }
 
@@ -1799,19 +1801,19 @@ impl<'a, 'tree> QueryMatch<'a, 'tree> {
         buffer2: &mut Vec<u8>,
         text_provider: &mut impl TextProvider<'a>,
     ) -> bool {
-        fn get_text<'a, 'b: 'a, I: Iterator<Item = &'b [u8]>>(
+        fn get_text<'a, 'b: 'a, I: Iterator<Item = Cow<'b, [u8]>>>(
             buffer: &'a mut Vec<u8>,
             mut chunks: I,
-        ) -> &'a [u8] {
-            let first_chunk = chunks.next().unwrap_or(&[]);
+        ) -> Cow<'a, [u8]> {
+            let first_chunk = chunks.next().unwrap_or(Cow::Owned(vec![0u8; 0]));
             if let Some(next_chunk) = chunks.next() {
                 buffer.clear();
-                buffer.extend_from_slice(first_chunk);
-                buffer.extend_from_slice(next_chunk);
+                buffer.extend_from_slice(&first_chunk);
+                buffer.extend_from_slice(&next_chunk);
                 for chunk in chunks {
-                    buffer.extend_from_slice(chunk);
+                    buffer.extend_from_slice(&chunk);
                 }
-                buffer.as_slice()
+                Cow::Borrowed(buffer.as_slice())
             } else {
                 first_chunk
             }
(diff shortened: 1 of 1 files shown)


ubolonton:
@maxbrunsfeld Have you been able to take a look at this yet?

ahlinc:
@ubolonton Have you consider an option to implement the `TextProvider` for your `Value<'e>` with some caching of requested text ranges in the `elisp-tree-sitter` library or have any concerns about it?

_In #488 there was the easy solution cause tree-sitter already used `X.as_ref()` for text retrieval in a function that was a consumer for a text callback result. I'm not sure that `Cow` is a good way to solve this case by adding one more container for the data that also don't need to be mutated by tree-sitter itself._

If I correctly understand your case, you don't have a buffer or buffers with a full source text that you can share to tree-sitter internals reliably and your `Value<'e>` dynamically retrieves text portions from the elisp side and wraps such portions into new `Value<'e>` objects.

ubolonton:
> @ubolonton Have you consider an option to implement the `TextProvider` for your `Value<'e>` with some caching of requested text ranges in the `elisp-tree-sitter` library or have any concerns about it?

What do you mean by caching of requested text ranges? A buffer to temporarily store text ranges? That would result in redundantly copying the generated owned texts. Avoiding that is an intended use case of  `Cow`.

> _In #488 there was the easy solution cause tree-sitter already used `X.as_ref()` for text retrieval in a function that was a consumer for a text callback result. I'm not sure that `Cow` is a good way to solve this case by adding one more container for the data that also don't need to be mutated by tree-sitter itself._

What are your specific concerns?

> If I correctly understand your case, you don't have a buffer or buffers with a full source text that you can share to tree-sitter internals reliably and your `Value<'e>` dynamically retrieves text portions from the elisp side and wraps such portions into new `Value<'e>` objects.

This is partially correct. The original source text is internal to Emacs, and cannot be shared. The `Value<'e>` object here is not a wrapper around text. It's a wrapped Lisp function that can be invoked to return `String`s.  Basically the problem boils down to implementing `TextProvider` for owned texts, not borrowed texts:

```rust
impl<'a> TextProvider<'a> for Vec<u8>

impl<'a, F, I> TextProvider<'a> for F
where
    F: FnMut(Node) -> I,
    I: Iterator<Item = Vec<u8>> + 'a
```

domq:
Hi, I'm new to tree-sitter (and I don't know Rust), so my apologies if I'm misinformed. It seems to me that because of this #1294 being pending, [elisp-tree-sitter](https://github.com/emacs-tree-sitter/elisp-tree-sitter) is [stuck](https://github.com/emacs-tree-sitter/elisp-tree-sitter/releases/tag/0.18.0) with what amounts to a forked version of tree-sitter from February, 2022; while the rest of the world (not the least of which being Emacs itself, which [merged `treesit` in November](https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01443.html)) is moving on.

Currently, for instance, I'm getting failures from the [`tree-sitter-langs`](https://github.com/emacs-tree-sitter/tree-sitter-langs) test suite because of ABI mismatch errors. It seems that my options are
- to upgrade elisp-tree-sitter, which (because of the way Cargo patches work) appears to require me to fork tree-sitter and rebase the PR myself;
- or downgrade the tree-sitter library on my system, which sooner or later will force me to rebuild Emacs with it.

Obviously I'd prefer the third option where this pull request gets merged and everything just works ☺

Respectfully yours,

domq:
>    to upgrade elisp-tree-sitter, which (because of the way Cargo patches work) appears to require me to fork tree-sitter and rebase the PR myself;

Update: I managed to do just that in #2011, which I tested by throwing https://github.com/emacs-tree-sitter/elisp-tree-sitter/pull/249 on top. By the power of these changes combined, plus just a small amount of cheating in the `.cask` subdirectory, I got the green bar in ([my fork of](https://github.com/domq/tree-sitter-langs)) `tree-sitter-langs` that I was after (as mentioned in [my earlier message](https://github.com/tree-sitter/tree-sitter/pull/1294#issuecomment-1372478755)).

@ubolonton, I'm not sure how to do a pull request on a pull request... In fact, I tried, and the diff wasn't pretty for obvious reasons. So feel free to reset your branch to mine (it was your work in the first place).

sogaiu:
Is this enough for elisp-tree-sitter + tree-sitter-langs to move to a newer version of tree-sitter?

I thought https://github.com/tree-sitter/tree-sitter/issues/1336 might need attention -- it's for tree-sitter-langs but that seems like it would be important for practical elisp-tree-sitter usage.

sogaiu:
It looks like @ahlinc has attended to #1336 in https://github.com/tree-sitter/tree-sitter/pull/2013.

Thanks!

domq:
> Is this enough for elisp-tree-sitter + tree-sitter-langs to move to a newer version of tree-sitter?

In  my experiments, elisp-tree-sitter won't build without either #1294 or #2011 applied.

sogaiu:
I gave it an attempt and mentioned it at https://github.com/tree-sitter/tree-sitter/pull/2011#issuecomment-1375153978.

Seems like it worked mostly :)

PR #1295, opened 2021-07-25
docs(Using parsers): Fix spelling, remove unusual keyword

nokome:
The `atch` in the example seems like a mistake (not in alphabetical order like the rest, not a keyword in any language that I'm aware of). Rather than assuming it refers to `match`, I just removed it.


frankeld:
I believe it may be `catch`

maxbrunsfeld:
Thanks!

PR #1304, opened 2021-07-29
Correct quiet description for playground

Razzeee:
Not sure how to best test this/my rust might be dangerous.

But as firefox didn't want to open the page and I wasn't getting the error either, I thought it might be helpful to always print the url.


maxbrunsfeld (approved):
Thanks for noticing that @Razzeee.

ahlinc:
@Razzeee It's merged. Thanks!

PR #1319, opened 2021-08-04
fix(cli): allow dead code in Logger; fixes CI failure

ahlinc:
This PR fixes a failure in the CI due to a dead code for some unfinished feature in the CLI `Logger`.
It seems `rustc` 1.54.0 improved its dead code detection.

### DIFF
--- a/cli/src/logger.rs
+++ b/cli/src/logger.rs
@@ -1,5 +1,6 @@
 use log::{LevelFilter, Log, Metadata, Record};
 
+#[allow(dead_code)]
 struct Logger {
     pub filter: Option<String>,
 }


ahlinc:
@maxbrunsfeld Could you look to this PR onto a piece of unused code that modern Rust compiler found? Could we delete that unused piece of code or do you interested to implement some logging filtering feature?

maxbrunsfeld:
Thanks for jumping on this @ahlinc. I'm ok with deleting the unused code, or just marking it as unused. I'm don't recall right now the origin of that code.

maxbrunsfeld (approved):

ahlinc:
I suppose when we would improve logging some filtering will make sense. I merge it as-is so it would serve as a notice.

PR #1329, opened 2021-08-08
Improve performance of `tree-sitter generate`

narpfel:
I spent some time profiling `tree-sitter generate` and found some spots that were easily improved:

* `HashMap` and the standard library hasher accounted for 33 % (half of that time spent hashing stuff) of wall time, so I switched the three hottest hash maps to [`IndexMap`s](https://github.com/bluss/indexmap) using [`FxHash`](https://github.com/rust-lang/rustc-hash) as the hashing algorithm. Unlike `SipHash13` (which `std::hash::DefaultHasher` currently uses), `FxHash` is not HashDoS resistant, but this is presumably not needed here.
* In `Generator::add_parse_table`, `find` was repeatedly used on a growing `Vec`, resulting in quadratic behaviour.
* A small gain results from improving the hashing implementation of `SmallBitVec`: https://github.com/servo/smallbitvec/pull/27

Benchmarks (on `tree-sitter-cpp`):
```
Summary
  './tree-sitter-3-smallvec generate' ran
    1.04 ± 0.01 times faster than './tree-sitter-2-hash-map generate'
    1.41 ± 0.02 times faster than './tree-sitter-1-indexmap generate'
    1.85 ± 0.04 times faster than './tree-sitter-0-master generate'
```


maxbrunsfeld:
This is great, thank you! Could you explain in a bit more detail about how you chose to use IndexMap? Is that data structure more suitable for this workload than the std hashmap? Or does IndexMap allow the use of different hashing algorithms than std HashMap?

narpfel:
The hash algorithm of `std::collections::HashMap` can be swapped as well, however `IndexMap` stores its keys and values densely in memory, so iterating is more cache efficient. I couldn’t measure whether this is the reason why `IndexMap` is faster here, but I chose to use it nonetheless as it was already a transitive dependency via `serde_json`. `HashMap` is just a couple percent slower (1  to 5 %, depending on the grammar), so it’s not a huge win in any case.

The other major difference between `HashMap` and `IndexMap` is that `IndexMap`s store items in insertion order, but other than that, `IndexMap` is basically a normal `HashMap`.

In conclusion, most of the effect of cf69a2c94cef61b8d07e80832e6e92cc78217801 results from using `FxHash`.

maxbrunsfeld:
Ok, thanks for the explanation, and pointing out that it was already a transitive dep.

maxbrunsfeld (approved):

PR #1337, opened 2021-08-14
Remove unnecessary borrows

narpfel:
This produces an `unused_must_use` warning on nightly:
https://github.com/rust-lang/rust/pull/86426


maxbrunsfeld:
Thanks!

PR #1346, opened 2021-08-19
Handle aliases in unicode property escapes in regexes

FnControlOption:
This allows using `\p{Space_Separator}` instead of `\p{Zs}` in regexes


maxbrunsfeld (commented):
Great work!

It looks like you added a unicode property alias to one of the test fixtures. Could you add an entry to the `corpus.txt` for that parser, which just verifies that the `letter_number` token is produced?

maxbrunsfeld (approved):

PR #1058, opened 2021-04-14
Corpus test files support arbitrary suffixes in headers and separators

ahelwer:
Some langues (example: TLA+) have language constructs which conflict with the corpus test file's header and separator format of `====` and `----`. Users can now work around this problem with the addition of separator suffixes, a string of arbitrary characters after the first `====` which must be appended to all subsequent `====` and `----`. This also enables inclusion of the `=` character in test names. For example:

```
=========|||
First test name
=========|||
<language input>
----------------|||
<s expression>
=========|||
Second = test = name
=========|||
<language input>
---------------|||
<s expression>
```

Fixes #982 and #1047

Issue #1047, opened 2021-04-06
Unexpected behaviour when test name contains equals = character


Sjord:
Having `=` in the title of a test makes the test before it fail.

## Reproduction

In [tree-sitter-php/test/corpus/class.txt](https://github.com/tree-sitter/tree-sitter-php/blob/master/test/corpus/class.txt#L38), change `Anonymous classes` to `Anonymous=classes` and run `tree-sitter test`.

## Actual behaviour

First, no test `Anonymous=classes` shows: 

```
  class:
    ✗ Abstract class
    ✓ Conditional class definition
    ✓ Class constant modifiers
    ✓ Final class
    ✓ Implicitly public properties and methods
    ✓ Property Types
```

Second, the test before that fails in a peculiar way:

```
  1. Abstract class:
[...]

PR #1348, opened 2021-08-19
Add support for arbitrary suffixes in test file headers and separators

### LINKED ISSUE #982: Workaround for tree-sitter test file separator conflict
Copied from [this](https://stackoverflow.com/q/66606818/2852699) stackoverflow question, since upon examining the source code I don't see a workaround.

TLA+ uses `----` and `====` in the top-level definition of a module:

```
---- MODULE Test ----
====
```

This conflicts with the tree-sitter test file format, so I can't have more than one test in a file. One solution to this issue is to expose a CL parameter for `tree-sitter test` which changes the character used in the separators. For example:

```
tree-sitter test -start "@" -end "%"
```

ahelwer:
Some langues (example: TLA+) have language constructs which conflict with the corpus test file's header and separator format of `====` and `----`. Users can now work around this problem with the addition of separator suffixes, a string of arbitrary characters after the first `====` which must be appended to all subsequent `====` and `----`. This also enables inclusion of the `=` character in test names. For example:

```
=========|||
First test name
=========|||
<language input>
----------------|||
<s expression>
=========|||
Second = test = name
=========|||
<language input>
---------------|||
<s expression>
```
These changes are currently available to use in https://www.npmjs.com/package/@tlaplus/tree-sitter-cli, which is in active use by the [tree-sitter-tlaplus grammar](https://github.com/tlaplus-community/tree-sitter-tlaplus/).

Fixes #982 and #1047

(This PR replaces https://github.com/tree-sitter/tree-sitter/pull/1058 since I moved the forked tree-sitter repo under the tlaplus-community organization)

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -5,7 +5,6 @@ use difference::{Changeset, Difference};
 use lazy_static::lazy_static;
 use regex::bytes::{Regex as ByteRegex, RegexBuilder as ByteRegexBuilder};
 use regex::Regex;
-use std::char;
 use std::ffi::OsStr;
 use std::fmt::Write as FmtWrite;
 use std::fs;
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Sorry for the delay. This looks great @ahelwer.

PR #1353, opened 2021-08-22
сhore: Add license files to npm packages

### LINKED ISSUE #979: web-tree-sitter npm package doesn't contain license file
Hi,

I've noticed that [web-tree-sitter npm package](https://www.npmjs.com/package/web-tree-sitter) is not distributed with a LICENSE file. As I understand it from the GitHub repository structure the code itself is licensed under MIT, but when it gets distributed the LICENSE file is not present. Unfortunately MIT requires that code is distributed with full license text. 

Would it be possible to include license into release process of web-tree-sitter?

Thanks!

ahlinc:
Closes #979

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2018 Max Brunsfeld
+Copyright (c) 2018-2021 Max Brunsfeld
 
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
Thanks! I like the way you used the prepack script to avoid duplication.

PR #1367, opened 2021-08-30
Allow to pass Module object to init-function

jrieken:
Builds on top of https://github.com/tree-sitter/tree-sitter/pull/1250#issuecomment-907421445

Adds the ability to optionally pass the Emscription Module-object to `Parser.init` so that things like `locateFile` can be controlled. 

Real world usage: https://github.com/microsoft/vscode-anycode/commit/1a6b8818177635cb911e8f085ccf03219992a320

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -17,24 +17,15 @@ var MIN_COMPATIBLE_VERSION;
 var TRANSFER_BUFFER;
 var currentParseCallback;
 var currentLogCallback;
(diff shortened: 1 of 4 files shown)


jrieken:
This PR is better than https://github.com/tree-sitter/tree-sitter/pull/1250 - same functionality, nicer (existing) API structure.

maxbrunsfeld:
Awesome. Thanks for getting this finished.

PR #1372, opened 2021-08-31
feat(query): add first_retun flag to TSQueryMatch

vigoux:
Allows for users of the API to determine if the match is returned for the first time, for example to run the predicates only once.

We encountered this little problem when working on tree-sitter intergration on neovim.

cc @bfredl

### DIFF
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -111,6 +111,7 @@ typedef struct {
   uint16_t pattern_index;
   uint16_t capture_count;
   const TSQueryCapture *captures;
+  bool first_return;
 } TSQueryMatch;
 
 typedef enum {
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -159,11 +159,12 @@ typedef struct {
   uint16_t start_depth;
   uint16_t step_index;
   uint16_t pattern_index;
-  uint16_t consumed_capture_count: 12;
+  uint16_t consumed_capture_count: 11;
   bool seeking_immediate_match: 1;
   bool has_in_progress_alternatives: 1;
   bool dead: 1;
   bool needs_parent: 1;
+  bool returned: 1;
 } QueryState;
 
 typedef Array(TSQueryCapture) CaptureList;
@@ -3160,6 +3161,8 @@ bool ts_query_cursor_next_match(
   }
 
   QueryState *state = &self->finished_states.contents[0];
+  state->returned = true;
+  match->first_return = false;
   match->id = state->id;
   match->pattern_index = state->pattern_index;
   const CaptureList *captures = capture_list_pool_get(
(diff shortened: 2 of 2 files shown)


maxbrunsfeld:
The `id` field on `TSQueryMatch` exists for a similar purpose (so you can check if the predicates have already been run), but maybe `first_return` is more useful than `id`. Did you consider using `id` to check if the predicates have run?

maxbrunsfeld:
I'm just remembering more about this problem...

The benefit of the `id` field is that if the predicates *fail* to match, you can get rid of the match by calling `ts_query_cursor_remove_match(cursor, id);`, which is good because then you never have to deal with that match again, and it doesn't occupy any storage space inside of the `TSQueryCursor`. Right now, though, it's probably inconvenient to have to keep *track* of the ids, because the matches are not guaranteed to be returned in ascending order of `id`. Is that why you're not using `id` currently?

I think we could actually change the implementation so that the `id` counter would only be incremented once a match was actually *returned*. That way, in order to avoid running predicates repeatedly, you could just keep track of a `max_match_id`. What do you think? Would that work for your use case? If so, I'd prefer that solution, because it avoids having two somewhat redundant fields (`id` and `first_return`).

maxbrunsfeld:
I pushed a small commit to `master` that ensures the match ids are sequential. Could you take a look and see if this would enable your use case in Neovim?

bfredl:
@maxbrunsfeld yes, I think that would work for us, we will give it a try :+1:

ubolonton:
Seems like [that commit](https://github.com/tree-sitter/tree-sitter/commit/22a5cfbe102056ff3d51b2b44a8666fd8f85fdc0) also fixed an issue I had recently, illustrated using the [HCL grammar](https://github.com/MichaHoffmann/tree-sitter-hcl/tree/b048a42):

With this code:
```hcl
resource "random_pet" "this" {
  length = 2
}

resource "tls_private_key" "this" {
  algorithm = "RSA"
}

resource "aws_key_pair" "this" {
  key_name   = random_pet.this.id
  public_key = tls_private_key.this.public_key_openssh
}
```

and this query:
```clojure
((block (identifier) @keyword
        . [(string_lit (template_literal) @function)
           (identifier) @function])
 (.eq? @keyword "output"))

((block (identifier) @keyword)
 (.match? @keyword "^(resource|data|output|locals|lifecycle)$"))
```

The command `tree-sitter query --captures` did not return the first `resource`. With that commit, it now returns all 3:
```
    pattern:  1, capture: 0 - keyword, start: (0, 0), end: (0, 8), text: `resource` # It did not return this.
    pattern:  1, capture: 0 - keyword, start: (4, 0), end: (4, 8), text: `resource`
    pattern:  1, capture: 0 - keyword, start: (8, 0), end: (8, 8), text: `resource`
```

vigoux:
Hey, sorry to bother you again, but the fix pushed to master effectively fixes our issue.
Following this, do you think it is possible to make a release following those changes ?

Thanks in advance !

maxbrunsfeld:
I published Tree-sitter 0.20.1.

PR #1373, opened 2021-09-03
fix(lib): fix segfault on ts_query_new with incompatible grammar version

ahlinc:
With this fix a `tree-sitter highlight` command outputs an error into stderr like:
```console
Failed to load property sheet for injection string 'jsdoc': Incompatible language version 11. Expected minimum 13, maximum 13
```

Screenshot from 2021-09-03 14-29-21


maxbrunsfeld:
Thanks for fixing this!

PR #1383, opened 2021-09-13
feat(cli): add a flag to compile a parser in debug mode with -O0 C/C++ compiler flag

ahlinc:
This PR adds a feature to compile parser files like `src/parser.c` and `src/scanner.c` with `-O0` flag.
This allows to easily use `tree-sitter parse` command with a debugger and have correct stepping by code lines of an external scanner in a debugging session.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -101,6 +101,7 @@ pub struct Loader {
     language_configuration_ids_by_file_type: HashMap<String, Vec<usize>>,
     highlight_names: Box<Mutex<Vec<String>>>,
     use_all_highlight_names: bool,
+    debug_build: bool,
 }
 
 unsafe impl Send for Loader {}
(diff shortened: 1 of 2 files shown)


cli/loader/src/lib.rs
```diff
@@ -347,7 +349,11 @@ impl Loader {
         parser_path: &Path,
         scanner_path: &Option<PathBuf>,
     ) -> Result<Language> {
-        let mut library_path = self.parser_lib_path.join(name);
+        let mut lib_name = name.to_string();
+        if self.debug_build {
+            lib_name.push_str(".debug._");
```
ahlinc, line 354: This is used to generate separate debug parser library versions like `~/.cache/tree-sitter/lib/javascript.debug.so` to don't affect regular versions when a user calls the `tree-sitter parse` without the `--debug-build` flag.

maxbrunsfeld:
This is a great idea. Thanks @ahlinc!

PR #1384, opened 2021-09-14
fix(dsl): Allow to use console.log() in grammar files, resolve #703

### LINKED ISSUE #703: console.log() in module causes some cryptic error
Hey, folks! First of all, I apologize for my ignorance... My guess is that this bug is happening because I'm a NodeJS noob. This is literally my first NodeJS project(and I've loved it so far!).

Anyways, the deets: I'm creating a tree-sitter for this language I use at work. We have a file which contains all the valid "instructions" and their arguments(very similar to functions). In my grammar.js, I've described the language very generally and I've required a module which parses this file containing all the instructions.

In the required module, I've been trying to use console.log() to debug and send some useful messages to the user as they're compiling the tree-sitter grammar, but console.log() appears to break the program.

Here's the output with console.log() in the required module:
`expected value at line 1 column 1`

And the output when I instead use console.error():
`Using CRBasic DEF file at /home/xeyler/crbasic-defs/CR6.DEF`

Naturally, the console.error() output is the output I'd expect when using console.log().

Any clue what could be causing this? Thanks!

ahlinc:
This PR fixes issue with `console.log()` usage in grammar files. It would be awesome to use a dedicated file descriptor for communication between CLI and an underling node.js sub process but such approach is not portable. So redirecting `console.log()` to stderr I think is enough solution.

Fixes #703

### Edit:
_There're other options on how to fix this if current solution wouldn't be appropriate for some reasons that I could miss:_
1.  _Don't use indented JSON and interpret only the last line of stdout as JSON grammar._
2. _Generate a long random token that would be unique and long enough, transfer it over environment variable and interpret only a part of the output that follows the token line._

### DIFF
--- a/cli/src/generate/dsl.js
+++ b/cli/src/generate/dsl.js
@@ -414,5 +414,7 @@ global.token = token;
 global.grammar = grammar;
 global.field = field;
 
+console.log = console.debug = console.info = console.error = console.warn;
 const result = require(process.env.TREE_SITTER_GRAMMAR_PATH);
-console.log(JSON.stringify(result, null, 2));
+process.stdout.write(JSON.stringify(result, null, 2));
+process.stdout.write('\n');


maxbrunsfeld:
I think the first alternative that you listed would be even better: formatting the JSON as a single line, and having [`load_js_grammar_file`](https://github.com/tree-sitter/tree-sitter/blob/561dfe3a2d228e5877f27b737e46628082bd7015/cli/src/generate/mod.rs#L198) only read the final line from the node process's stdout.

That way, if people used `console.log`, it would print to stdout as usual.

PR #1385, opened 2021-09-14
feat(rust): Add an id() method for QueryMatch

ahlinc:
This is follow up PR for https://github.com/tree-sitter/tree-sitter/commit/22a5cfbe102056ff3d51b2b44a8666fd8f85fdc0 commit and reasons described in https://github.com/tree-sitter/tree-sitter/pull/1372#issuecomment-918492254. It makes sense to make ids accessible for QueryMatch instances in the Rust binding.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -1753,6 +1753,10 @@ impl QueryCursor {
 }
 
 impl<'a, 'tree> QueryMatch<'a, 'tree> {
+    pub fn id(&self) -> u32 {
+        self.id
+    }
+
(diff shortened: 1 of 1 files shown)


maxbrunsfeld (approved):

ahlinc:
@maxbrunsfeld Thanks for the approval, this ids doesn't require for the [`remove()`](https://github.com/tree-sitter/tree-sitter/blob/22a5cfbe102056ff3d51b2b44a8666fd8f85fdc0/lib/binding_rust/lib.rs#L1756) method directly, but I found that they may be useful in custom logging for debug purposes.

PR #1391, opened 2021-09-17
Add WGSL WebGPU Shading Language

mehmetoguzderin:
This PR adds WebGPU Shading Language to the list of available grammars, the linked repository's grammar is periodically automatically extracted from the WGSL specification itself, which is actually extracted every single time where the specification gets a modification in itself to check the validity of both the syntax and the examples. Thank you very much for developing and maintaining tree-sitter, fantastic project!


maxbrunsfeld:
Thanks!

mehmetoguzderin:
Thank you for the merge! Just for reference, this is where the WGSL spec lives: https://github.com/gpuweb/gpuweb/tree/main/wgsl

PR #1397, opened 2021-09-21
binding(rust): Mark set_cancellation_flag self as mutable

ahlinc:
C lib definition is not **const**:
https://github.com/tree-sitter/tree-sitter/blob/a63c09375f76ed6dfe5e9a4f8384c805c009fb31/lib/include/tree_sitter/api.h#L312
So the Rust equivalent of the `&self ` receiver seems should be marked as mutable.


maxbrunsfeld:
Thanks!

PR #1399, opened 2021-09-21
fix(cli): Fix playground opening in a browser, regression in #1304

ahlinc:


maxbrunsfeld:
Thanks for fixing this!

PR #1411, opened 2021-09-24
Fix compiling C library with Clang on Windows

aminya:
Fixes a part of #1889

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1776,7 +1776,11 @@ void ts_parser_print_dot_graphs(TSParser *self, int fd) {
   }
 
   if (fd >= 0) {
+    #ifdef _WIN32
+    self->dot_graph_file = _fdopen(fd, "a");
+    #else
     self->dot_graph_file = fdopen(fd, "a");
+    #endif
   } else {
     self->dot_graph_file = NULL;
   }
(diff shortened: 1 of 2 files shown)


ahlinc:
I agree with the second warning but for the first warning I'd raiser add compiler flag than introduce preprocessor condition:
> If you need to use the existing functions or variables [for portability reasons, you can turn off the warning.](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-160#turn-off-the-warning)

> To fix this issue, we usually recommend you change your code to use the suggested function names instead. **However, the updated names are Microsoft-specific. If you need to use the existing function names [for portability reasons, you can turn off these warnings](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-160#posix-function-names). The functions are still available in the library under their original names.**

And I'd suggest to remove binding regeneration from this PR at all as it's not related and the bindgen on Windows produces slightly different results that on Linux or Macos, I think this will be included into CI process and wouldn't be a task for making such regeneration manually.

aminya:
> but for the first warning I'd raiser add a compiler flag than introduce the preprocessor condition:

This is more than a warning. Depending on the compiler installed on Windows and the linker includes, this can lead to a linker time error. For example, if someone uses the Clang compiler with Windows linker, Treesitter fails to compile due to the missing `fdopen` symbol.

Theoretically, this deprecation can become an actual compile-time error. The fact that MSVC runtime is providing a compatible `fdopen` symbol on a non POSIX system is just by chance.

Issue #940, opened 2021-02-25
Some external scanners used for testing cannot handler CRLF line endings


lukedan:
The following tests fail when using CRLF line endings:
- Language: `external_tokens`
  - `corpus - external tokens that require state`
- Language: `inverted_external_token`
  - `corpus - Expressions on one line`
  - `corpus - Line breaks followed by whitespace`

See also #913.

PR #1412, opened 2021-09-25
Fixed some test failures due to CRLF line endings

ahelwer:
Normalize line endings when comparing expected `tree-sitter generate` error output to actual output
Add `\r` skipping to some external scanners
Fixes https://github.com/tree-sitter/tree-sitter/issues/940
Partially addresses https://github.com/tree-sitter/tree-sitter/issues/913
The following issues in the bash and ruby grammars still must be resolved for all tests to pass on Windows:
- https://github.com/tree-sitter/tree-sitter-bash/issues/94
- https://github.com/tree-sitter/tree-sitter-ruby/issues/164


ahelwer:
@maxbrunsfeld this looks good

maxbrunsfeld:
Thanks!

PR #1414, opened 2021-09-26
Add tree-sitter-cuda, tree-sitter-glsl, tree-sitter-commonlisp

theHamsta:
Adding my parsers to the parser list. tree-sitter-cuda and tree-sitter-glsl are simple extensions of tree-sitter-cpp or tree-sitter-c respectively.


maxbrunsfeld:
These all look great. Thanks!

PR #1434, opened 2021-10-10
Update to unicode standard 14

razzeee:


razzeee:
No, I only saw that there's a new version, when trying to solve that problem

maxbrunsfeld:
Ah I see. We’ll this is still a good change; thanks!

PR #1435, opened 2021-10-11
Expose GrammarJSON to public

oovm:
I am trying to implement a workflow that does not require nodejs.

But I am missing some type information, the current workflow is

1. JS DSL
2. JSON (via NodeJS)
3. `GrammarJSON` (via serde_json)
4. `InputGrammar` (via crate::generate::parse_grammar)
5. c_code (via generate_parser_for_grammar_with_opts)

This means I need to expose at least one of `GrammarJSON` and `InputGrammar`.

(This will also cause a lot of related structures leak to public)


maxbrunsfeld:
Could you explain more about your alternative workflow? What is the javascript DSL you're using, and how are you running it? Could Tree-sitter be enhanced to allow that form of JS evaluation, so that this data type could still be encapsulated?

oovm:
Okay, my needs are a bit complicated. Give me about three days to write a document explaining it in detail.

PR #1473, opened 2021-11-05
Make SubtreeInlineData work on Big-Endian

### LINKED ISSUE #1466: SubtreeInlineData is broken on Big-Endian (specifically OpenBSD/macppc)
On OpenBSD/macppc (which is big-endian 32bit powerpc), even small examples like https://tree-sitter.github.io/tree-sitter/using-parsers#an-example-program fail (full paste: https://paste.sr.ht/~thestr4ng3r/65d668eac155ead1fd3a3cea74b3798a7ea3886a):
```
g5-openbsd$ egdb test-json-parser
...
(gdb) r
Starting program: /home/florian/dev/tstest/test-json-parser
...
assertion "!self.data.is_inline" failed: file "lib/src/subtree.c", line 375, function "ts_subtree_summarize_children"

Program received signal SIGABRT, Aborted.
thrkill () at /tmp/-:9
9	/tmp/-: No such file or directory.
(gdb) bt
#0  thrkill () at /tmp/-:9
#1  0xded35974 in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
#2  0xdecd7808 in _libc___assert2 (file=<optimized out>, line=<optimized out>, func=<optimized out>, failedexpr=<optimized out>) at /usr/src/lib/libc/gen/assert.c:52
#3  0x2c7b9fe4 in ts_subtree_summarize_children (self=..., language=0x2c7d5078 <tree_sitter_json.language>) at lib/src/subtree.c:375
#4  0x2c7bbac0 in ts_subtree_new_node (symbol=15, children=0xfffcdd58, production_id=0, language=0x2c7d5078 <tree_sitter_json.language>) at lib/src/subtree.c:526
#5  0x2c7a363c in ts_parser__reduce (self=0xc5193000, version=0, symbol=15, count=1, dynamic_precedence=0, production_id=0, is_fragile=false, end_of_non_terminal_extra=false) at lib/src/parser.c:853
#6  0x2c79f8b8 in ts_parser__advance (self=0xc5193000, version=0, allow_node_reuse=true) at lib/src/parser.c:1471
#7  0x2c79e96c in ts_parser_parse (self=0xc5193000, old_tree=0x0, input=...) at lib/src/parser.c:1890
#8  0x2c7a0668 in ts_parser_parse_string_encoding (self=0xc5193000, old_tree=0x0, string=0x2c783a05 "[1, null]", length=9, encoding=TSInputEncodingUTF8) at lib/src/parser.c:1943
#9  0x2c7a0570 in ts_parser_parse_string (self=0xc5193000, old_tree=0x0, string=0x2c783a05 "[1, null]", length=9) at lib/src/parser.c:1937
#10 0x2c794bc0 in main () at main.c:27
(gdb) up 3
#3  0x2c7b9fe4 in ts_subtree_summarize_children (self=..., language=0x2c7d5078 <tree_sitter_json.language>) at lib/src/subtree.c:375
375	  assert(!self.data.is_inline);
(gdb) p self
$1 = {data = {is_inline = true, visible = true, named = false, extra = false, has_changes = false, is_missing = true, is_keyword = false, symbol = 23 '\027', padding_bytes = 253 '\375', size_bytes = 8 '\b', padding_columns = 197 '\305', padding_rows = 1 '\001', lookahead_bytes = 9 '\t', parse_state = 13480},
  ptr = 0xc517fd08}
(gdb)
```

That is because of `SubtreeInlineData` being used as part of the `MutableSubtree` union:
```c
typedef struct {
  bool is_inline : 1;
  bool visible : 1;
  bool named : 1;
  bool extra : 1;
  bool has_changes : 1;
  bool is_missing : 1;
  bool is_keyword : 1;
  uint8_t symbol;
  uint8_t padding_bytes;
  uint8_t size_bytes;
  uint8_t padding_columns;
  uint8_t padding_rows : 4;
  uint8_t lookahead_bytes : 4;
  uint16_t parse_state;
} SubtreeInlineData;

typedef union {
  SubtreeInlineData data;
  SubtreeHeapData *ptr;
} MutableSubtree;
```

Apparently this tries to rely on the lowest bit of a pointer to always be 0 in combination with the byte order being little-endian. This fails on OpenBSD/macppc where you can get pointers like the above `0xc517fd08` where the `0xc5` byte overlaps with these flags.

I see two ways to fix this:
* Reorder SubtreeInlineData depending on the endian such that `is_inline` always falls into the right place. This can be tricky especially for 32bit.
* Move `is_inline` after the pointer or even outside the union, but that would make it bigger than 8 bytes.

If it's decided which solution to take, I can send a pr.

thestr4ng3r:
Fixes #1466

This orders the members of `SubtreeInlineData` such that the `is_inline` bit falls exactly into the least significant bit of the pointer in `Subtree`, while keeping a maximum size of 8 bytes. There are three cases that need different layouts:

* Little-endian, no matter which pointer size: as before, the bit is simply the first member in the struct
* Big-endian, 32-bit: it has to occupy the highest bit of the 4th byte in the struct
* Big-endian, 64-bit: it has to occupy the highest bit of the 8th byte in the struct

I am not a huge fan of bloating up the definition of the struct so much, but I don't really see another way. I used macros to not repeat all members multiple times.
Also unfortunately there is no standard way to detect the endian in the preprocessor, so I have only added the checks I could actually verify myself and if undetected it falls back to assuming little-endian as before.
The detection for setting `TS_BIG_ENDIAN` and `TS_PTR_SIZE` can be overridden by using `-D` compiler args so any project that embeds tree-sitter can supply this by themselves if they want.

A little program to check the exact layout is here: https://github.com/thestr4ng3r/ts-endian
It zero-initializes a `Subtree`, sets `is_inline` to 1 and then checks if the pointer is exactly 1, e.g.:
```
g5-openbsd:ts-endian$ uname -a
OpenBSD g5-openbsd.fritz.box 7.0 GENERIC#940 macppc
g5-openbsd:ts-endian$ ./ts-endian
sizeof(void *) = 4
TS_PTR_SIZE = 32
TS_BIG_ENDIAN = 1
sizeof(SubtreeInlineData) = 8
sizeof(Subtree) = 8
==> ptr = 0x1
```

I tested this on the following platforms:
* OpenBSD 7.0 macppc (32bit)
* Mac OS X 10.5 ppc 32bit
* Mac OS X 10.5 ppc 64bit
* macOS 12 arm64

As a sidenote, another idea I had was to not use an explicit `is_inline` member, but simply perform the check like `((size_t)ptr) & 1`, since in any case that is what we want to use. Then one would have to just make sure that for any of the above cases, `SubtreeInlineData` would not put any actual payload data at the position of that bit. Including the current `is_inline` there are two unused bits avaliable. But because we need 3 bits and because of other alignment restrictions, this wouldn't work.

Another possibility that could actually work would be something like this:
```c
union {
  uint64_t data;
  void *ptr;
} Subtree;
```
Then one would have to manually pull the members out of the `data` value. But that would require a lot of annoying bit-shifting and anding instead of accessing struct members.

### DIFF
--- a/lib/src/host.h
+++ b/lib/src/host.h
@@ -1,0 +1,21 @@
+
+// Determine endian and pointer size based on known defines.
+// TS_BIG_ENDIAN and TS_PTR_SIZE can be set as -D compiler arguments
+// to override this.
+
+#if !defined(TS_BIG_ENDIAN)
+#if (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) \
+  || (defined( __APPLE_CC__) && (defined(__ppc__) || defined(__ppc64__)))
+#define TS_BIG_ENDIAN 1
+#else
+#define TS_BIG_ENDIAN 0
+#endif
+#endif
+
+#if !defined(TS_PTR_SIZE)
+#if UINTPTR_MAX == 0xFFFFFFFF
+#define TS_PTR_SIZE 32
+#else
+#define TS_PTR_SIZE 64
+#endif
+#endif
--- a/lib/src/subtree.h
+++ b/lib/src/subtree.h
@@ -49,10 +49,9 @@ typedef struct {
 typedef struct {
 #define SUBTREE_3_BYTES \
   uint8_t symbol; \
-  uint8_t padding_bytes; \
-  uint8_t size_bytes; \
+  uint16_t parse_state;
 
-#define SUBTREE_6_BITS \
+#define SUBTREE_BITS \
   bool visible : 1; \
   bool named : 1; \
   bool extra : 1; \
@@ -60,36 +59,39 @@ typedef struct {
   bool is_missing : 1; \
   bool is_keyword : 1;
 
-#define SUBTREE_4_BYTES \
+#define SUBTREE_SIZE \
   uint8_t padding_columns; \
   uint8_t padding_rows : 4; \
   uint8_t lookahead_bytes : 4; \
-  uint16_t parse_state;
+  uint8_t padding_bytes; \
+  uint8_t size_bytes;
 
 #if TS_BIG_ENDIAN
 #if TS_PTR_SIZE == 32
-  SUBTREE_3_BYTES
-  SUBTREE_6_BITS
+  uint16_t parse_state;
+  uint8_t symbol;
+  SUBTREE_BITS
   bool unused : 1;
   bool is_inline : 1;
-  SUBTREE_4_BYTES
+  SUBTREE_SIZE
 #else
-  SUBTREE_4_BYTES
-  SUBTREE_3_BYTES
-  SUBTREE_6_BITS
+  SUBTREE_SIZE
+  uint16_t parse_state;
+  uint8_t symbol;
+  SUBTREE_BITS
   bool unused : 1;
   bool is_inline : 1;
 #endif
 #else
   bool is_inline : 1;
-  SUBTREE_6_BITS
-  SUBTREE_3_BYTES
-  SUBTREE_4_BYTES
+  SUBTREE_BITS
+  uint8_t symbol;
+  uint16_t parse_state;
+  SUBTREE_SIZE
 #endif
 
-#undef SUBTREE_3_BYTES
-#undef SUBTREE_6_BITS
-#undef SUBTREE_4_BYTES
+#undef SUBTREE_BITS
+#undef SUBTREE_SIZE
 } SubtreeInlineData;
 
 // A heap-allocated representation of a subtree.
--- a/lib/src/subtree.h
+++ b/lib/src/subtree.h
@@ -11,6 +11,7 @@ extern "C" {
 #include "./length.h"
 #include "./array.h"
 #include "./error_costs.h"
+#include "./host.h"
 #include "tree_sitter/api.h"
 #include "tree_sitter/parser.h"
 
@@ -39,21 +40,56 @@ typedef struct {
 //
 // This representation is used for small leaf nodes that are not
 // errors, and were not created by an external scanner.
+// The idea behind the layout of this struct is that the `is_inline`
+// bit will fall exactly into the same location as the least significant
+// bit of the pointer in `Subtree` or `MutableSubtree`, respectively.
+// Because of alignment, for any valid pointer this will be 0, giving
+// us the opportunity to make use of this bit to signify whether to use
+// the pointer or the inline struct.
 typedef struct {
-  bool is_inline : 1;
-  bool visible : 1;
-  bool named : 1;
-  bool extra : 1;
-  bool has_changes : 1;
-  bool is_missing : 1;
+#define SUBTREE_3_BYTES \
+  uint8_t symbol; \
+  uint8_t padding_bytes; \
+  uint8_t size_bytes; \
+
+#define SUBTREE_6_BITS \
+  bool visible : 1; \
+  bool named : 1; \
+  bool extra : 1; \
+  bool has_changes : 1; \
+  bool is_missing : 1; \
   bool is_keyword : 1;
-  uint8_t symbol;
-  uint8_t padding_bytes;
-  uint8_t size_bytes;
-  uint8_t padding_columns;
-  uint8_t padding_rows : 4;
-  uint8_t lookahead_bytes : 4;
+
+#define SUBTREE_4_BYTES \
+  uint8_t padding_columns; \
+  uint8_t padding_rows : 4; \
+  uint8_t lookahead_bytes : 4; \
   uint16_t parse_state;
+
+#if TS_BIG_ENDIAN
+#if TS_PTR_SIZE == 32
+  SUBTREE_3_BYTES
+  SUBTREE_6_BITS
+  bool unused : 1;
+  bool is_inline : 1;
+  SUBTREE_4_BYTES
+#else
+  SUBTREE_4_BYTES
+  SUBTREE_3_BYTES
+  SUBTREE_6_BITS
+  bool unused : 1;
+  bool is_inline : 1;
+#endif
+#else
+  bool is_inline : 1;
+  SUBTREE_6_BITS
+  SUBTREE_3_BYTES
+  SUBTREE_4_BYTES
+#endif
+
+#undef SUBTREE_3_BYTES
+#undef SUBTREE_6_BITS
+#undef SUBTREE_4_BYTES
 } SubtreeInlineData;
 
 // A heap-allocated representation of a subtree.


alemuller:
I think implementing the padding with [type attributes](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Common-Type-Attributes.html#Common-Type-Attributes) (in particular, `aligned`) is a more sane approach than with unused struct fields.

By the way, in link above, there are some attributes related to endianess. Can't they be used for this?

Also. in a quick look at the issue, the misaligned memory access (in your arch) seems to caused because:
* `sizeof(SubtreeInlineData)` is 64/8
* `sizeof(SubtreeHeapData *)` is 32/8
* `SubtreeHeapData *` is accessing the wrong half of union (?)

This's weird. Does this also happens with other compilers? With no optimization? Maybe there are some missing casting?

Something else, about the original code. Maybe enumeration could be used instead of bitfields?
```c
// Something like
typedef enum = {
  visible   = (1<<0),
  named     = (1<<1),
  extra     = (1<<2),
  missing   = (1<<3),
  keyword   = (1<<4),
} foo_t __attribute__(packed);
```

thestr4ng3r:
> I think implementing the padding with [type attributes](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Common-Type-Attributes.html#Common-Type-Attributes) (in particular, `aligned`) is a more sane approach than with unused struct fields.

The `aligned` attribute only specifies byte-alignment, but the unused member here only pads by a single bit. Moreover, `is_inline` here must be at the 7th position, which is a prime, so even if there was a bit-alignment attribute for bitfields, it wouldn't help us. We could make `is_missing` take 2 bits though, but I think it would be a bit less explicit than the `unused` member.

> By the way, in link above, there are some attributes related to endianess. Can't they be used for this?

I don't know if `scalar_storage_order` affects bitfields too, but it is only available in gcc >= 6 and not in clang. But in any case, it would only help us to get rid of the `unused` member and use a single ordering of the single bits for all archs. The other members would still need manual reordering.

> Also. in a quick look at the issue, the misaligned memory access (in your arch) seems to caused because:
> 
>     * `sizeof(SubtreeInlineData)` is 64/8
> 
>     * `sizeof(SubtreeHeapData *)` is 32/8
> 
>     * `SubtreeHeapData *` is accessing the wrong half of union (?)
> 
> 
> This's weird. Does this also happens with other compilers? With no optimization? Maybe there are some missing casting?

It is a bit hard to get across in words so I tried to visualize it. `0123456789abcdef` here is a single byte composed of its bits from lsb to hsb. This is the layout that was there before any of my changes:

```
little endian 32bit:

       is_inline
       |
mem: [ 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef ]
ptr:   ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^
       |
       lsb


little endian 32bit:

       is_inline
       |
mem: [ 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef ]
ptr:   ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^
       |
       lsb


big endian 64bit:

                      is_inline
                      |
mem: [ 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef ]
ptr:   ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^
                                                          |
                                                          lsb


big endian 64bit:

                      is_inline
                      |
mem: [ 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef 0123456789abcdef ]
ptr:   ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^
                                                                                                                              |
                                                                                                                              lsb
```

From my knowledge, this is the specified behavior regardless of the compiler. Optimization should also never affect the struct layouts.


> 
> Something else, about the original code. Maybe enumeration could be used instead of bitfields?
> 
> ```c
> // Something like
> typedef enum = {
>   visible   = (1<<0),
>   named     = (1<<1),
>   extra     = (1<<2),
>   missing   = (1<<3),
>   keyword   = (1<<4),
> } foo_t __attribute__(packed);
> ```

Using `uint8_t` instead of the bitfield and setting the bits manually would indeed work. For big-endian, the actual values would still have to be swapped though, so in the end it is a matter of taste I guess.

thestr4ng3r:
Another idea I just played around with a bit is for 32bit big-endian, to not use the specific lsb of the pointer, but just use any of the free bits after the pointer. This can re-use the layout of 64bit big-endian, so no distinction of bitness is necessary:
https://github.com/tree-sitter/tree-sitter/compare/master...thestr4ng3r:big-endian-no32
with adjusted testing: https://github.com/thestr4ng3r/ts-endian/blob/no32/main.c
**BUT**: assigning only a pointer will not assign the inline bit, so one has to always make sure to zero-initialize the entire struct and `NULL_SUBTREE` is broken too. So while this alternative is possible, I think it is very error-prone.

thestr4ng3r:
Rebased after 0.20.1
@maxbrunsfeld Any comments on this? We are currently shipping it as a custom patch: https://github.com/rizinorg/rizin/tree/dev/subprojects/packagefiles/tree-sitter-0.19.5

lib/src/subtree.h
```diff
@@ -39,21 +40,56 @@ typedef struct {
...
-  uint8_t size_bytes;
-  uint8_t padding_columns;
-  uint8_t padding_rows : 4;
-  uint8_t lookahead_bytes : 4;
+
+#define SUBTREE_4_BYTES \
+  uint8_t padding_columns; \
+  uint8_t padding_rows : 4; \
+  uint8_t lookahead_bytes : 4; \
   uint16_t parse_state;
```
maxbrunsfeld, line 67: Some minor naming suggestions:

1. If you switched the `uint16_t parse_state;` with `uint8_t padding_bytes; uint8_t size_bytes;`, then all of the "size"-related fields would be together in `SUBTREE_4_BYTES`, and you could rename that macro to `SUBTREE_SIZE`, which indicates more about its meaning.
2. Then, `SUBTREE_3_BYTES` would only contain two fields, and you could just inline them, instead of having that macro at all.
2. I think `SUBTREE_6_BITS` could be renamed to just `SUBTREE_BITS`.

maxbrunsfeld (commented):
Thanks for fixing this. I left some small suggestions.

lib/src/host.h
```diff
@@ -0,0 +1,21 @@
...
+#if (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) \
+  || (defined( __APPLE_CC__) && (defined(__ppc__) || defined(__ppc64__)))
+#define TS_BIG_ENDIAN 1
+#else
+#define TS_BIG_ENDIAN 0
+#endif
+#endif
+
+#if !defined(TS_PTR_SIZE)
+#if UINTPTR_MAX == 0xFFFFFFFF
```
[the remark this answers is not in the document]
thestr4ng3r: From what I know, `uintptr_t` is part of C99, but an optional type, so I think it would be better to keep the option to set `TS_PTR_SIZE` from outside.

lib/src/subtree.h
thestr4ng3r: The idea behind the names was that each indicates exactly how much space it takes so the layout is directly visible. But your suggestion works too, so I have updated it accordingly.

XVilka:
There seem to be Ruby corpus test failure on AppVeyor, @thestr4ng3r :
```
failures:
failures:
    tests::corpus_test::test_ruby_corpus
test result: FAILED. 219 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 57.88s
error: test failed, to rerun pass '-p tree-sitter-cli --lib'
Command exited with code 101
```

maxbrunsfeld:
I think that failure is unrelated to this change.

thanks so much for this work @thestr4ng3r.

Issue #1452, opened 2021-10-21
Return type for field IDs in Rust bindings could be `Option<NonZeroU16>`


Sakarah:
Currently in the Rust binding API, when a field ID is returned we encode the returning value as an `Option<u16>`, to reflect that when the C function returns zero it is because a children has no field ID, or a field name does not exist.
However in Rust the `Option<u16>` type allows for the value `Some(0)` disabling some optimizations and making the API return type less informative.

I suggest replacing such uses of `Option<u16>` with `Option<NonZeroU16>` which is converted from `u16` completely for free, and still retain the full special case expressiveness.
Probably we also should create a type alias `FieldId = NonZeroU16`, and replace all remaining usage of `u16` as `FieldId` when relevant.

Sadly, this is a breaking change from the current API, but I guess it is worthy while the library is not released in version 1.0.

Tell me what you think about that, and I can make a corresponding PR.

PR #1474, opened 2021-11-07
Rust bindings: Use NonZeroU16 for field id and add type alias

Sakarah:
This PR solves issue #1452.

It also adds two new type alias in the Rust bindings, `KindId` and `FieldId` to have a slightly more expressive API.

Remaining questions:
- [ ] Should type alias be replaced by new-types instead ?
- [ ] Should `id_for_node_kind` return an `Option<KindId>` ? Currently it returns the "end" symbol if the kind does not exist.

The PR can be merged as it is today, and questions can also be answered later, though we might want to answer them now since they are strongly related and might prevent further breaking changes.

This is a breaking change for Rust downstream crates but it only affects codes that do more than comparisons on node kind identifiers and field identifiers. It is not a braking change for the C library.


dundargoc:
> This PR solves issue https://github.com/tree-sitter/tree-sitter/issues/1452.

That issue is already marked as solved though. Is this PR still needed?

PR #1483, opened 2021-11-12
Don't use zero maxlen for snprintf in ts_subtree__write_to_string

furunkel:
It seems that (some implementations of?) `snprintf` returns -1 and sets `errno` to `EINVAL` if a `maxlen` of zero is passed. This causes the count to underflow and `ts_subtree__write_to_string` returns a gigantic size which the succeeding malloc will refuse to allocate.


maxbrunsfeld:
Thanks!

PR #1485, opened 2021-11-17
Add doc for cursor functions for rust binding

### LINKED ISSUE #567: Make it clear in cursor-related doc that they are scoped to the starting node
There seems to be a bug with `ts_tree_cursor_goto_parent` when the cursor is initialized on a non-root node.

I got it in the Emacs Lisp binding, but can reproduce it with the NodeJS binding (0.16.0):

```javascript
const Parser = require('tree-sitter');
const Rust = require('tree-sitter-rust');

parser = new Parser();
parser.setLanguage(Rust);

tree = parser.parse("fn foo() {}");

> c0 = tree.rootNode.walk(); c0.gotoFirstChild();
true
>
> c1 = c0.currentNode.walk(); c0.currentNode === c1.currentNode;
true
>
> c0.gotoParent();
true
>
> c1.gotoParent();
false
>
> c0.currentNode.type;
'source_file'
>
> c1.currentNode.type;
'function_item'
>
```

Enter-tainer:
See https://github.com/tree-sitter/tree-sitter/issues/567#issuecomment-595348892.

Make it clear that the cursor only "walks" within the node that it started from, and it can't go outside of it.

Closes #567


Enter-tainer:
I currently added doc for the rust binding, not pretty sure if other functions (like `goto_first_child_for_byte`) or other bindings (nodejs, web, etc) have the same issue.

Enter-tainer:
Seems that CI has failed. But what I did is just adding some document. Any suggestions? How to fix the CI?

Enter-tainer:
anyone approve ci run?

PR #1487, opened 2021-11-17
Add link to swift bindings package

mattmassicotte:
Hello!

I'm a long-time user of tree-sitter, but just finally got around to open sourcing the bindings I use. They are mostly wrappers, but do also include a few utilities to make interoperating with Swift and its string/data constructs easier.

I thought it would be nice to include it in the docs.


maxbrunsfeld:
This looks great. Thanks!

PR #1532, opened 2021-12-11
Added bindings for Java and Kotlin to `index.md`

### LINKED ISSUE #1531: Bindings for Java/Kotlin
Big fan of this project! We are currently trying to use tree-sitter instead of eclipse CDT in our Kotlin/Java-based project. For this purpose, I have created bindings for Kotlin (currently only for the JVM, but potentially also for native) in a separate repository
 (https://github.com/oxisto/kotlintree). It is basically using JNA to access the native functions and using the JNAs extensibility to create classes that have a more "Kotlin"-feel to it rather than just using the pure C api. 

Is there any interest in providing those bindings as an "official" part of the project? If yes, I would try to prepare a PR for it, otherwise, it will probably life as an independent project. Let me know what you think.

oxisto:
Closes #1531


maxbrunsfeld:
Thanks!

Issue #1535, opened 2021-12-14
Allow swapping malloc implementation at runtime


casouri:
Not being able to swap malloc implementation at runtime is blocking us from merging tree-sitter support into Emacs. It is quite reasonable to expect the system to provide tree-sitter library and matching language libraries. On the other hand, embedding tree-sitter into Emacs is both a maintenance burden and a potential risk: Emacs' tree-sitter might not match the language libraries provided by the system. So we really want to link tree-sitter dynamically, and we need to be able to change malloc at runtime to do that. Could you consider adding this feature to tree-sitter? TIA.

maxbrunsfeld:
I think this functionality can be built on *top* of Tree-sitter's statically-configurable malloc implementation. You could define your own dynamically-configurable malloc function that uses a mutable function pointer, and tell Tree-sitter to use that function by defining the [`ts_malloc`/`ts_free`](https://github.com/tree-sitter/tree-sitter/blob/836d753c73d41659a5c7f7958807dde4fabb2322/lib/src/alloc.h#L29-L40) macros.

casouri:
We want to use the tree-sitter library provided by the system in the form of a dynamic library, instead of pulling it into our code base and modify the `ts_malloc` and `ts_free` macros, because that spares us from maintaining tree-sitter in-tree. And it's no different from other libraries we use, eg, librsvg, libpng, etc. They all provide means to change malloc at runtime. I think it only makes sense for tree-sitter to also provide that feature if it wants to be used as a dynamic library.

Of course, if there is a way to change malloc implementation without modifying tree-sitter's source in its current form, I'm very interested to know.

alemuller:
Had this issue as well, but thought was too niche to open an issue. The standard way to change the allocator at runtime is with the  `LD_PRELOAD` envvar (see [mimalloc](https://github.com/microsoft/mimalloc) or any allocator doc).

Eli-Zaretskii:
The `LD_PRELOAD` trick only works on Posix platforms.  On MS-Windows, something like this cannot work at all, because when a library, such as tree-sitter, is compiled and linked into a DLL, it records the `malloc` functions the linker finds at that time, so it will use the Microsoft original runtime even if the program that calls tree-sitter overrides the `malloc` implementation by different functions.

[...]

PR #1544, opened 2021-12-17
Allow to change the allocator dynamically

mkvoya:
This is a draft implementation for #1535. Two things need mentioning:

1. Each malloc/free call will become a function call through the function pointer, which might make the program slightly (perhaps negligible) slower.
2. The clients have two approaches to use alternative allocators:
    1) define the macros at compilation time;
    2) reassign the global allocation function pointers (`ts_current_malloc`, `ts_current_calloc`, `ts_current_realloc`, and `ts_current_free`) by invoking `ts_set_allocator` at runtime. The function can be invoked more than once, as long as the applications take care of the memory allocated by the old allocator.

### DIFF
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -134,13 +134,6 @@ typedef enum {
   TSQueryErrorLanguage,
 } TSQueryError;
 
-typedef struct {
-  void *(*malloc)(size_t);
-  void *(*calloc)(size_t, size_t);
-  void *(*realloc)(void *, size_t);
-  void (*free)(void *);
-} TSAllocator;
-
 /********************/
 /* Section - Parser */
 /********************/
@@ -148,15 +141,18 @@ typedef struct {
 /**
  * Switch to a new allocator.
  *
- * Returns true iff the switch successes.
+ * This function can be invoked more than once. However, the application needs
+ * to pay attention to the memory allocated by the old allocator but might be
+ * freed by the new one.
  *
- * This function can only be invoked *once*, before tree-sitter has allocated
- * any memory via malloc/calloc/realloc. Otherwise, memory bugs could occur
- * since some memory is allocated by the old allocator, but freed by the new
- * one.
- *
+ * Specifically, the application either,
+ *  1. ensures all parsers and trees are freed before calling it;
+ *  2. provides an allocator that shares its state with the old allocator.
  */
-bool ts_set_allocator(TSAllocator *new_alloc);
+void ts_set_allocator(void *(*new_malloc)(size_t),
+		      void *(*new_calloc)(size_t, size_t),
+		      void *(*new_realloc)(void *, size_t),
+		      void (*new_free)(void *));
 
 /**
  * Create a new parser.
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -145,12 +145,6 @@ typedef struct {
 /* Section - Parser */
 /********************/
 
-
-/**
- * The allocator.
- */
-extern TSAllocator *ts_allocator;
-
 /**
  * Switch to a new allocator.
  *
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -148,10 +148,15 @@ typedef struct {
 /**
  * Switch to a new allocator.
  *
- * Returns the old allocator. The caller needs to take care of any allocated
- * memory managed by the old allocator.
+ * Returns true iff the switch successes.
+ *
+ * This function can only be invoked *once*, before tree-sitter has allocated
+ * any memory via malloc/calloc/realloc. Otherwise, memory bugs could occur
+ * since some memory is allocated by the old allocator, but freed by the new
+ * one.
+ *
  */
-TSAllocator *ts_set_allocator(TSAllocator *new_alloc);
+bool ts_set_allocator(TSAllocator *new_alloc);
 
 /**
  * Create a new parser.
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -134,9 +134,30 @@ typedef enum {
   TSQueryErrorLanguage,
 } TSQueryError;
 
+typedef struct {
+  void *(*malloc)(size_t);
+  void *(*calloc)(size_t, size_t);
+  void *(*realloc)(void *, size_t);
+  void (*free)(void *);
+} TSAllocator;
+
 /********************/
 /* Section - Parser */
 /********************/
+
+
+/**
+ * The allocator.
+ */
+extern TSAllocator *ts_allocator;
+
+/**
+ * Switch to a new allocator.
+ *
+ * Returns the old allocator. The caller needs to take care of any allocated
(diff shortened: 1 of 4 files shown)


casouri:
You didn't seem to modify api.h to include `ts_allocator`. Am I missing something?

mkvoya:
> You didn't seem to modify api.h to include `ts_allocator`. Am I missing something?

My bad. I have updated the PR by exposing `struct TSAllocator` and `TSAllocator *ts_set_allocator(TSAllocator *)` in `api.h`.

cireu:
Should we use mutex to protect it and ensure thread-safety?

mkvoya:
> Should we use mutex to protect it and ensure thread-safety?

I don't think we need to provide thread-safety in the `ts_set_allocator` function.
To provide thread-safety in `ts_set_allocator `, we also need to protect all paths that invoke `ts_allocator->XXX`, which will involve mutexes in all the allocation/deallocation paths.

Thus, I prefer to leave the interface thread-unsafe and let applications to guarantee the thread-safety.

ubolonton:
This implementation currently uses a global allocator that can be switched out *multiple times*, after it *has allocated some memory*. That will result in memory bugs, as tree-sitter will use the new allocator's `free` on memory allocated by the old allocator's `malloc`.

Setting a global allocator should be allowed only *once* (which can be enforced by `ts_set_allocator`), *before it is used to allocate any memory* (which should be documented, not enforced, for better performance).

If we really want to be able to change the allocator multiple times, we would need to modify other code to pass `TSAllocator` structs around, instead of using `ts_{malloc,calloc,realloc,free}` directly. I don't think we want that functionality, though.

mkvoya:
> This implementation currently uses a global allocator that can be switched out _multiple times_, after it _has allocated some memory_. That will result in memory bugs, as tree-sitter will use the new allocator's `free` on memory allocated by the old allocator's `malloc`.
> 

Yes. The current code leave the problem to the application. Originally I expected the applications could solve the issue since the old allocator was returned to the application. Now, I think you are right. The applications cannot do this well.

> Setting a global allocator should be allowed only _once_ (which can be enforced by `ts_set_allocator`), _before it is used to allocate any memory_ (which should be documented, not enforced, for better performance).
> 

Will do.

> If we really want to be able to change the allocator multiple times, we would need to modify other code to pass `TSAllocator` structs around, instead of using `ts_{malloc,calloc,realloc,free}` directly. I don't think we want that functionality, though.

Agreed.

lib/include/tree_sitter/api.h
```diff
@@ -134,10 +134,30 @@ typedef enum {
   TSQueryErrorLanguage,
 } TSQueryError;
 
+typedef struct {
+  void *(*malloc)(size_t);
+  void *(*calloc)(size_t, size_t);
+  void *(*realloc)(void *, size_t);
+  void (*free)(void *);
+} TSAllocator;
```
maxbrunsfeld, line 142: I'd prefer to remove this struct definition from the public API, and just pass the function pointers as 4 separate parameters to `ts_set_allocator`.

Then, I would just store each function in its own `static` variable, declared in `alloc.c`. That will remove a small amount of indirection when invoking the functions.

lib/include/tree_sitter/api.h
```diff
@@ -134,10 +134,30 @@ typedef enum {
...
 
+/**
+ * Switch to a new allocator.
+ *
+ * Returns true iff the switch successes.
+ *
+ * This function can only be invoked *once*, before tree-sitter has allocated
+ * any memory via malloc/calloc/realloc. Otherwise, memory bugs could occur
+ * since some memory is allocated by the old allocator, but freed by the new
+ * one.
```
maxbrunsfeld, line 156: I actually don't think we need to impose this "only once" restriction in the library. Applications can maintain this invariant themselves. There are valid use cases where the user may want to call this function more than once, and it should work provided they either:
1. Ensure all trees are freed before calling it
2. Provide an allocator that shares some state with the old allocator.

For instance, once we add this feature, I'd like to refactor the Tree-sitter test suite to use `ts_set_allocator` to implement the "allocation tracking" which is currently done using a compile time conditional. That should work fine, since the tracking allocator just *wraps* the normal allocator.

So could you just remove the logic for enforcing that this is only called once, and make this return `void`?

We could still include some kind of warning in the doc comment, but I also think it's reasonable to expect users to decide this for themselves.

lib/src/alloc.c
```diff
@@ -0,0 +1,84 @@
...
+TSAllocator *ts_allocator = &ts_tracking_allocator;
+
+#else
+
+#include <stdlib.h>
+
+bool ts_toggle_allocation_recording(bool value) {
+  (void)value;
+  return false;
+}
```
maxbrunsfeld, line 27: Like I mentioned above, I think we'll want to simplify this system of tracking allocations once we add this feature, but I don't think you need to do it as part of this PR (unless you want to).

maxbrunsfeld (commented):
Thanks for the PR! I left a couple of suggestions.

lib/include/tree_sitter/api.h
mkvoya: If the variable is defined as `static`, it cannot be invoked in other source files. Perhaps it should be a normal global variable?

lib/include/tree_sitter/api.h
mkvoya: > Applications can maintain this invariant themselves. There are valid use cases where applications may want to call this function more than once, and it should work, as long as the application either:
> 
> 1. Ensures all parsers and trees are freed before calling it
> 2. Provides an allocator that shares its state with the old allocator.
> 

Can I use these comments (with slight modifications) as the doc comment before the function?


> So could you just remove the logic for enforcing that this is only called once, and make this return `void`?

Do we need to provide an interface to return current allocator functions?

lib/src/alloc.c
mkvoya: I am not familiar with existing allocation tracking, so perhaps we can leave the refactoring in later PR.

lib/include/tree_sitter/api.h
mkvoya: > > Applications can maintain this invariant themselves. There are valid use cases where applications may want to call this function more than once, and it should work, as long as the application either:
> > 
> > 1. Ensures all parsers and trees are freed before calling it
> > 2. Provides an allocator that shares its state with the old allocator.
> 
> Can I use these comments (with slight modifications) as the doc comment before the function?
> 

I added the two cases in the doc comment before the function. If you think I need to rephrase in my own words, please let me know.

lib/include/tree_sitter/api.h
maxbrunsfeld: Oh, I was thinking that it'd be internal to `alloc.c`, and exposed via some functions in `alloc.h`, but that's not necessary. What you did with `ts_current_alloc` looks great. Thanks.

maxbrunsfeld (approved):
This looks great.

maxbrunsfeld:
Thanks @mkvoya for adding this, and to everyone who commented and helped design it.

cireu:
@maxbrunsfeld  Close #1535 ?

PR #1547, opened 2021-12-18
test tags queries in 'tree-sitter test'

the-mikedavis:
Adding `tags.scm` to the elixir grammar (see [tree-sitter-elixir#18](https://github.com/elixir-lang/tree-sitter-elixir/pull/18)), I thought it would be nice to be able to test expected `@reference.*` and `@definition.*` captures against fixtures the same way you can test highlighting captures.

This PR changes `tree-sitter test` to check for a `test/tags` directory similar to `test/highlight` and it uses the same format as highlighting. For example, testing the python `tags.scm`:

```py
def abc(d):
	# <- definition.function
	e = fgh(d)
	#    ^ reference.call
	return d(e)
	#      ^ reference.call
```

Except it uses the machinery in the `tree_sitter_tags` crate rather than `tree_sitter_highlight`. A run of `tree-sitter test` for a grammar with a `test/tags` directory looks like so when passing:

```
tags:
  ✓ fixture.exs (15 assertions)
```

And like so when failing:

```
tags:
  ✗ fixture.exs
    Failure - row: 8, column: 7, expected tag: 'definition.module', actual tag: 'definition.function'
```

Where this `tags:` section comes immediately after the highlighting tests, if they exist.

There's a lot of duplicated code now between the highlight testing and tags testing that should probably be refactored. I'm pretty new to rust though so I'm not too sure how to approach that.

What do you think?

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -8,6 +8,7 @@ pub mod query_testing;
 pub mod tags;
 pub mod test;
 pub mod test_highlight;
+pub mod test_tags;
 pub mod util;
 pub mod wasm;
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -4,7 +4,8 @@ use glob::glob;
 use std::path::Path;
 use std::{env, fs, u64};
 use tree_sitter_cli::{
-    generate, highlight, logger, parse, playground, query, tags, test, test_highlight, util, wasm,
+    generate, highlight, logger, parse, playground, query, tags, test, test_highlight, test_tags,
+    util, wasm,
 };
 use tree_sitter_config::Config;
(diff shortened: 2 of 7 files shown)


maxbrunsfeld:
This is a good idea; thanks for implementing it. There's a bit of duplication as you said, but it doesn't bother me too much.

maxbrunsfeld (approved):

the-mikedavis:
Sorry to ping you on this @maxbrunsfeld but is there anything I can do to push this PR along? I promise to add some tests to the code-search enabled repos if merged 😄

maxbrunsfeld:
Thanks @the-mikedavis.

razzeee:
Might also be nice to mention this in the docs?

PR #1548, opened 2021-12-18
remove unused iterate_assertions function in test_highlight

the-mikedavis:
While working on #1547 I noted that this block is duplicated in `iterate_assertions` and `test_highlight` and the return value is discarded. Maybe it makes more sense to keep this `iterate_assertions` function and remove the block in the `test_highlight` function instead?


dundargoc:
Still relevant? If so, mind rebasing?

amaanq:
No I removed the duplicate code below it a while ago, oops

PR #1549, opened 2021-12-19
Generate MJS file

Symbitic:
This generates an ES6 Module for the WASM binding in addition to the regular JavaScript file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,7 @@ jobs:
           files: |
             tree-sitter-*.gz
             lib/binding_web/tree-sitter.js
+            lib/binding_web/tree-sitter.mjs
             lib/binding_web/tree-sitter.wasm
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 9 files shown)


maxbrunsfeld (commented):
I don't understand the reasoning behind some of these changes. A few questions:
1. Why does there need to be a separate JavaScript file, as opposed to just pointing the `module` field in the package.json at the existing `tree-sitter.js` file?
2. If there *does* need to be a separate file on disk, is the `.msj` file different in its contents, or can we just make a copy of the `.js` file with a different name?
3. Why does the `playground` subcommand need to serve both JavaScript files to the browser?

Symbitic:
1. The `tree-sitter.js` file doesn't use the ESM module format (`import`/`export`). Emscripten only creates that when creating output files with the `.mjs` extension. The `module` field cannot point to a non-ESM file.
2. As above, `.mjs` is different. It uses `import`/`export`. Emscripten by default creates files that can be used like `<script src="tree-sitter.js>` but with ESM they can be used as `<script type="module" src="tree-sitter.mjs>`. Deno supports ESM, but not the default Emscripten output format.
3. It's not strictly necessary, but a common use case is for people is to include Tree-sitter from the Playground URL. This allows the same thing.

PR #1571, opened 2022-01-03
Add realloc to wasm exports

414owen:
`tree-sitter-haskell` uses `realloc` directly, and it seems like a small and crucial enough function to include in wasm builds.


maxbrunsfeld:
Yeah, it's a pretty common function, and the library already uses it internally.

wenkokke:
Thanks, @maxbrunsfeld! Would a minor release with this patch be possible?

PR #1573, opened 2022-01-04
cli: deduplicate query results in 'tree-sitter query --test'

the-mikedavis:
I noticed that `tree-sitter query --test` will give failures when the queries return multiple results, such as when you test highlight queries outside of `tree-sitter test` (see https://github.com/elixir-lang/tree-sitter-elixir/pull/18#issuecomment-996851772)

Somewhat minimally I have a grammar [tree-sitter-git-config](https://github.com/the-mikedavis/tree-sitter-git-config) and a foo.gitconfig file like so:

```gitconfig
[core]
# ^ tag
    path = /path/to/foo.inc ; include by absolute path
    ;       ^ string.special.path
    foo = bar
```

with [these highlight queries](https://github.com/the-mikedavis/tree-sitter-git-config/blob/5a12b9ebe685c9c546e244ff01bbe24d9f16d8dd/queries/highlights.scm). Running `tree-sitter query --test queries/highlights.scm foo.gitconfig` gives an error like so:

```
foo.gitconfig
  pattern: 8
    capture: 7 - punctuation.bracket, start: (0, 0), end: (0, 1), text: `[`
  pattern: 2
    capture: 1 - tag, start: (0, 1), end: (0, 5), text: `core`
  pattern: 8
    capture: 7 - punctuation.bracket, start: (0, 5), end: (0, 6), text: `]`
  pattern: 10
    capture: comment, start: (1, 0), end: (2, 0)
  pattern: 3
    capture: 2 - property, start: (2, 1), end: (2, 5), text: `path`
  pattern: 9
    capture: 8 - punctuation.delimiter, start: (2, 6), end: (2, 7), text: `=`
  pattern: 6
    capture: 5 - string.special.path, start: (2, 7), end: (2, 25), text: ` /path/to/foo.inc `
  pattern: 7
    capture: 6 - string, start: (2, 7), end: (2, 25), text: ` /path/to/foo.inc `
  pattern: 10
    capture: comment, start: (2, 25), end: (3, 0)
  pattern: 10
    capture: comment, start: (3, 4), end: (4, 0)
  pattern: 3
    capture: 2 - property, start: (4, 4), end: (4, 7), text: `foo`
  pattern: 9
    capture: 8 - punctuation.delimiter, start: (4, 8), end: (4, 9), text: `=`
  pattern: 7
    capture: 6 - string, start: (4, 9), end: (4, 13), text: ` bar`
Assertion failed: at (2, 7), found string.special.path, expected string
```

Because the query results overlap.

I think it makes sense to deduplicate these query results when testing to bring the behavior of `tree-sitter query --test` in line with `tree-sitter test` for highlights (and tags as in #1547). But maybe this is the expected behavior? What do you think?

### DIFF
--- a/cli/src/query.rs
+++ b/cli/src/query.rs
@@ -101,6 +101,7 @@ pub fn query_files_at_paths(
             )?;
         }
         if should_test {
+            results.dedup_by(|a, b| a.start == b.start && a.end == b.end);
             query_testing::assert_expected_captures(results, path, &mut parser, language)?
         }
     }


maxbrunsfeld:
I'm actually not totally clear on how how `query --test` should behave in this regard. @the-mikedavis With your change, it will take the *first* match for a given position, right?

@patrickt I think you added the `query --test` command. I can't remember - is the command still being used by the semantic code team at GitHub? If so, does it matter what it does when there are multiple matches containing a range?

the-mikedavis:
Yep, captures after the first one will be discarded for any highlights that exactly overlap. I think this fits well with how highlights and tags work, but there's the chance that you have multiple captures you intend on testing and those wouldn't work with this change.

I think the current behavior is that if you have two overlapping captures, you cannot write a test case that will pass because the matches are considered against the expectation, so you would need to expect both captures in the same expectation, which the syntax doesn't support. (I may be remembering wrong about this, let me test... :detective:)

the-mikedavis:
With the elixir grammar:

```
$ cat example.scm
((identifier) @constant.builtin
 (#match? @constant.builtin "^__(MODULE|ENV)__$"))

(identifier) @variable
$ cat example.exs
__MODULE__
# ^ constant.builtin
# ^ variable
$ tree-sitter query --captures example.scm example.exs --test
example.exs
    pattern:  0, capture: 0 - constant.builtin, start: (0, 0), end: (0, 10), text: `__MODULE__`
    pattern:  1, capture: 1 - variable, start: (0, 0), end: (0, 10), text: `__MODULE__`
Assertion failed: at (0, 0), found constant.builtin, expected variable
```

Even if you write tests for both matching queries, the assertion will fail.

I think this is because [`query_testing::assert_expected_captures`](https://github.com/tree-sitter/tree-sitter/blob/bf210f0c9ec7931c1a5f639461495db240aac149/cli/src/query_testing.rs#L126-L149) is looking through the actual captures and `.find`ing the first expectation comment that matches the range. Maybe it makes sense to refactor that function instead so it finds all matching expectation comments and checks if any match the actual capture?

amaanq:
This might be incorrect/outdated now given that the precedence for queries is now last-wins instead of first wins?

amaanq:
Closing as outdated probably...

the-mikedavis:
Ah sorry for the radio silence on this - it got lost in my notifications. I'm pretty sure you're correct that this is out of date with the precedence change and will need some work. I haven't tried or needed this workflow in a long time so I will leave updating this to anyone that's interested in the branch

amaanq:
no worries, thanks for replying!

PR #1578, opened 2022-01-10
Improve `ts_query_new` performance using pointer indirection

alex-pinkus:
Profiling the `ts_query__analyze_patterns` function shows that it spends a lot of time copying items in its various state sets. These state sets are kept sorted, and the items are fairly large, so any time that we insert new entries near the front of the array, a lot of large calls to memcpy must occur.

While experimenting with more drastic ways to reduce this, I noticed an easy win along the way: by dynamically allocating each `AnalysisState`, we can reduce the size of the actual _array_ since the items become smaller (68 bytes vs 8 bytes). This gave me a 15-20% performance improvement on my [`experimental-tree-sitter-swift`](https://github.com/alex-pinkus/experimental-tree-sitter-swift) benchmarks. According to the benchmarks in this repository, performance is mostly flat, with the added pointer-chasing _not usually_ causing a problem (although there are a few very quick tests that got 1ms slower).

In addition to the pointer-chasing, one downside of the additional dynamic allocation is that we must manage the memory and be more careful to avoid sharing. To enforce proper usage, I modified `AnalysisStateSet` to _wrap_ an array, rather than just being a typedef for an array. The `AnalysisStateSet` always owns its contents, and creates a clone of any items that are requested of it, and the wrappers enforce that and call it out.

### Benchmarks on my own grammar
Here is the output of `cargo bench` on `experimental-tree-sitter-swift`, where I saw the 20% improvement:
_Before:_
```
Benchmarking highlights/ts_query_new: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 124.6s, or reduce sample count to 10.
highlights/ts_query_new time:   [1.2444 s 1.2451 s 1.2459 s]                                       
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```
_After:_
```
Benchmarking highlights/ts_query_new: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 102.9s, or reduce sample count to 10.
highlights/ts_query_new time:   [1.0337 s 1.0350 s 1.0363 s]                                       
                        change: [-16.995% -16.878% -16.755%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high mild
```

Comparing the PDFs of the two, we see that even the outlier data points with dynamic allocation are faster than without.

_Before_:
current-master-branch
_After_:
with-analysis-state-set-ptrs

To understand the reason this changed so much, compare the flamegraph visualizations for each:
_Before_:
flamegraph-current-master-branch

_After_:
flamegraph-with-analysis-state-ptrs

You can see that the current master branch spends a solid chunk of its time on `array_splice` and the corresponding `memmove`, and that once the data is hidden behind the smaller pointers, that time almost entirely goes away.

### Benchmarks from this repository
To get these numbers, I ran a modified version of the `tree-sitter` benchmarks that skips non-query-related tests and collects measurements for queries. I then upped the repetition count to 500 to get more granular data. Most of the performance numbers are identical; the differences are:
```
# < is the current master branch, > is the performance I see with these changes
$ diff current-master-branch.txt with-analysis-state-set-ptrs.txt
40c40
<     tags.scm                                      	time 5 ms	speed 324 bytes/ms
---
>     tags.scm                                      	time 6 ms	speed 278 bytes/ms
65,67c65,67
<     locals.scm                                    	time 32 ms	speed 27 bytes/ms
<     highlights.scm                                	time 54 ms	speed 43 bytes/ms
<     tags.scm                                      	time 106 ms	speed 10 bytes/ms
---
>     locals.scm                                    	time 40 ms	speed 21 bytes/ms
>     highlights.scm                                	time 49 ms	speed 47 bytes/ms
>     tags.scm                                      	time 101 ms	speed 10 bytes/ms
81c81
<   Average Speed (normal): 234 bytes/ms
---
>   Average Speed (normal): 232 bytes/ms
```

## Conclusion
**I'd like to merge some form of this change - it seems to be a solid improvement to me**. I'd appreciate guidance on additional validation. All unit tests still pass, of course, but what else should I be doing to make sure this change didn't break anything? @maxbrunsfeld  mentioned [on twitter](https://twitter.com/maxbrunsfeld/status/1479539497125617665) that there are other languages this is slow for - what other grammars should I benchmark to see if this yields improvements (and to validate that it doesn't cause regressions)?
And while this was some obvious low-hanging fruit, I think there are more drastic changes that can do even better. I'm curious for others' thoughts on these **future improvements**:
1. **Adding some fancier data structures so that the query analysis logic doesn't have to maintain these sorted lists.** If this were Rust, of course, we could use a `BTreeSet` and be done with it. Is there appetite for some additional data structures defined as macros and implemented locally? Have you previously tried that and found that the overhead isn't worth the benefit?
2. **Running this logic at grammar compile time.** It seems like a lot of repositories have their queries checked into the same repository, and that a lot of this sorting and finding paths is reasonable as a compilation step rather than a runtime step. Perhaps this info could be embedded in `parser.c` with some checksum, and on trying to load a query with a compatible checksum, we could speed up the load time by using the preloaded data structures?

Of course, I wouldn't want to block this PR on any of those future improvements - but thought I might bring them up for discussion.

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -225,7 +225,9 @@ typedef struct {
   uint16_t step_index;
 } AnalysisState;
 
-typedef Array(AnalysisState) AnalysisStateSet;
+typedef Array(AnalysisState *) AnalysisStateSet;
+
+typedef Array(AnalysisState *) AnalysisStatePool;
 
 /*
  * AnalysisSubgraph - A subset of the states in the parse table that are used
@@ -627,33 +629,33 @@ static unsigned analysis_state__recursion_depth(const AnalysisState *self) {
 }
 
 static inline int analysis_state__compare_position(
-  const AnalysisState *self,
-  const AnalysisState *other
+  AnalysisState *const *self,
+  AnalysisState *const *other
 ) {
-  for (unsigned i = 0; i < self->depth; i++) {
-    if (i >= other->depth) return -1;
-    if (self->stack[i].child_index < other->stack[i].child_index) return -1;
-    if (self->stack[i].child_index > other->stack[i].child_index) return 1;
+  for (unsigned i = 0; i < (*self)->depth; i++) {
+    if (i >= (*other)->depth) return -1;
+    if ((*self)->stack[i].child_index < (*other)->stack[i].child_index) return -1;
+    if ((*self)->stack[i].child_index > (*other)->stack[i].child_index) return 1;
   }
-  if (self->depth < other->depth) return 1;
-  if (self->step_index < other->step_index) return -1;
-  if (self->step_index > other->step_index) return 1;
+  if ((*self)->depth < (*other)->depth) return 1;
+  if ((*self)->step_index < (*other)->step_index) return -1;
+  if ((*self)->step_index > (*other)->step_index) return 1;
   return 0;
 }
 
 static inline int analysis_state__compare(
-  const AnalysisState *self,
-  const AnalysisState *other
+  AnalysisState *const *self,
+  AnalysisState *const *other
 ) {
   int result = analysis_state__compare_position(self, other);
   if (result != 0) return result;
-  for (unsigned i = 0; i < self->depth; i++) {
-    if (self->stack[i].parent_symbol < other->stack[i].parent_symbol) return -1;
-    if (self->stack[i].parent_symbol > other->stack[i].parent_symbol) return 1;
-    if (self->stack[i].parse_state < other->stack[i].parse_state) return -1;
-    if (self->stack[i].parse_state > other->stack[i].parse_state) return 1;
-    if (self->stack[i].field_id < other->stack[i].field_id) return -1;
-    if (self->stack[i].field_id > other->stack[i].field_id) return 1;
+  for (unsigned i = 0; i < (*self)->depth; i++) {
+    if ((*self)->stack[i].parent_symbol < (*other)->stack[i].parent_symbol) return -1;
+    if ((*self)->stack[i].parent_symbol > (*other)->stack[i].parent_symbol) return 1;
+    if ((*self)->stack[i].parse_state < (*other)->stack[i].parse_state) return -1;
+    if ((*self)->stack[i].parse_state > (*other)->stack[i].parse_state) return 1;
+    if ((*self)->stack[i].field_id < (*other)->stack[i].field_id) return -1;
+    if ((*self)->stack[i].field_id > (*other)->stack[i].field_id) return 1;
   }
   return 0;
 }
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Wow, awesome work @alex-pinkus, and thanks for the wonderful writeup! This looks a big improvement; I'll give a bit more careful review tomorrow, and will also be glad to discuss your suggested follow-up directions.

Off the top of my head, another language that exhibited slowdowns during query analysis was the [elixir grammar](https://github.com/elixir-lang/tree-sitter-elixir), as reported [here](https://github.com/nvim-treesitter/nvim-treesitter/issues/1957). I don't remember the details right now, but I think that when I was last reproducing the slowness for Elixir, I used the following query file and Elixir source file:

```clj
; query file
(call target: (identifier) @function)
```

```ex
# elixir file
quote do
  A.B.c()
end
```

Aside from that, I'd just want to make sure it doesn't slow down query construction for other languages, such as [the ones](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/tests/corpus_test.rs) that Tree-sitter uses as part of its test suite.

There's a benchmark in this repo that profiles query construction. Maybe we should add your Swift grammar and the official Elixir grammar to the set of grammars that Tree-sitter uses in its own CI and [benchmark](https://github.com/tree-sitter/tree-sitter/blob/master/cli/benches/benchmark.rs), which also exercises query construction.

lib/src/query.c
```diff
@@ -669,6 +671,96 @@ static inline bool analysis_state__has_supertype(AnalysisState *self, TSSymbol s
...
+//
+// IMPORTANT: The caller MUST ENSURE that this item is larger (by the comparison function
+// `analysis_state__compare`) than largest item already in this set. If items are inserted in the
+// wrong order, the set will not function properly for future use.
+//
+// The caller retains ownership of the passed-in memory. However, the clone that is created by this
+// function will be managed by the state set.
+static inline void analysis_state_set__push_by_clone(
+  AnalysisStateSet *self,
+  AnalysisState *owned_item
```
alex-pinkus, line 741: This should be named `borrowed_item`, not `owned_item`

alex-pinkus:
It's a good thing you pointed me at `tree-sitter-elixir`—a quick microbenchmark had it performing _worse_ with this initial PR than without. Luckily, when I looked at the code with fresh eyes I realized that it would be very easy to bring in some basic object pooling for those `AnalysisState` items. With object pooling in place, I'm seeing a 15% reduction now for `tree-sitter-elixir` (1.475 secs with those two changes vs 1.738 secs without either) and a _28%_ for `experimental-tree-sitter-swift`!

New commit incoming that adds object pooling. Let me know if you need the commits squashed for hygiene.

maxbrunsfeld (commented):
What do you think about sharing one object pool across all three `AnalysisState` instances (`states`, `next_states`, and `deeper_states`)? So instead of storing the pool as a *field* on `AnalysisStateSet`, the pool would be a separate local variable, and we'd pass it as an additional parameter to functions that allocate and free states, like `analysis_state_set__clear`, `analysis_state_set__push_by_clone`, and `analysis_state_set__insert_sorted_by_clone`.

It's pretty minor, but I think it would let us allocate slightly fewer `AnalysisState` objects (and obviously two fewer arrays). And also aesthetically, the `AnalysisStateSet` would go back to being a single array, so we could return to using fields like `size` directly, just for brevity, and eliminate `analysis_state_set__get` and `analysis_state_set__size`. Let me know if you think that's reasonable or if I'm missing something there.

alex-pinkus:
Sure, we can share one object pool. My thinking was that even without the pool, the `AnalysisStateSet` would still be a struct. By having all usage go through the helper functions that enforce ownership invariants, we avoid getting into tricky states where two sets contain the same item and only one of them should free it. Accessing `states` directly and doing `array_*` operations on them becomes a code smell. That said, it’s your codebase — I’m happy to switch to the less-verbose approach if you feel strongly.

Will post an updated rev this evening!

maxbrunsfeld:
I think it definitely helps clarify the ownership to have have all of the *writes* go through the helper functions; I just think that specifically for reading the *length* of the set, and for borrowing an entry from the array by index, I'm fine with reading the fields directly.

alex-pinkus:
Yeah, my worry is that if it’s legal to _sometimes_ treat it as an array, it becomes harder to spot when someone _shouldn’t_ be treating it that way. Then again, I might be overthinking it :)

alex-pinkus:
OK! Latest changes are up (I squashed them for posterity, since otherwise the history would add the struct stuff and then undo it). I ran `criterion` benchmarks for swift, elixir, kotlin (noticing @theHamsta's having tagged this issue), and cpp (for the other end of the spectrum, being a grammar that takes 20ms to compile its highlight queries):

_Swift_:
```
highlights/ts_query_new time:   [1.4395 s 1.4422 s 1.4452 s]
                        change: [-41.550% -41.065% -40.487%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
```
pdf-swift

_Kotlin_:
```
highlights/ts_query_new time:   [318.05 ms 329.40 ms 342.65 ms]
                        change: [-23.845% -20.696% -17.323%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  2 (10.00%) high mild
```
pdf-kotlin

_Elixir_:
```
highlights/ts_query_new time:   [801.62 ms 822.99 ms 848.83 ms]                                  
                        change: [-15.002% -12.698% -9.9846%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
```
pdf-elixir


_C++_:
```
highlights/ts_query_new time:   [20.889 ms 21.138 ms 21.405 ms]                                    
                        change: [-8.1952% -6.2113% -4.2796%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 200 measurements (5.50%)
  11 (5.50%) high mild
```
pdf-cpp


These results are not apples-to-apples with the ones in the original post because I'm on my Mac and ran the original ones from my Linux desktop. The `change:` lines for each are all relative to the runtime of the current `master` branch running on my Mac as well (of course).

Benchmark results from this repo are available at: [old.txt](https://github.com/tree-sitter/tree-sitter/files/7843120/old.txt)  [new.txt](https://github.com/tree-sitter/tree-sitter/files/7843123/new.txt) but mostly show no meaningful change from `master`.

alex-pinkus:
(I got notified about the CI failure by email; so just published an amended commit that fixes it)

maxbrunsfeld:
In terms of next steps for optimizing this, I believe that there is room for optimization via some additional ahead-of-time computation, as you mentioned in your second suggestion.

There is definitely a lot of redundancy in the way that we walk the state graph. Many of the "starting" states for a given grammar rule have identical behavior, and we could save some work by *not* seeding the analysis with *all* of the starting states. Specifically, any parse state with the same [`core_id`](https://github.com/tree-sitter/tree-sitter/blob/bf210f0c9ec7931c1a5f639461495db240aac149/cli/src/generate/tables.rs#L49) is going to behave identically with respect to this analysis.

One thing I'd like to try is to add a new array to the generated parsers, which stores the `core_id` for each state. Then we could use this data during query analysis, when initializing the "subgraphs". We could refrain from including multiple starting states with the same `core_id`.

I actually explored this approach when I first added query analysis, but after some very cursory testing, I found that it didn't improve things enough to be worth including in the initial implementation. Every time, we add some piece of data to the generated parsers, we have to bump Tree-sitter's ABI version (that expresses compatibility between the runtime library and the generated parsers), which causes a headache for some users. But now, I'm thinking that maybe for *some* grammars (like Elixir, Swift, Kotlin), this data actually might make a bigger difference than it did in the grammars that I initially tested against. So I think it might be worth exploring this direction further.

alex-pinkus:
Thank you for the pointers! That all makes a ton of sense. I hadn't thought about the ABI version bump but of course we would have to do that.

To make sure I'm understanding you correctly: we wouldn't bump `MIN_COMPATIBLE` version, just the `TREE_SITTER_LANGUAGE_VERSION`, and query analysis would look at all starting states if it didn't have `core_id` to use to dedup. But that would still be painful for some users who are on older `tree-sitter-cli` versions because once parsers get regenerated with the newer language version, they stop being able to use them? Or is there anything I'm overlooking?

maxbrunsfeld:
Yeah, you’re exactly right; we could actually do it in a backwards compatible way, so this wouldn’t need to be that painful even.

alex-pinkus:
Great, I can take a stab at that later this week and draft a new PR!

PR #1579, opened 2022-01-10
Add my Smali parser to the WIP list

yotamN:
This is my first time creating a parser so I would love to get some feedback on it.


amaanq:
Hi @yotamN, I maintain a fork based on your grammar [here](https://github.com/amaanq/tree-sitter-smali) but I was thinking it should be potentially rewritten (at least partially) since the logic for identifiers is a bit wonky in the context of queries/having highlights, and there is quite a few things missing (e.g. many additional keywords, const-method-handles, invoke-custom call sites, and more that are semi-new). If you are interested I would appreciate collaborating with you to have a complete smali parser

Issue #1405, opened 2021-09-22
get_column seems to return byte count, not codepoint count


ahelwer:
Using tree-sitter-cli 0.20.0. From [the docs](https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners) (emphasis mine):
 > `uint32_t (*get_column)(TSLexer *)` - (Experimental) A function for querying the current column position of the lexer. **It returns the number of unicode code points (not bytes) since the start of the current line.**

I tested this with the following grammar, which will match any sequence of `x` (U+0078) and `∧` (U+2227), where the `external_token` consumes nothing but is used to print out `get_column` values:
```js
module.exports = grammar({
  name: 'test',
  
  externals: $ => [
    $.external_token
  ],

  rules: {
    source_file: $ => repeat(seq($.external_token, $.codepoint)),
    codepoint: $ => choice('x', '∧')
  }
})
```
and the following `scanner.cc`:
```c++
#include <tree_sitter/parser.h>
#include <cstdio>

enum TokenType {
  EXTERNAL_TOKEN
};

extern "C" {
  void * tree_sitter_test_external_scanner_create() { return NULL; }
  void tree_sitter_test_external_scanner_destroy(void *payload) {}
  unsigned tree_sitter_test_external_scanner_serialize(void *payload, char *buffer) { return 0; }
  void tree_sitter_test_external_scanner_deserialize(void *payload, const char *buffer, unsigned length) {}

  bool tree_sitter_test_external_scanner_scan(
    void *payload,
    TSLexer *lexer,
    const bool *valid_symbols
  ) {
    while (' ' == lexer->lookahead) { lexer->advance(lexer, true); }
    uint32_t col = lexer->get_column(lexer);
    printf("NEXT: [%X] COL: [%d]\n", lexer->lookahead, col);
    return !lexer->eof(lexer);
  }
```
When tested on a file containing `x∧x∧`, the output of `tree-sitter parse` is:
>NEXT: [78] COL: [0]
NEXT: [2227] COL: [1]
NEXT: [78] COL: [4]
NEXT: [2227] COL: [5]
NEXT: [0] COL: [8]
(source_file [0, 0] - [0, 8]
  (external_token [0, 0] - [0, 0])
  (codepoint [0, 0] - [0, 1])
  (external_token [0, 1] - [0, 1])
  (codepoint [0, 1] - [0, 4])
  (external_token [0, 4] - [0, 4])
[...]

PR #1581, opened 2022-01-11
`get_column` now counts codepoints instead of bytes

ahelwer:
As discussed in https://github.com/tree-sitter/tree-sitter/issues/1405
Relevant to https://github.com/tree-sitter/tree-sitter-haskell/issues/69
Will address https://github.com/tlaplus-community/tree-sitter-tlaplus/issues/46
Reverts column calculation logic to version prior to https://github.com/tree-sitter/tree-sitter/pull/978


lib/src/lexer.c
```diff
@@ -233,8 +238,25 @@ static void ts_lexer__mark_end(TSLexer *_self) {
 
 static uint32_t ts_lexer__get_column(TSLexer *_self) {
   Lexer *self = (Lexer *)_self;
+  
+  uint32_t goal_byte = self->current_position.bytes;
+  
   self->did_get_column = true;
-  return self->current_position.extent.column;
+  self->current_position.bytes -= self->current_position.extent.column;
+  self->current_position.extent.column = 0;
```
ahelwer, line 246: One thing I worry about: at the very start of the file, could this possibly be moved prior to the BOM and count the BOM as a codepoint?

maxbrunsfeld: I think that it could do that, but that would be ok. Currently, we lexer does count the BOM as a code point; it just automatically skips it as if it were leading whitespace.

test/fixtures/test_grammars/external_unicode_column_alignment/scanner.c
```diff
@@ -0,0 +1,85 @@
...
+}
+
+bool tree_sitter_external_unicode_column_alignment_external_scanner_scan(
+  void *payload,
+  TSLexer *lexer,
+  const bool *whitelist
+) {
+  Scanner *scanner = payload;
+  // 9633 is the int equivalent of □ (U+25A1)
+  while (iswspace(lexer->lookahead) || 9633 == lexer->lookahead) {
```
maxbrunsfeld, line 56: Maybe write this integer as `0x25A1`? It's probably still worth having a comment explaining that's the box character, but it would be slightly  more self-explanatory.

maxbrunsfeld (commented):
Thanks so much for the PR @ahelwer. I left some minor comments.

/cc @razzeee @tek @wenkokke If you all get a chance, could you test this branch against the Haskell and Elm parsers? This should be an improvement to those parsers, since the behavior of `get_column` will now be more intuitive, and be identical between Wasm and native.

test/fixtures/test_grammars/external_unicode_column_alignment/scanner.c
ahelwer: I actually ran into a lot of difficulty with this, weirdly. Neither `L'□'` nor `'□'` worked, nor `\x25A1`. Even stranger is that it works just fine in my grammar: https://github.com/tlaplus-community/tree-sitter-tlaplus/blob/26bbaf5d07356cacbf21defa9316988bc732c8b3/src/scanner.cc#L517

maxbrunsfeld: Not a single-quoted character literal; just the plain integer `0x25A1`.

ahelwer: Ah sure, that works. I'll need to figure out the various ways of writing a codepoint and when they work/don't work eventually.

ahelwer:
Possible solution if performance becomes an issue: best-effort caching the current column position as an `int32_t`, living as part of the scanner state that isn't serialized into the tree. The value is set to `-1` to indicate "unknown", then set to zero whenever a newline is encountered and incremented on subsequent calls to `ts_lexer__do_advance`. Every time `ts_lexer_goto` is called the column position cache is set to `-1` (unknown). Whenever `get_column` is called, if the value is `-1` then its position is recalculated by reading from the start of the line; otherwise, the cached value is returned.

maxbrunsfeld:
I love that idea @ahelwer. If you want to do it as part of this, I'm ok with that, but I also think it might make sense to hold off until we do a bit of testing on large haskell / elm files. I'm not sure how often `get_column` is called; I'm hoping we won't notice regressions.

wenkokke:
@maxbrunsfeld This requires #1571 to work with tree-sitter-haskell, but it fixes some of the observed errors in the wasm parser—significantly more root nodes have type `haskell` instead of `ERROR`.

(I've not performed a full comparison of the generated trees.)

maxbrunsfeld:
Ok, that sounds good.

PR #1589, opened 2022-01-14
Ignore duplicate states when initializing subgraphs in `ts_query__analyze_patterns`

alex-pinkus:
This change exposes a new `primary_state_ids` field on the `TSLanguage` struct, and populates it by tracking the first encountered state with a given `core_id`. (For posterity: the initial change just exposed `core_id` and deduplicated within `ts_analyze_query`)

With this `primary_state_ids` field in place, the `ts_query__analyze_patterns` function only needs to populate its subgraphs with starting states that are _primary_, since non-primary states behave identically to primary ones. This leads to large savings across the board, since most states are not primary (more narrative after these graphs):

_Swift_:
States: 7797
Primary states: 2273

stats-swift

_Kotlin_:
States: 9424
Primary states: 897
stats-kotlin

_Elixir_:
States: 5658
Primary states: 529
stats-elixir

_C++_:
States: 5460
Primary states: 1236
stats-cpp

It appears that pruning out non-primary states causes a mostly linear reduction in runtime, although the overhead for analysis becomes more impactful. Put another way, the runtime per (primary) state is usually close to 60% higher than it was before the reduction. In the Swift grammar, after pruning all but 29% of the states. query analysis takes 45% as long as it used to. Similarly, in the C++ grammar, having 22% of the states leads to 37% of the original runtime. The outlier here is Kotlin, where 9.5% of states are primary and analysis takes 11% of the time (only 17% additional overhead). This doesn’t mean we’re doing _more work_ per state that we weren’t before, it just suggests that the overhead has started to dominate.

The flamegraphs for Swift show that the remaining work happens in roughly the same proportions that it did before.

_Before_:  
flamegraph-master  
_After_:  
flamegraph-after-dedup

Note that since the runtime for the actual work has gone down, the benchmark machinery (`rayon::*`, `criterion::*`) takes up a larger proportion of the analysis.

## Benchmarks from `tree-sitter/tree-sitter`
Full benchmarks are available at: [benchmark-old.txt](https://github.com/tree-sitter/tree-sitter/files/7877635/benchmark-old.txt) [benchmark-new.txt](https://github.com/tree-sitter/tree-sitter/files/7877636/benchmark-new.txt). Confusingly, I'm going to perform my comparisons using throughput in the subsequent analysis since these benchmark files don't do well with small-ms time per iteration.

**1. If we ignore sampling noise, the only meaningful changes are in the "Constructing Queries" step (unsurprising)**:  
Look, for instance, at the `tree-sitter-cpp` diff:
```
 Language: cpp
   Constructing Queries
-    highlights.scm                                     time 13 ms      speed 82 bytes/ms
+    highlights.scm                                     time 5 ms       speed 192 bytes/ms
   Parsing Valid Code:
     rule.cc                                            time 0 ms       speed 8448 bytes/ms
     marker-index.h                                     time 0 ms       speed 4832 bytes/ms
@@ -111,7 +112,7 @@ Language: cpp
     test.sh                                            time 4 ms       speed 496 bytes/ms
     install.sh                                         time 7 ms       speed 779 bytes/ms
     release.sh                                         time 0 ms       speed 657 bytes/ms
-    atom.sh                                            time 4 ms       speed 682 bytes/ms
+    atom.sh                                            time 5 ms       speed 568 bytes/ms
     doc-build.sh                                       time 5 ms       speed 543 bytes/ms
     relocate.sh                                        time 0 ms       speed 665 bytes/ms
     malloc.c                                           time 1 ms       speed 6049 bytes/ms
@@ -119,12 +120,12 @@ Language: cpp
     parser.c                                           time 3 ms       speed 11119 bytes/ms
     no_newline_at_eof.go                               time 0 ms       speed 280 bytes/ms
     proc.go                                            time 85 ms      speed 1381 bytes/ms
-    letter_test.go                                     time 5 ms       speed 2073 bytes/ms
+    letter_test.go                                     time 6 ms       speed 1776 bytes/ms
     value.go                                           time 48 ms      speed 1508 bytes/ms
     deeply-nested.html                                 time 8 ms       speed 668 bytes/ms
     deeply-nested-custom.html                          time 4 ms       speed 1214 bytes/ms
-    text-editor-component.js                           time 87 ms      speed 1721 bytes/ms
-    jquery.js                                          time 96 ms      speed 2550 bytes/ms
+    text-editor-component.js                           time 88 ms      speed 1702 bytes/ms
+    jquery.js                                          time 97 ms      speed 2523 bytes/ms
     Logger.kt                                          time 3 ms       speed 1130 bytes/ms
     python3.8_grammar.py                               time 64 ms      speed 775 bytes/ms
     tabs.py                                            time 1 ms       speed 618 bytes/ms
@@ -141,13 +142,13 @@ Language: cpp
     ast.rs                                             time 23 ms      speed 2729 bytes/ms
   Average Speed (normal): 6640 bytes/ms
   Worst Speed (normal):   4832 bytes/ms
-  Average Speed (errors): 1742 bytes/ms
+  Average Speed (errors): 1728 bytes/ms
   Worst Speed (errors):   16 bytes/ms
```

**2. The language that sees the _least_ improvement in its "Constructing Queries" step is `python`, with only a 33% improvement in throughput:**  
```
 Language: python
   Constructing Queries
-    highlights.scm                                     time 3 ms       speed 458 bytes/ms
-    tags.scm                                           time 1 ms       speed 129 bytes/ms
+    highlights.scm                                     time 2 ms       speed 611 bytes/ms
+    tags.scm                                           time 0 ms       speed 258 bytes/ms
```

**3. The only language to take _more than 100ms_ (on my Linux box) after this change is Swift:**
```
 Language: swift
   Constructing Queries
-    highlights.scm                                     time 983 ms     speed 2 bytes/ms
-    locals.scm                                         time 144 ms     speed 2 bytes/ms
+    highlights.scm                                     time 458 ms     speed 6 bytes/ms
+    locals.scm                                         time 153 ms     speed 2 bytes/ms
```

## Conclusion
When we exclude non-primary states from subgraph initialization, we see huge reductions in `ts_query__analyze_patterns` runtime across the board. Although this is imperceptible for the faster grammars (`rust` going from 9ms to 6ms, for instance), it makes a significant difference for others, bringing `kotlin`, `ruby`, and `elixir` under the commonly-accepted 100ms threshold of perceptibility of delay.

At this point, I expect further optimization for `swift` to come from changes to that grammar itself, since its primary state count is a clear outlier among the grammars. Of course, primary state count isn't the only variable here: the C++ parser is 14x faster than Elixir (5ms vs 71ms), but has 133% more primary states. It may be worth documenting the properties of grammars that are more or less performant so that authors like me can find ways to restructure our own grammar definitions.

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -106,6 +106,7 @@ impl Generator {
         }
 
         self.add_non_terminal_alias_map();
+        self.add_primary_state_id_list();
 
         let mut main_lex_table = LexTable::default();
         swap(&mut main_lex_table, &mut self.main_lex_table);
@@ -560,6 +561,29 @@ impl Generator {
             dedent!(self);
         }
         add_line!(self, "0,");
+        dedent!(self);
+        add_line!(self, "}};");
+        add_line!(self, "");
+    }
+
+    /// Produces a list of the "primary state" for every state in the grammar.
+    ///
+    /// The "primary state" for a given state is the first encountered state that behaves
+    /// identically with respect to query analysis. We derive this by keeping track of the `core_id`
+    /// for each state and treating the first state with a given `core_id` as primary.
+    fn add_primary_state_id_list(&mut self) {
+        add_line!(
+            self,
+            "static const TSStateId ts_primary_state_ids[STATE_COUNT] = {{"
+        );
+        indent!(self);
+        let mut first_state_for_each_core_id = HashMap::new();
+        for (idx, state) in self.parse_table.states.iter().enumerate() {
+            let primary_state = first_state_for_each_core_id
+                .entry(state.core_id)
+                .or_insert(idx);
+            add_line!(self, "[{}] = {},", idx, primary_state);
+        }
         dedent!(self);
         add_line!(self, "}};");
         add_line!(self, "");
@@ -1369,6 +1393,7 @@ impl Generator {
         if !self.parse_table.production_infos.is_empty() {
             add_line!(self, ".alias_sequences = &ts_alias_sequences[0][0],");
         }
+        add_line!(self, ".ts_primary_state_ids = ts_primary_state_ids,");
 
         // Lexing
         add_line!(self, ".lex_modes = ts_lex_modes,");
--- a/lib/binding_rust/bindings.rs
+++ b/lib/binding_rust/bindings.rs
@@ -855,5 +855,5 @@ extern "C" {
     );
 }
 
-pub const TREE_SITTER_LANGUAGE_VERSION: usize = 13;
+pub const TREE_SITTER_LANGUAGE_VERSION: usize = 14;
 pub const TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION: usize = 13;
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -21,7 +21,7 @@ extern "C" {
  * The Tree-sitter library is generally backwards-compatible with languages
  * generated using older CLI versions, but is not forwards-compatible.
  */
-#define TREE_SITTER_LANGUAGE_VERSION 13
+#define TREE_SITTER_LANGUAGE_VERSION 14
 
 /**
  * The earliest ABI version that is supported by the current version of the
--- a/lib/include/tree_sitter/parser.h
+++ b/lib/include/tree_sitter/parser.h
@@ -110,6 +110,7 @@ struct TSLanguage {
   const TSSymbol *public_symbol_map;
   const uint16_t *alias_map;
   const TSSymbol *alias_sequences;
+  const TSStateId *ts_primary_state_ids;
   const TSLexMode *lex_modes;
   bool (*lex_fn)(TSLexer *, TSStateId);
   bool (*keyword_lex_fn)(TSLexer *, TSStateId);
--- a/lib/src/language.h
+++ b/lib/src/language.h
@@ -200,6 +200,19 @@ static inline TSStateId ts_language_next_state(
   }
 }
 
+// Whether the state is a "primary state". If this returns false, it indicates that there exists
+// another state that behaves identically to this one with respect to query analysis.
+static inline bool ts_language_state_is_primary(
+  const TSLanguage *self,
+  TSStateId state
+) {
+  if (self->version >= 14) {
+    return state == self->ts_primary_state_ids[state];
+  } else {
+    return true;
+  }
+}
+
 static inline const bool *ts_language_enabled_external_tokens(
   const TSLanguage *self,
   unsigned external_scanner_state
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -960,28 +960,30 @@ static bool ts_query__analyze_patterns(TSQuery *self, unsigned *error_offset) {
         if (lookahead_iterator.next_state != state) {
           state_predecessor_map_add(&predecessor_map, lookahead_iterator.next_state, state);
         }
-        const TSSymbol *aliases, *aliases_end;
-        ts_language_aliases_for_symbol(
-          self->language,
-          lookahead_iterator.symbol,
-          &aliases,
-          &aliases_end
-        );
-        for (const TSSymbol *symbol = aliases; symbol < aliases_end; symbol++) {
-          array_search_sorted_by(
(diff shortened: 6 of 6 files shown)


maxbrunsfeld:
These results look promising!

> is there a more descriptive name than `core_id` to communicate intent?

Yeah, great question. I think that instead of exposing each state's `core_id` directly, we should assign to each state a `primary_state_id`, which would refer directly to another earlier *parse state*. The `primary_state_id` for a given state `s` would be defined as the id of the *first* parse state that has the same `core_id` as `s`. So for some states, the `primary_state_id` would be the state's own `id`.

```c
static const TSStateId ts_primary_state_ids[STATE_COUNT] = {
  // Some states' primary ids are equal to their own ids
  [0] = 0,
  [1] = 1,

  // Other states' primary ids refer to earlier states in the table
  [5] = 1,
  [6] = 2,
  
  // ...
};
```

This would simplify the query construction code: it could check if a state is "primary" by comparing its *primary id* with its own id. In older versions, every state would be considered "primary".

```c
static inline bool ts_language_state_is_primary(
   const TSLanguage *self,
   TSStateId state
 ) {
   if (self->version >= 14) {
     return self->primary_state_ids[state] == state;
   } else {
     return true;
   }
 }
```

#### Questions

* We could simplify the static data even further and just store an array of *booleans* indicating which states are "primary". I have a vague feeling like storing the state ids has some potential future use though.
* I'm not sure if "primary" is the right adjective for this. Maybe there's something less vague we should use

> Add Swift, Kotlin, and Elixir to benchmarks

This is a great idea. Thanks for volunteering to make that improvement along with this.

> Add backwards-compatibility tests to prove that version 13 can still be parsed

In my opinion, you don't need to do anything around this. I'm ok with just adding the field to the end of the struct like you've done, without any explicit backward-compatibility testing.

alex-pinkus:
> I think that instead of exposing each state's core_id directly, we should assign to each state a primary_state_id ... I'm not sure if "primary" is the right adjective for this. Maybe there's something less vague we should use

Exposing `primary_state_id` makes a lot of sense, and would save us from having to keep a bitset of `core_ids` in `query.c`, which would be great!

For the sake of brainstorming, alternate names might be `canonical_state`, `dominating_state`, `preemptive_state`, `overriding_state`, but I don't think any of those are better than `primary`. **Q: Any of those alternates sound better to you or should I go with `primary_state_id`?**

> We could simplify the static data even further and just store an array of booleans indicating which states are "primary". I have a vague feeling like storing the state ids has some potential future use though.

I'm with you that exposing state IDs might be useful, and since a `TREE_SITTER_VERSION` bump still has _some_ pain associated with it, I'm inclined to go broad here. If the choice were between `core_id` and `boolean`, I could see the hesitancy of not wanting to introduce more vocabulary / leak internal state, but the `primary_state_id` simplifier is a great middle ground.

> > Add Swift, Kotlin, and Elixir to benchmarks
>
> This is a great idea. Thanks for volunteering to make that improvement along with this.

Well, it was your idea originally :)

I notice that we currently grab the latest `master` when we fetch repos today. This seems dangerous for repos outside the `tree-sitter` org: if someone makes a change that causes instability, you won't be able to revert it yourself to fix CI. At the same time, if you _don't_ fetch the latest master from that repo, their broken change potentially goes unnoticed.

**Q: Would you be supportive of fetching those repos at some pinned release tag / commit hash?** We can talk more (in a separate change) about some experiments I was planning on trying out on `tree-sitter-swift` for dependabot-style upgrades of pinned repository versions. I think that would get you the best of both worlds.

maxbrunsfeld:
Yeah, it was a good idea to move that benchmark change to a separate PR. This is great.

I'd like to give downstream applications a chance to upgrade the library before we start bumping the ABI versions of grammars. So I'm going to gate this new feature behind a flag to `tree-sitter generate` for a brief time.

/cc @bfredl, it would be good to upgrade the Tree-sitter library inside Neovim soon, so that Neovim will be able to load newly-generated parsers that have this additional data. It allows queries to load significantly faster.

bfredl:
cc @clason

maxbrunsfeld:
Ok, the next release of the Tree-sitter CLI will be able to generate this new data if you pass `--abi=14` or `--abi=latest`, but by default it will continue to generate parsers with ABI version 13. I'd like to do a follow-up release fairly soon though, that bumps the default ABI version.

PR #1592, opened 2022-01-16
Add Kotlin, Swift, and Elixir to benchmarks

alex-pinkus:
Initially suggested in https://github.com/tree-sitter/tree-sitter/pull/1578#issuecomment-1008512836. I decided to split this out from #1589 for the purpose of discussion.

~Because the Swift grammar generates its `grammar.json` from source every time, this also modifies the `generate-fixtures` scripts so that they attempt an `npm install` when they need to.~ Edit: this is no longer the case because the Swift grammar won’t build from source on the Appveyor 32-bit host. Instead this just points at a tag that has a checked-in `grammar.json`.

This change does not add these repositories into `corpus_test` because `elixir` seems to have a spurious failure there and `swift` runs forever without terminating.

### DIFF
--- a/script/fetch-fixtures
+++ b/script/fetch-fixtures
@@ -3,10 +3,11 @@
 GRAMMARS_DIR=$(dirname $0)/../test/fixtures/grammars
 
 fetch_grammar() {
-  local grammar=$1
-  local ref=$2
+  local organization=$1
+  local grammar=$2
+  local ref=$3
   local grammar_dir=${GRAMMARS_DIR}/${grammar}
-  local grammar_url=https://github.com/tree-sitter/tree-sitter-${grammar}
+  local grammar_url=https://github.com/${organization}/tree-sitter-${grammar}
 
   echo "Updating ${grammar} grammar..."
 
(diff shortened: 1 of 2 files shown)


alex-pinkus:
Seems like appveyor must have a really old version of npm which doesn’t play nicely with the swift parser’s typescript stuff.

alex-pinkus:
And indeed, trying to explicitly declare the Node version makes the issue quite clear:

```
Unsupported platform for node-win-x86@12.22.9: wanted {"os":"win32","arch":"x86"} (current: {"os":"win32","arch":"ia32"})
npm ERR! notsup Valid OS:    win32
npm ERR! notsup Valid Arch:  x86
npm ERR! notsup Actual OS:   win32
npm ERR! notsup Actual Arch: ia32
```

So I think the thing to do, then, is to make a special branch `1.1-with-static-grammar.json` for inclusion.

dundargoc:
Looks useful. Please rebase :)

dundargoc:
My sources tell me Max only allows testing parsers that are in tree-sitter organization.

PR #1599, opened 2022-01-17
Add `--abi` flag to the generate command, generate version 13 by default

maxbrunsfeld:
###  Background

In https://github.com/tree-sitter/tree-sitter/pull/1589, we've introduced a new language ABI version which allows queries to be loaded more efficiently. As with most ABI version bumps, new builds of the Tree-sitter *library* are compatible with old generated parsers, but the reverse is not true. Once we regenerate a parser with the new ABI version (14) old versions of the library won't be able to load it.

### Change

This PR adds an `--abi` flag to the `generate` subcommand. By default, Tree-sitter will generate the current language version that is commonly in use (right now version 13), but you can supply an alternative version number as an argument to `--abi`, or set `--abi=latest` to use the latest supported version. In CI, we generate the parsers with the latest ABI version.

### Notes

This also removes an undocumented `--prev-abi` flag, whose behavior was pretty unstable and I don't *think* anyone else was using besides me. If anyone was using it, it should be pretty easy to switch to using `--abi=13` (or whichever version is desired).

### DIFF
--- a/cli/src/generate/mod.rs
+++ b/cli/src/generate/mod.rs
@@ -40,7 +40,7 @@ struct GeneratedParser {
 pub fn generate_parser_in_directory(
     repo_path: &PathBuf,
     grammar_path: Option<&str>,
-    next_abi: bool,
+    abi_version: usize,
     generate_bindings: bool,
     report_symbol_name: Option<&str>,
 ) -> Result<()> {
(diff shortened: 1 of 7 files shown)


maxbrunsfeld:
Thanks for the report @clason; I’ll do another patch release this week and fix that.

PR #1601, opened 2022-01-18
Fix ABI 14 back compat by moving primary_field_ids to the end

alex-pinkus:
Due to an oversight in #1589, I added `primary_field_ids` into the `TSLanguage` struct in a place that wasn't the end. This is not actually backwards compatible and causes downstream failures :(
﻿
CC @maxbrunsfeld

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -1390,10 +1390,6 @@ impl Generator {
             add_line!(self, ".alias_sequences = &ts_alias_sequences[0][0],");
         }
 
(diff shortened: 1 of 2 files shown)


alex-pinkus:
Validated that without this change, I see the same segmentation fault that was initially reported in https://github.com/neovim/neovim/pull/17130 (which I believe is also causing their test failures). With this change, I see no segfault and highlighting works properly.

maxbrunsfeld:
Oh, wow, I totally glossed over that when reviewing it too. Thanks for following up with the fix!

PR #1602, opened 2022-01-18
prevent future captures for `#is-not? local` matches

### LINKED ISSUE #1597: allow `#is-not? local` to gate multiple captures in a stanza?
Currently the property predicate `(#is-not? local)` may only apply to a node that is a local reference, as in the ruby queries:

```tsq
((identifier) @function.method
 (#is-not? local))
(identifier) @variable
```

If you try to use `(#is-not? local)` on multiple captures, for example:

```tsq
((field_access
   record: (identifier) @module
   field: (label) @function)
 (#is-not? local))
```

then the first capture (`@module`) is correctly gated on whether or not `(identifier)` is local, but the second capture (`@function`) always matches, even when the `(identifier)` is local.

I think it is possible to fix this case while keeping the current "locals info can be parsed and used in a single pass" behavior like so (in [this](https://github.com/tree-sitter/tree-sitter/blob/e96ee19901947d8550abbcf425791c2a411c8d97/highlight/src/lib.rs#L834-L850) block):

```diff
diff --git a/highlight/src/lib.rs b/highlight/src/lib.rs
index 58d7e88c..0d097201 100644
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -835,6 +835,7 @@ where
             // highlighting patterns that are disabled for local variables.
             if definition_highlight.is_some() || reference_highlight.is_some() {
                 while layer.config.non_local_variable_patterns[match_.pattern_index] {
+                    match_.remove();
                     if let Some((next_match, next_capture_index)) = layer.captures.peek() {
                         let next_capture = next_match.captures[*next_capture_index];
                         if next_capture.node == capture.node {
```

When the capture that is not local is skipped over, we `.remove()` the match (which calls `ts_query_cursor_remove_match` from `lib/src/query.c`). This is used [higher in the same function](https://github.com/tree-sitter/tree-sitter/blob/e96ee19901947d8550abbcf425791c2a411c8d97/highlight/src/lib.rs#L691-L693) for injections. I tried this locally but despite the `.remove()`, I still see the `@function` capture being emitted by the `layer.captures` peekable :thinking:

Am I on the right track to fixing this? Is this something that even should be fixed, or should locals info only be used to highlight/not-highlight the node interpreted as a local definition/reference?

the-mikedavis:
closes #1597

~In this case I thought it would make sense to use `match.remove()`, but that appears to only clear out existing matches and does not prevent future captures for the removed match. It might make sense to change the C query api to ignore future captures for a removed match, but fixing it in the rust binding was so much easier :sweat_smile:~

~Here we store a set of match IDs to ignore when a capture for each match finds that it is not local. I suppose this could be used by other mechanism like injections, so I've left it named generically, but I'm open to suggestions on that.~

### DIFF
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -111,7 +111,7 @@ struct HighlightIterLayer<'a> {
     scope_stack: Vec<LocalScope<'a>>,
     ranges: Vec<Range>,
     depth: usize,
-    ignore_match_ids: Vec<u32>,
+    ignore_match_id: i32,
 }
 
 impl Highlighter {
@@ -414,7 +414,7 @@ impl<'a> HighlightIterLayer<'a> {
                     captures,
                     config,
                     ranges,
-                    ignore_match_ids: Vec::new(),
+                    ignore_match_id: -1,
                 });
             }
 
@@ -685,7 +685,7 @@ where
             let (mut match_, capture_index) = layer.captures.next().unwrap();
             let mut capture = match_.captures[capture_index];
 
-            if layer.ignore_match_ids.contains(&match_.id()) {
+            if layer.ignore_match_id >= match_.id() as i32 {
                 continue 'main;
             }
 
@@ -841,7 +841,7 @@ where
             // highlighting patterns that are disabled for local variables.
             if definition_highlight.is_some() || reference_highlight.is_some() {
                 while layer.config.non_local_variable_patterns[match_.pattern_index] {
-                    layer.ignore_match_ids.push(match_.id());
+                    layer.ignore_match_id = match_.id() as i32;
                     if let Some((next_match, next_capture_index)) = layer.captures.peek() {
                         let next_capture = next_match.captures[*next_capture_index];
                         if next_capture.node == capture.node {
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -111,7 +111,6 @@ struct HighlightIterLayer<'a> {
     scope_stack: Vec<LocalScope<'a>>,
     ranges: Vec<Range>,
     depth: usize,
-    ignore_match_id: i32,
 }
 
 impl Highlighter {
@@ -414,7 +413,6 @@ impl<'a> HighlightIterLayer<'a> {
                     captures,
                     config,
                     ranges,
-                    ignore_match_id: -1,
                 });
             }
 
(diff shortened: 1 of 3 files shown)


highlight/src/lib.rs
```diff
@@ -683,6 +685,10 @@ where
             let (mut match_, capture_index) = layer.captures.next().unwrap();
             let mut capture = match_.captures[capture_index];
 
+            if layer.ignore_match_ids.contains(&match_.id()) {
```
the-mikedavis, line 688: Since we're just doing contains lookups, a hashset would be better, no? Or actually I guess you could use just a u32 because match IDs are incrementing integers...

the-mikedavis:
Actually, I think it might be better to fix this in the C source instead. Currently `ts_query_cursor_remove_match` only removes finished QueryStates, so it won't remove a match that hasn't been finished. If we also try removing the match from `states`, I think we can prevent new captures on a match we want to remove

maxbrunsfeld:
Yeah, that makes sense @the-mikedavis. I think you're right that it should be fine to remove matches from `states`.

maxbrunsfeld:
@the-mikedavis Could you add a unit test in `query_test.rs` for this behavior? It should go next to [this test](https://github.com/tree-sitter/tree-sitter/blob/2346570901ef01517dad3e4a944a36d7b7237e4f/cli/src/tests/query_test.rs#L3056), and could be called something like `test_query_captures_with_matches_removed_before_they_finish`.

To set up the test, you'll need to use a pattern that is *guaranteed* to match after a certain point, at which *some* captures have already been returned, but there are still more captures. When Tree-sitter detects that a pattern is guaranteed to match, it will start to eagerly return the captures that it has found, even though it hasn't matched the entire pattern yet.

For example, in your example from #1597:

```clj
((field_access
   record: (identifier) @module
   field: (label) @function)
```

I'm not sure what language this is, but I'm guessing that a `field_access` node *always* has a `field: (label)` child, so the pattern is guaranteed to match successfully once the `record: (identifier)` is found. So Tree-sitter will return the `(identifier)` capture while the match is still unfinished.

You'll need to pick an example *similar* to this, but using one of the "fixture grammars" that Tree-sitter already uses as part of its test suite.

maxbrunsfeld:
Nice work @the-mikedavis!

PR #1612, opened 2022-01-22
Various updates to Creating Parsers docs

ahelwer:
This update addresses undocumented features, issues I've run into while writing my parser, and questions I see recur on the discussions page:
1. Documented test separator suffixes (https://github.com/tree-sitter/tree-sitter/pull/1348)
1. Documented partial precedences field (https://github.com/tree-sitter/tree-sitter/pull/939)
1. Documented lexer `eof` function
1. Added external scanning to conflicting tokens rules
1. Added earliest starting position to conflicting tokens rules (see note below)
1. Added note emphasizing difference between lexical precedence and parse precedence
1. Added note about tree-sitter calling external scanner with all tokens marked valid during error recovery
1. Added note about risk of infinite loops in external scanners
1. Added note about terminal keywords in the externals array

I'm not 100% sure I got (4) right, but it's meant to address things like https://github.com/tree-sitter/tree-sitter/discussions/1610 which I also ran into (regexes gobbling up everything). Maybe it should just be added to the section on regexes themselves. Happy to make changes to any of these, or remove them if they just don't fit in the docs as written.

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -543,7 +543,7 @@ Lexical precedence determines which token is chosen to interpret a given section
 It is a lower-level operation that is done first.
 The above list fully capture tree-sitter's lexical precedence rules, and you will probably refer back to this section of the documentation more often than any other.
 Most of the time when you really get stuck, you're dealing with a lexical precedence problem.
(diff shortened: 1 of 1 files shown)


ahlinc:
@ahelwer Thank you for the awesome documentation improvements!

Yherealtita (commented):
Reviewing

PR #1555, opened 2021-12-26
Add CMake support for tree_sitter C library

aminya:
- This adds cross-platform CMake support for the tree_sitter C library. 
- This also helps the IDEs to work correctly (autocomplete, showing the warnings in the code, etc.) by generating the required `compile_command.json` for clangd or MSVC. e.g. automatic refactoring was used in #1618
- Enables the compiler warnings. For example, see the shadowed variables revealed by PR #1618. The CMake is based on the cpp-best-practices project, which helps the project evolve. 
 
- This cmake file generates a tree-sitter package, which allows it to be used from other projects and potentially added to the C++/C package managers like vcpkg and Conan
```cmake
    find_package(tree-sitter CONFIG REQUIRED)
    target_link_libraries(main PRIVATE  tree-sitter::tree_sitter tree-sitter::project_options tree-sitter::project_warnings)
```
image

- optionally, you can remove the old Makefile as it doesn't have any of the above-mentioned points.

PR #1618, opened 2022-01-27
fix: fix variable shadows

### LINKED ISSUE #1888: Warnings when compiling with -Wshadow
## Issue

There is a vast number of warnings (currently 31) when tree-sitter is compiled with `-Wshadow`. This makes it very easy to use another variable from the upper scope mistakenly.

```
tree-sitter\lib\src\query.c:3551:23: warning: Local variable 'state' shadows outer variable [shadowVariable]
          QueryState *state = &self->states.contents[j];
                      ^
tree-sitter\lib\src\query.c:3395:21: note: Shadowed declaration
        QueryState *state = &self->states.contents[i];
                    ^
tree-sitter\lib\src\query.c:3551:23: note: Shadow variable
          QueryState *state = &self->states.contents[j];
```

## Solution

This is fixed by the pull request #1618

aminya:
Fixes #1888

This fixes 31 shadow warnings from `-Wshadow` in tree-sitter.

Detected using #1555 

The variables were automatically renamed using the IDE, so they are correct without human error. This became possible because #1555 generates `compile_commands.json` that makes clangd work inside the IDE, and so automatic refactoring becomes available.

### DIFF
--- a/lib/src/array.h
+++ b/lib/src/array.h
@@ -109,9 +109,9 @@ extern "C" {
 // callback to determine the order.
 #define array_insert_sorted_with(self, compare, value) \
   do { \
-    unsigned index, exists; \
-    array_search_sorted_with(self, compare, &(value), &index, &exists); \
-    if (!exists) array_insert(self, index, value); \
+    unsigned index__, exists__; \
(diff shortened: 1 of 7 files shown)


amaanq:
Superseded by #2370

PR #1624, opened 2022-01-29
feat: make logging a separate feature

aminya:


ObserverOfTime:
I don't think this is particularly useful. Logging is not enabled unless a logger is defined.

Issue #1626, opened 2022-01-29
Improve performance of `get_column`


ahelwer:
As sketched [here](https://github.com/tree-sitter/tree-sitter/pull/1581#issuecomment-1010512449) in the PR changing `get_column` to count codepoints instead of bytes, we can create a "best effort" cache of the current column position:

1. A `column_cache` field is added to the `Lexer` struct in `lib/src/lexer.h`
2. `column_cache` is updated as `ts_lexer__advance` is called in `lib/src/lexer.c`, incrementing with each character and being set to `0` upon consuming a newline
3. If `ts_lexer_goto` is called then the `column_cache` field is set to a value indicating "unknown", possibly `-1` if given type `int32_t` or with a separate boolean `column_unknown` field in the `Lexer` state if given type `uint32_t` to match the return type of `get_column`
4. If `get_column` is called when `column_cache` is in a "known" state, the `column_cache` value is returned; otherwise, the lexer reads from the beginning of the line, updates `column_cache`, and returns the calculated value

In what cases is `ts_lexer_goto` called? I can think of the following possibilities:

1. When an external scanner is called, reads some text, then returns `false`
2. When an external scanner is called, calls `mark_end`, reads some more text, then returns `true`
3. When the internal lexer reads an identifier twice for keyword extraction
3. During error recovery
4. At the start of parsing

Are there any others? Since (1) and (2) are likely to be quite common among external scanners using `get_column`, additional caches could be added to handle those cases so calling `ts_lexer_goto` would set the value of `column_cache` to a value recorded upon entering an external scanner or calling `mark_end`. However, that would likely require changes to `lib/src/parser.c` in addition to `lib/src/lexer.c`.

PR #1637, opened 2022-02-04
Implemented basic cache for `get_column`

ahelwer:
Implements the basic cache described in https://github.com/tree-sitter/tree-sitter/issues/1626
Unfortunately I don't see huge performance differences when parsing the ten largest files in the [tlaplus/examples](https://github.com/tlaplus/examples) repo. Here are results for 0.20.0 (byte-based `get_column`), 0.20.4 (codepoint-based `get_column`), and this patch, all of which seem to have variance of +/- 15 ms:
| File | 0.20.0 Parse Time | 0.20.4 Parse Time | This Patch Parse Time |
|-----|--------------------|---------------------|--------------------------|
| aba_asyn_byz.tla | 52 ms | 57 ms | 43 ms |
| AllocatorImplementation.tla | 39 ms | 58 ms | 38 ms |
| AllocatorRefinement.tla | 53 ms | 77 ms |  58 ms |
| SchedulingAllocator.tla | 62 ms | 64 ms | 60 ms |
| SimpleAllocator.tla | 59 ms | 76 ms | 47 ms |
| Bakery.tla | 78 ms | 78 ms | 57 ms |
| Boulanger.tla | 61 ms | 77 ms  | 62 ms |
| bcastByz.tla | 76 ms | 91 ms | 74 ms |
| bcastFolklore.tla | 57 ms | 77 ms | 61 ms |
| bosco.tla | 64 ms | 85 ms | 61 ms |

(previously I thought the differences were quite large but it turned out that's because I wasn't using the `tree-sitter.exe` file directly - important!)

For my future reference, here are the commands I used to quickly get these numbers:
<details><summary>pwsh</summary>
<p>

```pwsh
$specs = Get-ChildItem -Path .\test\examples\external\specifications -Filter "*.tla" -Exclude "Reals.tla","Naturals.tla" -Recurse
$largest = Sort-Object -InputObject $specs -Property "Length" -Descending
$largest | Select-Object -First 10 |% {Write-Host $_.Name; Measure-Command {.\node_modules\tree-sitter-cli\tree-sitter.exe parse -q $_.FullName}} |% {$_.Milliseconds}
```
</p>
</details>

These changes do not implement the `mark_end` nor cache-before-external-scanner optimizations suggested in the linked work item.

If the basic approach is approved I will write tests. Currently all tests pass in my grammar, which makes heavy use of `get_column`.

These changes also fix an extremely minor bug where on the first line of the file the BOM could be counted as a column item.

### DIFF
--- a/lib/src/lexer.c
+++ b/lib/src/lexer.c
@@ -36,6 +36,48 @@ static const TSRange DEFAULT_RANGE = {
   .end_byte = UINT32_MAX
 };
 
+/**
+ * If column cache is valid, return true and set value of out_val to
+ * cached current column position; otherwise return false and set out_val
+ * to 0.
+ * @param self The lexer state.
+ * @param out_val Out parameter; cached column position.
+ * @return Whether the column cache is valid.
+ */
+static bool ts_lexer__try_get_cached_column_value(const Lexer *self, uint32_t *out_val) {
+  *out_val = self->column_cache.value;
+  return self->column_cache.valid;
+}
+
+/**
+ * Sets the column cache to the given value and marks it valid.
+ * @param self The lexer state.
+ * @param val The new value of the column cache.
+ */
+static bool ts_lexer__set_column_cache(Lexer *self, uint32_t val) {
+  self->column_cache.valid = true;
+  self->column_cache.value = val;
+}
+
+/**
+ * Increments the value of the column cache; no-op if invalid.
+ * @param self The lexer state.
+ */
+static void ts_lexer__increment_column_cache(Lexer *self) {
+  if (self->column_cache.valid) {
+    self->column_cache.value++;
+  }
+}
+
+/**
+ * Marks the column cache as invalid.
+ * @param self The lexer state.
+ */
+static void ts_lexer__invalidate_column_cache(Lexer *self) {
+  self->column_cache.valid = false;
+  self->column_cache.value = 0;
+}
+
 // Check if the lexer has reached EOF. This state is stored
 // by setting the lexer's `current_included_range_index` such that
 // it has consumed all of its available ranges.
@@ -105,6 +147,7 @@ static void ts_lexer__get_lookahead(Lexer *self) {
 static void ts_lexer_goto(Lexer *self, Length position) {
   self->current_position = position;
   bool found_included_range = false;
+  ts_lexer__invalidate_column_cache(self);
 
   // Move to the first valid position at or after the given position.
   for (unsigned i = 0; i < self->included_range_count; i++) {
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
This looks good!

lib/src/lexer.c
```diff
@@ -239,23 +289,36 @@ static void ts_lexer__mark_end(TSLexer *_self) {
...
+      self->data.lookahead == BYTE_ORDER_MARK
+    ) ts_lexer__do_advance(self, false, false);
+
+    while (self->current_position.bytes < goal_byte && !ts_lexer__eof(_self) && self->chunk) {
+      ts_lexer__do_advance(self, false, false);
+      result++;
+    }
+
+    ts_lexer__set_column_cache(self, result);
+  }
```
maxbrunsfeld, line 320: I think that you simplify this code a little bit. I would suggest:

1. Eliminating the `update_column_cache` parameter to `ts_lexer__do_advance`. Just have that method *always* update the column cache:

2. Restructure this `ts_lexer__get_column` function so that it just populates the column cache if necessary, using the normal code path, and then returns its value:

```c
static uint32_t ts_lexer__get_column(TSLexer *_self) {
  Lexer *self = (Lexer *)_self;
  self->did_get_column = true;

  if (!self->column_cache.valid) {
    uint32_t goal_byte = self->current_position.bytes;
    self->current_position.bytes -= self->current_position.extent.column;
    self->current_position.extent.column = 0;
    self->column_cache.valid = true;
    self->column_cache.value = 0;

    // ... (advance back to the initial position)
  }
  
  return self->column_cache.value;
}
```

ahelwer:
@maxbrunsfeld restructured the `get_column` function as requested. Added logging, which has been useful for debugging, but can easily be removed. Was looking at plugging up some cache invalidations but there are so many possible sources and it doesn't seem to kill performance that much so maybe it's best to keep it simple.

maxbrunsfeld:
I think I'd prefer to leave out the column cache logging from the final result, since it feels a bit more "internal" of a concept than the rest of the logging. But totally makes sense to use that as a strategy when debugging.

Other than that, this looks great. Thanks for making that stylistic tweak, and documenting the performance improvements so well. Glad to see that this speeds up the TLA+ parser so much!

ahlinc:
@ahelwer @maxbrunsfeld Could you tell what more is required for this PR? Is it ready for merging?

amaanq:
@ahelwer can you rebase to fix the conflicts?

ahlinc:
@ahelwer for the next time, please don't disable permission for maintainers to update a PR branch.

ahelwer:
@ahlinc @amaanq looking at the conflicts I think they're related to the fix for that infinite loop when calling get_column at end of file. Did either of you work on that? I don't want to cause a regression.

amaanq:
Yeah I believe @ahlinc did,

ahlinc:
> Yeah I believe @ahlinc did,

Unfortunately it was several months ago and I don't remember all details, I'll look on it again.

dundargoc:
Superseded by https://github.com/tree-sitter/tree-sitter/pull/2503

PR #1638, opened 2022-02-04
Rust loader library: Make dependencies optional

Sakarah:
The `tree-sitter-loader` crate currently always pull `tree-sitter-highlight` and tree-sitter-tags as dependencies.
However, apart from the tree-sitter CLI, most clients will not need both of these libraries.

This commit makes the dependencies optional, but still include them by default in order not to break the backward compatibility.
For the next breaking change, we probably should think of removing them from default.

### DIFF
--- a/cli/loader/Cargo.toml
+++ b/cli/loader/Cargo.toml
@@ -14,6 +14,9 @@ categories.workspace = true
 
 [features]
 wasm = ["tree-sitter/wasm"]
+# TODO: For backward compatibility these must be enabled by default,
+# consider removing for the next semver incompatible release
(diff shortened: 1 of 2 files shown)


dcreager (approved):
This looks good to me, thanks!  I'm working on the deeper detangling of these crates in a couple of feature branches ([`clean-up-loader`](https://github.com/dcreager/tree-sitter/tree/clean-up-loader) and [`multipart-query`](https://github.com/dcreager/tree-sitter/tree/multipart-query)).  That will hopefully make this API much easier to use and not dependent on highlight/tags like it currently is.

Issue #888, opened 2021-01-21
Install.js does not work behind a proxy


dnns92:
Dear devs, 
I cant install the package on my ubuntu 1804 machine being behind a proxy using npm. The bug occurs because the download of the github file does not use system proxy environment variables.

I have a workaround, but It may lead to differnet results (see below)

I have http_proxy, https_proxy, HTTP_PROXY, HTTPS_PROXY set as env variables on my machine. 

I think this happens because the https package of node has no default settings for the proxy. I can download the file using wget (wget uses the proxy environment variables named above).

This is the error:
```
Downloading https://github.com/tree-sitter/tree-sitter/releases/download/0.18.0/tree-sitter-linux-x64.gz
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: connect ETIMEDOUT  <YOUR SERVER IP>
```

The file gets downloaded here in the install.js:
```
get(assetURL, response => {
  if (response.statusCode > 299) {
    throw new Error([
      'Download failed',
      '',
      `url: ${assetURL}`,
      `status: ${response.statusCode}`,
      `headers: ${JSON.stringify(response.headers, null, 2)}`,
      '',
    ].join('\n'));
  }
  response.pipe(zlib.createGunzip()).pipe(file);
});

// FUNCTION DEFINITION
function get(url, callback) {
  https.get(url, response => {
    if (response.statusCode === 301 || response.statusCode === 302) {
      get(response.headers.location, callback);
    } else {
      callback(response);
    }
  });
}
```

-------------------- WORKAROUND --------------
1.) try to install using npm and let it fail
2.) go to your node_modules/tree-sitter-cli/ folder and open install.js
3.) comment out the get function call "get(bla)", "fs.createWriteStream(executableName);" and the "file.on(bla)" completely and save the file
4.) download the github file behind proxy using wget
5.) copy file to node_modules/tree-sitter-cli/<your-filename.gz> and "gzip -d" on your file 
6.) chmod 755 <your-file>
7.) echo 'hello' > example_file
8.) tree-sitter generate
9.) tree-sitter parse example_file  returns  (source_file [0, 0] - [1, 0])

_**This should have returned "(source_file [1, 0] - [1, 5])". Is this a problem ?**_

maxbrunsfeld:
Thanks for the report. If you have a fix you'd like to submit as a PR, to respect these proxy-related environment variables, I'd be open to merging that.

[...]

PR #1643, opened 2022-02-06
Fix: Set https proxy for npm cli install script

fengT-T:
fix https://github.com/tree-sitter/tree-sitter/issues/888


gavares:
I work behind a corp proxy and would love to see this change merged in so that I can install tree-sitter

PR #1659, opened 2022-02-19
Name anonymous types in api.h

MatthewGentoo:
Anonymous structs cannot be forward declared.  This change names anonymous types in api.h so that consumers can forward declare them.

### DIFF
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -41,46 +41,46 @@ typedef struct TSTree TSTree;
 typedef struct TSQuery TSQuery;
 typedef struct TSQueryCursor TSQueryCursor;
 
-typedef enum {
+typedef enum TSInputEncoding {
   TSInputEncodingUTF8,
   TSInputEncodingUTF16,
 } TSInputEncoding;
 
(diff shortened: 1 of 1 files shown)


MatthewGentoo:
ping

ahlinc:
@MatthewGentoo Could you explain what sense to have forward declarations for structs from the `api.h` file if the `api.h` file is protected by `#ifndef TREE_SITTER_API_H_` and can be included many times in every place where it's needed?

dundargoc:
What problem does this PR solve? I don't get it. I think the C code should work as is.

MatthewGentoo:
Hi, sorry for the delayed response and lack of explanation.

In C and C++, anonymous types can't be forward declared. Forward declaring the types allows for headers in the application to skip including `tree_sitter/api.h`. While the impact to compile time is quite small (some libraries, especially C++ ones, are a lot worse...) every little helps in larger applications.

dundargoc:
Eh, I feel like the structs here are so small that I find it hard to believe it'd make much difference. That said, I don't see a problem with including this anyway since it's such a non-intrusive change. Thanks for the explanation.

Issue #451, opened 2019-09-24
Unicode Plane0 surrogates not supported


blinker29:
Probably not critical, but regex values \uD800 to \uDFFF fail (see below), preventing a search for any non-BMP value pairs, i.e.,  /\uD834\uDD1E/.  There could be good reasons to avoid the complexities, but I thought I would point out the gap in case it was helpful.

> Error processing rule Pex_token1
> Details:
>   regex parse error:
>     \uD834\uDD1E
>       ^^^^
> error: hexadecimal literal is not a Unicode scalar value

maxbrunsfeld:
Did you try writing the character itself in the regex (`/𝄞/`)?

Unlike JavaScript, Tree-sitter does not assume strings are represented as UTF16, so regex syntax that is UTF16-specific currently doesn't work. The regex engine itself should work fine for non BMP characters though.

blinker29:
That worked, thank you.  I would prefer that the parser handle valid \u specs, but handling D800-DFFF is probably an edge case and not a priority.  I'm going to close.  At least there's a record for anyone else who happens to run into this limitation to resurrect as needed.  Cheers.

PR #1660, opened 2022-02-19
Expand regex support to include emojis and binary ops

alex-pinkus:
Certain programming languages allow emojis to be used as identifiers for [very serious and good reasons](https://www.swiftbysundell.com/special/emoji-driven-development-in-swift/). This pull request makes it possible for tree-sitter grammars to match that behavior without writing their own very long regexes or resorting to custom scanners by enabling the use of `\p{Emoji}`. As noted in [this issue](https://github.com/alex-pinkus/tree-sitter-swift/issues/60), there is, in fact, real code in the world that makes use of emoji identifiers!

The `Emoji` property alias is already present, but the actual property is not available since it lives in a new file. This adds that file to the `generate-unicode-categories-json`.

The `emoji-data` file follows the same format as the ones we already consume in `generate-unicode-categories-json`, so adding emoji support is fairly easy. Without this, grammars would need to hard-code a set of unicode ranges in their own regex. The Javascript library `emoji-regex` cannot be used because of #451.

For unclear reasons, the characters #, *, and 0-9 are marked as `Emoji=Yes` by `emoji-data.txt`. Because of this, a grammar that wishes to use emojis is likely to want to exclude those characters, so this change also adds support for binary operations in regexes, e.g. `[\p{Emoji}&&[^#*0-9]]`.

Lastly (and perhaps controversially), this change introduces new variables available at grammar compile time, for the major, minor, and patch versions of the tree-sitter CLI used to compile the grammar. This will allow grammars to conditionally adopt these new regex features while remaining backward compatible with older versions of the CLI. Without this part of the change, grammar authors who do not precompile and check-in their `grammar.json` would need to wait for downstream systems to adopt a newer tree-sitter CLI version before they could begin to use these features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -554,6 +554,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "semver"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0486718e92ec9a68fbed73bb5ef687d71103b142595b406835649bebd33f72c7"
+
+[[package]]
 name = "serde"
 version = "1.0.130"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -737,6 +743,7 @@ dependencies = [
  "regex",
  "regex-syntax",
  "rustc-hash",
+ "semver",
  "serde",
  "serde_json",
  "smallbitvec",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -32,6 +32,7 @@ lazy_static = "1.2.0"
 regex = "1"
 regex-syntax = "0.6.4"
 rustc-hash = "1"
+semver = "1.0"
 serde = { version = "1.0.130", features = ["derive"] }
 smallbitvec = "2.5.1"
 tiny_http = "0.8"
--- a/cli/src/generate/mod.rs
+++ b/cli/src/generate/mod.rs
@@ -20,6 +20,7 @@ use self::rules::AliasMap;
 use anyhow::{anyhow, Context, Result};
 use lazy_static::lazy_static;
 use regex::{Regex, RegexBuilder};
+use semver::Version;
 use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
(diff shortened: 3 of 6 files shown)


alex-pinkus:
Honestly not sure what’s going on with that [Appveyor x86 build failure](https://ci.appveyor.com/project/maxbrunsfeld/tree-sitter/builds/42640595/job/0cfwk766a3pdr5fh). Seems to be something about the Python grammar, which I haven’t changed.

maxbrunsfeld:
> For unclear reasons, the characters `#`, `*`, and `0-9` are marked as `Emoji=Yes` by emoji-data.txt.

That is very weird, and good to know. JavaScript regexes seem to behave that way too:

```js
/\p{Emoji}/u.test('#') // => true
```

> Honestly not sure what’s going on with that Appveyor x86 build failure

Sorry for the noise. That's a genuine bug that I only recently learned about after an innocuous change to the Python grammar caused it to start failing our randomized tests. I'm working on a fix.

maxbrunsfeld:
> Lastly (and perhaps controversially), this change introduces new variables available at grammar compile time, for the major, minor, and patch versions of the tree-sitter CLI used to compile the grammar. This will allow grammars to conditionally adopt these new regex features while remaining backward compatible with older versions of the CLI.

That seems  pretty reasonable. Thanks for the explanation!

blankname:
> For unclear reasons, the characters #, *, and 0-9 are marked as `Emoji=Yes` by `emoji-data.txt`. Because of this, a grammar that wishes to use emojis is likely to want to exclude those characters, so this change also adds support for binary operations in regexes, e.g. `[\p{Emoji}&&[^#*0-9]]`.

[Here's](https://github.com/BurntSushi/ripgrep/discussions/1623#discussioncomment-28827) an explanation from the author of ripgrep in regards to why those characters are matched by `\p{Emoji}` (tldr-ish: `\p{Emoji}` matches a single unicode codepoint that can make up an emoji, but a emoji may be made up of multiple codepoints and some are made of a combination of those characters and other code points).

alex-pinkus:
Thanks @blankname for the super interesting link—I figured it might be something like that. As always, burntsushi's explanation makes it very clear.

Since I'm using this to improve my Swift grammar, it's relevant that `swiftc` doesn't recognize identifiers that start with 0️⃣ (it parses them as invalid integer literals):
```swift
let 0️⃣ = "0" // fails with: '�' is not a valid digit in integer literal
let 👻0️⃣ = "spooky-0" // works
```

PR #1662, opened 2022-02-20
Update section-4-syntax-highlighting.md

hickford:
Add link to scope names and colours on GitHub


dundargoc:
I don't get it. Does https://github.com/Alhadis/language-etc/blob/master/samples/lists/scope-previews.nanorc have scope names that treesitter uses or what?

PR #1692, opened 2022-03-26
Fix build and tests on alternative architectures

alebastr:
Fix build and most of the tests on aarch64, ppc64le and s390x (big-endian).

Note: this is a minimal set of changes just to make things compile and tests pass. There are more mismatched types, but none of those break the build or affect the ABI. Case in point: `tags/include/tree_sitter/tags.h` doesn't have a single uint8_t parameter, while there's a plenty of `u8`s in the rust counterpart of the code.

There's one remaining test failure on s390x which I didn't have enough time to debug - #1693


maxbrunsfeld:
Thank you!

Issue #1718, opened 2022-04-24
RFC: Better typescript definition support for checking/completing external rules.


LumaKernel:
## Problem

Like following example, can't check external rules.

```js
// @ts-check
/// <reference path="./node_modules/tree-sitter-cli/dsl.d.ts" />

module.exports = grammar({
  name: 'mylang',
  externals: ($) => [$.comment],

  rules: {
    source_file: ($) => repeat($.comment),
  },
});
```

![image](https://user-images.githubusercontent.com/29811106/164952635-1a02b7ee-c840-4c9e-beff-93eed9209734.png)

## Sketch of idea

```js
// @ts-check
/// <reference path="./node_modules/tree-sitter-cli/dsl.d.ts" />

/** @type any */
const EXTERNAL = Symbol('EXTERNAL RULE');

/** @type typeof grammar */
const grammar0 = (/** @type any */ e) => {
  return grammar({
    ...e,
    rules: Object.fromEntries(Object.entries(e.rules).filter(([, v]) => v !== EXTERNAL)),
  });
};

module.exports = grammar0({
  name: 'mylang',
  extras: ($) => [$.comment],
  externals: (/** @type any */ $) => [$.comment],

  rules: {
    comment: EXTERNAL,

    source_file: ($) => repeat($.comment),
  },
});
```

Allow adding external rules as special specifier, we can check and complete names of these.

You can try completing external sources in your editor which has typescript language server.

![image](https://user-images.githubusercontent.com/29811106/164952712-bea54c20-7c85-4269-aeb0-6c848efce606.png)

By changing typeof `externals` to typeof `extras`, no need to add `/** @type any */`.

## What to do

  I'll make the PR soon.
  
  - Change type def of externals.
  - Add type def for EXTERNAL.
  - Drop EXTERNAL in grammar
    - grammar is currently throwing error for values other than functions, so it is backward-compatible change.
  - Add export for EXTERNAL.

SKalt:
Hey @LumaKernel, what would you think of altering the type definition of `grammar` to be generic over inferred external rule names? I used a similar trick in a typescript port of the tree-sitter DSL that I wrote a while back: https://github.com/SKalt/ts-dsl-tree-sitter/blob/master/src/legacy/dsl.ts#L25

PR #1719, opened 2022-04-24
Add EXTERNAL specifier to check types of them and setup test for type definitinos.

LumaKernel:
Resolves https://github.com/tree-sitter/tree-sitter/issues/1718

Details for decisions:

- Change type def of externals.
- Add type def for EXTERNAL.
- Drop EXTERNAL in grammar
  - grammar is currently throwing error for values other than functions, so it is backward-compatible change.
- Add export for EXTERNAL.

- Setup tests for type definition the project own .
- Setup CI for types as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,11 @@ jobs:
       - name: Run main tests
         run: cargo test
 
+      - name: Run typecheck tests
+        run: |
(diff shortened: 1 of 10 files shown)


LumaKernel:
Test passed.
https://github.com/LumaKernel/tree-sitter/actions/runs/2214504501

PR #1736, opened 2022-05-03
Add support for usage from Swift code

AntonioCandinho:
### :tophat: What is the goal?

Make TreeSitter available on Swift projects using the Swift package manager. 

### :page_facing_up: How is it being implemented?

Since version 3.0, Swift Package Manager supports building [C Language Targets. ](https://github.com/apple/swift-evolution/blob/master/proposals/0038-swiftpm-c-language-targets.md), making it possible to use these projects from Swift with just the addition of a few files. 

I've created a `Package.swift` file that configures a project name and a list of files to compile. I've excluded `lib.c`, avoiding duplicate symbols and `bindings` for `web` and `rust`. Swift Package Manager will automatically collect and compile all the C files in the provided `path`, and will automatically include everything in the `include` directory.

I've also added support for running the swift build in CI for making sure the project compiles. 

### :white_check_mark: How can it be tested?

I've created a [small repo](https://github.com/AntonioCandinho/TreeSitterSwiftExample) showing how the code can be used from Swift code. Please take a look.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,9 @@ jobs:
           RUSTFLAGS="-D warnings"
           cargo build --release
 
+      - name: Build as Swift package
+        run: swift build
+
       - name: Set up fixture parsers
         run: |
           script/fetch-fixtures
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ docs/assets/js/tree-sitter.js
 *.exp
 *.lib
 *.wasm
+
+.build
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,26 @@
+// swift-tools-version: 5.6
+import PackageDescription
+
+let package = Package(
+    name: "TreeSitter",
+    products: [
+        .library(
+            name: "TreeSitter",
+            targets: ["TreeSitter"]
+        ),
+    ],
+    dependencies: [ ],
+    targets: [
+        .target(
+            name: "TreeSitter",
+            dependencies: [],
+            path: "lib",
+            exclude: [
+               "src/lib.c",
+               "binding_web",
+               "binding_rust",
+            ]
+        ),
+    ]
+)
+
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.6
+// swift-tools-version:5.3
 import PackageDescription
 
 let package = Package(
--- a/Package.swift
+++ b/Package.swift
@@ -9,18 +9,20 @@ let package = Package(
             targets: ["TreeSitter"]
         ),
     ],
-    dependencies: [ ],
+    dependencies: [],
     targets: [
         .target(
             name: "TreeSitter",
             dependencies: [],
             path: "lib",
             exclude: [
-               "src/lib.c",
-               "binding_web",
-               "binding_rust",
+                "src/lib.c",
+                "binding_web",
+                "binding_rust",
+            ],
+            cSettings: [
+                .headerSearchPath("./src"),
             ]
         ),
     ]
 )
-


fjtrujy:
Is nice to have `tree-sitter` in Siwft, well done!

fjtrujy (approved):

mattmassicotte:
@AntonioCandinho wow this is amazing! I was *just* about to open an identical PR! I'm planning on using this for https://github.com/chimehq/swifttreesitter

simonbs:
This would be great! I'm doing the exact some thing in [Runestone](https://github.com/simonbs/runestone) but I'm including Tree-sitter as a submodule. This would let me remove the submodule and add Tree-sitter as a dependency using Swift Package Manager.

AntonioCandinho:
> @AntonioCandinho wow this is amazing! I was just about to open an identical PR! I'm planning on using this for https://github.com/chimehq/swifttreesitter
Great job with that project! I had written something quite similar myself, but it's very nice to know there are already some Swift wrappers available. I could probably get rid of a bunch of code just by using your wrapper. Thanks for the good work 🙏

mattmassicotte:
@AntonioCandinho that's great! I hope it's helpful. I'm also working on adding SPM support to the parsers themselves. It's a little more tricky, but not too bad.

mattmassicotte:
@AntonioCandinho oh one other thing I just noticed! Your package requires swift 5.6, but I believe that can be lowered to 5.3. This is minor, but it will improve compatibility with older installs. Might be worth considering.

mattmassicotte:
@maxbrunsfeld hello! Really don't like bothering you, but this has been sitting for a while. Please let us know if there's anything we can do to help move this along! I'm just really excited to get this one in.

maxbrunsfeld:
I'm open to this, but I don't quite understand. Shouldn't this file live in a project like `SwiftTreeSitter`, which also provides idiomatic Swift bindings to the Tree-sitter API?

mattmassicotte:
This is a great question! Today, there are only three ways to use an external dependency from Swift with SPM: local code, a locally-installed binary, or another Swift package. `SwiftTreeSitter` was relying on the locally-installed binary. This meant it was up the user to build and install the runtime. It wasn't ideal, and has proven to be a high bar.

With this PR, you still ultimately just have a C interface exposed to Swift. As you noted, all of the idiomatic stuff lives elsewhere. But, the build and install steps are automated in a way that is totally transparent to SPM users. I wish something like this wasn't necessary. I'd love to see SPM enhanced to make this kind of (somewhat) externally-invasive change unnecessary. But, today, it really does enhance the ergonomics for Swift clients.

maxbrunsfeld:
I would think that `SwiftTreeSitter` should be structured similarly to [py-tree-sitter](https://github.com/tree-sitter/py-tree-sitter). Why not include in that project all of:
1. Swift bindings to the Tree-sitter APIs
2. A git submodule pointing to a specific commit from this repository with which these bindings are known to work perfectly (or equivalently, vendored copies of the Tree-sitter runtime's C source files, whichever is more convenient).
3. This Swift config file.

That way, all of the Swift related concerns would live in one repository.

mattmassicotte:
Actually - after doing a little more research, I'm not 100% sure this is correct. I'm investigating submodules right now and will report back. That would be the ideal set up.

mattmassicotte:
@maxbrunsfeld I really appreciate a) your patience with me here and b) your question.

As it turns out, my understanding of SPM and submodules was incorrect 😰. SPM does indeed support submodules *for dependencies*. But it will never alter the root project. So, when I was experimenting with submodules, my root was SwiftTreeSitter. And SPM was failing to check out submodules. So, I just assumed that it didn't work at all. It did not occur to me to just try it in a real project. But that does actually work. (I feel somewhat better discovering that others have been tripped up by this behavior as well. But only a little.)

I know you've put the work in here, @AntonioCandinho , but with this new information, I don't think a Package.swift really belongs here. I've already updated SwiftTreeSitter to use submodules and verified that it works correctly for client projects. If you want to use the runtime directly, submodules + local package target is the way to go. This is how @simonbs does it, and I'm pretty sure it is the best option.

maxbrunsfeld:
@AntonioCandinho @simonbs Would it make sense for you all to use SwiftTreeSitter, rather than consuming the C API directly?

AntonioCandinho:
> @AntonioCandinho @simonbs Would it make sense for you all to use SwiftTreeSitter, rather than consuming the C API directly?

That's cool for me @maxbrunsfeld! Thank you both, @mattmassicotte and @maxbrunsfeld, for your time 😄 .

maxbrunsfeld:
Thanks for the contribution and discussion @AntonioCandinho.

PR #1738, opened 2022-05-03
Add link to JSON parser in document

rhysd:
I noticed [the JSON parser](https://github.com/tree-sitter/tree-sitter-json) is not included in the list of parsers in document. This PR adds it.


maxbrunsfeld:
Thanks!

PR #1748, opened 2022-05-17
cli: Bump tree-sitter dependency to 0.20.3

jamessan:
tree-sitter/tree-sitter#1504 added the capture_quantifiers API and its
usage in tree-sitter-cli, so the version should express the dependency.


jamessan:
@maxbrunsfeld ping

maxbrunsfeld:
:1+:

PR #1749, opened 2022-05-18
CMake, CPack, CDash support

SamuelMarks:
If you merge this I'll send similar PR's to all the other tree-sitter projects. Then contribute them all to vcpkg and/or similar.


maxbrunsfeld:
I'd prefer to not add all of these files to the project. Could you explain what problem these files solve? Can applications that use CMake just build Tree-sitter via the Makefile?

SamuelMarks:
Big advantages of this PR:
- Cross-platform configure (including package-management);
- Cross-platform build;
- Cross-platform test integration (commented out, but easy to integrate even with `cargo test`);
- Cross-platform test dashboard;
- Cross-platform packaging (multiple for Windows, Linux, macOS, BSD, &etc.).

Additionally IDE support, and easy integration with cross-platform package managers like [`vcpkg`](https://vcpkg.io), [`conan`](https://conan.io), and friends.

In terms of whether these CMake and related files are required to be in https://github.com/tree-sitter/tree-sitter… the answer is no. However, keeping them here will solve a number of issues, namely:
- When new users find this project, they will already be aware of CMake support;
- When `master` changes here your own CI will fail if you forget to update the CMakeLists.txt [correctly].

That last point is especially important, as otherwise my tree can lag behind yours, see, e.g., https://github.com/python-cmake-buildsystem/python-cmake-buildsystem [oh! - actually they're pretty much up-to-date with official trunk now… well at least release versions]

SamuelMarks:
Big advantages of this PR:
- Cross-platform configure (including package-management);
- Cross-platform build;
- Cross-platform test integration (commented out, but easy to integrate even with `cargo test`);
- Cross-platform test dashboard;
- Cross-platform packaging (multiple for Windows, Linux, macOS, BSD, &etc.).

Additionally IDE support, and easy integration with cross-platform package managers like [`vcpkg`](https://vcpkg.io), [`conan`](https://conan.io), and friends.

In terms of whether these CMake and related files are required to be in https://github.com/tree-sitter/tree-sitter… the answer is no. However, keeping them here will solve a number of issues, namely:
- When new users find this project, they will already be aware of CMake support;
- When `master` changes here your own CI will fail if you forget to update the CMakeLists.txt [correctly].

That last point is especially important, as otherwise my tree can lag behind yours, see, e.g., https://github.com/python-cmake-buildsystem/python-cmake-buildsystem

maxbrunsfeld:
Thanks for the explanation.

I don't think I want to take on the additional job of maintaining these files. Since there are a huge number of competing standards for how to build C projects, my preference is to maintain only the most minimal of those (a simple configuration-free Makefile) and allow others to adapt the project to other build systems based on that example.

PR #1759, opened 2022-05-30
Add C API names as document aliases in Rust docs

rhysd:
This PR adds C API names as [`doc(alias)`](https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search). `doc(alias)` defines alias for the API, for example [`chdir` is an alias of `std::env::set_current_dir`](https://doc.rust-lang.org/stable/std/iter/?search=chdir).

By this PR, searching C APIs directly on Rust document site can show the corresponding Rust APIs. Here is a screenshot where `ts_node_type` is searched. It shows `tree_sitter::Node::kind` corresponds to `ts_node_type` in Rust.

screenshot

Searching Rust APIs by C API names easily would be useful for beginners like me.


maxbrunsfeld:
Nice idea.

PR #1783, opened 2022-06-25
Allow empty external tokens during error recovery, if they change the scanner's state

maxbrunsfeld:
Previously, all empty external tokens were ignored when recovering from a syntax error, in order to guarantee forward progress and prevent infinite loops. But this prevented good error recovery in languages like Python, where empty external tokens like `indent` and `dedent` are very important to the code's structure.

This PR relaxes the criteria for allowing an empty external token to be used: they are now allowed only if the external scanner's state changed in the process of parsing them (as is the case for tokens like `indent` and `dedent`). Error recovery is inherently not very well defined, but this change (together with https://github.com/tree-sitter/tree-sitter-python/commit/188b6b062d8cb256e7dfe76b5ad5089bbdcb7014) certainly leads to better behavior in Python, and I think the same might apply in other languages. Many external tokens will still be ignored during error recovery.

### DIFF
--- a/cli/src/tests/helpers/scope_sequence.rs
+++ b/cli/src/tests/helpers/scope_sequence.rs
@@ -58,7 +58,7 @@ impl ScopeSequence {
         let mut position = Point { row: 0, column: 0 };
         for (i, stack) in self.0.iter().enumerate() {
             let other_stack = &other.0[i];
-            if *stack != *other_stack {
+            if *stack != *other_stack && ![b'\r', b'\n'].contains(&text[i]) {
                 let containing_range = known_changed_ranges
                     .iter()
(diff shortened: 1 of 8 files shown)


razzeee:
This seems to cause a freeze downstream and some changes to the parsed tree in some places (probably expercted, still not sure how to get the old behavior again)

https://github.com/elm-tooling/tree-sitter-elm/issues/126

maxbrunsfeld:
If there's an infinite loop, I believe it means that tree-sitter-elm's external scanner is returning an empty token during error recovery. You can detect error recovery in the external scanner by the fact that *all* tokens are considered valid at once.

Issue #1818, opened 2022-07-26
ts_query_new fails for large grammars / production_id > 255


slackner:
**TLDR:**

`AnalysisSubgraphNode` uses an 8-bit value for `production_id`, but this limit is too small for large grammars. In my case, it caused `ts_query__analyze_patterns` to fail unexpectedly.

**Background:**

I was just using tree-sitter to write / extend a parser, and at first everything was working fine. Then I added just one more `field()` in grammar.js, and `ts_query_new()` suddenly failed with error_type 5 (`TSQueryErrorStructure`).

Some debugging releaved that `ts_query__analyze_patterns` thought the pattern can no longer match (i.e., `can_finish_pattern` was false). After some more debugging, I noticed that apparently `ts_query__analyze_patterns` did not find all production rules, and seemed to skip some of them. The first being 256.

Then it clicked, and I checked the data types used throughout the code. And indeed, `AnalysisSubgraphNode` uses an 8-bit value to store the `production_id`. Other parts of the code use larger data types.

I quickly did the following change, and it indeed seems to fix all the issues.

```diff
diff --git a/lib/src/query.c b/lib/src/query.c
index 1033f1c3..a19df484 100644
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -242,7 +242,7 @@ typedef Array(AnalysisState *) AnalysisStatePool;
  */
 typedef struct {
   TSStateId state;
-  uint8_t production_id;
+  uint16_t production_id;
   uint8_t child_index: 7;
   bool done: 1;
 } AnalysisSubgraphNode;
```

maxbrunsfeld:
Yes, using `uint8_t` there was definitely a mistake; production IDs are represented as `uint16_t` in other places in the codebase. Nice debugging.

Want to open a PR with that patch?

slackner:
Thanks for the confirmation! I've opened a merge request with the diff above.

Note that this makes the struct packing unfortunately a bit less efficient. The values no longer fit into 4 bytes, and the compiler will probably add some padding afterwards...

maxbrunsfeld:
Yeah, it's probably not worth using bitfields for `child_index` and `done` anymore. Oh well.

PR #1819, opened 2022-07-26
query: Use uint16_t for production_id in AnalysisSubgraphNode struct

slackner:
Production IDs are represented as `uint16_t` in other places in the codebase.

Fixes https://github.com/tree-sitter/tree-sitter/issues/1818


maxbrunsfeld:
Thank you!

Issue #1737, opened 2022-05-03
Native Apple Silicon (M1, .etc) support?


imWildCat:
As of now, we only have x86 build for M1 Macs: 

```js
// ARM macs can run x64 binaries via Rosetta. Rely on that for now.
if (platformName === 'macos' && process.arch === 'arm64') {
  archName = 'x64';
}
```

Commit: <https://github.com/tree-sitter/tree-sitter/issues/942>

Any reason?

typesanitizer:
I don't know the reason but I suspect it is because GitHub Actions don't have M1 Mac runners. Cross-compiling from x86_64 to arm64 is generally more tedious with C code IME. Running the tests on the arm64 binary would also require QEMU or similar.

imWildCat:
Thanks!

Did we try something like https://github.com/messense/cargo-zigbuild ?

If not, I can start to try it!

typesanitizer:
I don't know (I just saw this issue as I was passing by) but that would only resolve the build aspect (which might be OK, I'm not sure if that was the main blocker or if the ability to run tests was).

amaanq:
I'm gonna safely close this as it seems to work now in our CI and works on my Mac

Issue #1732, opened 2022-05-01
Fails to build on ubuntu 20.04.4


xloem:
Build log: https://dweb.link/ipfs/bafybeifnhfeoo6lj4pnk7xudqf7m3k5gxzgvhghc7n2vzajwdkqtctrfmy/make.log

gcc version: 7.5.0

ahelwer:
I think this is because of -wall -wextra -werror

frexsdev:
> I think this is because of -wall -wextra -werror

NO! It is for error reporting. When you don't know, don't say anything plz

frexsdev:
What is the github of the project ?

xloem:
@frexsdev YOUR MESSAGE WAS SENT ON THE GITHUB OF THE PROJECT.

dundargoc:
Is this still a problem? If so we need more information. The provided link also does not work.

xloem:
I am no longer using either ubuntu or this project at this time, but here is more conventional ipfs.io link to the build log: https://ipfs.io/ipfs/bafybeifnhfeoo6lj4pnk7xudqf7m3k5gxzgvhghc7n2vzajwdkqtctrfmy/make.log . It's just an ipfs file; it's still on ipfs.

xloem:
Here's the content. Do IPFS links from other people work for you?
```

cc -O3 -Wall -Wextra -Werror -std=gnu99 -fPIC -Ilib/src -Ilib/include   -c -o lib/src/subtree.o lib/src/subtree.c
lib/src/subtree.c: In function ‘ts_subtree_string’:
lib/src/subtree.c:852:25: error: ‘(NULL)’ directive output truncated writing 6 bytes into a region of size 1 [-Werror=format-truncation=]
   if (!self.ptr) return snprintf(string, limit, "(NULL)");
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/stdio.h:867:0,
                 from lib/src/subtree.c:6:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:67:10: note: ‘__builtin_snprintf’ output 7 bytes into a destination of size 1
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

Issue #1686, opened 2022-03-14
Allow configurable toolchain binaries


sluongng:
Currently tree-sitter infer the `node` binary location using of the PATH environment variable.
However in Build System(such as Bazel) who are (1) strict in how environment variable is managed and (2) manage their own NodeJS toolchain (i.e. Bazel's rules_nodejs), separate from the host, there may be a need to bring your own NodeJS binary.

Allowing passing the absolute path of the nodejs library through a flag of `tree-sitter` CLI would be much desirable for this use case.

Taking a look at current implementation, I see there are other tools which might be needed as well https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/tree-sitter/tree-sitter%24+Command::new&patternType=literal (docker, dot, id, emcc...). I can imagine we would also want to do the same for these toolchains in the future.

If it's not possible/too hard to use a flag, I can make do with a specific environment variable (i.e. `TREE_SITTER_NODE_PATH`) so that I avoid having to override PATH.

Let me know what you think.

jakesarjeant:
FYI, I've implemented a (partial) solution to this in #2403, specifically the ability to specify a custom Node.js binary.

PR #1821, opened 2022-07-27
feat: use cmake for cross-platform support

kylo252:
- use cmake by default to build the core library to support all platforms
- if cmake is not found, then fallback to using cc directly
- pkg-config is supported as well as find_package

fixes: #1366 #1518 #1686 #1737 #1732 #1262 #1246
related: #1555 #1749


sogaiu:
May be related?

* https://github.com/tree-sitter/tree-sitter/pull/1749
* https://github.com/tree-sitter/tree-sitter/pull/1555

kylo252:
@sogaiu, thanks for the reminder!

the main difference is that this actually lets cargo invoke cmake, which solves various problems when building tree-sitter-cli as well.

another benefit is removing the need for a few files:
- [`Makefile`](https://github.com/tree-sitter/tree-sitter/pull/1821/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52): superseded by `CMakeLists.txt`
-  [`.ccls`](https://github.com/tree-sitter/tree-sitter/pull/1821/files#diff-a1cf5f28b1bd6b58d29cf8bbde170fa2609eec01235628da57d9501c70ff6bc9) and [`compile_flags.txt`](https://github.com/tree-sitter/tree-sitter/pull/1821/files#diff-2aa54b3f37d506cf422efaa520bf70c9cd4573b6e51b222385641f263392e0ab): superseded by `compile_commands.json` which is generated by CMake

aminya:
Well, I can add `cmake` to build.rs in #1555 and remove the makefile. But I don't think the maintainers are open to such changes. They have not even reviewed #1555 after one year.

RichardHyde:
I used the [kyloi252/tree-sitter](https://github.com/kylo252/tree-sitter/tree/cmake-refactor/cli) and was able to successfully build tree-sitter cli on a Raspberry Pi 4 with Raspberry PI OS 64-bit if that's any help to anyone.

PR #1823, opened 2022-07-27
Fix crash in ts_node_field_name_for_child for inline nodes

slackner:
The function only worked when `self.data.is_inline` is false.


maxbrunsfeld:
Thanks for the fix!

PR #1831, opened 2022-08-02
Add env vars for port and address

shaded-enmity:
Make it possible to pass port and address via pair of environmental variables `TREE_SITTER_PLAYGROUND_PORT` and `TREE_SITTER_PLAYGROUND_ADDR`, respectively.

What would be the best place where to document this


maxbrunsfeld:
Thank you!

PR #1835, opened 2022-08-06
fix: fix detection of the compiler on windows (allow Clang + MinGW)

aminya:
Fixes a part of #1889

The previous detection assumed that the compiler on Windows is MSVC, which is not correct all the time.

Also, related to #1411

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -377,7 +377,8 @@ impl Loader {
                 command.env(key, value);
             }
 
-            if cfg!(windows) {
+            let compiler = config.get_compiler();
+            if compiler.is_like_msvc() {
                 command.args(&["/nologo", "/LD", "/I"]).arg(header_path);
                 if self.debug_build {
                     command.arg("/Od");
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -395,13 +395,16 @@ impl Loader {
             } else {
                 command
                     .arg("-shared")
-                    .arg("-fPIC")
                     .arg("-fno-exceptions")
                     .arg("-g")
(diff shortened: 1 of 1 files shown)


sogaiu:
@aminya Thanks for this fix.  I was compiling on Windows using gcc with msys / mingw64 and this was just what was needed :+1: 

On a side note, it seems like the instance of:
```rust
 let compiler = config.get_compiler();
```
on line 376 is redundant, given what's on line 370 above.

I left line 376 out here and it seemed to work ok.

cli/loader/src/lib.rs
```diff
@@ -373,7 +373,8 @@ impl Loader {
                 command.env(key, value);
             }
 
-            if cfg!(windows) {
+            let compiler = config.get_compiler();
```
aminya, line 376:
```suggestion
```

aminya:
You're welcome!

> on line 376 is redundant, given what's on line 370 above.

Fixed in [`405f2ef` (#1835)](https://github.com/tree-sitter/tree-sitter/pull/1835/commits/405f2ef1185561f4a71ca829558b71f9e15b32f2)

sogaiu:
@ahlinc Do you think it would be practical / reasonable to merge this?

It would be great to be able to use clang and gcc via `tree-sitter` on Windows.  I'm not sure there is another good way without resorting to WSL-ish or Docker-ish kinds of things (which aren't so great for certain situations).

ahlinc:
@sogaiu yes, I'll merge this in a couple of days.

amaanq:
@aminya can you rebase? Mainly just want the new CI to run on this and then I'll merge

aminya:
@amaanq rebased the branch! Thanks

PR #1836, opened 2022-08-10
Avoid extracting default alias for extras

### LINKED ISSUE #1834: Alias of an extra affects all appearances of the extra
Aliasing an extra in any rule in the grammar (even an unused rule) changes the name of the extra everywhere it is reported. My understanding is that an alias is meant to apply locally, to only the part of the grammar where it is used, and not to make any global changes to the name of the rule being aliased.

I came across this because I am addressing https://github.com/returntocorp/ocaml-tree-sitter-core/issues/36. One of the extras in the Swift grammar also appears in another rule. Currently, ocaml-tree-sitter-core cannot handle this, because to construct a typed CST it needs to discard extras first. I am adding a pass which rewrites the grammar to alias any extras that appear in ordinary rules so ocaml-tree-sitter-core can distinguish those extras from others. Unfortunately, due to this issue, the alias causes all instances of that extra to be renamed, so it's still impossible to tell which extras need to be discarded.

Here is a concrete example:

`grammar.js`:

```
module.exports = grammar({
  name: "extra_test",
  extras: $ => [
    $.comma,
    "\n"
  ],
  rules: {
    program: $ => "foo",
    comma: $ => ",",
    unused: $ => alias($.comma, $.comma_other),
  }
});
```

`testfile`:
```
,foo,
```

Output of `tree-sitter parse testfile`:

```
(program [0, 0] - [1, 0]
  (comma_other [0, 0] - [0, 1])
  (comma_other [0, 4] - [0, 5]))
```

Expected output of `tree-sitter parse testfile`:

```
(program [0, 0] - [1, 0]
  (comma [0, 0] - [0, 1])
  (comma [0, 4] - [0, 5]))
```

nmote:
Tree sitter extracts default aliases for rules that only appear in aliases. The reasoning is explained in the comments in `extract_default_aliases.rs`. The default alias is used in place of the original name during parsing. However, the logic to determine whether it appears only in aliases overlooks the fact that extras may appear in the tree despite not being mentioned in any rules.

This adds an additional step which traverses the list of extras and notes that they all appear unaliased. This prevents the extraction of a default alias for such rules.

The code to get a reference to the symbol status was copy/pasted from above. I tried pulling it out into a closure or helper function but quickly got bogged down by the borrow checker. I'm not sure if there's an idiomatic way to do something like this, but if there is, it didn't seem like it was going to result in clearer code than just copying and pasting this small block.

Fixes #1834

### DIFF
--- a/cli/src/generate/prepare_grammar/extract_default_aliases.rs
+++ b/cli/src/generate/prepare_grammar/extract_default_aliases.rs
@@ -62,6 +62,18 @@ pub(super) fn extract_default_aliases(
         }
     }
 
+    for symbol in syntax_grammar.extra_symbols.iter() {
+        let mut status = match symbol.kind {
+            SymbolType::External => &mut external_status_list[symbol.index],
+            SymbolType::NonTerminal => &mut non_terminal_status_list[symbol.index],
+            SymbolType::Terminal => &mut terminal_status_list[symbol.index],
+            SymbolType::End | SymbolType::EndOfNonTerminalExtra => {
+                panic!("Unexpected end token")
+            }
+        };
+        status.appears_unaliased = true;
+    }
+
     let symbols_with_statuses = (terminal_status_list
         .iter_mut()
         .enumerate()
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Nice work, thanks so much for the fix.

PR #1846, opened 2022-08-19
feat(test): test all queries

### LINKED ISSUE #1790: Testing multiple queries
Hi,

When developping a tree-sitter parser for [viml](https://github.com/vigoux/tree-sitter-viml) (especially https://github.com/vigoux/tree-sitter-viml/pull/107), I wanted to start testing mutiple queries in a format similar to highligh tests.

To do this I had to come up with [this ugly hack](https://github.com/vigoux/tree-sitter-viml/commit/4e1f656a69310a1603b78089f8ce83d33801a4b5) and I wonder if there is a way to make the testing a bit more elegant, maybe by specifying queries in a directory, and tests machting the name of a given query in another directory.

If this is not planned, I would be glad to implement it as, from what I see, the code should be pretty much ready for something like this.

vigoux:
Fallback to default testing for all queries present in the parser'squeries directory.

For a given query <QUERY>.scm, the test files are searched in test/<QUERY>/*

Also mimic the output of other test-running subcommands when testing queries.

Fixes #1790

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -405,6 +405,45 @@ fn run() -> Result<()> {
             if test_tag_dir.is_dir() {
                 test_tags::test_tags(&loader, &test_tag_dir)?;
             }
+
+            // For the rest of the queries, find their tests and run them
+            for entry in walkdir::WalkDir::new(&current_dir.join("queries"))
+                .into_iter()
+                .filter_map(|e| e.ok())
+                .filter(|e| e.file_type().is_file())
+            {
+                let stem = entry
+                    .path()
+                    .file_stem()
+                    .map(|s| s.to_str().unwrap_or(""))
+                    .unwrap_or("");
+                if stem != "highlights" && stem != "tags" {
+                    println!("{} query:", stem);
+
+                    let paths = walkdir::WalkDir::new(&test_dir.join(stem))
+                        .into_iter()
+                        .filter_map(|e| {
+                            let entry = e.ok()?;
+                            if entry.file_type().is_file() {
+                                Some(String::from(entry.path().to_string_lossy()))
+                            } else {
+                                None
+                            }
+                        })
+                        .collect::<Vec<String>>();
+                    query::query_files_at_paths(
+                        *language,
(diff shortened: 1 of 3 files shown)


vigoux:
Just realized I did not test that properly and it does not quite work yet
Going to test that better.

vigoux:
This now also detects when an assertion fails to match.

vigoux:
I fixed a little problem related to nodes and captures spanning on multiple lines.

An assertion above an empty (or rather, too short) line will match the line above it, for example the following would work (if `foo` is captured by `bar`):
```vim
foo

" ^ bar
```

```vim
      foo
foobar
"     ^ bar
```

maxbrunsfeld:
Thanks for working on this! Could you write up a little doc about how the folder is intended to be structured, and how the tests are run?

vigoux:
Basically the directory structure is like the following (taken from viml parser):
```
queries/
├── highlights.scm
└── injections.scm
test/
├── highlight/
│   ├── comment.vim
│   ├── loops.vim
│   └── statements.vim
└── injections/
    ├── lua.vim
    ├── python.vim
    └── ruby.vim
```

The idea is that `tree-sitter` searches for test cases in the directory under `test` with the same basename than the query.

vigoux:
Do you want me to add anything more than what is currently in this PR ? Like docs, or anything ?

vigoux:
Any update on this ?

glfmn:
Any progress on this feature?  I am using the `test.py` script from the viml parser as a stopgap.

dundargoc:
@vigoux rebase you nerd so I can merge this

clason:
Bumping -- in fact, what would already be **very** useful is just to run the queries against the grammar to check that they're valid (i.e., only refer to nodes in the grammar and do not contain invalid patterns). This would allow grammar authors to run PR verifying that changes to the grammar are not breaking for query writers (or mark commits as such -- important for semantic versioning). 

(Basically, it would be enough to wire up `ts_query_new` from the API and process and pretty-print the output.)

PR #1849, opened 2022-08-21
docs: Added documentation on web-tree-sitter edge cases

NicholasLYang:
Added documentation on two annoying issues I had while trying to get web-tree-sitter working with a modern web stack, i.e. bundlers and server side rendering.


maxbrunsfeld:
Thanks for taking the time to write this up. Some day, I'd love to embrace any standard practices that will make bundlers 'just work' with our approach, but for now, this will probably help people problem-solve.

Issue #1628, opened 2022-01-30
Push `tree-sitter` version 0.20.4 to npm


alex-pinkus:
The [`npm` package `tree-sitter`](https://www.npmjs.com/package/tree-sitter) doesn't yet have an ABI-14-compatible version published--its latest version is 0.20.0. If, let's say, [someone were to publish](https://github.com/alex-pinkus/tree-sitter-swift/pull/104) a grammar that's explicitly compiled with `--abi 14`, anyone trying to depend on it from NPM would get a message like:

```
/home/runner/work/tree-sitter-swift/tree-sitter-swift/test-npm-package/node_modules/tree-sitter/index.js:259
  setLanguage.call(this, language);
              ^

RangeError: Incompatible language version. Compatible range: 13 - 13. Got: 14
    at Parser.setLanguage (/home/runner/work/tree-sitter-swift/tree-sitter-swift/test-npm-package/node_modules/tree-sitter/index.js:259:15)
    at Object.<anonymous> (/home/runner/work/tree-sitter-swift/tree-sitter-swift/test-npm-package/index.js:5:8)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
```

There has been a `0.20.4` published for `tree-sitter-cli`, but unless I'm missing something, we need to update the library too.

ahelwer:
Ref https://github.com/tree-sitter/tree-sitter/issues/1223

verhovsky:
This should be fixed by node-tree-sitter (simply "tree-sitter" on npm) version 0.20.3

https://github.com/tree-sitter/node-tree-sitter/pull/146

PR #1852, opened 2022-09-02
Generate parsers with ABI version 14 by default

maxbrunsfeld:
This makes `tree-sitter generate` use the newest ABI version by default. This will include the new `primary_state_ids` field introduced in https://github.com/tree-sitter/tree-sitter/pull/1589, which significantly speeds up construction of queries in some languages.

I had initially held off on bumping the generated ABI version, because I didn't want generated parsers to become incompatible with applications using older versions of the library. But the newer version of the library supporting ABI version 14 has been out for several months now.

Note that you can still do `tree-sitter generate --abi=13` if you need to regenerate a parser, but keep it compatible with an older version of the library.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -12,7 +12,7 @@ use tree_sitter_loader as loader;
 
 const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 const BUILD_SHA: Option<&'static str> = option_env!("BUILD_SHA");
-const DEFAULT_GENERATE_ABI_VERSION: usize = 13;
+const DEFAULT_GENERATE_ABI_VERSION: usize = 14;
 
 fn main() {
     let result = run();


dcreager (approved):

alex-pinkus:
There's an unfortunate side effect to this since the [npm `tree-sitter` package](https://www.npmjs.com/package/tree-sitter) hasn't been updated in a while: anyone trying to use this from `node-tree-sitter` is broken the next time the parser they use is regenerated. I just tested it to be absolutely certain and got:

```
/Users/alexpinkus/code/tree-sitter-swift/test-npm-package/node_modules/tree-sitter/index.js:259
  setLanguage.call(this, language);
              ^

RangeError: Incompatible language version. Compatible range: 13 - 13. Got: 14
```

@maxbrunsfeld, what do you think it would take to publish a new version of `node-tree-sitter` to NPM?

Ref: https://github.com/tree-sitter/node-tree-sitter/issues/111, #1628

Issue #949, opened 2021-02-27
Problem with statically predefined symbols in `binding_web/exports.json` for wasm language files


ahlinc:
Hello :wave: 

I've stumbled over situation that a predefined list of symbols in [lib/binding_web/exports.json](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/exports.json) isn't enough for my case when I worked on external scanner implementation.

Here is a list of all available library functions that can be used in `scanner.cc` for languages in `wasm` files without patches to the `exports.json` file:
```bash
$ curl -s https://raw.githubusercontent.com/tree-sitter/tree-sitter/master/lib/binding_web/exports.json | jq -r '.[]|sub("^_";"")' | c++filt
calloc
free
malloc
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::copy(char*, unsigned long, unsigned long) const
std::__2::__vector_base_common<true>::__throw_length_error() const
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::__init(char const*, unsigned long)
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::reserve(unsigned long)
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::__grow_by(unsigned long, unsigned long, unsigned long, unsigned long, unsigned long, unsigned long)
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::push_back(char)
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::basic_string(std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> > const&)
std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char> >::~basic_string()
std::__2::basic_string<wchar_t, std::__2::char_traits<wchar_t>, std::__2::allocator<wchar_t> >::push_back(wchar_t)
std::__2::basic_string<wchar_t, std::__2::char_traits<wchar_t>, std::__2::allocator<wchar_t> >::~basic_string()
operator delete(void*)
operator new(unsigned long)
abort
iswalnum
iswalpha
iswdigit
iswlower
iswspace
memchr
memcmp
memcpy
strlen
towupper
abort
.....................................
```

I found that `emsdk` is shipped with `llvm-objdump` that is able to extract required symbols from a wasm library:

[...]

PR #1864, opened 2022-09-08
Add optional WASM feature to the native library, allowing it to run wasm-compiled parsers via wasmtime

maxbrunsfeld:
### Background

Currently, Tree-sitter parsers can be compiled to WebAssembly (aka 'WASM') and run within a web browser, via the `web-tree-sitter` JavaScript library, which contains a WASM build of the Tree-sitter library.

In some applications that use the *native* Tree-sitter library, it would also be very useful to be able to load these same WASM builds of parsers, in order to support adding parsers as 🔌  **plugins** 🔌 , without requiring users to download platform-specific native binaries or to compile C code on their own machines.

### Change

This PR adds a new optional WASM feature to the core library, which can be enabled in Rust via the `wasm` cargo feature, and in C via the `TREE_SITTER_FEATURE_WASM` macro.

This feature allows you to build a native `TSLanguage` object from a WASM buffer. You can then use this language object just like any other (native-compiled) language object: parsing text that lives in native memory, constructing syntax trees on the native heap, sending it between threads, etc. Tree-sitter languages are *mostly* just plain immutable data, so they're easy to unmarshal from a compiled wasm module.

The only difference is that when using a wasm-based language with a `TSParser`, you must first provide the parser with a  `TSWasmStore`. This wasm store object wraps a `wasmtime::Store` object, which is used by the Tree-sitter library to invoke the language's lexing functions which are code, not data, so they require a WASM runtime to execute.

### Notes

**Wasmtime Dependency** - I originally thought that it'd be cool to code against WebAssembly's standard [`wasm.h` C interface](https://github.com/WebAssembly/wasm-c-api/blob/master/include/wasm.h), which is supposedly implemented by multiple runtimes (mainly wasmtime and V8). This way, applications using Tree-sitter would have multiple choices of which WASM runtime to link Tree-sitter against.

As I got into the details, I ended finding that some of the C APIs that I needed were currently *unimplemented* in wasmtime, but that Wasmtime provides its own custom C interface which *is* fully implemented, and which is supposedly better-designed from a performance perspective. So I ended up using Wasmtime's own C API. This means you can't use V8 as the wasm runtime when using this Tree-sitter feature, at least for now.

### Tasks

* build
    * [x] Open `wasmtime` PR for tweaks to the `wasmtime-c-api` `Cargo.toml`
* library
    * [x] Support parsers without external scanners
    * [x] Support parsers whose external scanners don't use `libc` or `libc++`
    * [x] Support parsers that use libc and libc++ by bundling a small WASM module exporting all of the same stdlib symbols that as `web-tree-sitter`
    * [x] Generalize logic for copying parse tables out of wasm memory, removing hard-coded stuff
    * [x] Make sure wasm stores work correctly when loading multiple languages
    * [x] Make error handling robust
       * return an informative error from `ts_wasm_store_load_language`
* cli
    * [x] Add a `--wasm` flag to the `tree-sitter parse` CLI command, which causes any parsers to be compiled to WASM instead of native shared libraries, and loaded via the new logic.
    * [x] Add a `--wasm` flag to the `tree-sitter test` command that works the same way
    * [x] Exercise this new behavior in the test suite. Maybe just run the entire test suite in both WASM and non-WASM modes?


lambdadog:
> - [ ] Generalize logic for copying parse tables out of wasm memory, removing hard-coded stuff

I haven't had a chance to read through all of your changes yet, but I had been hacking on this a bit prior to you posting this PR. Could you point me to where you've implemented the parse table lifting?

RubixDev:
>   * [ ]  Open `wasmtime` PR for tweaks to the `wasmtime-c-api` `Cargo.toml`

I don't exactly understand why the Rust dependency on `wasmtime-c-api` is necessary. The only place it is used is the `test()` function in `lib/binding_rust/wasm_language.rs` which appears to be unused, and even that just seems to call `wasmtime::Engine::default()` internally. So is it really necessary to create a PR for allowing use of that crate from Rust?

RubixDev:
Okay it seems I completely missed the point. Depending on the C API from Rust allows the tree-sitter C lib to be compiled and linked easily without having to manually vendor wasmtime. I created a PR on wasmtime here: https://github.com/bytecodealliance/wasmtime/pull/6765.

maxbrunsfeld:
Thanks for doing that @RubixDev. This branch is now using the upstream version of wasmtime.

maxbrunsfeld:
@ahlinc I'm getting strange errors on CI, trying to build for the mipsel targets:

```
  $BUILD_CMD test --target=mips-unknown-linux-gnu
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    CARGO_TERM_COLOR: always
    RUSTFLAGS: -D warnings
    CROSS_DEBUG: 1
    BUILD_CMD: cross
    EMSCRIPTEN_VERSION: 3.1.37
    EXE: 
    CARGO_INCREMENTAL: 0
    CROSS_IMAGE: ghcr.io/cross-rs/mips-unknown-linux-gnu:custom
    CROSS_CONFIG: /home/runner/work/tree-sitter/Cross.toml
    TREE_SITTER: tree-sitter.sh
    TARGET: mips-unknown-linux-gnu
    ROOT: /home/runner/work/tree-sitter/tree-sitter
    CC: mips-linux-gnu-gcc
    AR: mips-linux-gnu-ar
    CFLAGS: -Werror
    CROSS_RUNNER: /qemu-runner mips
   Compiling libc v0.2.149
   Compiling proc-macro2 v1.0.69
   Compiling unicode-ident v1.0.12
   Compiling quote v1.0.33
   Compiling syn v2.0.38
   Compiling cfg-if v1.0.0
   Compiling serde v1.0.190
   Compiling memchr v2.6.4
   Compiling aho-corasick v1.1.2
   Compiling cc v1.0.83
   Compiling regex-syntax v0.8.2
   Compiling bitflags v2.4.1
   Compiling regex-automata v0.4.3
   Compiling tree-sitter v0.20.10 (/project/lib)
error: Undefined temporary symbol $BB120_1

error: Undefined temporary symbol $BB121_1

error: Undefined temporary symbol $BB122_1

error: could not compile `regex-automata` (lib) due to 3 previous errors
```

It seems like MIPS isn't a Tier-1 target for Rust. Rustc has [known issues](https://github.com/rust-lang/rust/issues/108835) compiling other popular crates. Do we need to build for these targets on every branch?

ahlinc:
@maxbrunsfeld thank you for the notice, I'll reorganize CI and by default disable all non widely used targets. While it was beneficial to compile and test against so many targets and fixing related bugs I see that it's impractical to build for all of them on a daily basis development.

Nameitsattari (approved):

maxbrunsfeld:
Ok, this experimental feature is complete enough for people to try using in downstream applications. There will probably be bugs, but the feature basically works.

panekj:
Thanks a lot for the work! We will be happy to test it in [Lapce](https://github.com/lapce/lapce) (hopefully soon)

clason:
Huge step towards a truly platform-independent parser ecosystem. Out of interest (and I apologize if it's obvious), how does this relate to https://github.com/tree-sitter/tree-sitter/issues/949?

maxbrunsfeld:
> Out of interest (and I apologize if it's obvious), how does this relate to https://github.com/tree-sitter/tree-sitter/issues/949?

Yeah, that's a good question. That same limitation applies when using wasm-compiled parsers in this mode: there is a fixed (small) subset of the C and C++ standard libraries available that is compiled into the library, and any external scanners that rely on symbols that aren't in this subset will not work.

There may be some way to change how parsers are compiled to wasm, such that they can include code that they need from the standard library, but I don't know how to do that with Emscripten (or Clang) thus far. I feel like this design, though somewhat limiting, mostly works in practice.

But as part of stabilizing this feature, it would probably be good to add some tooling around detecting when external scanners use functions that are unavailable in a wasm context, and emitting warnings.

clason:
Ah, pity. That is a dealbreaker for Neovim (as the Markdown parser is now required infrastructure, and many parsers in nvim-treesitter have a scanner.c).

But still, great feature!

maxbrunsfeld:
Also, two big limitations to call out:

* Right now, there's no way to *delete* a `Language`, because they previously consisted only of static data. Now, with wasm-based languages, they are allocated dynamically at runtime. I plan to add an API for *deleting* a Wasm-based language, so that you can dynamically reload a given language after changing its wasm (enabling a hot-reloading workflow for parsers). That'll happen in a future PR.
* There's not yet any docs for how to use this feature.
* I haven't used this feature in an actual editor yet, so I repeat, there are probably bugs.

maxbrunsfeld:
> Ah, pity. That is a dealbreaker for Neovim (as the Markdown parser is now required infrastructure, and many parsers in nvim-treesitter have a scanner.c).

Just to be clear, parsers with a `scanner.c` will still work fine, as long as the scanner only uses certain functions from the C and C++ standard libraries. If there is some critical function that the Markdown scanner uses that I haven't thought to include, it's pretty easy to add it to the core.

PyaePhyo8612:
🐍🕊🤴

savetheclocktower:
> But as part of stabilizing this feature, it would probably be good to add some tooling around detecting when external scanners use functions that are unavailable in a wasm context, and emitting warnings.

This would be fantastic. We've tried to do so in Pulsar by [adding a runtime check](https://github.com/pulsar-edit/pulsar/blob/master/vendor/web-tree-sitter/README.md), but that requires a user to hit the specific code path first, and serves only to present a more understandable error message. If there were (or if there _is_) a way to analyze a `wasm` file and know which externals it wants, that'd be a major improvement.

clason:
Another question, just to be sure: If this mode is enabled, will you still be able to use native parsers (i.e., support both `.so` and `.wasm` in an editor)?

maxbrunsfeld:
Yes, you can still use the native parsers.

icp1994:
Can (should?) the latest tree-sitter release be built against a published `wasmtime` version instead of git rev?

maxbrunsfeld:
The challenge right now is that `wasmtime-c-api` is not available on crates.io. I'm trying to get that resolved (https://github.com/bytecodealliance/wasmtime/pull/7837), but if there are blockers, I may need to work around it by temporarily removing the `wasm` feature from the version that we publish to crates.io, or vendoring the `wasmtime-c-api` sources somehow.

PR #1876, opened 2022-09-18
Add tree-sitter-gleam to list of available parsers

J3RN:
Hello!  I'm the original author and maintainer of [tree-sitter-gleam](https://github.com/gleam-lang/tree-sitter-gleam).  At this stage it's relatively battle tested as it's used by GitHub for Gleam syntax highlighting as well as by the [official gleam-mode for Emacs](https://github.com/gleam-lang/gleam-mode), so I would like to add it to the list of "Available Parsers" in the docs.


J3RN:
@maxbrunsfeld Mind giving this a quick look?

PR #1879, opened 2022-09-21
change integer type in ts_parser__set_cached_token

mattmassicotte:
The use of `size_t` here causes a warning. Pretty much everywhere else, byte offsets are explicitly `uint32_t`, so this seems like the better type choice here.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -605,7 +605,7 @@ static Subtree ts_parser__get_cached_token(
 
 static void ts_parser__set_cached_token(
   TSParser *self,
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks!

Earlycrusher (approved):

however23:
> The use of `size_t` here causes a warning. Pretty much everywhere else, byte offsets are explicitly `uint32_t`, so this seems like the better type choice here.

however23:
> Thanks!

Welcome

Issue #1880, opened 2022-09-21
A number of integer type mismatches are producing warnings


mattmassicotte:
On my machine, I see a number of warnings about integer sizings. I've kinda just been living with them, but today I had a little time, and started looking.

`Array.size` and `Array.capacity` are defined as `uint32_t`, but frequently interact with `size_t`, which will not be 32 bits on all platforms.

Because of the underlying allocator implementation, I was tempted to change the `Array` definition to use `size_t`. This is potentially the "most correct" change, but it introduces a lot of knock-on warnings that would have to be addressed. It also change the in-memory layout in a way that I wasn't sure was safe.

Next, the query system performs a number of pointer arithmetic operations with explicit `uint32_t`. That also will cause issues on platforms with non-32 bit pointers. This is what `ptrdiff_t` and `intptr_t` were introduced to address.

I was particularly hesitant to make changes here, because I always mess this stuff up.

So, I decided to just open up a [PR](https://github.com/tree-sitter/tree-sitter/pull/1879) for the most straightforward fix for one of the warnings. I thought we could use this issue to discuss the other issues. The least work is just to add explicit casts. But, these warnings are here for a reason - the existing types allow potentially incorrect truncation.

maxbrunsfeld:
What compiler and flags are you using to see these warnings?

mattmassicotte:
This come up when building with clang, targeting arm64 (and x86_64 too I'm sure). The flags being used are:

```
-Wno-trigraphs -fpascal-strings -O0 -Wno-missing-field-initializers -Wno-missing-prototypes -Wno-return-type -Wno-missing-braces -Wparentheses -Wswitch -Wno-unused-function -Wno-unused-label -Wno-unused-parameter -Wno-unused-variable -Wunused-value -Wno-empty-body -Wno-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wno-constant-conversion -Wno-int-conversion -Wno-bool-conversion -Wno-enum-conversion -Wno-float-conversion -Wno-non-literal-null-conversion -Wno-objc-literal-conversion -Wshorten-64-to-32 -Wpointer-sign -Wno-newline-eof -Wno-implicit-fallthrough
```

The flag that catches these particular errors is `-Wshorten-64-to-32`. Adding this to the Makefile flags will catch it.

[...]

PR #1881, opened 2022-09-21
Explicit integer sizes for array capacity

mattmassicotte:
Addresses some integer precision losses related to the `Array` type. Slight API change to make explicit the `uint32_t` capacity limitation.

Discussion: https://github.com/tree-sitter/tree-sitter/issues/1880

### DIFF
--- a/lib/src/array.h
+++ b/lib/src/array.h
@@ -154,7 +154,7 @@ static inline void array__reserve(VoidArray *self, size_t element_size, uint32_t
     } else {
       self->contents = ts_malloc(new_capacity * element_size);
     }
-    self->capacity = (uint32_t)new_capacity;
+    self->capacity = new_capacity;
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/src/array.h
```diff
@@ -154,7 +154,7 @@ static inline void array__reserve(VoidArray *self, size_t element_size, uint32_t
     } else {
       self->contents = ts_malloc(new_capacity * element_size);
     }
-    self->capacity = new_capacity;
+    self->capacity = (uint32_t)new_capacity;
```
maxbrunsfeld, line 157: Is this cast necessary? It looks like `new_capacity` is already a `uint32_t`

maxbrunsfeld (commented):
Thanks! I just left one question.

lib/src/array.h
mattmassicotte: This is why we do code review! 😅

maxbrunsfeld:
Thanks for these fixes!

PR #1884, opened 2022-09-22
Fix/subtree and query integer sizing

mattmassicotte:
Almost of these were length calculations, which as discussed, aren't of concern. With this, all of the warnings I see are fixed.

There was one use of the `unsigned` type. This matches the API being used, so I felt ok keeping it as is instead of using an explicitly-sized type.

Discussion: https://github.com/tree-sitter/tree-sitter/issues/1880

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -331,7 +331,7 @@ static bool stream_advance(Stream *self) {
   if (self->input < self->end) {
     uint32_t size = ts_decode_utf8(
       (const uint8_t *)self->input,
-      self->end - self->input,
+      (uint32_t)(self->end - self->input),
       &self->next
     );
     if (size > 0) {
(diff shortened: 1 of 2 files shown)


mattmassicotte:
I know I probably should have done this all one on PR, but I was hoping to keep the changes targeted. Baby steps :)

Does this look ok? Would love to get this in!

mattmassicotte:
Just popping in again, would be cool to get these warnings resolved!

ahlinc:
@mattmassicotte Thanks, I marked this PR to don't forget and will review it soon.

maxbrunsfeld:
I'm ok with all of these, I guess. The fact that some of these conversions were considered "implicit" is a bit surprising to me, especially in declarations of the form

```
uint32_t x = some_64_bit_value
```

It seems pretty explicit to me 😅 . But if people want to enable those warnings in downstream projects, I guess it makes the library more useful to comply with them.

@mattmassicotte Are  you interested in tweaking the CI action to enable those when building the C library, maybe via an environment variable like CFLAGS that's set specifically on CI?

PR #1891, opened 2022-09-27
Add node position to XML output

flavioheleno:
This PR adds 4 new attributes to XML tags (when `--xml` option is passed):
* `srow`: start row position
* `scol`: start column position
* `erow`: end row position
* `ecol`: end column position

These new attributes allows users to work based on the XML output in a more flexible way with relevant information from the original source code.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -158,6 +158,7 @@ pub fn parse_file_at_path(
             println!("");
         }
 
+        // Render a XML if `--xml` was passed
         if matches!(output, ParseOutput::Xml) {
             let mut needs_newline = false;
             let mut indent_level = 0;
@@ -192,6 +193,12 @@ pub fn parse_file_at_path(
                         if let Some(field_name) = cursor.field_name() {
                             write!(&mut stdout, " type=\"{}\"", field_name)?;
                         }
+                        let start = node.start_position();
+                        let end = node.end_position();
+                        write!(&mut stdout, " srow=\"{}\"", start.row)?;
+                        write!(&mut stdout, " scol=\"{}\"", start.column)?;
+                        write!(&mut stdout, " erow=\"{}\"", end.row)?;
+                        write!(&mut stdout, " ecol=\"{}\"", end.column)?;
                         write!(&mut stdout, ">")?;
                         tags.push(node.kind());
                         needs_newline = true;


ahlinc:
Similar to: #1080

flavioheleno:
hey @ahlinc can I do anything to help?

ahlinc:
I'm not sure that the current XML output implementation should exist in the current form. I've looked into git history, how it was added and it looks it was added hastily and doesn't get into account many factors for example it should properly indented for normal readability, make control symbols escaping, etc. Also I'm not sure that it's suitable to process current xml output programmatically and may be it can be improved. Using xml as a debug output also isn't comfortable because xml markup by itself produces a lot of redundant info that complicates it visual perception.

ahlinc:
All above are reasons why I didn't fix previous PR by myself and didn't merge it.

ahlinc:
@flavioheleno could you describe your reasons and purposes to have XML format for tree-sitter trees?
It may help to move in a right direction on improvement of existing output implementation.

flavioheleno:
@ahlinc thank you for pointing out some reasons as to why the XML output is as-is right now! my use-case is to process it programmatically as you mentioned, although it could be improved, its current state is already good enough for my use case, apart from the lack of start/end positions.

ahlinc:
@flavioheleno could you tell what XML library do you use for consuming tree-sitter's xml tree? And if it's not hard provide a snippet of code how do you use for processing a tree? Do you consume node's text?

flavioheleno:
@ahlinc I'm using PHP's [SimpleXML](https://www.php.net/simplexml) which is based on [libxml](https://gitlab.gnome.org/GNOME/libxml2/-/wikis/home). I'm currently traversing the tree to extract nodes of interest (eg. function call expressions).

to be honest with you, I'm also working on a PHP wrapper for tree-sitter, but that takes a lot more time than simply relying on the cli's output to do my analysis :-)

code sample:
```php
$xml = new SimpleXMLElement(file_get_contents('/path/to/output.xml'));

foreach (traverseXml($xml) as $node) {
  if ($node->getName() === 'function_call_expression') {
    // work on $node here
  }
}

function traverseXml(SimpleXMLElement $root): iterable {
  yield $root;
  foreach ($root->children() as $node) {
    yield from traverseXml($node);
  }
}

```

andrew-bedford:
For what it's worth, I'm also interested in having positions in the xml output. I'd like to be able to use XPath to select nodes and report their positions back to the user when found.

amaanq:
superseded by #2970

Issue #1094, opened 2021-05-06
Quality indicators for grammars


julianthome:
Except for the differentiation between _"fairly complete"_ and _"in development"_ grammars on the website (https://tree-sitter.github.io/tree-sitter/#available-parsers), I could not find any quality indicators for grammars. If there are already discussions around quality metrics for tree-sitter grammars, I would be grateful if you could point me to them. 

One possible quality indicator that could be helpful to understand the maturity of a given grammar could be language coverage (i.e., how many constructs of the target language are (un)-supported). 

This type of information would be very useful to identify, track and eliminate weak spots certain grammars may have.

ahlinc:
I don't know about such discussions unfortunately but I'd like to describe my thoughts how this may work:
- Firstly, the simple list in documentation don't represent well any quality metrics.
- At this point amount of grammar implementations achieved a point where there're concurrent implementations for the same language exists.
- I think  several additional records may be introduced in the grammars `package.json` file like:
  - `specification_urls` - point to a language specification sources used for grammar implementation.
  - `completeness.percentage` - a percentage of a grammar completeness estimation provided by a grammar author.
  - `completeness.notes` - an array of notes about completeness provided by a grammar author.
  - `rule_naming_style` - `original_spec` - used naming based on a language spec, `tree_sitter_convention` - a snake  case style used in tree-sitter by convention. This may be useful to understand how much work needed to switch from a one grammar implementation to some another.

Such notes may be provided by grammar authors to track a grammar implementation state.
Most grammars hosted on Github can be fetched from the `tree-sitter` package [`dependants`](https://github.com/tree-sitter/tree-sitter/network/dependents?package_id=UGFja2FnZS0zMjE4MDA0NzQ1) and their `package.json` can be parsed for above mentioned fields automatically.
Also automatically may be computed some metrics like:
- Peak memory usage on grammar generation by `tree-sitter generate` command.
- Grammar binary library size.
- Does it have an external scanner.
- How much terminals handled by an external scanner.
- What fields about grammar filled in `package.json` like `scope`, `file_types` and so on.

All above information may be a source for automatically generated table of grammars instead of manually managed list.

> One possible quality indicator that could be helpful to understand the maturity of a given grammar could be language coverage (i.e., how many constructs of the target language are (un)-supported).

[...]

PR #1894, opened 2022-09-28
Unified parser list

mattmassicotte:
There are a bunch of parsers not in the docs list. I started looking at updating them, but it's kind of a pain because of the done/in-progress distinction.

I'd like to propose a single list. I'm not sure the distinction helps users that much, and also means it needs to be updated as parsers mature.


verhovsky:
The existing sort order makes more sense than sorting them with `uniq` or whatever. "C" appearing with and before "C++" and "C#" and not way after all the other "C*" languages makes sense. Ordering "Verilog" after "VHDL" makes sense for computers but not people, and so on.

mattmassicotte:
I'm happy to change the ordering, if that's really wanted. But, let's at least get down to a single list.

mattmassicotte:
Got stale. Going to close and possibly redo it.

ahlinc:
@mattmassicotte There was a question about grammars quality #1094, I left my thoughts there: https://github.com/tree-sitter/tree-sitter/issues/1094#issuecomment-1382654832

PR #1899, opened 2022-10-03
Add support for vim quickfix

EmilOhlsson:
Adds support outputting matching queries as a quickfix list that is for example loadable by vim. This can be useful when searching larger projects for certain code patterns. Also adds support for ignoring query names starting with leading underscore, as those might not be interesting for the end result when producing a quickfix list.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -168,6 +168,18 @@ fn run() -> Result<()> {
                         .takes_value(true),
                 )
                 .arg(&scope_arg)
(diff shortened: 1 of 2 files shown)


amaanq:
I don't think this makes sense to have here, it feels like a bodge at best and it is probably better suited to have vim natively integrate this via a plugin of sorts. Note that neovim has first-class support for tree-sitter so I'm sure it's even easier to implement there.

PR #1913, opened 2022-10-16
Allow web-tree-sitter to work with Emscripten 3

### LINKED ISSUE #1560: Grammars built with Emscripten 3 don't load in browser: RuntimeError: abort(Assertion failed: undefined)
I am trying to use tree-sitter in the browser (using Webpack) and when I load the grammar I built, I get this error

```
Uncaught (in promise) RuntimeError: abort(Assertion failed: undefined). Build with -s ASSERTIONS=1 for more info.
    at se (main.js:2)
    at N (main.js:2)
    at ye (main.js:2)
    at Le (main.js:2)
    at main.js:2
    at async main.js:2
```

It happens on this line

```js
const Python = await Parser.Language.load('tree-sitter-python.wasm')
```

I can see that the wasm file gets requested and is successfully downloaded. This happens with other grammars, not just the Python grammar. I am able to load wasm files I built a couple months ago.

I am using Node 16 and emcc 3.0.2 installed with Homebrew on an M1 Macbook Air with macOS 12.1

```sh
$ node --version
v16.13.1
$ emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.0.2-git
$
$ which node
/opt/homebrew/opt/node@16/bin/node
$ which emcc 
/opt/homebrew/bin/emcc
```

I am building the grammar with

```sh
npx tree-sitter build-wasm node_modules/tree-sitter-python/
```

You can download a broken `tree-sitter-bash.wasm` file I built if you do 

```sh
npm install curlconverter@4.0.0-alpha.8
```

In fact, the fastest way to reproduce this issue would be to

```sh
git clone git@github.com:curlconverter/curlconverter.github.io.git
cd curlconverter.github.io
git checkout c2d12aebf58c83f20a766edde46e875b5c976b6b
```

then in package.json change `"curlconverter": "^4.0.0-alpha.7",` to `"curlconverter": "=4.0.0-alpha.8",` before running `npm install` then running `npm start` and you should see the error in the browser's JS console. The grammar file shipped in the previous version (i.e. 4.0.0-alpha.7) should work though.

I also created a gist of a simpler project that I can reproduce this issue with: https://gist.github.com/verhovsky/a04c7256da13797e312acd0a8afd9023

J3RN:
Changes are explained in-line.

Originally I tried [implementing the "Basic Usage"](https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_web#basic-usage) from inside a React application, and encountered this error message:
Screenshot from 2022-10-16 17-51-22

Per its advice, I changed the emscripten flags in both `script/build-wasm` (for tree-sitter's WASM) and `cli/src/wasm.rs` (for the grammar's WASM) to include `-s ASSERTIONS=1` and made the following changes based on the issues surfaced.

Fixes #1560


cli/src/wasm.rs
```diff
@@ -83,6 +83,8 @@ pub fn compile_language_to_wasm(language_dir: &Path, force_docker: bool) -> Resu
         "-s",
         "NODEJS_CATCH_EXIT=0",
         "-s",
+        "NODEJS_CATCH_REJECTION=0",
```
J3RN, line 86: After enabling assertions, my first error was the generated tree-sitter grammar's WASM catching a rejection.  Sadly, I have not been able to recreate this for a screenshot.

[This flag (which disables that behavior) is already used in `script/build-wasm`](https://github.com/tree-sitter/tree-sitter/blob/cabdcfb41d780c760a2d0bea32f743b7d9ac8e4e/script/build-wasm#L89)

script/build-wasm
```diff
@@ -77,6 +77,8 @@ fi
 
 mkdir -p target/scratch
 
+runtime_methods='stringToUTF16','AsciiToString'
```
J3RN, line 80: Next, I encountered this error:
Screenshot from 2022-10-16 18-17-01

After some googling I discovered that [`stringToUTF16` is from emscripten's preamble](https://emscripten.org/docs/api_reference/preamble.js.html#stringToUTF16) and just needed exporting.

Next, the line `console.log(tree.rootNode.toString());` threw this error:
Screenshot from 2022-10-16 18-25-15

I quickly determined that the cause is the same and added it as well.

verhovsky:
#1560

J3RN:
@verhovsky I was able to replicate this issue. Building `web-tree-sitter`'s WASM in debug mode yields the following error instead:
image

This isn't something that can be exported like `stringToUTF16`, and so I'm digging a bit deeper now, starting with trying the suggestion.

lib/binding_web/exports.json
```diff
@@ -24,6 +24,7 @@
   "_memcmp",
   "_memcpy",
   "_memmove",
+  "_memset",
```
J3RN, line 27: This fixes the issue from running `tree-sitter-python`, see [my comment](https://github.com/tree-sitter/tree-sitter/pull/1913#issuecomment-1281130984)

verhovsky:
I tried a few more languages (tree-sitter-bash, -c, -cpp, -java, -javascript, -php and -python) and I can confirm this works. Great work

verhovsky:
I tried compiling tree-sitter-html (this time I did `build-wasm --debug` for tree-sitter.wasm like you suggested) and got this `Assertion failed: undefined symbol '__cxa_atexit'.` error in the console when trying to load it:

Screen Shot 2022-10-23 at 00 25 20

J3RN:
@verhovsky I don't have a good setup for testing this right now, but I just pushed a change that I believe will fix the issue you encountered with `tree-sitter-html`.  Would you mind testing it and reporting back?

verhovsky:
@J3RN that works. I did try this but I didn't realize I needed to keep adding underscores until it starts working lol

verhovsky:
Can you change cli/emscripten-version to "`3.1.24`", it's currently set to "`2.0.24`" and anything older than 3.1.1 fails if you run `cargo build --release && ./script/build-wasm --debug --docker` with this error:

```
emcc: error: undefined exported symbol: "___cxa_atexit" [-Wundefined] [-Werror]
```

maxbrunsfeld:
Thanks for the PR @J3RN, and for helping out with it @verhovsky. It looks like the issues that were blocking us from upgrading to Emscripten 3 are fixed now. It'll be great to be on the latest version.

PR #1914, opened 2022-10-17
Add towlower to wasm exports

dstoc:
tree-sitter/tree-sitter#1906

### DIFF
--- a/lib/binding_web/exports.json
+++ b/lib/binding_web/exports.json
@@ -35,6 +35,7 @@
   "_strcmp",
   "_strncpy",
   "_tolower",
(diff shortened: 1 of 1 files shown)


dstoc:
@maxbrunsfeld anything I should do here? The issues with the check seem unrelated to the change.

sogaiu:
FWIW, I think `strcmp` made it in [here](https://github.com/tree-sitter/tree-sitter/commit/1982e48568f4cd8badf565ee345989b6f82008fa).

amaanq:
That added tolower not towlower :grin:

PR #1918, opened 2022-10-20
Add Erlang to list of Available Parsers

alanz:
Switching from under-development

  https://github.com/AbstractMachinesLab/tree-sitter-erlang

to

  https://github.com/WhatsApp/tree-sitter-erlang/

which is based originally on it.


alanz:
cc @ostera, please do not merge without input from him.

ostera:
LGTM! 🚀

Decodetalkers:
Are they the same parse? if they are the same, I think we need to make a pr to nvim-treesitter

maxbrunsfeld:
Thank you!

PR #1936, opened 2022-10-31
Add note about including an external scanner in Rust crate

J3RN:
There's an already-existing note about modifying the `binding.gyp` file for compiling with Node.js; I just extend it to mention uncommenting a relevant section in `bindings/rust/build.rs`.


maxbrunsfeld:
Thanks!

PR #1938, opened 2022-10-31
Resolving a crash when freeing a null pointer when cancelling parses

logancollins:
This pull request resolves a potential crash we've seen in parser.c.

If a parse is in progress, in certain cases, the `lookahead.ptr` pointer can be `NULL` when the code reaches [src/parser.c#L](https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/parser.c#L1464). This will only actually crash if the cancellation token happens to have been set to a non-zero value.

This request contains a check to ensure that `lookahead.ptr` is not `NULL` before releasing the subtree.

I'm not _entirely_ up on all of the potential branch cases and other code with which this might interact, so I'm happy for a review to ensure this is an acceptable fix!

Thank you!

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1461,7 +1461,9 @@ static bool ts_parser__advance(
       ((self->cancellation_flag && atomic_load(self->cancellation_flag)) ||
        (!clock_is_null(self->end_clock) && clock_is_gt(clock_now(), self->end_clock)))
     ) {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks for the fix @logancollins; this makes a lot of sense.

Aman-preetam:
panicinc:merge/parse-cancel-pointer-crash

Aman-preetam:
Instagram hackimg

PR #1951, opened 2022-11-08
feat(rust): Load languages as untied from the Rust type system

ahlinc:
This PR fixes errors like:
```    
    error[E0308]: mismatched types
       --> src/main.rs:9:27
        |
        |             .set_language(tree_sitter_java::language())
        |              ------------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `tree_sitter::Language`, found a different struct `tree_sitter::Language`
        |              |
        |              arguments to this function are incorrect
        |
        = note: perhaps two different versions of crate `tree_sitter` are being used?
```

Such errors happen pretty often when user wants to use latest or some specific version of tree-sitter library in the Rust language but some language library uses other version in its `Cargo.toml` file.

This updates Rust binding wrapper around C API and it starts to return `*const ()` instead of `Language` instance tied to some specific version of tree-sitter library. The `*const ()` needs to be converted to a `Language` instance by using `Language::new()` method.
Also this PR completely removes language libraries dependency from the tree-sitter library, because language libraries may not require it at all. Currently such dependency exist for importing just one type, the `Language` symbol.
Another note is that language compatibility managed on runtime on the C level so Rust type system doesn't need to be involved for.

__It's the breaking change that would require from all languages to regenerate Rust bindings in the language repositories by previously manually remove related files like:__
```
rm bindings/rust/*.rs Cargo.*
```
__And run `tree-sitter generate` after that.__

### DIFF
--- a/cli/src/generate/templates/cargo.toml
+++ b/cli/src/generate/templates/cargo.toml
@@ -5,7 +5,7 @@ version = "0.0.1"
 keywords = ["incremental", "parsing", "PARSER_NAME"]
 categories = ["parsing", "text-editors"]
 repository = "https://github.com/tree-sitter/tree-sitter-PARSER_NAME"
-edition = "2018"
+edition = "2021"
 license = "MIT"
 
 build = "bindings/rust/build.rs"
@@ -19,8 +19,8 @@ include = [
 [lib]
 path = "bindings/rust/lib.rs"
 
-[dependencies]
-tree-sitter = "~RUST_BINDING_VERSION"
+[dev-dependencies]
+tree-sitter = ">= RUST_BINDING_VERSION"
 
 [build-dependencies]
 cc = "1.0"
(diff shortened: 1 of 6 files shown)


maxbrunsfeld:
> Such errors happen pretty often when user wants to use latest or some specific version of tree-sitter library in the Rust language but some language library uses other version in its Cargo.toml file.

I think we should probably avoid this error by just putting `tree-sitter = "*"` in the `dependencies` of the language crates. What do you think about that solution?

The reason I prefer that is that I don't like the unsafety of the `Language::new` approach. You could pass any arbitrary  `* const ()` pointer to the `Language::new` function, and get crashes.

ahlinc:
I found that [Cargo doc](https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies) states:
> While libraries can, strictly speaking, they should not. A version requirement of * says “This will work with every version ever”, which is never going to be true. Libraries should always specify the range that they do work with, even if it’s something as general as “every 1.x.y version”.

So it's not allowed to set `*` in a version.

In any way I think you are right and having `* const ()` in a safe api is the bad idea.

But even using `*` as a version locally doesn't solves the problem if to point all `tree-sitter` deps to a local fork of the `tree-sitter` repo.

PR #1953, opened 2022-11-09
fix(cli): Racing on playground webserver port binding

ahlinc:


maxbrunsfeld (approved):

cli/src/playground.rs
```diff
@@ -135,10 +122,24 @@ fn response<'a>(data: &'a [u8], header: &Header) -> Response<&'a [u8]> {
...
+        Some(port) => bind_to(&*addr, port).expect("Can't bind to the specified port"),
+        None => {
+            get_listener_on_available_port(&*addr).expect("Can't find a free port to bind to it")
+        }
+    };
+    Server::from_listener(listener, None).expect("Failed to start web server")
+}
+
+fn get_listener_on_available_port(addr: &str) -> Option<TcpListener> {
+    (8000..12000).find_map(|port| bind_to(addr, port))
```
webknjaz, line 140: @ahlinc have you thought of using an ephemeral port? If the port isn't important, set it to `0` and the kernel will bind to a free one. That will also be guaranteed to be free of race conditions.

ahlinc: The purpose of this port is to connect to it from a browser locally and browsers can't connect to ephemeral ports from the url address bar.
Random ports aren't good for this case because this will make all already open playground tabs unreloadable what also isn't good because this will not allow to recover the opened previously tab from the hard parser errors. The current approach isn't ideal too, but it's portable and works good most of the time.

webknjaz: I was thinking along the lines of getting a port from `0`, saving it and then reusing it.

ahlinc: `tree-sitter` cli is the standalone tool that doesn't store any state between calls except small piece of configuration. Also using random ports would be less convenient because it will be hard to just type a remembered address in the browsers' url bar to open a new playground instance.

ngirejw (commented):
AliExpress

PR #1956, opened 2022-11-10
Fix versioning in Makefile

### LINKED ISSUE #1158: `pkg-config --modversion tree-sitter` always reports `0.6.3` version
In https://github.com/rizinorg/rizin/issues/1194 we noticed that `pkg-config --modversion tree-sitter` always reports `0.6.3` version. We suspect version does not change even for releases that differ in API. That makes it different to have `./configure`-time version checks to guide users on minimum expected version for `rizin`.

Looks like the version is hardcoded in `Makefile`: https://github.com/tree-sitter/tree-sitter/blob/master/Makefile#L1

```Makefile
VERSION := 0.6.3
```

WDYT of bumping the version with releases?

Thanks!

casouri:
Currently tree-sitter libraries are versioned 0.6.3 in pkg-config files and 0.0 in /usr/local/lib. This prevented us from requiring for a specific version of tree-sitter, specifically any version >= 0.20.2, which is the first version that can change malloc on runtime.

Closes #1158

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := 0.6.3
+VERSION := 0.20.7
 
 # install directory layout
 PREFIX ?= /usr/local
@@ -22,8 +22,8 @@ CFLAGS ?= -O3 -Wall -Wextra -Werror
 override CFLAGS += -std=gnu99 -fPIC -Ilib/src -Ilib/include
 
 # ABI versioning
-SONAME_MAJOR := 0
-SONAME_MINOR := 0
+SONAME_MAJOR := 20
+SONAME_MINOR := 7
 
 # OS-specific bits
 ifeq ($(shell uname),Darwin)


ahlinc:
@casouri It's correct that the `VERSION` variable should be changed with every release of tree-siter's C library but SO versions don't need to be changed without ABI changes, see: https://github.com/tree-sitter/tree-sitter/issues/930#issuecomment-782924093

casouri:
Oh cool. TIL, thanks. I don't intend this PR to be merged, since you probably need to release a new version to fix the Makefile anyway.

maxbrunsfeld:
> I don't intend this PR to be merged

I'm open to changing the version. What should the SO ABI version numbers be?

casouri:
Updating `VERSION` to the actual version is good enough for us. (In the meantime we directly test for the existence of ts_set_allocator in the so file, instead of relying on the version number.) According to the information @ahlinc linked to, the ABI version don't need to change.

ahlinc:
Closes #1158

ahlinc:
Done.

PR #1990, opened 2022-12-14
fix: possible rollover of nanoseconds in clock.h

paf31:
Setting a timeout on a parser can result in a non-deterministic behavior, where parsers time out when they should not. An incorrect timeout is more likely for timeout intervals closer to (but less than) 1s. We noticed that the distribution of these incorrect timeouts didn't seem to be correlated with the complexity of the inputs.

The cause seems to be this possible rollover of the nanoseconds component of `timespec` in `clock.h`. Patching this fixes the timeout issue.

I'm not sure how I would go about adding tests for this, but if you think they are necessary and can provide some pointers, I'd be happy to include them.

Thanks!


maxbrunsfeld:
Sorry for the delay, and thanks for the fix!

PR #2019, opened 2023-01-08
Make error message more specific

verhovsky:
I forgot I was using `--docker` and the current error message made me waste a few minutes installing emcc


ahlinc:
@verhovsky thanks!

PR #2032, opened 2023-01-15
Reliable CI tests

ahlinc:
Some Tree-sitter tests are flaky and this from time to time affects CI with random fails in unexpected places.
At this moment I know two such cases:
- Tests that use timeouts may fail.
- Corpus tests that test random edits may fail on some random seed that produces an edit that changes parsing tree structure.

This PR marks such tests as `#[ignored]` by default and uses another `nextest` test runner to run such marked test separately with 10 retries.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,7 +24,7 @@ test_script:
   # Fetch and regenerate the fixture parsers
   - script\fetch-fixtures.cmd
   - cargo build --release
-  - script\generate-fixtures.cmd
+  - script\generate-fixtures.cmd --build
 
   # Run main tests
   - cargo nextest run --workspace --lib --tests --bins --examples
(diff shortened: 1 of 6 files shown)


sogaiu:
I tried out [43a44fc](https://github.com/tree-sitter/tree-sitter/pull/2032/commits/43a44fcf1c94161b61b8e015bed183a08b037078) a bit.

In the 15+ invocations I tried of `./script/test` I didn't see a single failure :)

ahlinc:
@sogaiu I didn't patched the `./script/test` and with regular `cargo test` runner it just skips flaky tests.
I found the approach in the PR not very useful as CI and `cargo test` works differently and there is no much more benefit from `cargo nextest` in CI.  I'll rework this PR in another way.

PR #1980, opened 2022-12-06
Add clojure to list of available parsers

dannyfreeman:
After some discussion in 
https://github.com/sogaiu/tree-sitter-clojure/issues/28

I decided to submit an issue to the tree-sitter organization  to include [tree-sitter-clojure](https://github.com/sogaiu/tree-sitter-clojure) in their list of available parsers. This repository is used by a couple of editors and tools, and  I am using it to build a tree-sitter based programming mode for Clojure in Emacs.
Since there are a couple of tools that depend on it, it seems  worthy of inclusion in an official list of tree-sitter grammars.

PR #2037, opened 2023-01-17
Include many new parsers in a unified list

mattmassicotte:
A single parser list and includes many missing parsers I'm aware of it. It closes #1541, #1980, and #1123.

It also partially addresses #1862 and #1623, but I really wanted to first get this change in. After, I can go through and re-create a PR that closes the rest of the pending parser list additions.

However, I just want to point out that I really don't like taking over the contributions of others like this. I know that this is a bit of a special circumstance, because many of these now have conflicts. But I just don't like doing that as a matter of etiquette. So, my hope is this is a one-time event.

I also want to be really careful with any form of curation.  I'd prefer not to exclude a parser project unless it explicitly links to a more suitable replacement. I've included links to two Markdown parsers for this reason - these two projects have extremely different designs, and it's hard to say which would be most useful to a client.

Should this be merged, I'll follow up with a second change that changes this to a table. This will handy to use for any quality metrics that come out as part of #1094. And, it will also be a nice place to put the information that @clason suggested. We can have discussion around that in that forthcoming PR.


mattmassicotte:
@ahlinc here's the parser list change!

ahlinc:
@mattmassicotte what do you think about to add some unicode emoji symbols as marks for parsers / bindings maintained in the Tree-sitter organization and by community members like I've demonstrated there: https://ahlinc.github.io/tree-sitter/#language-bindings ?

clason:
Slightly combative question: what would be the point? In my personal experience, parsers maintained by this org are not necessarily better (and, to be perfectly frank, often quite the opposite due to lack of maintenance).

clason:
And is that even something the org -- or the community -- wants?

ahlinc:
> Slightly combative question: what would be the point? In my personal experience, parsers maintained by this org are not necessarily better (and, to be perfectly frank, often quite the opposite due to lack of maintenance).

Good points, I agree, parsers maintenance is a time heavy thing and in the org there no enough people to support all org parsers in the actual state.

> Are more "first-party" parsers coming, or is the organization open to adopting ones from the community?
> And is that even something the org -- or the community -- wants?

Unfortunately at this moment I don't have answers for this questions.

Lets merge the new list in the state of current PR, as it's better to have it in a more actual state.
@mattmassicotte Thank you for the PR!

maxbrunsfeld:
Thanks!

maxbrunsfeld:
> is the organization open to adopting ones from the community?

I've come to the conclusion that we shouldn't, since the owners of this org already don't have a lot of spare time for routine maintenance of repos.

ahlinc:
@maxbrunsfeld there was an unresolved review question, but never mind I've fixed it on master.

PR #2048, opened 2023-01-24
docs: add capnp, smali, thrift

amaanq:
Title


amaanq:
I also have a couple of in-progress ones (kdl, nginx) if whoever sees this is ok with adding that to the in development section

Issue #2063, opened 2023-01-31
External scanner is not serialized unless `scan()` returns `true`


ben-clayton:
Currently the external scanner state is only serialized if `scan` returns `true`. This makes modification of the state highly restrictive.

External scanners may wish to perform look-ahead computations while the scanner has more look-ahead context, without forcing the current token to resolve to an external token.

I propose adding a `bool force_serialize_state;` field to `TSLexer`, which can be set to `true` in `scan` to signal that the state should be serialized, even if `scan` returns `false`. This field would be set to `false` before each call to `scan`.

This makes the new behavior (and potentially costly) serialization opt-in.

I have a PR that prototypes this proposal which I will post shortly.

ahelwer:
I return a zero-width token. It works pretty well and keeps the door from being opened to very weird bugs.

ben-clayton:
Unfortunately in my case, I need to have the scanner fall back to the 'internal' token scanner while performing look-ahead computations. Attempts to return a placeholder external token adds substantial, unwanted complexity to my grammar.

ahelwer:
So it's more about not wanting to sprinkle hidden zero-width external tokens throughout your grammar rather than lack of capability?

maxbrunsfeld:
I left some comments on #2064. I don't think we can make that particular change, but open to discussing other options.

ben-clayton:
Maybe - I haven't seriously considered adding hidden zero-width tokens in all the places I need to mutate state. There's a bunch of them.

Some background: I'm experimenting with adding template-argument-list vs less-than / greater-than disambiguation to WebGPU's shading language, WGSL.

The look-ahead needs to trigger on the `identifier` of an `identifier` `<` pair of tokens, and markup future `<` and `>` token pairs. Later the `<` and `>` tokens are either preserved (`scan` returns `false`), or replaced with external *template* tokens - either way the scanner state needs to mutate (we're popping from a queue created by the initial look-ahead scan).

My very rough scanner so far is: https://github.com/ben-clayton/gpuweb/blob/custom-template-scanner/wgsl/scanner.cc. This uses the new functionality in https://github.com/tree-sitter/tree-sitter/pull/2064, and works well.

[...]

PR #2064, opened 2023-01-31
Add `TSLexer::force_serialize_state`

ben-clayton:
A new lexer flag to ensure that an external scanner's state is serialized even when the `scan` function returns false.

This is useful for scanners that perform lookahead scanning, but do not want to override the internal scanner for the current token.

Issue: #2063


lib/src/parser.c
```diff
@@ -424,14 +424,15 @@ static Subtree ts_parser__lex(
...
+      self->lexer.data.force_serialize_state = false;
       bool found_token = self->language->external_scanner.scan(
         self->external_scanner_payload,
         &self->lexer.data,
         valid_external_tokens
       );
       ts_lexer_finish(&self->lexer, &lookahead_end_byte);
 
-      if (found_token) {
+      if (found_token || self->lexer.data.force_serialize_state) {
```
ben-clayton, line 435: Note: `ts_parser__lex` is a complex beast, and I'm far from certain that this is perfect. Tests seem happy with this - but I'm more than willing to take any suggestions here.

maxbrunsfeld:
I don’t think incremental parsing could work correctly with this behavior. It’s an important invariant that the scanner’s state can be assumed to be fixed unless external tokens are produced and used in the tree.

maxbrunsfeld (changes requested):
I definitely appreciate the good work that you put into this, especially the integration test that you added. But I don't think we want this option, because I'm pretty sure that it's incompatible with the way that Tree-sitter performs incremental parsing in the *presence* of external scanners.

When reusing a subtree during incremental parsing, we need to make sure the external scanner is in the same state as it was before that subtree was originally produced, and then we need to *reset* the scanner's state to match the state *after* that subtree. Both of those steps rely on the external scanner's states being reflected in the syntax tree itself. That is one important purpose of the serialization.

In the broad sense, I'm open to improvements to this model, but we would need to reason very carefully about how they work with incremental parsing.

ben-clayton:
Thanks, and no worries. I suspected this would be more subtle and difficult than I understood.

PR #2065, opened 2023-02-02
highlight/README: fix highlight_names list

artagnon:


artagnon:
I believe the the correct list for highlighting is the following, as this is what will get highlighted by default:

https://github.com/tree-sitter/tree-sitter/blob/e021d6e9799903312dfa3a1ed39e160008fe52ac/cli/src/highlight.rs#L150-L171

The following list has been extended with custom nodes:

https://github.com/tree-sitter/tree-sitter/blob/e021d6e9799903312dfa3a1ed39e160008fe52ac/cli/src/tests/highlight_test.rs#L26-L48

I've updated the sort order in both files, and added `number` to README.

artagnon:
EDIT: I've updated the sort order in `cli/highlight` and removed `variable` from README, as it is not highlighted by default.

PR #2080, opened 2023-02-09
Derive Hash for Language

Ekleog:
Given Language implements Eq, it can probably implement Hash too.

This makes it easy to derive a random identifier from a `Language`

For some context, I'm currently writing [a creduce-like based on tree-sitter](https://github.com/Ekleog/tree-sitter-reduce) (not yet ready for prime time, though `cargo run --bin rsreduce` already kinda works), and am trying to derive a random identifier for passes so that the user can know which log lines refer to the same pass. Among the things in a pass is the `Language`, and though I could just not hash it, I don't see any reason not to :)


maxbrunsfeld:
Thanks!

Ekleog:
Well, thank _you_ for the quick merge and work on tree-sitter! :)

PR #2090, opened 2023-02-20
Add tree-sitter-ungrammar and tree-sitter-yuck to the parser list

Philipp-M:
Both parsers should be complete as of now. They are both modeled based on the original parsers of their respective repos. Each contains enough syntax-highlighting queries to highlight the whole source code.


amaanq:

PR #2091, opened 2023-02-20
Fix: Fixed undefined property access in js binding for matches (refutedProperties) #2045

### LINKED ISSUE #2045: Cannot set property 'refutedProperties' of undefined on web-tree-sitter
Maybe related to https://github.com/tree-sitter/tree-sitter/issues/1968?

Basically, I installed `web-tree-sitter` and tried to parse a Ruby grammar. I exported the WASM file with: `tree-sitter build-wasm --docker node_modules/tree-sitter-javascript/`

Then I moved the resulting WASM file into `/tmp/grammars/ruby/grammar.wasm` and all queries to `/tmp/grammars/ruby/queries`, and tried to run this code:

```js
const Parser = require('web-tree-sitter');

let parser
Parser.init().then(() => {
  parser = new Parser()
  return Parser.Language.load('/tmp/grammars/ruby/grammar.wasm')
}).then(language => {
  let syntaxQuery = fs.readFileSync('/tmp/grammars/ruby/queries/highlights.scm', 'utf-8')
  parser.setLanguage(language)
  syntaxQuery = language.query(syntaxQuery)
  const tree = parser.parse('a = 1')
  global.highlights = syntaxQuery.matches(tree.rootNode)
})
```

This last line gave me the error `Cannot set property 'refutedProperties' of undefined`. The code is minified, so I'm not sure what should I have it there, but when I try to open on devtools (I'm running this code on Electron, but it fails on Node the same) I see this code:

```js
const e = this.setProperties[r];
e && (i[t].setProperties = e);
const n = this.assertedProperties[r];
n && (i[t].assertedProperties = n);
const s = this.refutedProperties[r];
s && (i[t].refutedProperties = s)
```

`s` is basically `{local: null}` and `i` is an array of 1 element. Problem is that `t` is `3`, so it's trying to find the 3rd element of a 1-size array.

Nimaoth:
Fixes #2045


ahlinc:
**TODO:**
- [ ] Check that Rust version doesn't affected.

jeff-hykin (commented):
Idk about the `typeof callback === 'object' && callback.constructor === Array` checks and the whitespace changes (I'm not a tree sitter contributor)

But I was also finding/fixing the `filteredCount` vs `i` fix and that part looks good/correct to me

Nimaoth:
Ah, yeah, those last few commits were only intended for my personal project, but maybe I can make it nicer so it can be merged upstream, otherwise I'll remove it from this branch

Issue #1526, opened 2021-12-09
ts_node_child_by_field_* problematic in combination with anonymous nodes


Cu3PO42:
I have produced a grammar and corresponding input for which `ts_node_child_by_field_{name,id}` works *incorrectly*. I derive my understanding of the expected behavior not only from intuition, but also from the output of `tree-sitter parse`.

In order to investigate this issue, I have created a fairly minimal grammar with which the behavior can be observed. It is as follows:
 ```js
module.exports = grammar({
    name: 'expressions',
    rules: {
        source_file: $ => $._expr,
        _whitespace: $ => /\s+/,

        _expr: $ => choice(
            $.and,
            $.identifier,
            $._brackets,
        ),

        and: $ => prec.left(seq(field("left", $._expr), "AND", field("right", $._expr))),
        identifier: $ => /[_\w][\w_\d]*/,

        _brackets: $ => seq("(", $._expr, ")"),
    },

    extras: $ => [$._whitespace],
})
```

On the input `( (FOO) AND BAR )`, the output of `tree-sitter parse` is
```
(source_file [0, 0] - [0, 17]
  (and [0, 2] - [0, 15]
    left: (identifier [0, 3] - [0, 6])
    right: (identifier [0, 12] - [0, 15])))
```

This also corresponds to my intuition on the expected behavior. However, when programmatically querying for the child labeled `left` of the `and` node, I get the token corresponding to the second opening bracket, i.e., the first child of the hidden  `_brackets` node rather than its only named child.

I believe that I have tracked down the incorrect code to [a snipped in `node.c`](https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/node.c#L551-L559). My understanding is that since `_brackets` is anonymous, it is not *relevant*. Therefore the code returns the first child, I believe this code should be changed to return the first non-anonymous child instead.

I have compiled a [demo repo](https://github.com/Cu3PO42/tree-sitter-issue-demo) including all necessary parts to observe this behavior.

* The Tree-Sitter parser lives in the `parser` subdirectory.
* `parser/examples/foo` contains the input I used
* The main directory includes minimal Rust code accessing the node in question and logging its start and end position.

maxbrunsfeld:
[...]

PR #2104, opened 2023-02-25
lib: fix `ts_node_field_name_for_child` implementation

ahlinc:
The original implementation in #1109 was not correct and didn't take into account that there may be a hidden nodes that should be transparently extend an indexing range of a parent node. Also the fact that field names are [inherited](https://github.com/tree-sitter/tree-sitter/issues/1526#issuecomment-990382515) from parent nodes to child nodes wasn't took into account.

Closes: #1878
Closes: #1642

_**Edit:** There is an additional related PR that extends API #1264 that needs to be improved like this PR and also merged._

### DIFF
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -569,24 +569,58 @@ recur:
   return ts_node__null();
 }
 
-const char *ts_node_field_name_for_child(TSNode self, uint32_t child_index) {
-  const TSFieldMapEntry *field_map_start = NULL, *field_map_end = NULL;
-  if (!ts_node_child_count(self)) {
+static inline const char *ts_node__field_name_from_language(TSNode self, uint32_t structural_child_index) {
+    const TSFieldMapEntry *field_map, *field_map_end;
+    ts_language_field_map(
+      self.tree->language,
+      ts_node__subtree(self).ptr->production_id,
+      &field_map,
+      &field_map_end
+    );
+    for (; field_map != field_map_end; field_map++) {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks so much for fixing this @ahlinc. Would you be up for adding a test for this method, which fails with the current implementation?

ahlinc:
@maxbrunsfeld sure, I'll add tests, I found that current implementation also don't work properly in some cases so I'll fix it.

ahlinc:
I've pushed a final fixed implementation that behaves the same way like retrieving field names over `cursor.field_name()` API. After adding tests this PR can be merged.

ilevyor:
Thanks so much for working on this fix! could I help get this merged by adding the tests? What would be the place to do that?

ahlinc:
@ilevyor you are welcome if you would like to make a try to write complete tests that would cover all edge cases.
A test that needs to be improved is here: [#1109/files](https://github.com/tree-sitter/tree-sitter/pull/1109/files#diff-dee3c4e82b172a0582dfe27ac708f7f225e0525a3468977e7127b3c4e331fe56R251-R275) But implementing a proper test needs a grammar that would be crafted to demonstrate all edge cases because it's unusual case where one nested fields override another parent fields on hidden rules.

_I think I'll add a separate todo issue for the tests and would merge this PR as is because it's better to have a working but not fully test covered API than completely broken for most cases._

ahlinc:
* The issue for tests: #2170

PR #2107, opened 2023-03-01
Fix bug with first child group anchor (anchor had no effect)

Philipp-M:
I'm not sure if this is intended behavior, but it feels to me that this is likely a bug (took me some time to figure that out, as  I was just learning about anchors with that first query so...)

Honestly I have no idea about the query codebase and this fix was just a hit and miss (but a hit it seems). But no tests have degraded and my issue is resolved with this change, see:

tree-sitter-group-anchors-bug

with the following query:
```scheme
(list .
  ((symbol) @string
	  (#match? @string "^(defwindow|defwidget|defvar|defpoll|deflisten|geometry|children|struts)$")))
```

and this parser: https://github.com/Philipp-M/tree-sitter-yuck/tree/607bf03cc78912cbaa85a3d373af21ddeaef5b41

on the following yuck code:

```yuck
(defwidget)
(defwidget other-symbol defwidget 23)
```

### DIFF
--- a/cli/src/tests/github_issue_test.rs
+++ b/cli/src/tests/github_issue_test.rs
@@ -6,11 +6,37 @@
 // cargo test --target $(rustc -vV | sed -nr 's/^host: //p') -- --test-threads 1
 // ```
 
+use super::helpers::query_helpers::assert_query_matches;
 use crate::tests::helpers::fixtures::get_language;
+use indoc::indoc;
 use tree_sitter::Query;
 
 #[test]
 fn issue_2162_out_of_bound() {
     let language = get_language("java");
     assert!(Query::new(language, "(package_declaration _ (_) @name _)").is_ok());
+}
+
+#[test]
+fn issue_2107_first_child_group_anchor_had_no_effect() {
+    let language = get_language("c");
+    let source_code = indoc! {r#"
+        void fun(int a, char b, int c) { };
+    "#};
+    let query = indoc! {r#"
+        (parameter_list
+            .
+            (
+                (parameter_declaration) @constant
+                (#match? @constant "^int")
+            )
+        )
+    "#};
+    let query = Query::new(language, query).unwrap();
+    assert_query_matches(
+        language,
+        &query,
+        source_code,
+        &[(0, vec![("constant", "int a")])],
+    );
 }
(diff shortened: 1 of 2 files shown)


amaanq:
@ahlinc would appreciate you checking this out :)

clason:
@ahlinc friendly ping -- looks like an obvious bugfix, and would be good to have fixed in Neovim :)

ahlinc:
_From Discord's #general channel_
> **amaanq:** @ahlinc do check this pr..its important imo :) https://github.com/tree-sitter/tree-sitter/pull/2107#issuecomment-1528748144
> **ahlinc:** Sorry, but I don't have a time for now to implement a test for it with any corpus grammar that currently used for tests.
> **amaanq:** ahhh ok, should I ask him to add tests for it if he can?
> **ahlinc:** Without a good test that would fail first I'm not open to blindly push the merge button for things that I didn't wrote and don't understand their implementation in all details.
> **ahlinc:** Yes, sure, with any corpus language mentioned in https://github.com/tree-sitter/tree-sitter/blob/321a652626c63bfea3ea320083a4b14863b80270/script/fetch-fixtures#L24-L38 
> **ahlinc:** An example of such test: https://github.com/tree-sitter/tree-sitter/blob/321a652626c63bfea3ea320083a4b14863b80270/cli/src/tests/query_test.rs#L2746-L2748C3
> **ahlinc:** @amaanq Anyone who want to push a stale PR with some Tree-sitter functionality improvement or a bugfix can implement good tests for it in an own fork an put a reference to an own branch in a comment for a PR and if tests would be ok then I'll update a PR's branch with proposed improvements and merge it.
> **amaanq:** kk
> **amaanq:** ty 🙂

clason:
Here's a similar example with the `tree-sitter-c` grammar (v0.20.2):
```c
void fun(int a, char b, int c) {
};
```
query
```lisp
(parameter_list
  .
  ((parameter_declaration) @constant
  (#match? @constant "^int")))
```
Screenshot 2023-05-09 at 17 54 19

clason:
@ahlinc is this enough to move this PR forward? or do you require a branch with an actual tree-sitter test? (If so, where should this go? I looked over the `test/` dir, but none of the subdirectories looked like the right place for this.)

ahlinc:
@clason thank you it's enough I'll convert it to a test and merge this PR soon.

ahlinc:
I've studied a bit the query parser code and ...
> On master, the `#match?` overrides the anchor

The problem isn't in the `#match?` statement the problem is that the `.` dot anchor immediately followed by a group represented by `( )` parenses what leads to a recursive call for the parsing function and where the `is_immediate` parameter wasn't propagated.

_@Philipp-M Thank you for the nice catch and the fix!_

PR #2126, opened 2023-03-14
[feat]: Expose the `ffi` interface via `into_raw()`

vhyrro:
This PR exposes the `ffi` module and the underlying data structures provided by the `bindings.rs` file and allows bidirectional interoperability between C and Rust, instead of simply `C` -> `Rust`.

The motivation behind it is somewhat niche, however I think it's a nice thing to have, especially when dealing with C bindings from Rust. I have a project that [provides such C bindings from Rust](https://github.com/nvim-neorg/neorg-dirman/blob/main/src/c_functions.rs), and I would love to be able to expose the C treesitter types as return values for some of the functions, such that one can use all of the C treesitter API without me having to write intermediate wrappers manually. For example, if I return a `ffi::TSTree` in my C function, the user can use all of the native `ts_tree_*` functions to operate on that tree from the C side.

To my knowledge there is no other way to access these data types from rust without managing a personal `bindings.rs`, which introduces its own plethora of issues, like having to link the C library while also having the Rust bindings in a single project, plus the overhead of making sure that both of the `tree-sitter` versions are in sync.

If there are any issues then let me know!


ahlinc:
If to provide support of this niche feature I'd suggest to split all such methods into separated `impl`s stored directlly in `ffi.rs` like:
<details><summary> `ffi.rs` example </summary> 

```rust
#![allow(dead_code)]
#![allow(non_upper_case_globals)]
#![allow(non_camel_case_types)]

include!("./bindings.rs");

extern "C" {
    pub(crate) fn dup(fd: std::os::raw::c_int) -> std::os::raw::c_int;
}

use crate::Parser;
use std::mem::ManuallyDrop;

impl Parser {
    /// Consumes the [Parser], returning a raw pointer to the underlying C structure.
    pub fn into_raw(mut self) -> *mut TSParser {
        self.stop_printing_dot_graphs();
        self.set_logger(None);

        ManuallyDrop::new(self).0.as_ptr()
    }
}
```

</details>

Also if to support just `.into_raw()` methods this looks like a one-side feature, so it makes sense to provide a counterparts  like `unsafe T::from_raw(...) -> T`.

vhyrro:
Good call, done that!

Regarding the `from_raw()` function - I'm contemplating on its implementation. Currently all of the tuple structs can just be manually constructed from the raw value by the user (e.g. `Language(ptr)` and `Node(NonNull:new_unchecked(ptr))`), so perhaps the better solution would be to make the `from_raw` wrappers safe? I.e.:
```rs
pub fn from_raw(ptr: *mut TSParser) -> Option<Parser> {}
// or
pub fn from_raw(ptr: *mut TSParser) -> Result<Parser, NullPointerError> {}
```

Although this seems a little redundant when the user can make these checks on their end super easily. Perhaps skip the `from_raw()` functions altogether? Happy to hear any feedback on this :)

ahlinc:
> Currently all of the tuple structs can just be manually constructed from the raw value by the user (e.g. Language(ptr) and Node(NonNull:new_unchecked(ptr)))

Actually this is not allowed:
```rust
error[E0423]: cannot initialize a tuple struct which contains private fields
  --> cli/src/parse.rs:89:29
   |
89 |         let tree = unsafe { Tree(NonNull::new(c_tree).unwrap()) };
   |                             ^^^^ help: a local variable with a similar name exists: `tree`
   |
note: constructor is not visible here due to private fields
  --> /home/ahlinc/DEV/MTF/JS/SAMPLES/TREE-SITTER/tree-sitter.origin/lib/binding_rust/lib.rs:47:17
   |
47 | pub struct Tree(NonNull<ffi::TSTree>);
   |                 ^^^^^^^^^^^^^^^^^^^^ private field
```

Because to allow this even an inner tuple members should be marked as `pub`.

> so perhaps the better solution would be to make the from_raw wrappers safe?

No, the Rust idea requires that all explicit operations with raw pointers should be unsafe and it's bad idea to broke this contract.

vhyrro:
Ah yeah true, my bad. `unsafe fn from_raw` it is then :smile:

vhyrro:
Okay, we're in the endgame now

A statement and a question to make sure everything is up to standard, and then I believe this PR is ready!:
- For the `Query::from_raw()` to work, a portion of the code had to be extracted, which naturally increased the complexity of `ffi.rs`. If this is not fine we can consider some alternate changes.
- Given that the `from_raw()` functions are marked `unsafe` I've taken the liberty of calling `NonNull::new_unchecked()`, but a non-undefined-behaviour alternative is `NonNull::new(...).unwrap()` or `NonNull::new(...).expect("Nice error message")`. From my understanding the current implementation is fine because this is semantically consistent with the meaning of an "unsafe function" and UB on incorrect input parameters is fine, can I get a confirmation? Soz, as you can see I'm not super well versed in the absolute best practices here :p

If both of those things are alright though then I await any potential further tips :)

ahlinc:
_Note: The PR changes become noisy after a rebase on top of the master branch, this due to latest massive CI/CD changes but which required to run CI properly. And it seems Github web interface doesn't show correctly the real state on a git tree._

Screenshot from 2023-04-04 14-22-48

vhyrro:
:purple_heart:

PR #2137, opened 2023-03-16
chore: remove all dead badges

ahlinc:


maxbrunsfeld:
thanks for doing this.

PR #2142, opened 2023-03-21
docs: alphabetise language bindings on front page

katrinafyi:
This helps with readability when scanning for a particular language. It is, strangely, almost sorted with the exception of Java and Kotlin.


ahlinc:
Thanks!

PR #2162, opened 2023-03-30
:skull: OOB access in query compiler

tomc1998:
In the case that `depth == 1`, the loop decrements depth to 0 & then accesses `self->stack[-1]` in `analysis_state__top`. I'm not sure if the fix I applied was correct, but it stopped it crashing for me & seems to work for me now.

Thought I'd send a PR just so this doesn't go unnoticed - someone who knows their way around probably wants a deeper look rather than just merging this.

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -1379,8 +1379,8 @@ static void ts_query__perform_analysis(
 
           // Pop from the stack when this state reached the end of its current syntax node.
           while (next_state.depth > 0 && next_state_top->done) {
-            next_state.depth--;
             next_state_top = analysis_state__top(&next_state);
+            next_state.depth--;
           }
 
           // If this hypothetical child did match the current step of the query pattern,


ahlinc:
@tomc1998 could you provide a link to a grammar, a source code sample and a query example for the `tree-sitter query ...` command that triggers OOB for an issue that you saw?

tomc1998:
It was the [java treesitter](https://github.com/tree-sitter/tree-sitter-java) w/ this query

```
(package_declaration _ (_) @name _)
```

But it should happen with any query that hits this loop

ahlinc:
> It was the [java treesitter](https://github.com/tree-sitter/tree-sitter-java) w/ this query
> 
> ```scheme
> (package_declaration _ (_) @name _)
> ```
> 
> But it should happen with any query that hits this loop

@tomc1998 I tried your query example and it doesn't reproducible:

Screenshot from 2023-04-04 15-34-03

tomc1998:
It's not going to segfault all the time, but it *will* always do a negative access, which is UB

E.g. if you compile with -fsanitize=undefined, it'll crash consistently

ahlinc:
> It's not going to segfault all the time, but it _will_ always do a negative access, which is UB
> 
> E.g. if you compile with -fsanitize=undefined, it'll crash consistently

Thanks for the note, it may have sense to add such flag in CI build to make it stricter.

ahlinc:
Reporoduced:
```sh
# cd tree-sitter/
# sudo dnf install libubsan.x86_64
# CFLAGS="-fsanitize=undefined" RUSTFLAGS="-lubsan" cargo run query --scope source.java <(echo (package_declaration _ (_) @name _)') <(echo 'package a; public interface A {}')

src/./query.c:941:22: runtime error: index -1 out of bounds for type 'AnalysisStateEntry [8]'
/dev/fd/62
  pattern: 0
    capture: 0 - name, start: (0, 8), end: (0, 9), text: `a`

# CFLAGS="-fsanitize=undefined" RUSTFLAGS="-lubsan" cargo test -- --test-threads 1

test tests::highlight_test::test_highlighting_cancellation ... src/./query.c:941:22: runtime error: index -1 out of bounds for type 'AnalysisStateEntry [8]'
ok

```

ahlinc:
I've added a reproducing test in:
- https://github.com/tree-sitter/tree-sitter/pull/2176

ahlinc:
@tomc1998 thank you for the report! It also helped to improve a CI/CD to catch OOB's in the future.

PR #2178, opened 2023-04-06
ci: build CLI for macos-arm64

### LINKED ISSUE #2009: node version of CLI fails on my M1 MacBook Pro
unfortunately was not able to reproduce on  other similar machines maybe unique to me?
On my machine running
`npm install --save-dev tree-sitter-cli`
followed by
`./node_modules/.bin/tree-sitter`
resulting in the following error
```
node:internal/child_process:413
    throw errnoException(err, 'spawn');
    ^

Error: spawn Unknown system error -86
    at ChildProcess.spawn (node:internal/child_process:413:11)
    at spawn (node:child_process:743:9)
    at Object.<anonymous> (~/tree-sitter-hello/node_modules/tree-sitter-cli/cli.js:8:1)
    at Module._compile (node:internal/modules/cjs/loader:1119:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1173:10)
    at Module.load (node:internal/modules/cjs/loader:997:32)
    at Module._load (node:internal/modules/cjs/loader:838:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:18:47 {
  errno: -86,
  code: 'Unknown system error -86',
  syscall: 'spawn'
}
```
expected output is a help menu from the cli. Other commands produce the same output.

clason:
closes #2009 

Caveat: can't run tests on `macos-arm64` as we can't use the built CLI on the host due to the wrong architecture.

Tested the build workflow on my fork; you can checkout the artefacts here: https://github.com/clason/tree-sitter/actions/runs/4629484854

I skipped building WASM for any target where it's not tested (and uploaded) to save resources, but that can of course be reverted (except for `macos-arm64`). More savings are possible by restricting this to only the target for which the build is actually uploaded (`linux-x64`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ env:
   CROSS_DEBUG: 1
 
 on:
+  workflow_dispatch:
   workflow_call:
     inputs:
       ref:
@@ -27,6 +28,7 @@ jobs:
           - { name: windows-x64   , target: x86_64-pc-windows-msvc      , os: windows-latest                   }
           - { name: windows-x86   , target: i686-pc-windows-msvc        , os: windows-latest                   }
           - { name: macos-x64     , target: x86_64-apple-darwin         , os: macos-latest                     }
+          - { name: macos-arm64   , target: aarch64-apple-darwin        , os: macos-latest                     }
 
     env:
       BUILD_CMD: cargo
@@ -125,30 +127,34 @@ jobs:
       run: make.sh CFLAGS="-Werror" -j
 
     - name: Build wasm library
+      if: ${{ !matrix.job.use-cross && matrix.job.name != 'macos-arm64' }} # Not used
       run: script/build-wasm
 
     - name: Build CLI
       run: $BUILD_CMD build --release --target=${{ matrix.job.target }}
 
     - name: Fetch fixtures
+      if: ${{ matrix.job.name != 'macos-arm64' }} # Not used
       run: script/fetch-fixtures
 
     - name: Generate fixtures
+      if: ${{ matrix.job.name != 'macos-arm64' }} # Can't run CLI on host
       run: script/generate-fixtures
 
     - name: Generate WASM fixtures
-      if: "!matrix.job.use-cross"
+      if: ${{ !matrix.job.use-cross && matrix.job.name != 'macos-arm64' }} # Not used
       run: script/generate-fixtures-wasm
 
     - name: Run main tests
+      if: ${{ matrix.job.name != 'macos-arm64' }} # Can't run CLI on host
       run: $BUILD_CMD test --target=${{ matrix.job.target }}
 
     - name: Run wasm tests
-      if: "!matrix.job.use-cross" # TODO: Install Emscripten into custom cross images
+      if: ${{ !matrix.job.use-cross && matrix.job.name != 'macos-arm64' }} # Not used
       run: script/test-wasm
 
     - name: Run benchmarks
-      if: "!matrix.job.use-cross" # It doesn't make sense to benchmark something in an emulator
+      if: ${{ !matrix.job.use-cross && matrix.job.name != 'macos-arm64' }} # Cross-compiled benchmarks make no sense
       run: $BUILD_CMD bench benchmark -p tree-sitter-cli --target=${{ matrix.job.target }}
 
     - name: Upload CLI artifact
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,6 @@ env:
   CROSS_DEBUG: 1
 
 on:
-  workflow_dispatch:
   workflow_call:
     inputs:
       ref:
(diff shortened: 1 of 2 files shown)


ahlinc:
Thank you for the PR but cross compile for ARM isn't simple due to Apple license restrictions for its components redistribution.
As a workaround I thought about to take an xcode libs from the macos runner and store them in an actions cache then use them in a custom cross image.

ahlinc:
Thank you for the PR but cross compile for ARM isn't simple due to Apple license restrictions for its components redistribution.
As a workaround I thought about to take an xcode libs from the macos runner and store them in an actions cache then use them in a custom cross image.

ahlinc:
https://github.com/cross-rs/cross-toolchains#apple-targets

clason:
I don't see how that applies here, as you are (presumably) using the legally installed SDKs of the macOS runner, and not redistributing anything?

To be clear, what I'm doing is compiling for M1 on an x86 host, which is a standard Rust feature. I'm not using a full cross-compilation environment (which would presumably allow _running_ and testing the cross-compiled build).

ahlinc:
Did you tried to run this artifact https://github.com/tree-sitter/tree-sitter/suites/12073390579/artifacts/635380239 on your M1?

ahlinc:
Maybe you are right and for building CLI it would be enough to compile for target only on a standard macos-x64 runner.

clason:
> Did you tried to run this artifact https://github.com/tree-sitter/tree-sitter/suites/12073390579/artifacts/635380239 on your M1?

```
~/Downloads ❯❯❯ file ./tree-sitter
./tree-sitter: Mach-O 64-bit executable arm64
~/Downloads ❯❯❯ ./tree-sitter
tree-sitter 0.20.8 (dddcb9cfc9a85d915a23ea9d1ee916ebdd8f7657)
Max Brunsfeld <maxbrunsfeld@gmail.com>
Generates and tests parsers

USAGE:
    tree-sitter <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    init-config       Generate a default config file
    generate          Generate a parser
    parse             Parse files
    query             Search files using a syntax tree query
    tags              Generate a list of tags
    test              Run a parser's tests
    highlight         Highlight a file
    build-wasm        Compile a parser to WASM
    playground        Start local playground for a parser in the browser
    dump-languages    Print info about all known language parsers
```

ahlinc:
> I would understand if the lack of testing for this target was a blocker for you

Yes, but I'm fine with just available built CLI for now.

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ env:
   CROSS_DEBUG: 1
 
 on:
+  workflow_dispatch:
```
ahlinc, line 9: Makes no sense because there is additional parameter for source code checkout on https://github.com/tree-sitter/tree-sitter/pull/2178/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L42
The main CICID workflow pretty quickly calls this workflow with correct additional inputs and should be used instead as a main entry point and it has `workflow_dispatch` trigger.

clason: `workflow_dispatch` is for manually triggering a workflow; we've found it quite useful on occasion, and it doesn't hurt otherwise. but here it's mainly left over from my testing on my fork; feel free to remove, of course.

ahlinc:
Thank you for the great addition! :tada:

clason:
My pleasure; happy I can contribute!

(But note that `npm install` still won't work until you've made a new release from master, or manually added a macos-arm64 binary to the 0.20.8 release. If you trust me, I can send you the binary I installed from homebrew, or can try to dig up how to download it directly from them -- which is not trivial.)

ahlinc:
@clason unfortunatelly I cannot do it right now, I still doesn't have a permission to the `tree-sitter-cli` package on `npmjs.com` yet and waiting for a reply from @maxbrunsfeld on my emails. I'll update and complete all things when it would become possible.

clason:
I was referring to that bit in the code:
https://github.com/tree-sitter/tree-sitter/blob/978a2db239c553e80c7a1baecede1a8fdc9f34a8/cli/npm/install.js#L30-L32

But that of course only covers local installation (`npm install .`), not from the global registry.

ahlinc:
> But that of course only covers local installation (`npm install .`), not from the global registry.

If I don't miss something it exactly works in a scenario:
```sh
# mkdir tree-sitter-foobar
# cd tree-sitter-foobar/
# npm add -D tree-sitter-cli

added 1 package in 1s
# tree -a node_modules/
node_modules/
├── .bin
│   └── tree-sitter -> ../tree-sitter-cli/cli.js
├── .package-lock.json
└── tree-sitter-cli
    ├── cli.js
    ├── dsl.d.ts
    ├── install.js
    ├── LICENSE
    ├── package.json
    └── tree-sitter                # <-- this was downloaded by `install.js`

3 directories, 8 files
```

clason:
Yup, that was my understanding as well. But on macos-64, that won't work, since it tries to download a binary that does not exist (yet).

ahlinc:
I've published `tree-sitter-cli` [v0.20.8](https://www.npmjs.com/package/tree-sitter-cli/v/0.20.8) on npm so the `npm i -D tree-sitter-cli` command should work fine on M1.

PR #2182, opened 2023-04-07
Docs: Add tree-sitter-just to the parser list

futile:
This PR adds [tree-sitter-just](https://github.com/IndianBoy42/tree-sitter-just) to the list of parser.

`just` is the format used by `justfile`s: https://github.com/casey/just/

This parser is also mentioned in the `README.md` of `casey/just`, so it is also "blessed" by the project itself.

Since highlighting for `just`-code-blocks seems to already work on github, I'm wondering if this/another parser for `justfile`s is already known? But maybe github just uses a parser that's not yet listed in the docs.


amaanq:

PR #2183, opened 2023-04-07
fix: Use / paths when building WASM

### LINKED ISSUE #532: The build-wasm subcommand does not work within Docker on windows
Hi, I am trying to build the .wasm file for a custom parser, on Windows 10 with Docker. I failed to find proper documentation on that but from what I saw it should be pretty straightforward, just running `tree-sitter build-wasm` after `tree-sitter generate` in the language folder. Generate command works perfectly, the tests are running without an error, but when it comes to the building part I get this error:

`emcc command failed - shared:ERROR: src\parser.c: No such file or directory ("src\parser.c" was expected to be an input file, based on the commandline arguments provided)`

Beware that this is not a file or directory issue as I would also encounter the error during the generate or test step, and also the error is different if I enter the path in a wrong way, e.g.:

`Failed to read grammar file "C:\\Users\\---\\Documents\\GitHub\\---\\src\\server\\---_server\\---_tree_sitter\\error\\src\\grammar.json"
Details:
  The system cannot find the path specified. (os error 3)`

I suspect that it has something to do with emcc but I can't be sure. Any help would be appreciated.

joelspadin:
Changed the build-wasm command to always use forward slashes in paths, since using Windows style paths breaks if the build is run with Docker.

Fixes #532


joelspadin:
I should probably note that I don't really know what I'm doing with Rust, so please double check that I haven't done anything dumb here, but I've tested this on Windows + Docker and WSL + Docker.

ahlinc:
@joelspadin Great job! Thank you!
I'll merge this after #2185 will be completed.

Issue #1540, opened 2021-12-15
Incremental highlighting


dickmao:
https://github.com/tree-sitter/tree-sitter/blob/8bbf75a067cf5c52d0d11609db375945dfdd2a74/highlight/src/lib.rs#L353

The extant `HighlightIterLayer::new` reparses the source, which is suitable for stateless cli invocation of `tree-sitter highlight`, but is less apropos for an incremental highlighter, say a text editor.

In my fork, I introduced a new `HighlightIterLayer` method  which accepts and queries a `Tree` already being kept up to date in Commercial Emacs.

Perhaps the `highlight` submodule was only ever intended for the cli, which begs the question how Atom is leveraging tree-sitter for incremental highlighting.

izzyreal:
> In my fork, I introduced a new HighlightIterLayer method which accepts and queries a Tree already being kept up to date

@dickmao  I'm working on a small IDE written in C/C++. Can I use your approach there to achieve incremental highlighting? Can you point to the branch and code location where you made changes in your fork?

Many thanks in advance.

dickmao:
The critical function is `ts_highlighter_return_highlights`:
https://github.com/commercial-emacs/tree-sitter/blob/5b282dac8012d55103eb55ce8cf55726efa1cfaf/highlight/src/c_lib.rs#L194

It is used in Commercial Emacs here:
https://github.com/commercial-emacs/commercial-emacs/blob/d34154f1fec9c3fed8b74875572edfa4e2f15d70/src/tree-sitter.c#L364

izzyreal:
Awesome, I'll take a look

dundargoc:
PRs welcome.

PR #2200, opened 2023-04-11
Highlighting Existing Trees

liamrosenfeld:
Allows using the highlight interface to query an existing tree. Does not support injections because the idea is to have it not parse anything.

This enables incremental highlighting because you can modify the tree outside of the highlighter (Fixes: #1540)

### DIFF
--- a/highlight/src/lib.rs
+++ b/highlight/src/lib.rs
@@ -103,7 +103,7 @@ where
 }
 
 struct HighlightIterLayer<'a> {
-    _tree: Tree,
+    _tree: Option<Tree>, // needed to keep tree in memory
     cursor: QueryCursor,
     captures: iter::Peekable<QueryCaptures<'a, 'a, &'a [u8]>>,
     config: &'a HighlightConfiguration,
@@ -155,7 +155,32 @@ impl Highlighter {
             cancellation_flag,
             highlighter: self,
             iter_count: 0,
-            layers: layers,
+            layers,
+            next_event: None,
+            last_highlight_range: None,
+        };
+        result.sort_layers();
+        Ok(result)
+    }
+
+    /// Iterate over the highlighted regions for a given node.
+    /// Does not parse anything (and therefore does not support injections).
+    pub fn highlight_existing_tree<'a>(
+        &'a mut self,
+        source: &'a [u8],
+        node: &'a Node,
+        config: &'a HighlightConfiguration,
+        cancellation_flag: Option<&'a AtomicUsize>,
+    ) -> Result<impl Iterator<Item = Result<HighlightEvent, Error>> + 'a, Error> {
+        let layers = vec![HighlightIterLayer::new_from_tree(source, node, config)];
+        let mut result = HighlightIter {
+            source,
+            byte_offset: 0,
+            injection_callback: |_| None,
+            cancellation_flag,
+            highlighter: self,
+            iter_count: 0,
+            layers,
             next_event: None,
             last_highlight_range: None,
         };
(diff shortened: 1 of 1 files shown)


ahlinc:
@liamrosenfeld thank you for the proposed contribution, but I'll postpone it yet. I'm not sure that moving into direction of having multiple niche case specific methods is a good idea for the library. I thought before about an another approach of significantly change and improve the highlighting API by introducing something like `trait TreeProvider` that will allow to pass a current and an old trees and been called on every injection processing. But such changes also need to take into account reasons why other Tree-sitter consumers like editors implement totally custom own highlighting implementations instead of using this highlighting library or contributing in it by improving its API and facilities.

liamrosenfeld:
That makes sense. This is definitely a stopgap, and a more uniform solution would be preferable.

I implemented this because my project already used the tree directly, so highlighting requiring an entire second parse (which additionally didn't support incremental parsing) was causing a performance bottleneck. I feel like this is a reasonably common use case (since some editors that use tree sitter for highlighting additionally use the tree for other things such as breadcrumbs). If there is anything that could make this less of a special case, just let me know and I could try to implement it.

As for other editors not using the official highlighting library, my guess is that not supporting the tree sitter TextProvider is a barrier because many of them use text ropes (where it isn't performant to get a &[u8]). The main difference I could spot in helix's highlighting code is [this line](https://github.com/helix-editor/helix/blob/406c5c38a1bb0475579c7d4cd1fd7a90f09d543c/helix-core/src/syntax.rs#L1111). That is just my guess, the opinion of someone who actually worked on that would probably be much more informative.

ahlinc:
> If there is anything that could make this less of a special case, just let me know and I could try to implement it.

Unfortunately, I don't have an additional time for now to review all changes made in different editors to get a whole and solid picture of reasons about why and what customizations in highlighting approaches they made and after that write a plan of what changes are needed to make this lib more useful and be used as a one good common solution for all consumers. As I understand for now it just used by CLI as a highlighting demo implementation and not more.

liamrosenfeld:
That's all good. Now that this has piqued my curiosity, I'll probably do a more in-depth review of other editors on my own. If a solid plan arises from that, I'll let you know.

clason:
>  But such changes also need to take into account reasons why other Tree-sitter consumers like editors implement totally custom own highlighting implementations instead of using this highlighting library or contributing in it by improving its API and facilities.

1. Tree-sitter is not the only source of highlighting (e.g., LSP semantic tokens)
2. Not every project uses Rust (yet)...

clason:
>> Not every project uses Rust (yet)...

> The lib already exposes C API.

Yes, but the implementation is still in Rust, so this adds a whole new toolchain to the build system for a C project. At least for Neovim, this is unacceptable (our current build system is horrid enough, even after heroic efforts to drag it kicking and screaming into the twentieth century).

> At least I'd like to see most of the features that that can be generalized and API changes that would allow incremental highlighting instead of a current limited approach of a one pass over full input that is usable only in CLI tools and cases where only static highlighting is enough.

Sure, but that still means we'd have to integrate _two_ different highlighters instead of adding tree-sitter parsing and querying into the already existing highlighter. For an existing project with a lot of legacy code, the latter can be significantly easier.

In both cases, the cost simply outweighs the (perceived) benefits for us. 

(I'm just _explaining_ why specifically Neovim doesn't use the tree-sitter highlighter, since you asked; I'm not trying to convince you that this is the only possibility. I do appreciate the chicken-and-egg situation here, where increasing the benefits -- and thus the likelihood of using it -- happens more easily with more editors contributing.)

I would specifically ask the Helix devs in chat about this, though, as this project is 1. pretty young and tree-sitter first, 2. written in Rust, and 3. appears to vendor or at least very closely follow the tree-sitter highlighter...

ObserverOfTime:
Doesn't the [highlight crate](https://github.com/tree-sitter/tree-sitter/tree/master/highlight) handle this?

PR #2230, opened 2023-04-25
discussion: allow external scanners to resolve references to already parsed nodes

3p3r:
Hello! I'm opening this PR to get some feedback from maintainers about the direction I am going with this. Some context:

There are some DSLs that have "references" embedded in them. For example [Terraform HCL](https://developer.hashicorp.com/terraform/language/expressions/references) or CloudFormation JSON both have this capability. You can for example do [`{ "Ref": "SomeOtherNodeId" }`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html) where that particular static string is always 100% mentioned in some way or form in another Node in the same tree and its uniqueness is guaranteed by the DSL.

So to allow scanners to figure these Nodes out, scanners fill their debug buffer with placeholders of where these "references" are located, and then once the `TSTree` is done parsing, scanners get a chance to look at the entire tree and figure out what those references actually refer to. If they need to, they can update their own buffer that's embedded in the tree to have metadata back to users about these references.

There is a fine line here where you might argue that this needs to be inside a compiler implementation and not a parser/lexer, however I figured if this can be done quickly within a few microseconds inside the scanner, then why not?

I am also curious to see if maintainers have an alternative approach to solving this issue in these sorts of grammars.


ahlinc:
You've replied by yourself:
> There is a fine line here where you might argue that this needs to be inside a compiler implementation and not a parser/lexer,

> however I figured if this can be done quickly within a few microseconds inside the scanner, then why not?

It can't be resolved in microseconds, many of such references in Terraform and Cloudformation targets complex values, for Terraform they can be on other files and for Cloudformation some of them may be computed in a context that can't be exist for the external scanner. And in any way all such machinery wouldn't work in microseconds and a the caller code that expects the parser result for ex. for highlighting rendering would stuck and wait all such complex resolution tasks in the external scanner. Such things are out of scope for the Tree-sitter project.

_More info in: https://github.com/tree-sitter/tree-sitter/discussions/2520#discussioncomment-6753870_

3p3r:
FWIW the scope of this solution was limited to simple strings and not more complicated dynamic references. For example a direct `Fn::Ref` to a string would work, but if that's wrapped in a `Fn::Base64`, it would parse as a simple object pair or a base Expression rule.

I also ended up hitting a wall with what you suggested, and that's external references to other documents. It was absolutely fun to fork tree sitter and work on it though!

amaanq:
> I also ended up hitting a wall with what you suggested, and that's external references to other documents. It was absolutely fun to fork tree sitter and work on it though!

I can imagine :) I'm glad you feel that way, I do as well when working on it (psst, there are of course other issues you can help out with if you love tree-sitter :grin:  )

PR #2232, opened 2023-04-29
Add build.zig for easy use as a Zig library

SuperAuguste:
Hey tree-sitters, I'm a Zig community member working on editor tooling for the language and was experimenting with tree-sitter and using it in Zig.

Adding a build.zig would make it seamless to use via the Zig package manager.

I understand if you don't merge this because it's quite a (comparatively) niche usecase, I can just maintain a fork with this change rebased instead :)

Thanks!


TiAmo91 (approved):

ahlinc:
There is a simple entrypoint file for Swift's PM so I don't see objections to don't add a similar simple entrypoint for Zig's PM.

ahlinc:
@SuperAuguste for the next time, please don't disable permission for maintainers to update a PR branch.

ahlinc:
I've opened an another PR with a rebased commit to avoid long branch merges.

PR #2238, opened 2023-05-02
fix: use SIGINT instead of stdin for interruption to don't mess up piped commands

### LINKED ISSUE #2133: `tree-sitter highlight` cli is reading from stdin when not appropriate
I'm trying to do this:

```shell-session
$ tree-sitter highlight myfile.whatever | less -R
```

The symptom I'm seeing is that it `less` doesn't reliably respond to keystrokes. It takes multiple presses of a key to get it to respond. I can reproduce this on macOS (tree-sitter coming from either Homebrew or Nix) and my Debian box (debian package).

I tracked it down and it seems that tree-sitter is reading from stdin in certain situations and that ends up stealing the keystrokes from `less` more or less randomly.

Is this intended behavior? I didn't see it documented and a quick skim through the highlight code didn't reveal anything obvious to me, but I know it's happening.

It's easier to debug on linux:

```shell-session
$ strace -f -o /tmp/strace.out bash -c 'tree-sitter highlight some_file | less -R'
```

Then in another terminal I do
```shell-session
$ pstree -lp 4079155
bash(4079155)─┬─less(4079157)
              └─tree-sitter(4079156)───{tree-sitter}(4079158)
$ lsof 4079157
COMMAND     PID  USER   FD   TYPE DEVICE SIZE/OFF       NODE NAME
... irrelevant lines snipped ... 
less    4079157 david  txt    REG  259,1   198960   19834528 /usr/bin/less
less    4079157 david    0r  FIFO   0,13      0t0 3077288649 pipe
less    4079157 david    1u   CHR 136,35      0t0         38 /dev/pts/35
less    4079157 david    2u   CHR 136,35      0t0         38 /dev/pts/35
less    4079157 david    3r   CHR 136,35      0t0         38 /dev/pts/35
$ lsof 4079156
COMMAND       PID  USER   FD   TYPE DEVICE SIZE/OFF       NODE NAME
... irrelevant lines snipped ... 
tree-sitt 4079156 david  txt    REG  259,1 12360272   19832904 /usr/bin/tree-sitter
tree-sitt 4079156 david    0u   CHR 136,35      0t0         38 /dev/pts/35
tree-sitt 4079156 david    1w  FIFO   0,13      0t0 3077288649 pipe
tree-sitt 4079156 david    2u   CHR 136,35      0t0         38 /dev/pts/35
```

Then I check the `/tmp/strace.out` and I see:

```
4079158 read(0,  <unfinished ...>
... lots of stuff cut ...
4079157 read(3,  <unfinished ...>
... lots of stuff cut, I press space in the terminal running less ...
4079158 <... read resumed>" ", 8192)    = 1
4079158 read(0,  <unfinished ...>
```

That shows pid 4079158 (a tree-sitter thread) reading stdin (which bash left open as a pty since it was the first command in the pipeline) and pid 4079157 (less) reading from fd 3, which is the pty that it manually opened to take input from the user. We can see the space I pressed got read by tree-sitter and not less.

It also seems to have to do with the size of the output.

This create some test javascript files. The first set has higher syntax density than the 2nd set, but both input files in each set are the same length:
```shell
for k in $(seq 10 20); do \
  yes 'x="abcd";' | head -${k}00 > /tmp/high-${k}k.js; \
  echo -n "high-${k}k.js: "; \
  tree-sitter highlight /tmp/high-${k}k.js | wc -c; \
done
for k in $(seq 20 30); do \
  yes '//comment' | head -${k}00 > /tmp/low-${k}k.js; \
  echo -n "low-${k}k.js: "; \
  tree-sitter highlight /tmp/low-${k}k.js | wc -c; \
done
```
Those are labeled by their input size, and the loops run tree-sitter to print the output sizes. This will probably vary by theme. With my config file I see this:
```
high-10k.js: 57000
high-11k.js: 62700
high-12k.js: 68400
high-13k.js: 74100
high-14k.js: 79800
high-15k.js: 85500
high-16k.js: 91200
high-17k.js: 96900
high-18k.js: 102600
high-19k.js: 108300
high-20k.js: 114000
low-20k.js: 50000
low-21k.js: 52500
low-22k.js: 55000
low-23k.js: 57500
low-24k.js: 60000
low-25k.js: 62500
low-26k.js: 65000
low-27k.js: 67500
low-28k.js: 70000
low-29k.js: 72500
low-30k.js: 75000
```

If I try those out, `high-12k.js` and `low-27k.js` consistently work, but `high-13k.js` and `low-28k.js` consistently fail.
So it seems like the failure boundary is somewhere between 68,400 bytes and 70,000 bytes.

The results are consistent for me on both Mac and Linux.

ahlinc:
Fixes #2133 

This makes Tree-sitter CLI usable with the _`less -r`_ command to page long outputs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
+checksum = "67fc08ce920c31afb70f013dcce1bfc3a3195de6a228474e45e1f145b36f8d04"
 dependencies = [
  "memchr",
 ]
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.70"
+version = "1.0.71"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7de8ce5e0f9f8d88245311066a578d72b7af3e7088f32783804676302df237e4"
+checksum = "9c7d0618f0e0b7e8ff11427422b64564d5fb0be1940354bfe2e0529b18a9d9b8"
 
 [[package]]
 name = "ascii"
@@ -57,9 +57,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bumpalo"
-version = "3.12.0"
+version = "3.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d261e256854913907f67ed06efbc3338dfe6179796deefc1ff763fc1aee5535"
+checksum = "9b1ce199063694f33ffb7dd4e0ee620741495c32833cde5aa08f02a0bf96f0c8"
 
 [[package]]
 name = "bytes"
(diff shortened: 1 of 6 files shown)


limck301:
include/spirv/unified1/OpenCL.std.h

PR #2257, opened 2023-05-12
add the parser for apex language

jsuarez-chipiron:
Hi, long time ago I created a grammar for Salesforce Apex language based on the java existing grammar. I've been using it mostly from neovim but also with the rust and python bindings and used to perform static analysis in very large code base of Apex.

Here is the repo where my grammar is: https://github.com/jsuarez-chipiron/tree-sitter-apex

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -35,6 +35,7 @@ There are currently bindings that allow Tree-sitter to be used from the followin
 * [Ada](https://github.com/briot/tree-sitter-ada)
 * [Agda](https://github.com/tree-sitter/tree-sitter-agda)
 * [Apex](https://github.com/aheber/tree-sitter-sfapex)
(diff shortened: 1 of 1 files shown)


rody:
Thanks for that, I'll try it asap.

For info, there is another repo for the Apex language, as well as SOQL and SOSL located at: https://github.com/aheber/tree-sitter-sfapex

I've used this one in my Apex mode for emacs and it seems to be working quite well too.
It could be interesting to merge the efforts of maintaining the grammar between the two projects.

Note: I'm not the maintainer of the other grammar, I just use it in my projects.

jsuarez-chipiron:
Thanks @rody ,

I just tryid the grammar from aheber and looks better than mine. I need to do a regression on all the functionally that I have build on top of mine grammar but I agree with you that is pointless to have two different grammars for apex.

amaanq:

Issue #1922, opened 2022-10-20
Lifetimes are too restrictive in `QueryCursor::matches`


oberblastmeister:
In https://docs.rs/tree-sitter/latest/tree_sitter/struct.QueryCursor.html#method.matches the `text_provider` gets bound as
```rust
T: TextProvider<'a> + 'a
```
However, the method also takes 
```rust
&'a mut self
```
So the cursor is required to have the same lifetime as the `TextProvider`, but it may not be able to. I made this change https://github.com/oberblastmeister/tree-sitter/blob/lifetime/lib/binding_rust/lib.rs#L175, but I wanted to ask first.

amaanq:
Is this still an issue on master? TextProvider got reworked

ahlinc:
I think this can be closed because `TextProvider` doesn't depend from lifetimes directly anymore.
Related:
* #2431 
* #2464

PR #2273, opened 2023-05-18
Rework lifetimes

smoelius:
This PR proposes to rework how `QueryMatches`, `QueryMatch`, and `QueryCaptures` use lifetimes.

With the current lifetimes, this function does not compile: https://github.com/smoelius/tree-sitter/blob/19c0a5591e2ca7cebb5a75323a56d6a3127522fc/cli/src/tests/query_test.rs#L4590-L4610

With the changes proposed in this PR, the function compiles.

The current lifetimes have other adverse consequences. For example, both [`QueryMatches`](https://github.com/smoelius/tree-sitter/blob/2a9d951cd6faf8367a402210c406f0f64818aa24/lib/binding_rust/lib.rs#L175) and [`QueryCaptures`](https://github.com/smoelius/tree-sitter/blob/2a9d951cd6faf8367a402210c406f0f64818aa24/lib/binding_rust/lib.rs#L185) require that the tree outlive the query. But there is no obvious reason for such a requirement. Moreover, it prevents applications from building a set of queries then operating on multiple trees, for example.

Additional reasons/discussion are given in #1922.

To state the obvious, merging this PR would be a breaking change.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -4591,3 +4591,36 @@ fn test_query_error_does_not_oob() {
         }
     );
 }
+
+#[test]
+fn test_query_and_cursor_have_distinct_lifetimes() {
+    let language = get_language("javascript");
+    let query = Query::new(language, "(if_statement)").unwrap();
+    let source = "return a + b - c;";
+    let mut parser = Parser::new();
+    parser.set_language(language).unwrap();
+    let tree = parser.parse(source, None).unwrap();
+    let _captures = is_match(&query, tree.root_node(), source.as_bytes());
+}
+
+fn is_match<'query, 'source, 'tree, T>(
+    query: &'query Query,
+    node: Node<'tree>,
+    text_provider: T,
+) -> Option<Vec<tree_sitter::QueryCapture<'tree>>>
+where
+    T: tree_sitter::TextProvider<'source> + 'source,
+{
+    let mut cursor = QueryCursor::new();
+
+    cursor.set_max_start_depth(1);
+
+    let mut query_matches = cursor.matches(query, node, text_provider);
+
+    if let Some(query_match) = query_matches.next() {
+        assert!(query_matches.next().is_none());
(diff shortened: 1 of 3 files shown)


smoelius:
Can I politely ask what your thoughts are on this?

ahlinc:
It looks good and I agree that current lifetimes aren't perfectly modeled but I'd like to make some experiments first because recently I figured out other issues in related placed and would like to check all things before merging this PR.

ahlinc:
@smoelius could you recheck does this still actual for you on the master?

_A fixed code:_
```rust
#[test]
fn test_query_and_cursor_have_distinct_lifetimes() {
    let language = get_language("javascript");
    let query = Query::new(language, "(if_statement)").unwrap();
    let source = "return a + b - c;";
    parser.set_language(language).unwrap();
    let tree = parser.parse(source, None).unwrap();
    let _captures = is_match(&query, tree.root_node(), source.as_bytes());
}

fn is_match<'query, 'source, 'tree, T>(
    query: &'query Query,
    node: Node<'tree>,
    text_provider: T,
) -> Option<Vec<tree_sitter::QueryCapture<'tree>>>
where
    T: tree_sitter::TextProvider<&'source [u8]>,
{

    cursor.set_max_start_depth(1);

    let mut query_matches = cursor.matches(query, node, text_provider);

    if let Some(query_match) = query_matches.next() {
        assert!(query_matches.next().is_none());
        Some(query_match.captures.to_vec())
    } else {
        None
    }
}
```
_Compiles without issues on the master, I suppose it's after #2431_

ahlinc:
I still believe that lifetimes should be improved but that test seems isn't relevant anymore.

ahlinc:
I'd suggest to rework lifetimes schema as in the following PR:
* #2464

ahlinc:
I'm closing this because after changes to lifetimes schema and to `TextProvider` this PR is totally incompatible in its changes with the master.

PR #2274, opened 2023-05-18
Work around tree-sitter-typescript#244

xeger:
This allows `tree-sitter-typescript` and numerous other NPM modules to work correctly under Electron 22.

I'm not sure if it's the right fix. I'm not sure if it has unintended consequences. But it lets my [VS Code extension](https://github.com/cucumber/vscode) load and run again!

(Note: I've published a forked build of `web-tree-sitter` to prove the concept of this workaround, and the alterations to `package.json` are present on this branch. Don't merge as-is -- not that I'd expect anyone to be so bold without a thorough understanding of the issue!)

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -908,7 +908,7 @@ class Language {
       : loadWebAssemblyModule;
 
     return bytes
-      .then(bytes => loadModule(bytes, {loadAsync: true}))
+      .then(bytes => loadModule(bytes, {allowUndefined: true, loadAsync: true}))
       .then(mod => {
         const symbolNames = Object.keys(mod)
         const functionName = symbolNames.find(key =>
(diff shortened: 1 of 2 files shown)


ahlinc:
I don't think that allowing _**undefined** symbols_ would resolve the issue, I suppose it just postponing a crash until the code would try to call undefined function.

xeger:
That's likely the case -- in which case it seems like my use case is just lucky, and doesn't call the undefined exports (or guards against undefined, etc).

If the export table is implicit to the WASM module -- which I would expect it to be -- then I am _very_ confused about the error. The exports clearly exist in the disassembled WASM I've checked. So, it's something about the WASM loader in Electron 22 (and whatever underlying v8 is being used).

sogaiu:
I think this is related to #2338.

FWIW, in addition to the `allowUndefined: true` method, another work-around I found that worked for my setup was:

```javascript
   .then(bytes => loadModule(bytes, {}))
```

PR #2281, opened 2023-05-20
Handle edge cases involving consecutive "zero or" modifiers

smoelius:
This PR addresses some edge cases that could result in infinite loops/invalid memory references.

Note that while those problems are gone, and the tests pass, I am not certain that the fixes are correct. Please scrutinize.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -4674,3 +4674,49 @@ fn test_query_error_does_not_oob() {
         }
     );
 }
+
+#[test]
+fn test_consecutive_zero_or_modifiers() {
+    let language = get_language("javascript");
+    let mut parser = Parser::new();
+    parser.set_language(language).unwrap();
+
+    let zero_source = "";
+    let three_source = "/**/ /**/ /**/";
+
+    let zero_tree = parser.parse(zero_source, None).unwrap();
+    let three_tree = parser.parse(three_source, None).unwrap();
+
+    let tests = [
+        "(comment)*** @capture",
+        "(comment)??? @capture",
+        "(comment)*?* @capture",
+        "(comment)?*? @capture",
+    ];
+
+    for test in tests {
+        let query = Query::new(language, test).unwrap();
+
+        let mut cursor = QueryCursor::new();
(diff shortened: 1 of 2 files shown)


cli/src/tests/query_test.rs
```diff
@@ -4591,3 +4591,48 @@ fn test_query_error_does_not_oob() {
...
+                    if m.captures.len() == 1 {
+                        len_1 = true;
+                    }
+                }
+
+                assert_eq!(len_3, x == '*' || y == '*' || z == '*');
+                assert_eq!(len_1, x == '?' && y == '?' && z == '?');
+            }
+        }
+    }
```
maxbrunsfeld, line 4637: This is a stylistic thing, but personally, I'd prefer to see this test written in a more concrete way, with two or three different hand-written queries, instead of the queries being generated via the for loops. It'd be a bit longer, but that way, we could make specific assertions about which matches we expect, for each query.

I think just `(comment)*** @capture` and `(comment)??? @capture` would be enough.

lib/src/query.c
```diff
@@ -2596,8 +2596,11 @@ static TSQueryError ts_query__parse_pattern(
       repeat_step.alternative_is_immediate = true;
       array_push(&self->steps, repeat_step);
 
+      // Stop when `step->alternative_index` is `NONE` or it points to
+      // `repeat_step` or beyond. Note that having just been pushed,
+      // `repeat_step` occupies slot `self->steps.size - 1`.
       QueryStep *step = &self->steps.contents[starting_step_index];
-      while (step->alternative_index != NONE) {
+      while (step->alternative_index != NONE && step->alternative_index < self->steps.size - 1) {
```
maxbrunsfeld, line 2603: This fix seems reasonable to me.

amaanq (approved):

samirbani:
في الأربعاء، ١٩ يوليو ٢٠٢٣ ١٠:٢٩ ص Amaan Qureshi ***@***.***>
كتب:

smoelius:
Sorry, I just noticed the comments on this PR.

@amaanq It looks like you addressed them. Thank you very much for doing that.

maxbrunsfeld (approved):

PR #2289, opened 2023-05-25
Return early if syntax errors are encountered

criminosis:
Hi Tree-Sitter team!

In my use case of Tree-Sitter I've been parsing source code samples but was wanting to early exit if the file had an error during the parsing. I followed in the footsteps of the other parser settings (exit flag and timeout) for the rust binding.

Full disclosure I'm not completely sure if I got all the web asm binding side of things, I'm primarily coming in through the Rust binding side. But I tried to mimic everywhere I saw the timeout exit being exposed.

Happy to make changes!

### DIFF
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -592,6 +592,28 @@ fn test_parsing_on_multiple_threads() {
 }
 
 #[test]
+fn test_early_out_on_syntax_error() {
+    let mut parser = Parser::new();
+    parser.set_language(get_language("python")).unwrap();
+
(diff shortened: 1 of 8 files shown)


ahlinc:
@criminosis Thank you for the PR and the idea, but I added a proposal #2541 with a better and more generic design.

Allan-Clements:
@ahlinc Glad to hear someone else liked the idea, I was beginning to worry I had fundamentally misunderstood my approach in my PR 😅 .

Checked out your proposal, it makes a lot of sense. Hope it gets some momentum!

criminosis:
@ahlinc Glad to hear someone else liked the idea, I was beginning to worry I had fundamentally misunderstood my approach in my PR 😅 .

Checked out your proposal, it makes a lot of sense. Hope it gets some momentum!

PR #2309, opened 2023-06-08
Add ts_node_language() that returns the language of the node

Jille:
Since an input might have a nested sublanguage (like Javascript inside Vue) and symbols are per-language, we need to know which language a node is.


patrickt (approved):
Makes sense to me given the existence of injections and the like.

amaanq:
@Jille is there a difficult problem that exposing this helps solve? An example would be nice

ahlinc:
The same method exists in Rust binding API `Node::language()` so it makes sense to add it on the C level at least for a feature parity and that other high level bindings would be able to use it. Also it make sense to use it directly in the Rust binding.

Jille:
@amaanq I want to parse languages (like HTML) which contain other languages (like JS) and still get useful Symbol numbers, which are specific to a language. (Each language uses Symbol numbers starting from 1)

ahlinc (approved):

PR #2311, opened 2023-06-08
Add Support for Swift Package Manager

ActuallyTaylor:
This pull request adds support for Swift Package Manager. This allows Swift projects to integrate with Tree-Sitter in a quicker and much more efficient manner. You no longer have to manually include the source, it can just be included as a SPM dependency.

This branch makes only two changes. The first is adding a `Package.swift` file that defines the swift package and allows this repository to be added as a SPM dependency. The second is adding `.swiftpm` to the `.gitignore` this is so workspace files generated by Xcode are not automatically included in git pushes.

This branch also passes all [tests](https://tree-sitter.github.io/tree-sitter/contributing#testing) outlined in the documentation.


ahlinc:
The `Package.swift` already exists in a Swift binding: https://github.com/ChimeHQ/SwiftTreeSitter/blob/main/Package.swift
@ActuallyTaylor could you explain reasons to add it to the main repo for the  C code  that is foreign for Swift?

ActuallyTaylor:
> The `Package.swift` already exists in a Swift binding: https://github.com/ChimeHQ/SwiftTreeSitter/blob/main/Package.swift
> 
> @ActuallyTaylor could you explain reasons to add it to the main repo for the  C code  that is foreign for Swift?

Including `Package.swift` in the main repository would make project maintainers much more comfortable with using TreeSitter in their swift projects. Not having to rely on a third party to update their fork every time the main repository updates would provide security benefits and peace of mind.

On top of that, Swift itself supports C and C++ natively: https://developer.apple.com/documentation/swift/using-imported-c-structs-and-unions-in-swift

You are able to call C functions directly from within swift making this library extremely easy to use and include into Swift projects. 

Even though the project itself does not contain any Swift code, the trivial interoperability between Swift and C makes TreeSitter an amazing choice to use in Swift based projects. Offering first party support for Swift Package Manager is only a singular file and would provide both security and piece of mind to anyone currently relying on third party repositories that add this file.

ahlinc:
Ok, as I see it was added to grammar repos https://github.com/tree-sitter/tree-sitter-cpp/pull/160 and it's just a package manager entry point (not an alternative build system) so there is no reasons to don't add it.

I would only ask to eliminate non meaningful commits by before merging this.
_The `git rebase --interactive` is helpful for this to reorder commits and mark as fixups commits that needs to be included to others._

ActuallyTaylor:
> I would only ask to eliminate non meaningful commits by before merging this

All of the commits have been squashed into a new single commit.

PR #2316, opened 2023-06-12
Add APIs for retrieving descendant counts, seeking to nodes by their descendant index

maxbrunsfeld:
### Background

When rendering a syntax tree, nodes are typically displayed in depth-first traversal order. In order to render an arbitrary *portion* of a depth-first walk of a syntax tree, it's useful to be able to start walking a tree at a given *index* within a depth-first walk.

### Change

This PR adds the following related C APIs, which expand the ability to work with depth-first walks of a syntax tree.

* `ts_node_descendant_count` - How many total descendant nodes does this node contain, including the node itself?
* `ts_tree_cursor_goto_descendant` - Move the cursor to the node that would appear at the given index in a depth-first walk of the tree
* `ts_tree_cursor_current_descendant_index` - Retrieve the index of the cursor's current node in a depth-first walk of the tree
* `ts_tree_cursor_current_depth`  - Retrieve the current depth of the tree cursor in the tree.

### Implementation Notes

This PR doesn't add any overhead in the syntax tree. It was already necessary to maintain on each syntax node a field called `node_count`, which was similar to `descendant_count`, and was used during error recovery as a measure of how much parsing progress had been made since discovering an error.

I've slightly changed the meaning of that field, so that it stores a node's visible descendant count (it now excludes *hidden* nodes from the count), which can serve the same purpose, for error recovery, but can also be used as the basis for the new APIs.

### Future Follow-up Steps

* Add WASM bindings for these new APIs
* Use these APIs to make the rendering of the Tree-sitter playground more efficient.


maxbrunsfeld:
@ahlinc I'm noticing that with the new CI workflow, the build appears as a failure, even when the tests pass on all platforms. Do you have a sense of what causes that?

ahlinc:
@maxbrunsfeld I suppose it's due to a change in some fixture grammar because for now we don't use commits and point to master branches. Let me see more detailed tomorrow and fix it, after the fix I'll rebase your PR branch on a fixed master.

Thank you for this great improvement I saw your branch yesterday.

ahlinc:
This PR is more performant version of what I previously wrote in Rust: https://github.com/ahlinc/tree-sitter/blob/alpha/cli/src/render.rs#L633-L661

Tisha5252:
> ### Background
> When rendering a syntax tree, nodes are typically displayed in depth-first traversal order. In order to render an arbitrary _portion_ of a depth-first walk of a syntax tree, it's useful to be able to start walking a tree at a given _index_ within a depth-first walk.
> 
> ### Change
> This PR adds the following related C APIs, which expand the ability to work with depth-first walks of a syntax tree.
> 
> * `ts_node_descendant_count` - How many total descendant nodes does this node contain, including the node itself?
> * `ts_tree_cursor_goto_descendant` - Move the cursor to the node that would appear at the given index in a depth-first walk of the tree
> * `ts_tree_cursor_current_descendant_index` - Retrieve the index of the cursor's current node in a depth-first walk of the tree
> * `ts_tree_cursor_current_depth`  - Retrieve the current depth of the tree cursor in the tree.
> 
> ### Implementation Notes
> This PR doesn't add any overhead in the syntax tree. It was already necessary to maintain on each syntax node a field called `node_count`, which was similar to `descendant_count`, and was used during error recovery as a measure of how much parsing progress had been made since discovering an error.
> 
> I've slightly changed the meaning of that field, so that it stores a node's visible descendant count (it now excludes _hidden_ nodes from the count), which can serve the same purpose, for error recovery, but can also be used as the basis for the new APIs.
> 
> ### Future Follow-up Steps
> * Add WASM bindings for these new APIs
> * Use these APIs to make the rendering of the Tree-sitter playground more efficient.

sarahsmile7:
> ### Background
> 
> When rendering a syntax tree, nodes are typically displayed in depth-first traversal order. In order to render an arbitrary *portion* of a depth-first walk of a syntax tree, it's useful to be able to start walking a tree at a given *index* within a depth-first walk.
> 
> ### Change
> 
> This PR adds the following related C APIs, which expand the ability to work with depth-first walks of a syntax tree.
> 
> * `ts_node_descendant_count` - How many total descendant nodes does this node contain, including the node itself?
> * `ts_tree_cursor_goto_descendant` - Move the cursor to the node that would appear at the given index in a depth-first walk of the tree
> * `ts_tree_cursor_current_descendant_index` - Retrieve the index of the cursor's current node in a depth-first walk of the tree
> * `ts_tree_cursor_current_depth`  - Retrieve the current depth of the tree cursor in the tree.
> 
> ### Implementation Notes
> 
> This PR doesn't add any overhead in the syntax tree. It was already necessary to maintain on each syntax node a field called `node_count`, which was similar to `descendant_count`, and was used during error recovery as a measure of how much parsing progress had been made since discovering an error.
> 
> I've slightly changed the meaning of that field, so that it stores a node's visible descendant count (it now excludes *hidden* nodes from the count), which can serve the same purpose, for error recovery, but can also be used as the basis for the new APIs.
> 
> ### Future Follow-up Steps
> 
> * Add WASM bindings for these new APIs
> * Use these APIs to make the rendering of the Tree-sitter playground more efficient.

9980b9524ad31313dad3958f09cb8df36702c60b

lib/binding_rust/bindings.rs
```diff
@@ -148,35 +148,102 @@ extern "C" {
     pub fn ts_parser_delete(parser: *mut TSParser);
 }
 extern "C" {
-    #[doc = " Set the language that the parser should use for parsing.\n\n Returns a boolean indicating whether or not the language was successfully\n assigned. True means assignment succeeded. False means there was a version\n mismatch: the language was generated with an incompatible version of the\n Tree-sitter CLI. Check the language's version using `ts_language_version`\n and compare it to this library's `TREE_SITTER_LANGUAGE_VERSION` and\n `TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION` constants."]
```
sarahsmile7, line 151: ?

sarahsmile7 (approved):

lib/binding_rust/bindings.rs
sarahsmile7: > ?

https://github.com/tree-sitter/tree-sitter/actions/runs/5249743481/job/14205457639

PR #2172, opened 2023-04-04
Linting, Formatting, & Tidying

amaanq:
This PR ~tidies up the ci scripts,~ fixes a few warnings/clippy lints in the autogenerated Rust bindings, and applies clippy lints throughout the project where I found reasonable and useful.

Issue #2003, opened 2022-12-27
Get position of child ts_builtin_sym_error_repeat in node


SrTobi:
Hi, I'm trying to implement a way to get error messages for a parsed syntax tree (like #255).
Unfortunately it is not trivial to get the correct error location from the syntax tree (#833).
There are ERROR nodes, but the error could be anywhere inside that tree.

By going through the source code of tree-sitter, I discovered a private tree symbol `ts_builtin_sym_error_repeat`, which seems to indicate the exact location where the parser failed and recovered in the parse version that corresponds to the returned syntax tree.

Unfortunately I cannot seem to get hold of this subtree by means of the official api.
Is there a way? Would a function alá 

```
// In an ERROR node, returns the offset of the position where the parser first encountered an error
uint32_t ts_node_first_error_recovery_position(TSNode);
```

make sense?

jaboatman:
I looked into this a bit, because I'm using tree-sitter as the backend for a language grammar. The inner error nodes are hidden, which is sensible for the editor case. To expose them for this use case though, I just changed this function:
```c

TSSymbolMetadata ts_language_symbol_metadata(
  const TSLanguage *self,
  TSSymbol symbol
) {
  if (symbol == ts_builtin_sym_error)  {
    return (TSSymbolMetadata) {.visible = true, .named = true};
  } else if (symbol == ts_builtin_sym_error_repeat) {
    // Changed from .visible = false, .named = false
    return (TSSymbolMetadata) {.visible = true, .named = true};
  } else {
    return self->symbol_metadata[symbol];
  }
}
```
and now my error nodes container inner `_ERROR` nodes, which correspond to the portion of the node which failed to parse.

If there is general interest, I can see about creating a proper pull request for it. Of course this would need to be configurable, I can see a few ways to handle that:
* Make it a configuration on the language - this is probably the easiest to implement since it keeps everything local to `language.c`, although this feels odd to me from an end user perspective that some _languages_ would be configured to show errors and others wouldn't. Although I suppose we could just have it be set at the API level, and not in your actual `grammar.json`.
* Make it a parser option - in this case we would maybe want to introduce a new symbol with a different name like `ERROR_INTERNAL` (to keep the `_ERROR` being hidden convention), which the parser would then produce instead of the `_ERROR` nodes.
* Make it `.visible = true, .named = false` instead and now they are accessible but still hidden from the regular tree viewer.

Issue #1631, opened 2022-02-01
Unable to use tree sitter as parser for compiler


seanyoung:
I've attempted to use tree-sitter for the [Solang Solidity Compiler](https://github.com/hyperledger-labs/solang), however I found it impossible to use because:

- When walking the parse tree, only *named* or *visible* nodes can found. *missing* nodes cannot, which makes it possible to find out what nodes are missing in an error node (needed for nice error message, more than just "parse error").
- When there is a parse error, we need to know what tokens would be acceptable, so we can give a nice compiler message saying.

As fas as I can see, tree sitter has a great properties (GLR, great error recovery, context sensitive keywords) but it is not usable in its current form, unless you can assume that the input is already valid.

ahelwer:
Tree-sitter is designed for use with language tooling (highlighting, code folding, formatting, code navigation, etc.) rather than to be the backbone of a compiler or interpreter. Thus it is permissive and does not allow for error messages.

razzeee:
https://github.com/tree-sitter/tree-sitter/issues/255#issuecomment-786268102

seanyoung:
> [#255 (comment)](https://github.com/tree-sitter/tree-sitter/issues/255#issuecomment-786268102)

That isn't going to cut it. You want a human-readable error string with a position. Ideally you want to be able to traverse the tree and get the missing nodes, so that you know exactly where the error happened, so you can do your best to resolve the rest of the parse tree and provide as many errors and warning as possible.

As far as I know, there is no other parser generator available for rust which is GLR with parse error recovery and context sensitive keywords.

stephe-ada-guru:
Andrew Helwer ***@***.***> writes:

> Tree-sitter is designed for use with language tooling (highlighting,
> code folding, formatting, code navigation, etc.) rather than to be the
> backbone of a compiler or interpreter. Thus it is permissive and does
> not allow for error messages.

But one feature of language tooling is providing code diagnostics to
help the user fix syntax errors without having to run the full compiler.
LSP provides for that; most IDEs I've used support it (Android
Studio, Emacs).

So it would be useful for tree-sitter to provide some form of error
messages.

-- 
-- Stephe

seanyoung:
Considering the tree sitter runtime is written in C, I think it might be time for a new pure rust runtime with a better api.

ahelwer:
Usually tree-sitter is used as a fast first-pass parser before the slower actual parser comes up with the syntax errors.

[...]

PR #2324, opened 2023-06-16
Api extensions: previous sibling, last child, lookahead iterator

daumantas-kavolis-sensmetry:
_This PR is follow up for https://github.com/tree-sitter/tree-sitter/discussions/2266#discussioncomment-5924434_

Extends the API with functions that can help generate better error messages and suggestions:

- tree cursor now supports walking back with `ts_tree_cursor_goto_previous_sibling` and going straight to the last child with `ts_tree_cursor_goto_last_child`
- tree cursor can be reset to another cursor's state so that parents are not lost, and cursors can be reused without allocating again
- exposed parse state and and added next parse state on node which can be used for generating valid symbols in place or after the node
- exposed non-aliased symbol id and name which are used by `ts_language_next_parse_state`
- exposed lookahead iterator for generating valid symbols at a given parse state

`symbol_iterator` may be a more appropriate name instead of `lookahead_iterator` that was already used internally.

I have also noticed that lookahead iterator can generate symbol aliases instead of non-aliased symbols, such as `primitive_type` instead of `u64` etc. in Rust.


lib/src/tree_cursor.c
```diff
@@ -258,6 +350,50 @@ bool ts_tree_cursor_goto_next_sibling(TSTreeCursor *self) {
...
+
+  // for that, save current position before traversing
+  Length position = array_back(&self->stack)->position;
+  TreeCursorStep step = ts_tree_cursor_goto_sibling_internal(
+      _self, ts_tree_cursor_child_iterator_previous);
+  if (step == TreeCursorStepNone)
+    return step;
+
+  // if row has not changed, column is still valid
+  if (array_back(&self->stack)->position.extent.row == position.extent.row)
```
maxbrunsfeld, line 366: If we made the change I suggested above, this would become:

```c
if (!length_is_undefined(array_back(&self->stack)->position)) return step;
```

script/generate-bindings
```diff
@@ -4,7 +4,6 @@ output_path=lib/binding_rust/bindings.rs
 header_path='lib/include/tree_sitter/api.h'
 
 bindgen                                        \
-  --size_t-is-usize                            \
```
maxbrunsfeld, line 7: Just curious why you removed this one.

lib/binding_rust/lib.rs
```diff
@@ -1318,6 +1400,15 @@ impl<'a> TreeCursor<'a> {
         };
     }
 
+    /// Move this cursor to the previous sibling of its current node.
+    ///
+    /// This returns `true` if the cursor successfully moved, and returns
+    /// `false` if there was no previous sibling node.
+    #[doc(alias = "ts_tree_cursor_goto_previous_sibling")]
```
maxbrunsfeld, line 1407: Could you add a note to this documentation that this method is slower than `goto_next_sibling` (because of the need to recalculate positions)?

maxbrunsfeld (commented):
Great work on all of this. I left some minor suggestions.

maxbrunsfeld:
Also, I'd love to hear more about your use case for some of these functions? Are you successfully using the `LookaheadIterator` for some interesting application feature?

script/generate-bindings
ahlinc: This is because it's default behavior for the latest `bindgen-cli` and it doesn't have such CLI parameter anymore.
I also removed it on master and added protection to don't allow to use older version of the bindgen than the last used one.

daumantas-kavolis-sensmetry:
We have two main uses in mind which `tree-sitter` currently lacks:

1. User-friendly parser error messages.  
   Just having `MISSING` and `ERROR` nodes is very limiting. The `lookahead_iterator` can be used to generate valid symbols in a language agnostic way without having to resort to `grammar.json`. `ts_language_next_state` is needed to generate symbols in place of `ERROR` nodes. Bidirectional cursors can be useful for `MISSING` nodes, i.e. in JS an identifier followed by a `MISSING` node may a misspelled keyword. Here, `ts_tree_cursor_reset_to` may be useful for having a scratch cursor that can iterate through siblings and parents without allocating in most cases. We haven't decided on the exact error message generation yet but this should make them far better than generic (and often useless) `ERROR` and `MISSING` messages.
2. Language server suggestions.  
   A `lookahead_iterator` constructed for the next parse state can be used to generate valid symbols on a whitespace character, and when constructed for the current parse state - for existing words. Keywords can be returned as-is while identifiers would trigger reference completion.

Also, bidirectional cursors can be used to implement `goto_previous_leaf_node` which can be useful for qualified references split over multiple composite nodes.

`lookahead_iterator` returning aliased symbols is likely related to #2299.

ahlinc:
Overall I like this implementation, but I'd like to improve doc comments with a followup commit.

Also I see a tendency that there become more integer ids and they seep more and more from C to Rust what makes harder to control what things mean and where they can be used. Rust provides a concept  of zero cost _newtype_ pattern that can improve this situation and control where different ids appears and where they can be applied. This would reduce complexity for new users and may be there will be less questions and misunderstanding.

ahlinc:
Also I think we would need to review the following issues:
* #833
* #255 
* #1631
* #2003 

And decide what of them we can close by this PR if it fully covers their needs.

daumantas-kavolis-sensmetry:
Wrapping integers in _newtypes_ may be better left as a separate PR. Would it also make sense to wrap integers in C in structs since they are only used as ids and not arithmetic types?

ahlinc:
> Would it also make sense to wrap integers in C in structs since they are only used as ids and not arithmetic types?

I think on the C level it's better to left it as is.
But for Rust apply benefits provided by its type system.

> Wrapping integers in newtypes may be better left as a separate PR.

Agree.

lib/src/tree_cursor.c
```diff
@@ -97,6 +97,51 @@ static inline bool ts_tree_cursor_child_iterator_next(
...
 
+static inline Length length_sub_zero(Length a, Length b) {
+  // length_sub doesn't account for 0 row subtraction, i.e. only columns
+  // should be subtracted, but changing point_sub breaks other tests
+  Length result = length_sub(a, b);
+  if (b.extent.row == 0 && a.extent.row != 0) {
+    result.extent.column -= b.extent.column;
+  }
+  return result;
+}
```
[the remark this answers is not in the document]
daumantas-kavolis-sensmetry: That's much clearer, thanks. Changed in the latest commit.

lib/binding_rust/lib.rs
daumantas-kavolis-sensmetry: Also added a similar note to `ts_tree_cursor_goto_last_child`.

maxbrunsfeld (approved):

lib/binding_rust/ffi.rs
```diff
@@ -130,3 +132,19 @@ impl QueryCursor {
         ManuallyDrop::new(self).ptr.as_ptr()
     }
 }
+
+impl LookaheadIterator {
+    /// Reconstructs a [LookaheadIterator] from a raw pointer.
```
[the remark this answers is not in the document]
daumantas-kavolis-sensmetry: Current format without backticks is consistent with the remaining docs in `ffi.rs`, for consistency all of them should be changed or none.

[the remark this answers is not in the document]
amaanq: yeah my PR [here](https://github.com/tree-sitter/tree-sitter/pull/2172) does that everywhere else - we (Ahlinc and I) think it'll be ready to merge next week, though I don't mind much if you don't do it here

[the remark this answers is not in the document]
daumantas-kavolis-sensmetry: In addition, `from_raw` here is missing `const` and `LookaheadIterator` should be replaced with `Self` to match the PR.

maxbrunsfeld (approved):

ahlinc (approved):

ahlinc (approved):

PR #2352, opened 2023-06-28
Docs: Added a link to tree-sitter-usd

ColinKennedy:
This PR adds [tree-sitter-usd](https://github.com/ColinKennedy/tree-sitter-usd) to the list of parser.

USD is the format used by .usd and .usda files: https://github.com/PixarAnimationStudios/OpenUSD

I've verified that this parser is robust but running it on several hundreds of FOSS production files.


amaanq:

PR #2370, opened 2023-07-04
fix -Wshadow warnings, some clang-tidy lints

amaanq:


Realprach (commented):
Licenses

amaanq:
> Licenses

what? wdym

clason:
Spambot, just ignore.

lib/src/parser.c
```diff
@@ -185,7 +185,7 @@ static bool ts_parser__breakdown_top_of_stack(
       TSStateId state = ts_stack_state(self->stack, slice.version);
       Subtree parent = *array_front(&slice.subtrees);
 
-      for (uint32_t j = 0, n = ts_subtree_child_count(parent); j < n; j++) {
+      for (uint32_t j = 0, cnt = ts_subtree_child_count(parent); j < cnt; j++) {
```
maxbrunsfeld, line 188: What is the rationale for some of these renames? `n` seems pretty standard for a count. `cnt` is a bit weird IMO. If we need an abbreviated word I'd go w/ `len`.

maxbrunsfeld (commented):
Left some minor questions.

lib/src/array.h
```diff
@@ -65,19 +65,19 @@ extern "C" {
...
 // Insert one `element` into the array at the given `index`.
-#define array_insert(self, index, element) \
-  array__splice((VoidArray *)(self), array__elem_size(self), index, 0, 1, &element)
+#define array_insert(self, _index, element) \
+  array__splice((VoidArray *)(self), array__elem_size(self), _index, 0, 1, &(element))
 
 // Remove one `element` from the array at the given `index`.
-#define array_erase(self, index) \
-  array__erase((VoidArray *)(self), array__elem_size(self), index)
+#define array_erase(self, _index) \
```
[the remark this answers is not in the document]
amaanq: Exists shadowed a previous declaration in query.c, and I renamed index as well for consistency
image

lib/src/parser.c
amaanq: Honestly it's because clang-tidy suggests to not use variables less than 3 chars long, but for a loop variable (like i, j, k) it's accepted, might be a bug where multiple loop variables aren't picked up. I'd agree n is fine here

maxbrunsfeld (approved):

Scutua (approved):

PR #2371, opened 2023-07-04
chore: update exports to include some string.h functions

amaanq:
@ahlinc this and #2359 would fix CI :sweat_smile:


patrickt (approved):
This seems sensible. I’m sure we could have a holy war about which C string copying function to use, but I’d rather not. 😄

Issue #1942, opened 2022-11-02
Caching or Serializing a `TSQuery`


SergioBenitez:
Hello!

I'm working on a static site generator in Rust which I hope uses tree-sitter based syntax highlighting. As is stands, creating a `HighlightConfiguration` is an expensive operation. For some languages, like Ruby, this can take upwards of 500ms. But even for "smaller" languages, like Bash, time to create the configuration structure is measured on the order of 50ms. Compared to the total time to generate my site (~15ms) _without_ syntax highlighting, this is astronomical. In comparison, loading a `syntect` `SynaxSet` takes on the order of ~25ms, and this is for the entire set, not per language, as the ~50ms measures.

In an effort to reduce the time to create a `HighlightConfiguration`, I've [started work on a fork](https://github.com/SergioBenitez/tree-sitter/tree/dehydrate) which introduces the concept of "hydration", in essence allowing a `HighlightConfiguration` structure to be serialized and then deserialized, affording most of the structure's creation time. Please see https://github.com/SergioBenitez/tree-sitter/tree/dehydrate for the exact implementation.

The net effect of this patch is that a `HighlightConfiguration` can be created in about half the time. The hope is to decrease this further, but to do so requires being able to serialize/deserialize a `TsQuery`, removing the need for [hydration to call `ts_query_new`](https://github.com/SergioBenitez/tree-sitter/blob/2cd141f651fae5e9bdfeff4867f57e017722a9d8/lib/binding_rust/lib.rs#L143-L149) entirely, which should reduce the `HighlightConfiguration` creation time to nearly zero.

Looking at `TSQuery`, it _appears_ to be serializable, but seeing as the structure exists only in C and isn't reproduced in the bindings, I'm not certain that efforts to serialize it would go anywhere. So, my issue amounts to the following:

 * Are there any other efforts to decrease the time to create a `HighlightConfiguration` and/or cache/serialize a `TSQuery`?
 * Is the representation of a `TSQuery` stable enough for serialization? IE, would an implementation of a serializer in C be accepted?
 * Are there any other APIs I may be missing that would make it significantly less expensive to create a `HighlightConfiguration`?

Thank you!

ahelwer:
One alternative (that I'm sure you've considered) is to have separate modes of building your site, so for iterative building when editing you can use a simple fast syntax highlighter (or none at all) and for the rare occurrence that the site is actually updated and published you can use the slower tree-sitter highlighter.

Keats:
[...]

PR #2374, opened 2023-07-06
TSQuery serialization and deserialization

ErinvanderVeen:
At [Topiary](https://github.com/tweag/topiary), we've run into [issues](https://github.com/tweag/topiary/issues/443) with Query compilation, with it taking up about 80-90% of Topiary's total runtime.
In an effort to reduce this number we were made aware of https://github.com/tree-sitter/tree-sitter/issues/1942.

In that issue, an approach is proposed of compiling the Query structure to a linear memory region, which we attempted to do.
As a first (and easy) prototype we, at Topiary, experimented with creating a buffer from the Query structure that explicitly doesn't include the language field.
Users are instead expected to provide the language again during deserialization.

With this draft PR I want to open a discussion surrounding this implementation.
Is this sufficient? What could we do more/instead? Does this cover the use cases described in #1942?

This PR is far from being ready to be merged. For starters, there are only Rust bindings to the new functions.

### DIFF
--- a/lib/binding_rust/bindings.rs
+++ b/lib/binding_rust/bindings.rs
@@ -839,6 +839,21 @@ extern "C" {
     pub fn ts_query_cursor_set_max_start_depth(arg1: *mut TSQueryCursor, arg2: u32);
 }
 extern "C" {
+    #[doc = " Serializes a Query to a bytestring."]
+    pub fn ts_query_serialize(
+        arg1: *const TSQuery,
+        arg2: *mut usize,
+    ) -> *const ::std::os::raw::c_char;
+}
+extern "C" {
+    #[doc = " Deserializes a bytestring to a TSQuery."]
+    pub fn ts_query_deserialize(
+        arg1: *const ::std::os::raw::c_char,
+        arg2: *const usize,
+        arg3: *const TSLanguage,
+    ) -> *mut TSQuery;
+}
+extern "C" {
     #[doc = " Get the number of distinct node types in the language."]
     pub fn ts_language_symbol_count(arg1: *const TSLanguage) -> u32;
 }
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -210,6 +210,14 @@ pub struct LanguageError {
     version: usize,
 }
 
+/// An error that occured during (de)serialization
+#[derive(Debug, PartialEq, Eq)]
+pub enum SerializationError {
+    LanguageError(LanguageError),
+    AllocationError,
+    DeserializationError(QueryError),
+}
+
 /// An error that occurred in `Parser::set_included_ranges`.
 #[derive(Debug, PartialEq, Eq)]
 pub struct IncludedRangesError(pub usize);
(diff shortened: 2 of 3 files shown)


Keats:
Were you able to run this branch with Topiary? Any benchmark coming out of it?

ErinvanderVeen:
> Were you able to run this branch with Topiary? Any benchmark coming out of it?

Sure! Our criterion OCaml benchmark (which formats a large file) goes from:
```
format_ocaml            time:   [565.08 ms 565.78 ms 566.57 ms]
```
To
```
format_ocaml            time:   [127.36 ms 127.86 ms 128.37 ms]
```
on my machine.

Of note: this was a very quick and dirty implementation in Topiary, some additional overhead might be required to properly handle deserialization of the queries, but I expect it to be relatively small.

ahlinc:
@ErinvanderVeen  It looks like you forget to commit a file `bits/stdint-uintn.h`.

amaanq:
> @ErinvanderVeen  It looks like you forget to commit a file `bits/stdint-uintn.h`.

it's a glibc header

ahlinc:
> > @ErinvanderVeen  It looks like you forget to commit a file `bits/stdint-uintn.h`.
> 
> it's a glibc header

It seems it isn't portable and don't exists for Emscripten `emcc` compiler.

ErinvanderVeen:
> @ErinvanderVeen It looks like you forget to commit a file `bits/stdint-uintn.h`.

That must've been my language server, sorry about that noise. All those imports are irrelevant.

ErinvanderVeen:
> 1. It would be great if it would be possible to improve queries compilation process that it would produce compiled query as a continuous block of memory that would be interpreted as a set of structs by a query running logic and used directly and  in the same time it would be possible to dump it to a file very simply. This would make serialization and deserialization very trivial and fast.
> 2. Provide a way to store compiled queries directly in `*.so` library files so queries compilation would happened at once with a language compilation. And provide an additional API to use such queries _stored_ in the library.

It doesn't look possible to determine the length of all the `TSQuery` struct arrays ahead-of-time. So in my mind an implementation of this would just be:
1. Create the `TSQuery` struct (or at least the containing arrays)
2. "Serialize"/"Compile" the `TSQuery` struct to a single region of memory.
Where this single region of memory will have to start with an header containing the offsets for all the fields/arrays for O(1) access.

To me, it feels like, at that point, you are basically always performing serialization. You're just moving the O(n) cost of it from the `ts_query_serialize` function to the `ts_query_new` function. All the while complicating the `TSQuery` field access, which would have to be wrapped in either a macro or a function.

The big advantage you get in return is the simplification of restoring from serialization. But even then you have to either
1) Still supply the language, again complicating the restoration
2) Find a way to serialize the language field

I could definitely investigate this further, if this approach is desirable.

ahlinc:
> To me, it feels like, at that point, you are basically always performing serialization. You're just moving the O(n) cost of it from the ts_query_serialize function to the ts_query_new function.

I agree and such additional cost shouldn't happen for queries compiled in runtime but may provide a benefit for fastest deserialization and query object construction for precompiled queries (that's what users expect and want to achieve to speedup their code startup).

> I could definitely investigate this further, if this approach is desirable.

I'd ask @maxbrunsfeld first, may be he can give some advises.

samirbani:
في الأربعاء، ١٢ يوليو ٢٠٢٣ ٤:١٦ م Andrew Hlynskyi ***@***.***>
كتب:

RubixDev:
I have integrated this in my crate [`syntastica`](https://github.com/RubixDev/syntastica/tree/precomp) now to improve the startup time. This is the result on my machine:

```
old behaviour: 2.803549488s
with pre-compiled queries: 21.155867ms
```

<details>
<summary>(using this Rust code)</summary>

```rust
use std::time::{Duration, Instant};

use syntastica_parsers_git::{LanguageSetImpl, RawLanguageSetImpl};

fn main() {
    const ITERATIONS: usize = 10;

    let mut durations = Vec::with_capacity(ITERATIONS);
    for _ in 0..ITERATIONS {
        let start = Instant::now();
        let mut set = RawLanguageSetImpl::new();
        set.preload_all();
        durations.push(start.elapsed());
    }
    println!(
        "old behaviour: {:?}",
        durations.iter().sum::<Duration>() / ITERATIONS as u32,
    );

    let mut durations = Vec::with_capacity(ITERATIONS);
    for _ in 0..ITERATIONS {
        let start = Instant::now();
        let mut set = LanguageSetImpl::new();
        set.preload_all();
        durations.push(start.elapsed());
    }
    println!(
        "with pre-compiled queries: {:?}",
        durations.iter().sum::<Duration>() / ITERATIONS as u32,
    );
}
```

</details>

The performance improvement is obvious and definitely desired.

Just a few nitpicks that I think should also be resolved before this is ready for merging:

<ul>
<li><code>SerializationError</code> does not implement the <code>Error</code> trait</li>
<li><code>Query::deserialize</code> takes a <code>&Vec&lt;u8&gt;</code>, should be <code>&[u8]</code> </li>
<li><details><summary>this clippy warning</summary>

```
warning: manual `!RangeInclusive::contains` implementation
    --> lib/binding_rust/lib.rs:1889:12
     |
1889 |         if version < MIN_COMPATIBLE_LANGUAGE_VERSION || version > LANGUAGE_VERSION {
     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `!(MIN_COMPATIBLE_LANGUAGE_VERSION..=LANGUAGE_VERSION).contains(&version)`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains
```

</details></li>
</ul>

amaanq:
last clippy warning should be untouched in this PR

ObserverOfTime:
Superseded by #2594.

PR #2382, opened 2023-07-13
Restore Rust bindings generation with newer bindgen 0.65.1

ahlinc:


amaanq:
Latest bindgen is 0.66.1 but this should be fine for a while

amaanq (approved):

ahlinc:
> Latest bindgen is 0.66.1 but this should be fine for a while

I know, as a first step I just restore what was there before.

PR #2394, opened 2023-07-16
Add more architectures to ci

amaanq:
This PR also fixes portability issue in tests for the `mipsel` target:
```
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> cli/src/tests/helpers/allocations.rs:5:30
  |
5 |         atomic::{AtomicBool, AtomicU64, Ordering::SeqCst},
  |                              ^^^^^^^^^
  |                              |
  |                              no `AtomicU64` in `sync::atomic`
  |                              help: a similar name exists in the module: `AtomicU32`

For more information about this error, try `rustc --explain E0432`.
```

Closes: #1366

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,7 @@ jobs:
         runtime:
         - linux-arm64     #
         - linux-arm       #
+        - linux-armv7l    #
         - linux-x64       #
         - linux-x86       #
         - linux-i586      #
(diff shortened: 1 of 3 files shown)


amaanq:
Btw @ahlinc release's write permission still errors on my fork's actions (which is why I had it as read then force pushed) 🫠 no idea how to deal with that

ahlinc:
> Btw @ahlinc release's write permission still errors on my fork's actions (which is why I had it as read then force pushed) 🫠 no idea how to deal with that

@amaanq could you invite me in your fork with maintainer level, so I'd be able to test it in your fork's environment (can't do in in the main repo for obvious reasons)  and in my fork it's not reproducible because I'm it's owner.

ahlinc:
@amaanq thank you for spending time and adding all of that new targets!

ahlinc (approved):

PR #2400, opened 2023-07-17
feat: add the ability to specify a custom decode function

### LINKED ISSUE #717: Support custom text decoding function
I've been working on [a text editor](https://github.com/lukedan/codepad) that handles raw binary data of arbitrary character encoding, and I'm trying to integrate tree-sitter into it for syntax highlighting. However, it seems that tree-sitter only supports UTF-8 and UTF-16. I did a bit of digging and discovered that, in `ts_lexer__get_lookahead` in [lexer.c](https://github.com/tree-sitter/tree-sitter/tree/master/lib/src/lexer.c), a `UnicodeDecoderFunction` is used to decode text which is actually general enough to be extended to a lot of other encodings. It would be great (and probably fairly simple) to expose this as part of the interface instead, so that users can use whichever encoding they want by providing custom decode functions.

amaanq:
Closes #717

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -18,6 +18,7 @@
 #include "./stack.h"
 #include "./subtree.h"
 #include "./tree.h"
+#include "unicode.h"
 
 #define LOG(...)                                                                            \
   if (self->lexer.logger.log || self->dot_graph_file) {                                     \
@@ -1990,7 +1991,7 @@ TSTree *ts_parser_parse_string(
   const char *string,
   uint32_t length
 ) {
-  return ts_parser_parse_string_encoding(self, old_tree, string, length, TSInputEncodingUTF8);
+  return ts_parser_parse_string_encoding(self, old_tree, string, length, TSInputEncodingUTF8, NULL);
 }
 
 TSTree *ts_parser_parse_string_encoding(
@@ -1998,13 +1999,15 @@ TSTree *ts_parser_parse_string_encoding(
   const TSTree *old_tree,
   const char *string,
   uint32_t length,
-  TSInputEncoding encoding
+  TSInputEncoding encoding,
+  UnicodeDecodeFunction decode_function
 ) {
   TSStringInput input = {string, length};
   return ts_parser_parse(self, old_tree, (TSInput) {
     &input,
     ts_string_input_read,
     encoding,
+    decode_function,
   });
 }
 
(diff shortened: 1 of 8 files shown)


amaanq:
Some tests could be useful, but I wonder what encodings we could/should test on

ahlinc:
An approach with a global pointer to a decoding function doesn't allow to work with different parser instances and files in various encodings in a multi threaded fashion.

amaanq:
> An approach with a global pointer to a decoding function doesn't allow to work with different parser instances and ~files~ input streams with various custom encodings in a multi threaded fashion.

Yep, I know. I wasn't sure if it would be better to have it be global or not (I more or less followed the approach of the tree-sitter allocator)

ahlinc:
> > An approach with a global pointer to a decoding function doesn't allow to work with different parser instances and ~files~ input streams with various custom encodings in a multi threaded fashion.
> 
> Yep, I know. I wasn't sure if it would be better to have it be global or not (I more or less followed the approach of the tree-sitter allocator)

Allocator is a different story

amaanq:
reworked it to be local to the parse instance - but the Rust side seems a bit rough to implement :sweat_smile:

soz-ai (commented):
[]()

amaanq:
implemented the rust side and tested it on some utf-32 and windows cp1252 😁

lib/src/parser.c
```diff
@@ -1990,21 +1991,23 @@ TSTree *ts_parser_parse_string(
...
 
 TSTree *ts_parser_parse_string_encoding(
   TSParser *self,
   const TSTree *old_tree,
   const char *string,
   uint32_t length,
-  TSInputEncoding encoding
+  TSInputEncoding encoding,
+  UnicodeDecodeFunction decode_function
 ) {
```
theHamsta, line 2004: Isn't this a breaking change that requires increasing the minor version?

ahlinc:
After reviewing this again I think this may be implemented in a less intrusive way:
- First of all keep all existing C APIs untouched.
- An addition of `TSInputEncoding_TSInputEncodingCustom` doesn't needed at all or it may be made hidden.
- Introduce a one `ts_parser_parse_decode` C API function that would receive a pointer to a decoding fn and store it in the `payload` field of `TSInput` struct so the last may be kept unchanged. And set the encoding field to a hidden sentinel.
- The decoding fn may actually return not a buffer of bytes but a buffer of u32 decoded code points and an internal `decode` function would be noop in this case.

amaanq:
Supersedes by #3833

Issue #768, opened 2020-10-19
Main thread panicks with 'no entry found for key'


Cu3PO42:
I have created a grammar which causes `tree-sitter generate` to fail with the following error:
```
$ /home/cu3po42/development/tree-sitter-minerva/node_modules/.bin/tree-sitter generate
thread 'main' panicked at 'no entry found for key', cli/src/generate/render.rs:1049:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error Command failed with exit code 101.
```
The verbose backtrace is as follows:
```
$ /home/cu3po42/development/tree-sitter-minerva/node_modules/.bin/tree-sitter generate
thread 'main' panicked at 'no entry found for key', cli/src/generate/render.rs:1049:21
stack backtrace:
   0:     0x564c8314c495 - backtrace::backtrace::libunwind::trace::h14d338b30b3ea0a7
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1:     0x564c8314c495 - backtrace::backtrace::trace_unsynchronized::h73ea91d74a3fd67f
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2:     0x564c8314c495 - std::sys_common::backtrace::_print_fmt::hd42948c952866e12
                               at src/libstd/sys_common/backtrace.rs:78
   3:     0x564c8314c495 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::ha8f928866ff7571e
                               at src/libstd/sys_common/backtrace.rs:59
   4:     0x564c83177f0c - core::fmt::write::he0c1e5f7426d2718
                               at src/libcore/fmt/mod.rs:1076
   5:     0x564c83144622 - std::io::Write::write_fmt::hf3afc6cfd57d0033
                               at src/libstd/io/mod.rs:1537
   6:     0x564c8314ecd0 - std::sys_common::backtrace::_print::hfc0110703f3696fd
                               at src/libstd/sys_common/backtrace.rs:62
   7:     0x564c8314ecd0 - std::sys_common::backtrace::print::h3f77c6990ddfaa22
                               at src/libstd/sys_common/backtrace.rs:49
   8:     0x564c8314ecd0 - std::panicking::default_hook::{{closure}}::heae49580a8d62d75
[...]

PR #2401, opened 2023-07-18
fix: disallow inlining the first rule

amaanq:
This prevents a panic when indexing symbol_ids during the generation process

This solves one case in #768, namely
```js
module.exports = grammar({
  name: 'mylang',

  inline: $ => [
    $.block,
  ],

  rules: {
    block: $ => $.macr,
    macr: $ => choice("M", "m"),
    source_file: $ => "hello"
  }
});
```

The others are more difficult, and seem to have to do w/ the extractor replacing/modifying symbols


maxbrunsfeld (commented):
Minor, but can you replace “Token” with “Rule” in the message?

ahlinc (approved):

PR #2402, opened 2023-07-18
fix: pass a value_id the same size of predicate_capture_ids's elements to avoid big-endian integer narrowing

### LINKED ISSUE #2395: Big-endian bug in tests
Big endian architectures currently fail in 1 test, tests::query_test::test_query_is_pattern_guaranteed_at_step

Output from powerpc 32/64, mips, s390x, sparc64:

```
 query example: "no guaranteed steps"
  query example: "all guaranteed steps"
  query example: "a fallible step that is optional"
  query example: "multiple fallible steps that are optional"
  query example: "guaranteed step after fallibe step"
  query example: "fallible step in between two guaranteed steps"
  query example: "one guaranteed step after a repetition"
  query example: "guaranteed steps after multiple repetitions"
  query example: "a guaranteed step with a field"
  query example: "multiple guaranteed steps with fields"
  query example: "nesting, one guaranteed step"
  query example: "a guaranteed step after some deeply nested hidden nodes"
  query example: "nesting, no guaranteed steps"
  query example: "a guaranteed step after a nested node"
  query example: "a step that is fallible due to a predicate"
thread 'tests::query_test::test_query_is_pattern_guaranteed_at_step' panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`: Description: a step that is fallible due to a predicate, Pattern: "(subscript_expression object: (member_expression object: (identifier) @obj property: (property_identifier) @prop) \"[\" (#match? @prop \"foo\"))", substring: "property_identifier",
 expected is_definite to be false', cli/src/tests/query_test.rs:4068:17    
```

_Related to:_ 
* #2394

amaanq:
Closes #2395 

This solves a bug on big-endian architectures where the value would be later passed by reference as an elements "view" before being inserted. The bug is in `ts_query__analyze_patterns`, specifically when the macro `array_inserted_sorted_by` is called on line 1791. The issue is that the step's value_id is a uint32_t, and is casted as a void pointer eventually when the value (if it doesn't exist) is inserted in the array in `array__splice`. Then when writing elements that are the size of the input array's elements which is 2 for `predicate_capture_ids` (uint16_t), only two of the bytes of the original `uint32_t step->value_id` are written. This is okay for little-endian systems, which is largely why it was undetected for so long, but not big-endian. The reason this fails is because the number will be in one "data byte" - think of the data byte being 01 in 00 00 00 01. Since we're only writing two bytes, the `00 01` part isn't written and we lose out on data. I'm not sure of other spots that can contain this kind of bug - but I do think better care could be taken (maybe an assert for the elem size vs input size?)

For the scope of this PR, the change I made should be sufficient though.

TLDR; casting is messy, C is hard

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,52 +33,79 @@ jobs:
       fail-fast: false
       matrix:
         runtime:
-        - linux-arm64     #
-        - linux-arm       #
-        - linux-armv7l    #
-        - linux-x64       #
-        - linux-x86       #
-        - linux-i586      #
-        - linux-mipsel    #
-        - linux-mips64el  #
-        - linux-ppc64el   #
-        - linux-riscv64gc #
-        - windows-arm64   #
-        - windows-x64     # <-- No C library build - requires an additional adapted Makefile for `cl.exe` compiler
-        - windows-x86     #     -- // --
-        - macos-arm64     # <-- MacOS M1/M2 - no tests, only CLI build to be published on release artifacts
-        - macos-x64       #
+        - linux-arm64       #
+        - linux-arm         #
+        - linux-armhf       #
+        - linux-armv5te     #
+        - linux-armv7l      #
+        - linux-x64         #
+        - linux-x86         #
+        - linux-i586        #
+        - linux-mips        #
+        - linux-mips64      #
+        - linux-mipsel      #
+        - linux-mips64el    #
+        - linux-powerpc     #
+        - linux-powerpc64   #
+        - linux-powerpc64el #
+        - linux-riscv64gc   #
+        - linux-s390x       #
+        - linux-sparc64     #
+        - linux-thumbv7neon #
+        - windows-arm64     #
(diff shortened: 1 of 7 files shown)


ahlinc:
Thank you very much for digging this issue with Big-endians!!! :trophy:  

I think this can be merged with excluded `sparc64` target.

maxbrunsfeld (approved):
Nice fix. Thanks @amaanq.

ahlinc:
> maybe we can bump the timeout time?

Then on normal platforms timeout tests may fail permanently because normal parsing would happen first.

I think we need to and an env var `TREE_SITTER_TEST_PARSER_TIMEOUT` with default value set a current timeout value and for the spark target tune it accordingly over the build matrix.

ahlinc (approved):

samirbani:
في الأربعاء، ١٩ يوليو ٢٠٢٣ ٣:٥٠ ص Amaan Qureshi ***@***.***>
كتب:

PR #2406, opened 2023-07-19
Fix miscompilation of queries that start with repetitions followed by alternatives

maxbrunsfeld:
This fixes a bug in the compilation of queries like the following:

```clj
(
  (line_comment)* @comment
  .
  [
    (function_item) @func
    (struct_item) @struct
  ]
)
```

Previously, the query would fail to match the repeated nodes preceding any of the alternatives except for the first alternative.

The fix was a one line deletion, but it took me a while to find. Along the way, I improved the query debugging code.


amaanq:
oh oops.. I just fixed this in my big endian branch

amaanq (approved):

amaanq:
Awesome fix - looks good (the query code really is complex...)

PR #2407, opened 2023-07-19
Bump deps

amaanq:


Cargo.toml
```diff
@@ -4,7 +4,10 @@ members = ["cli", "lib"]
 resolver = "2"
 
 [workspace.package]
-rust-version = "1.65"
+rust-version = "1.71"
```
ahlinc, line 7: This shouldn't be raised without strong reasons, it's a minimum supported rust version.

ahlinc: And it may be different for different crates and maximally lowest for the lib binding crate, but no one requested that or complained on it yet.

Cargo.toml
```diff
@@ -4,7 +4,10 @@ members = ["cli", "lib"]
...
 
 [profile.release]
-strip = true
+strip = true      # Automatically strip symbols from the binary.
+lto = true        # Link-time optimization.
+opt-level = "s"   # Optimize for speed.
+codegen-units = 1 # Maximum size reduction optimizations.
```
ahlinc, line 13: Do this things produce a visible / comparable impact?

ahlinc (changes requested):

samirbani:
في الأربعاء، ١٩ يوليو ٢٠٢٣ ١:٥٤ م Andrew Hlynskyi ***@***.***>
كتب:

Cargo.toml
amaanq: For me the binary release goes from 14mb to 12mb

Cargo.toml
```diff
@@ -7,4 +7,7 @@ resolver = "2"
 rust-version = "1.65"
 
 [profile.release]
-strip = true
+strip = true      # Automatically strip symbols from the binary.
+lto = true        # Link-time optimization.
+opt-level = "s"   # Optimize for speed.
```
mati865, line 12: `s` stands for "optimise for size" and is typically slightly slower than the default `3` and marginally slower than `2`.
In rate scenarios when `2` is faster than `3` (optimisation shenanigans), `s` will be faster than `3` but still slower than `2`.

narpfel: I just measured this, for me the default opt level (i. e. 3) is ~10% faster for `tree-sitter generate` for all grammars that I tested with.

amaanq: can you try out w/ 2?

we can make the fastest the default profile, and have one for size

amaanq: Thanks - I think 2 is a good balance then of compile time and speed

PR #2411, opened 2023-07-20
update fuzz script, some minor fixes

amaanq:


script/build-fuzzers
```diff
@@ -39,34 +39,39 @@ for lang in ${languages[@]}; do
...
+  ts_lang="tree_sitter_$(echo "$lang" | tr -- - _)"
+  $CXX "$CXXFLAGS" -std=c++11 -I lib/include -D TS_LANG="$ts_lang" -D TS_LANG_QUERY_FILENAME="\"${ts_lang_query_filename}\"" \
     "test/fuzz/fuzzer.cc" "${objects[@]}" \
-    libtree-sitter.a "$LIB_FUZZER_PATH" \
-    -o "out/${lang}_fuzzer"
+    libtree-sitter.a \
+    -o "test/fuzz/out/${lang}_fuzzer"
 
-  python test/fuzz/gen-dict.py "${lang_dir}/src/grammar.json" > "out/$lang.dict"
+  # python test/fuzz/gen-dict.py "${lang_dir}/src/grammar.json" > "out/$lang.dict"
```
ahlinc, line 71: If you delete the file it doesn't make sense to left this commented line.

script/build-fuzzers
```diff
@@ -6,30 +6,30 @@ if [[ "$(uname -s)" != Linux ]]; then
...
-if [[ -z "$LIB_FUZZER_PATH" ]]; then
-  echo "LIB_FUZZER_PATH not set"
-  exit 1
-fi
-
-CC=${CC:-clang}
-CXX=${CXX:-clang++}
-LINK=${LINK:-clang++}
+CC=clang
+CXX=clang++
```
ahlinc, line 10: Don't remove possibility to override from env

script/build-fuzzers
```diff
@@ -6,30 +6,30 @@ if [[ "$(uname -s)" != Linux ]]; then
...
 
 default_fuzz_flags="-fsanitize=fuzzer,address,undefined"
 
 CFLAGS=${CFLAGS:-"$default_fuzz_flags"}
 CXXFLAGS=${CXXFLAGS:-"$default_fuzz_flags"}
 
 export CFLAGS
-make
+make CC="$CC" CXX="$CXX"
```
ahlinc, line 18: This change doesn't needed

script/build-fuzzers
```diff
@@ -39,34 +39,39 @@ for lang in ${languages[@]}; do
...
   elif [ -e "${lang_scanner}.c" ]; then
-    $CC $CFLAGS -std=c99 -g -O1 "-I${lang_dir}/src" -c "${lang_scanner}.c" -o "${lang_scanner}.o"
+    $CC "$CFLAGS" -std=c99 -g -O1 "-I${lang_dir}/src" -c "${lang_scanner}.c" -o "${lang_scanner}.o"
     objects+=("${lang_scanner}.o")
   fi
 
 
   # Compiling with -O0 speeds up the build dramatically
-  $CC $CFLAGS -g -O0 "-I${lang_dir}/src" "${lang_dir}/src/parser.c" -c -o "${lang_dir}/src/parser.o"
+  $CC "$CFLAGS" -g -O0 "-I${lang_dir}/src" "${lang_dir}/src/parser.c" -c -o "${lang_dir}/src/parser.o"
```
ahlinc, line 51: Wrapping vars in double quotes breaks variables expansion in the shell in case if there are multiple parameters delimited with spaces. Please remove all added double quote wrappings.

script/build-fuzzers
```diff
@@ -39,34 +39,39 @@ for lang in ${languages[@]}; do
...
     ts_lang_query_filename=""
   fi
 
   # FIXME: We should extract the grammar name from grammar.js. Use the name of
   # the directory instead. Also, the grammar name needs to be a valid C
   # identifier so replace any '-' characters
-  ts_lang="tree_sitter_$(echo $lang | tr -- - _)"
-  $CXX $CXXFLAGS -std=c++11 -I lib/include -D TS_LANG="$ts_lang" -D TS_LANG_QUERY_FILENAME="\"${ts_lang_query_filename}\"" \
+  ts_lang="tree_sitter_$(echo "$lang" | tr -- - _)"
+  $CXX "$CXXFLAGS" -std=c++11 -I lib/include -D TS_LANG="$ts_lang" -D TS_LANG_QUERY_FILENAME="\"${ts_lang_query_filename}\"" \
```
ahlinc, line 66: The same as above, remove added dquotes

script/build-fuzzers
```diff
@@ -39,34 +39,39 @@ for lang in ${languages[@]}; do
...
+    libtree-sitter.a \
+    -o "test/fuzz/out/${lang}_fuzzer"
 
-  python test/fuzz/gen-dict.py "${lang_dir}/src/grammar.json" > "out/$lang.dict"
+  # python test/fuzz/gen-dict.py "${lang_dir}/src/grammar.json" > "out/$lang.dict"
+  JQ_FILTER='.. | if .type? == "STRING" or (.type? == "ALIAS" and .named? == false) then .value else null end'
+  jq "$JQ_FILTER" < "${lang_dir}/src/grammar.json" |
+		grep -v "\\\\" | grep -v null |
+		iconv -c -f UTF-8 -t ASCII//TRANSLIT |
+		awk '!/^""$/' >"test/fuzz/out/$lang.dict"
```
ahlinc, line 76: Are you sure that this shell pipeline makes the same as the deleted script?

ahlinc: The python script made output sorted and reencoded strings to escaped byte strings like:
```
"\x7b"
"\x7c"
"\x7c\x3d"
"\x7c\x7c"
```

ahlinc (changes requested):

script/build-fuzzers
amaanq: right, was debugging why clang wasnt selected by the makefile

script/build-fuzzers
amaanq: sure

script/build-fuzzers
amaanq: Yeah it's used here https://github.com/vigoux/tree-sitter-fuzz-action/blob/main/entrypoint.sh

The jq filter here is better as it removes empty strings and libfuzzer doesn't allow utf8 input, just ascii, for core languages it's fine but for grammars that use funky symbols (julia, dhall), it won't work

ahlinc: I don't know what's wrong with two backslash tokens, so why it's needed to drop them `grep -v "\\\\"` and what didn't   do with the python script.

But all above shell pipeline can be replaces with a single `jq` call:
```
< ../tree-sitter-c/src/grammar.json jq '
  [ ..
    | if .type? == "STRING" or (.type? == "ALIAS" and .named? == false) then .value else null end
    | select(. != null and test("\\S"))
  ] | unique | .[]
'
```

amaanq: To be honest I can't remember

But does your version strip special utf8 chars?

ahlinc: The bellow would filter all non ascii:
```
< ../tree-sitter-dhall/src/grammar.json jq '
  [ ..
    | if .type? == "STRING" or (.type? == "ALIAS" and .named? == false) then .value else empty end
    | select(test("\\S") and length == utf8bytelength)
  ] | unique | .[]
'
```

ahlinc: The original pipelined version doesn't remove duplicates and don't sort an input.

amaanq: that's better, thanks!

Issue #2396, opened 2023-07-16
Improve `tree-sitter highlight --check` for downstream use


clason:
#2340 added a  `--check` flag that ensures that highlight queries only use captures from an official list. This is also useful for downstream projects (Neovim, Helix, Emacs) that have a custom(ized) set of captures, to ensure they at least align on a "common core". To be able to use this test in their CI, however, requires that the test can be customized by

* [x] allowing to provide a list of custom captures in a config file that extends (or overrides?) the hard-coded "common core" list; 
* [x] providing a flag (also for `tree-sitter highlight` itself) that disables early termination on matching query (since Neovim -- [and Zed?](https://github.com/nvim-treesitter/nvim-treesitter/issues/3944#issuecomment-1458782497) -- always process and apply all queries in order);
* [ ] optionally also checking predicates and directives against a "common core" (to be agreed on, and similarly extensible with a custom list).

@patrickt

ahlinc:
> that disables early termination on matching query

If I understand you right and you means a behavior that for highlighting logic the last match wins instead of the first match like in Tree-sitter's current highlighting implementation. Then I think Tree-sitter's implementation just needs to migrate to the same logic like in Neovim because the last match win strategy:
1. Allows to combine queries in order that intuitively is in sync with overriding concept familiar for many.
1. Allows easily combine queries that capture a context before other context consuming queries.
1. Verified on practice in Neovim and Zed too.

clason:
Yes, exactly! I wouldn't advocate unconditionally changing this since it would be quite a breaking change for current users of the crate and CLI, but I understand that you'd rather not support two different strategies (and am happy that you seem to prefer "our" strategy, too 😄 ).

I'm assuming (hopefully correctly) that my suggestion is relatively straightforward to implement: just wrap the lines that break out early in a guard: https://github.com/tree-sitter/tree-sitter/blob/2a277879ab38ca6d61c5558ea870116b7fab9653/highlight/src/lib.rs#L929-L936

If you want to switch, I'd recommend a three-step deprecation strategy:
1. add the "last match wins" behavior as a flag, default `false`
2. make the default `true`
3. remove the flag (and the code).

ahlinc:
> If you want to switch, I'd recommend a three-step deprecation strategy:

It's hard to follow for such plan because potentially all `highlight.scm` files may need changes and we wouldn't be able to maintain two copies.

[...]

PR #2412, opened 2023-07-20
feat: add an `--apply-all-captures` argument to highlight & test

amaanq:
Related to #2396

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -78,6 +78,10 @@ fn run() -> Result<()> {
         .long("quiet")
         .short("q");
 
+    let apply_all_captures_arg = Arg::with_name("apply-all-captures")
+        .help("Apply all captures to highlights")
+        .long("apply-all-captures");
+
     let matches = App::new("tree-sitter")
         .author("Max Brunsfeld <maxbrunsfeld@gmail.com>")
         .about("Generates and tests parsers")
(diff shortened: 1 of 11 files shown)


cli/src/test_highlight.rs
```diff
@@ -162,68 +162,7 @@ pub fn test_highlight(
     let assertions =
         parse_position_comments(highlighter.parser(), highlight_config.language, source)?;
 
-    iterate_assertions(&assertions, &highlights, &highlight_names)?;
```
amaanq, line 165: I deleted all this since the code is the same as `iterate_assertions`, seemed to be an accidental duplicate that got refactored into a function

cli/src/main.rs
```diff
@@ -78,6 +79,11 @@ fn run() -> Result<()> {
         .long("quiet")
         .short("q");
 
+    let apply_all_captures_arg = Arg::with_name("apply-all-captures")
+        .help("Apply all captures to highlights")
+        .long("apply-all-captures")
+        .short("apply-all");
```
ahlinc, line 85: Short usually a one char length parameter and if I remember correctly latest clap version accept only single chars here. I wouldn't assign short name here at all.

amaanq: Yeah I added that a while ago before refactoring the cli, removed

amaanq:
A couple notes:

Queries by default are still used from either the one specified in package.json or tree-sitter-LANG/queries, this makes it hard for downstream to specify their own queries - I can add another flag for this if that'd be ok

I made the convention style for custom files simple, newline separated strings, but comments are allowed starting with `;`.

amaanq:
@patrickt @dcreager (or matter of fact anyone on @tree-sitter/semantic-code), it'd be nice to get some thoughts on the linked issue, Max's words, and this PR. Currently this PR is not a breaking change but a followup one to make last-wins precedence the default would be, as Max and I (and others) agree that it's definitely much better to work with than short-circuiting.

maxbrunsfeld (approved):

PR #2422, opened 2023-07-24
fix: take the last range element by default if the query error isn't found

### LINKED ISSUE #1753: CLI panics when parsing queries containing unclosed parenthesis error
Steps to reproduce:
1. `git clone https://github.com/tree-sitter/tree-sitter-python`
2. `tree-sitter highlight ./tree-sitter-python/examples/compound-statement-without-trailing-newline.py`
  Observe that the parser and queries work successfully.
3. Edit the file `./tree-sitter-python/queries/highlights.scm`, adding an unclosed parenthesis `(` at any blank line.
4. Repeat the highlight command.
  Observe that tree-sitter crashes.

I expect tree-sitter to report an error in the query file, just like other syntax errors.

Seems the bug is from the CLI, and it failed to locate the found error.

Stacktrace:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/loader/src/lib.rs:777:14
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: tree_sitter_loader::LanguageConfiguration::include_path_in_query_error
   4: once_cell::unsync::OnceCell<T>::get_or_try_init
   5: tree_sitter_loader::LanguageConfiguration::highlight_config
   6: tree_sitter::run
   7: tree_sitter::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Version:
tree-sitter 0.20.6

amaanq:
Closes #1753 

In the screenshot we can see the new range end is 1901 (from the example in the linked issue), `contains` will not include the last index, so an error at 1900 previously where the range was from `0..1900` would panic

image

Typical off-by-one bug

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -782,7 +782,7 @@ impl<'a> LanguageConfiguration<'a> {
         let (path, range) = ranges
             .iter()
             .find(|(_, range)| range.contains(&offset_within_section))
-            .unwrap();
+            .unwrap_or(ranges.last().unwrap());
         error.offset = offset_within_section - range.start;
         error.row = source[range.start..offset_within_section]
             .chars()


amaanq:
Ah you're right - I should've tested on multiple query files knowing that they get concatenated. I think your fix makes a lot more sense :sweat_smile:

maxbrunsfeld (approved):

PR #2423, opened 2023-07-24
feat!: use `Option<NonZeroU16>` for TSFieldIds

### LINKED ISSUE #1452: Return type for field IDs in Rust bindings could be `Option<NonZeroU16>`

amaanq:
Closes #1452

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -10,6 +10,7 @@ use std::{
     fmt, hash, iter,
     marker::PhantomData,
     mem::MaybeUninit,
+    num::NonZeroU16,
     ops,
     os::raw::{c_char, c_void},
     ptr::{self, NonNull},
@@ -92,6 +93,8 @@ pub enum LogType {
     Parse,
     Lex,
 }
+
+type FieldId = NonZeroU16;
 
 /// A callback that receives log messages during parser.
 type Logger<'a> = Box<dyn FnMut(LogType, &str) + 'a>;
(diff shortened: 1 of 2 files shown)


lib/binding_rust/lib.rs
```diff
@@ -93,6 +94,8 @@ pub enum LogType {
     Lex,
 }
 
+type TSFieldId = Option<NonZeroU16>;
```
maxbrunsfeld, line 97: Let's just call this `FieldId`, instead of `TSFieldId`; I think that matches the normal rust style a bit better.

amaanq: agreed

lib/binding_rust/lib.rs
```diff
@@ -93,6 +94,8 @@ pub enum LogType {
     Lex,
 }
 
+type FieldId = Option<NonZeroU16>;
```
maxbrunsfeld, line 97: Oh, also, I think the `FieldId` should be *inside* of the `Option`, so the type of `field_id_for_name` would be `Option<FieldId>`.

maxbrunsfeld (approved):

PR #2413, opened 2023-07-20
feat: Use lookup tables in ts_lex

osiewicz:
Summary: This pull request augments `tree-sitter-cli` with ability to generate lookup tables instead of if-else chains. This should reduce compile times of `tree-sitter-rust` and improve binary sizes all around. 

Detailed summary [is available here](https://github.com/osiewicz/ramblings/blob/tree-sitter-lut/posts/tree-sitter/001-lookup-tables/0001-lookup-tables.md) - it's also WIP.

PR #2428, opened 2023-07-24
fix(generate): add Vertical Tab and Form Feed to the whitespace class

amaanq:
Motivation: https://github.com/tree-sitter/tree-sitter-c/issues/117

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -475,7 +475,9 @@ impl NfaBuilder {
                 .add_char(' ')
                 .add_char('\t')
                 .add_char('\r')
-                .add_char('\n'),
+                .add_char('\n')
+                .add_char('\x0B')
+                .add_char('\x0C'),
             ClassPerlKind::Word => CharacterSet::empty()
                 .add_char('_')
                 .add_range('A', 'Z')


aryx (approved):

aryx:
What do other regexp engines do for this like Perl PCRE? Does \s cover form feed in PCRE?

aryx:
Is there a spec we can refer too? I think \s was introduced by Perl (and PCRE) so probably it's specified there.

amaanq:
Sure, though JS and Perl are a bit different I think 

Perl:
https://www.pcre.org/original/doc/html/pcrepattern.html

For compatibility with Perl, \s did not used to match the VT character (code 11), which made it different from the the POSIX "space" class. However, Perl added VT at release 5.18, and PCRE followed suit at release 8.34. The default \s characters are now HT (9), LF (10), VT (11), FF (12), CR (13), and space (32), which are defined as white space in the "C" locale. This list may vary if locale-specific matching is taking place. For example, in some locales the "non-breaking space" character (\xA0) is recognized as white space, and in others the VT character is not.

JS:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape
\s
Matches any [whitespace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space) or [line terminator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#line_terminators) character.

whitespace are \t, \vt, \ff, space, \xa0, \u{FEFF}, and a few other bizarre unicode ones linked in the site
line terminators are \u{200C}, \u{200D}, and \u{FEFF} (BOM)


Maybe we can just keep it as is, since both seem to agree on adding VT and FF as of late

aryx:
Your call (or maybe @maxbrunsfeld has strong opinion on this).

amaanq:
It'd be nice to get Max's thoughts on this

maxbrunsfeld:
I think we want to match ECMAScript's syntax, because the grammars are JS.

amaanq:
Gotcha - I've updated it now to fully cover the ranges that ECMAScript covers, the Mozilla page and the V8 engine both agree on this set.

amaanq:
Well, maybe we need to remove Byte-Order Mark (BOM, FEFF)...

amaanq:
Mind double-checking if this is good now @maxbrunsfeld... though this is a decent enough change to perhaps signal bumping the ABI to 15 soon, and subsequently bumping the tree-sitter version ;) (potentially after the api extension, lifetimes rework, cli clap upgrade to v4, and lookup tables PRs)

ahlinc:
Some more thoughts, it's needed to check how the BOM already handled by the parser because it shouldn't impact next token column offset but have impact on the bytes offset. If I remember correctly it's the current behavior and the BOM shouldn't be added in any case because it's actually not a space but a technical symbol.

Hmm, all above are valid for the BOM at the beginning of the input but what if it occurs somewhere in the input and not as a first bytes? How editors show it? Silently as a space or as a surrogate symbol?
If as a space then it makes sense to add it but also check that it will still be correctly skipped.

ahlinc:
> though this is a decent enough change to perhaps signal bumping the ABI to 15 soon, and subsequently bumping the tree-sitter version ;) (potentially after the api extension, lifetimes rework, cli clap upgrade to v4, and lookup tables PRs if any of those are ready/bear fruit)

From all above only the _"lookup tables PRs"_ is related to ABI version.

maxbrunsfeld:
The only concern I have about this is that it might make parsers larger across the board (in terms of code size), since this character class is used so commonly. On the other hand, https://github.com/tree-sitter/tree-sitter/pull/2413 might help mitigate this problem.

@amaanq Could you the binary size and compile time of a large example parser (like `tree-sitter-typescript`, or `tree-sitter-rust`) before and after this change?

If there's a significant size or compile-time regression, maybe we should start by just adding the form feed character, or whichever ones we think are common.

amaanq:
Rust - 794kb vs 811kb, but the compile time was 28s vs 1m19s on my machine with -O3 😬 
CPP - 2.4mb for both, same compile times
Python - 481kb for both, same compile times
Verilog: 18mb for both, same compile times


So, I think this problem scales exponentially the more where using lookup tables would apply (like Rust)

maxbrunsfeld:
@amaanq What do you think about reducing the characters we add for the time being, so as not to increase compile times too much?

amaanq:
That's fine with me, we can stick with VT and FF for now as they are more commonly used than the others

PR #2429, opened 2023-07-24
ts_node__prev_child: Reset earlier node when backtracking

### LINKED ISSUE #2421: Infinite loop in prev_sibling
Originally reported in https://github.com/helix-editor/helix/issues/7689

Under some conditions the `prev_sibling` function seems to cause an infinite loop inside tree sitter when called for the first sibling (so null should be returned). The original reproduction case in helix is somewhat complex. To make reproduction easier I have implemented a minimized reproduction case: 

https://github.com/pascalkuthe/tree-sitter-bug

The issue occurs with the markdown grammar in particular (present as submodule in the repo) but there have been reports of similar freezes for other grammars too.

# Reproduction Steps

* Requirements: unix operating system (tested on fedora linux 38), c compiler, rust 1.65 or newer (and matching cargo installation), git
* clone: https://github.com/pascalkuthe/tree-sitter-bug into a new directory
* inside that directory execute `cargo run`

# Expected Behavior

Program terminates with the following outupt:

``` text
root: {Node document (0, 0) - (1, 4)}
{Node < (1, 0) - (1, 1)}
{Node block_continuation (1, 0) - (1, 0)}
done: ... 
```
# Observed Behavior

Program prints the folloing output and then gets stuck in an infinite loop (never terminates):

``` text
root: {Node document (0, 0) - (1, 4)}
{Node < (1, 0) - (1, 1)}
{Node block_continuation (1, 0) - (1, 0)}
```

It seems that https://github.com/tree-sitter/tree-sitter/issues/1111 is somewhat related as the `block_continuation` token has lengh 0. However, this issue is different (as far as I understand) in that the parser terminates correctly and produces a valid tree. The infinite loop only occurs inside `prev_sibling`.

the-mikedavis:
`earlier_node` must be reset when backtracking up the tree or else `ts_node__prev_child` may loop infinitely on a subtree which contains zero-width tokens.

Fixes #2421

### DIFF
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -237,6 +237,8 @@ static inline TSNode ts_node__prev_sibling(TSNode self, bool include_anonymous) 
       return earlier_node;
     } else {
       node = earlier_node;
+      earlier_node = ts_node__null();
+      earlier_node_is_relevant = false;
     }
   }
 


amaanq:
Fix looks good, would like Max to check it out though

maxbrunsfeld:
It’d be great to add or enhance a unit test to reproduce the previous failure.

Thanks for the fix!

the-mikedavis:
Thank you!

PR #2431, opened 2023-07-25
Improve TextProvider implementation to allow owned text

ahlinc:
* Closes #1294
* Closes #2011 
* Closes #2432 

[elisp-tree-sitter](https://github.com/emacs-tree-sitter/elisp-tree-sitter) would be able to migrate on the latest Tree-sitter with a one line fix:
```diff
diff --git c/core/src/query.rs w/core/src/query.rs
index e2d5586..179f1dc 100644
--- c/core/src/query.rs
+++ w/core/src/query.rs
@@ -137,7 +137,7 @@ fn make_query_cursor() -> Result<QueryCursor> {
 fn text_callback<'e>(
     text_function: Value<'e>,
     error: &'e RefCell<Option<Error>>,
-) -> impl TextProvider<'e> {
+) -> impl TextProvider<Vec<u8>> + 'e {
     move |child: Node| {
         let beg = child.lisp_start_byte();
         let end = child.lisp_end_byte();
```

PR #2432, opened 2023-07-25
Separate text-provider lifetime

helixbass:
Hi, I ran into a lifetime issue with the signatures for `QueryCursor::matches()`/`QueryCursor::captures()`

Specifically the current signature is equating the lifetime of the text provider parameter with the lifetime of other parameters eg the query

But that caused a problem for me, I can try and make a small repro case if you want but basically I was doing something like:
```
fn outer<'a>(text_provider: impl TextProvider<'a>) {
    let mut query_cursor = QueryCursor::new();
    let query = &something;
    ...
    query_cursor.captures(query, ..., text_provider)....
}
```
and the compiler expected eg the `query` reference to live as long as `'a` which it doesn't

But so the changes in this PR did get rid of the lifetime issue for me

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -172,23 +172,25 @@ pub struct QueryMatch<'cursor, 'tree> {
 }
 
 /// A sequence of `QueryMatch`es associated with a given `QueryCursor`.
-pub struct QueryMatches<'a, 'tree: 'a, T: TextProvider<'a>> {
+pub struct QueryMatches<'a, 'tree: 'a, 'b: 'a, T: TextProvider<'b>> {
     ptr: *mut ffi::TSQueryCursor,
     query: &'a Query,
     text_provider: T,
     buffer1: Vec<u8>,
     buffer2: Vec<u8>,
     _tree: PhantomData<&'tree ()>,
+    _text_provider: PhantomData<&'b ()>,
 }
 
 /// A sequence of `QueryCapture`s associated with a given `QueryCursor`.
-pub struct QueryCaptures<'a, 'tree: 'a, T: TextProvider<'a>> {
+pub struct QueryCaptures<'a, 'tree: 'a, 'b: 'a, T: TextProvider<'b>> {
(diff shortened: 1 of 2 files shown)


lib/binding_rust/lib.rs
```diff
@@ -172,23 +172,25 @@ pub struct QueryMatch<'cursor, 'tree> {
 }
 
 /// A sequence of `QueryMatch`es associated with a given `QueryCursor`.
-pub struct QueryMatches<'a, 'tree: 'a, T: TextProvider<'a>> {
+pub struct QueryMatches<'a, 'tree: 'a, 'b: 'a, T: TextProvider<'b>> {
```
helixbass, line 175: Maybe a more specific name than `'b`? Eg `'text_provider`?

lib/binding_rust/lib.rs
```diff
@@ -172,23 +172,25 @@ pub struct QueryMatch<'cursor, 'tree> {
...
     ptr: *mut ffi::TSQueryCursor,
     query: &'a Query,
     text_provider: T,
     buffer1: Vec<u8>,
     buffer2: Vec<u8>,
     _tree: PhantomData<&'tree ()>,
+    _text_provider: PhantomData<&'b ()>,
```
helixbass, line 182: I'm not entirely clear on why this was necessary but without it the compiler was complaining about `'b` being unused (I would've thought that the fact that the struct stores a `T` which is associated with the lifetime `'b` would make it consider it "used"?)

ahlinc: Rust compiler considered a lifetime as bound only if it exists in a structure body.

ahlinc:
@helixbass thank you for contribution and the solution is correct but there already exist a #2431 PR that makes `TextProvider` generic over a return type so the lifetime defines a client code.

helixbass:
Ok I'm able to use the PR branch you linked to successfully (I'm able to use the new `TextProvider` etc type signatures without any lifetime issues), thanks

PR #2399, opened 2023-07-17
feat(cli)!: upgrade clap to v4 and rework cli args into structs

amaanq:

PR #2435, opened 2023-07-27
feat: add option to display context around parse errors

ZedThree:
Strawperson proposal for an option to `tree-sitter parse` that displays some context around errors. Example output:

```
==========
Parsing error:

   !> Possible characters encountered in a lexeme
   type :: enum_char
      character(1, tfc) :: space = tfc_" "
                                   ^~~~
                                   ERROR
      character(1, tfc) :: hash = tfc_"#"
      character(1, tfc) :: squote = tfc_"'"
```

Obviously there's a few things that could be done to polish this, for example including line numbers, optional size of the context, and so on.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -140,6 +140,7 @@ fn run() -> Result<()> {
                 .arg(&debug_graph_arg)
                 .arg(Arg::with_name("output-dot").long("dot"))
                 .arg(Arg::with_name("output-xml").long("xml").short("x"))
+                .arg(Arg::with_name("output-context").long("context"))
                 .arg(
                     Arg::with_name("stat")
                         .help("Show parsing statistic")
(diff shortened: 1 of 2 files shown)


amaanq:
It looks like a rudimentary implementation of ahlinc's fork with better cli output. I think that route is better to go.

ZedThree:
Ah, I had a look through PRs and issues here and didn't see anything. This was just quickly thrown together as I found it useful myself when writing grammars.

Is that feature from @ahlinc's fork likely to make it into the main repo?

amaanq:
Sooner or later, yeah. It's not perfect and has a few problems w/ edge cases that need to be solved (and ideally we'd want #2399 in first to make the successive PR more readable), but otherwise you can use it w/ this command:
`cargo install --git https://github.com/ahlinc/tree-sitter --branch alpha`

ZedThree:
Thanks! I'll close this and I'll build on top of that then

amaanq:
Awesome!

PR #2437, opened 2023-07-27
feat: warn when unused conflicts are present in a grammar

### LINKED ISSUE #267: grammar conflicts usage report
I have a big grammar that was created incrementally, so I collected ~200 line of conflicts.
https://github.com/tree-sitter/tree-sitter-verilog/blob/master/grammar.js#L4553
I feel that some of then not used anymore, but I have no way of telling what was used and what not.
It would be nice having some sort of use covegarge report produces by the generator. Is it possible?

amaanq:
Closes #267 

Sample output w/ c++:

image

### DIFF
--- a/cli/src/generate/build_tables/build_parse_table.rs
+++ b/cli/src/generate/build_tables/build_parse_table.rs
@@ -57,6 +57,7 @@ struct ParseTableBuilder<'a> {
     parse_state_info_by_id: Vec<ParseStateInfo<'a>>,
     parse_state_queue: VecDeque<ParseStateQueueEntry>,
     non_terminal_extra_states: Vec<(Symbol, usize)>,
+    actual_conflicts: HashSet<Vec<Symbol>>,
     parse_table: ParseTable,
 }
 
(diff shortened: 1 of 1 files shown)


amaanq:
I wonder if this is okay as default behavior, we could add a flag if not

cli/src/generate/build_tables/build_parse_table.rs
```diff
@@ -972,13 +988,19 @@ pub(crate) fn build_parse_table<'a>(
         inlines,
         &item_set_builder,
     );
+    let actual_conflicts = syntax_grammar
+        .expected_conflicts
+        .clone()
+        .into_iter()
+        .collect();
```
ahlinc, line 995: For this place I'd suggest to rewrite it like in the commit: https://github.com/amaanq/tree-sitter/commit/9aa0b08ccd073782256dd0d2948edf0dad21c464

ahlinc (changes requested):

ahlinc:
I think it's okay to be default behavior because it would lead to less commited errors and if an author during a grammar development would see such warning it would be less of confuse of why added conflicts aren't work.

Visual representation makes sense to improve in a future PR where improve styling for all emitable messages.

PR #2438, opened 2023-07-28
feat: add bindings for C, Go, Python, and Swift

amaanq:
# Background

The haphazard levels of support for various bindings across parser repos is an issue, with first-class support practically only existing for rust and wasm thus far. This is an issue for many downstream consumers, such as users of our python bindings, users wanting to use a makefile for C, Go support being nonexistent, and Swift users having spotty support at best. 

# Changes

This PR addresses this variation in bindings support/quality across the ecosystem to ensure that every decently sized language that we would like support (More are welcome to be added via a PR! Just no memelangs) are actually supported properly. Thus running `tree-sitter generate` in a repo will add these bindings if they're missing, and users don't have to touch them at all to keep them maintained. This is a win-win for the languages getting proper support in this PR

# Future

We'd like users of various languages to have the proper support they need when using tree-sitter, and we're obviously missing some languages that are huge (Java, Kotlin, Ruby, C#, etc). If you are interested in having support, at the very minimum there needs to be some stable bindings for the target language. Afterwards, a PR adding support to auto-generate binding-stub files that handle the build process/FFI interop with your language would be most welcome.


cli/src/generate/templates/Makefile
```diff
@@ -0,0 +1,114 @@
...
+ifeq ($(shell uname),Darwin)
+	SOEXT = dylib
+	SOEXTVER_MAJOR = $(SONAME_MAJOR).dylib
+	SOEXTVER = $(SONAME_MAJOR).$(SONAME_MINOR).dylib
+	LINKSHARED := $(LINKSHARED)-dynamiclib -Wl,
+	ifneq ($(ADDITIONALLIBS),)
+	LINKSHARED := $(LINKSHARED)$(ADDITIONALLIBS),
+	endif
+	LINKSHARED := $(LINKSHARED)-install_name,$(LIBDIR)/libtree-sitter-$(PARSER_NAME).$(SONAME_MAJOR).dylib,-rpath,@executable_path/../Frameworks
+else
```
ahlinc, line 61: I'd add here explicit check for Linuxes and bellow an additional section that would abort with a message that platform is not supported to reject Windows users from attempt to use this makefile.

cli/src/generate/templates/go.sum
```diff
@@ -0,0 +1,12 @@
...
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/smacker/go-tree-sitter v0.0.0-20230720070738-0d0a9f78d8f8 h1:DxgjlvWYsb80WEN2Zv3WqJFAg2DKjUQJO6URGdf1x6Y=
+github.com/smacker/go-tree-sitter v0.0.0-20230720070738-0d0a9f78d8f8/go.mod h1:q99oHDsbP0xRwmn7Vmob8gbSMNyvJ83OauXPSuHQuKE=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.4.0/go.mod h1:YvHI0jy2hoMjB+UWwv71VJQ9isScKT/TqJzVSSt89Yw=
+github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/stretchr/testify v1.7.4/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
```
ahlinc, line 12: Is it possible to avoid all of this hash pins?

I feel a future stream of complains and PRs that someone needs a newer version.

ahlinc (changes requested):

cli/src/generate/templates/cargo.toml
```diff
@@ -23,4 +19,4 @@ path = "bindings/rust/lib.rs"
 tree-sitter = "~RUST_BINDING_VERSION"
 
 [build-dependencies]
-cc = "1.0"
+cc = "1.0.79"
```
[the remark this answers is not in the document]
amaanq: Hm, I'll just do ^1.0 then for compatibility

cli/src/generate/templates/cargo.toml
```diff
@@ -1,20 +1,16 @@
...
-keywords = ["incremental", "parsing", "PARSER_NAME"]
+description = "PARSER_NAME grammar for tree-sitter"
+autoexamples = false
 categories = ["parsing", "text-editors"]
-repository = "https://github.com/tree-sitter/tree-sitter-PARSER_NAME"
-edition = "2018"
+edition = "2021"
+keywords = ["incremental", "parsing", "PARSER_NAME"]
 license = "MIT"
+repository = "https://github.com/tree-sitter/tree-sitter-PARSER_NAME"
```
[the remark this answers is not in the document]
amaanq: Can we bump edition to 2021 and add PARSER_NAME to keywords though?

cli/src/generate/templates/go.sum
amaanq: Yeah, it's like Cargo.lock, it's annoying to build go without it since you'd have to run `go mod tidy` at first
image

ahlinc: How fast does the tidy work?
I think it's possible to silently run it in the folder as a generation step.

amaanq: a few seconds, it's not slow but just adds hidden steps for the user to do/know about

XVilka:
This "`eof` is set but not used" warning is quite annoying all the time BTW (irrelevant to that particular PR though).

amaanq:
Agreed, fixed

cli/src/generate/templates/go.sum
maxbrunsfeld: Can we avoid depending on any of this stuff, and export the raw language pointer, in a way that can be consumed by `go-tree-sitter`?

ahlinc:
* Added population of an empty dir to use `tree-sitter generate tree-sitter-foobar` as quick kickstarter.
* Added creation of a `grammar.js` template.
* Added creation of a `dsl.d.ts` file (excluded in .gitignore) and mentioned it on top of `grammar.js` template. This makes for no need to use `npm i tree-sitter-cli` to access DSL spec if someone already have Tree-sitter installed in some other way and it's preferred to them.

ahlinc (changes requested):
Still I'm not sure about the `Makefile` template correctness.

ahlinc:
TODO:
- [ ] Show a warning when some generated files like `dsl.d.ts` is out of sync and need to be deleted manually to be regenerated on updated version on the next `tree-sitter generate` call.
  - This is not related to binding files like `bindings/rust/build.rs` where users may do changes manually.
    - We may consider an option to manage such files automatically like adding an external scanner compilation step or including query file references. But this also would require to add some opt-out mechanism in the `package.json` like a list of `manually-managed-files`.

cli/src/generate/templates/Package.swift
```diff
@@ -0,0 +1,42 @@
...
+                    "test",
+                    "grammar.js",
+                    "LICENSE",
+                    "Makefile",
+                    "package.json",
+                    "README.md",
+                    "script",
+                    "src/grammar.json",
+                    "src/node-types.json",
+                    "*.wasm",
```
mattmassicotte, line 29: An exclude list is unnecessary for recent Swift versions. It is safe to just leave this array empty.

cli/src/generate/templates/gitignore
```diff
@@ -0,0 +1,18 @@
...
+
+# Zig artifacts
+/zig-cache/
+/zig-out/
+
+# Example dirs
+/examples/*/
+
+# Grammar extras
+/dsl.d.ts
```
mattmassicotte, line 18: Swift will produce a `.build` directory at the top level. Including a `\.build` ignore rule can catch that one.

amaanq:
@arthurzam given our recent collaboration in Bash, and to my knowledge your extensive usage of py-tree-sitter for parsing, can I ask you to check out the Python bindings I've introduced here? This should allow easier integration of a parser without requiring py-tree-sitter to load the library itself (which should be updated after this)

arthurzam:
> @arthurzam given our recent collaboration in Bash, and to my knowledge, your extensive usage of py-tree-sitter for parsing, can I ask you to check out the Python bindings I've introduced here? This should allow easier integration of a parser without requiring py-tree-sitter to load the library itself (which should be updated after this)

I'll be happy to help, but I'm unsure from where to enter this PR. Should I run the `tree-sitter-cli` from this branch on `tree-sitter-bash` and see the bindings? Can I request you to create a branch for tree-sitter-bash from where I can use and test.

I also want to ask some questions:

1. Should this be integrated in the git repository of each grammar repository? Or should it be generated for release artifacts only? (if you ask me, I'm unsure what is better)
2. Is the idea to generate a self containing grammar artifact (wheel for python, crate for rust, and such) so you can import it, import the main `tree-sitter` library and use it?
3. Would those artifacts, for example Python's wheel, be uploaded to supply website, like PyPI for python? I think you should, since it would give best ease of use to the users (well, left to grammar repo maintainer of course).

Also good news, is that in Gentoo project, I help maintaining big amount of python packages packaged for Gentoo, and I also maintain some python projects, so for Python bindings I can help plan the best way. I can also try to add the CI for generating the wheels on release (since we want to do for multiple python targets, multiple CPU arches, glibc/musl).
For issue where I'm unsure, I can ask for help from other strong Gentoo developers that handle Python.

cli/src/generate/templates/setup.py
```diff
@@ -0,0 +1,37 @@
...
+    ffibuilder.compile(verbose=True)
+    setup(
+        name='tree-sitter-PARSER_NAME',
+        version='0.0.1',
+        description='CAMEL_PARSER_NAME grammar for tree-sitter',
+        url='http://github.com/tree-sitter/tree-sitter-PARSER_NAME',
+        license='MIT',
+        packages=find_packages(),
+        package_data={
+            'tree_sitter_PARSER_NAME': ['_tree_sitter_PARSER_NAME.cpython-311-darwin.so'],
```
arthurzam, line 34: this name feels very bad, since you assume a lot of things (OS, python version).

theHamsta: You can get this suffix for the current Python interpreter via `distutils`
```
import distutils.sysconfig
distutils.sysconfig.get_config_var('EXT_SUFFIX')
```

arthurzam: You should be very careful of using `distutils`, since it was removed in python 3.12. There is a continuation as part of `setuptools`, so I recommend to verify it carefully that it works on python 3.12.

theHamsta: As package data, this `.so` shouldn't require any suffix as it is not thought to be importable as Python module but `dlopen`ed by python-tree-sitter.

cli/src/generate/templates/Makefile
```diff
@@ -0,0 +1,116 @@
...
+	PARSER_NAME := $(shell basename $(PARSER_REPO_URL))
+	PARSER_NAME := $(subst tree-sitter-,,$(PARSER_NAME))
+	PARSER_NAME := $(subst .git,,$(PARSER_NAME))
+endif
+
+ifeq (, $(PARSER_URL))
+	PARSER_URL := $(subst :,/,$(PARSER_REPO_URL))
+	PARSER_URL := $(subst git@,https://,$(PARSER_URL))
+	PARSER_URL := $(subst .git,,$(PARSER_URL))
+endif
```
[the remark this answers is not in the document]
amaanq: will just drop by for anyone else watching, it was to get the parser name via the git url but that was bad imo, so the first one was removed but parser url is kept for the pc file

cli/src/generate/templates/setup.py
amaanq: > this name feels very bad, since you assume a lot of things (OS, python version).

Hm, yeah, how should I go about this to be platform-agnostic?

amaanq:
> > @arthurzam given our recent collaboration in Bash, and to my knowledge, your extensive usage of py-tree-sitter for parsing, can I ask you to check out the Python bindings I've introduced here? This should allow easier integration of a parser without requiring py-tree-sitter to load the library itself (which should be updated after this)
> 
> I'll be happy to help, but I'm unsure from where to enter this PR. Should I run the `tree-sitter-cli` from this branch on `tree-sitter-bash` and see the bindings? Can I request you to create a branch for tree-sitter-bash from where I can use and test.
> 
> I also want to ask some questions:
> 
> 1. Should this be integrated in the git repository of each grammar repository? Or should it be generated for release artifacts only? (if you ask me, I'm unsure what is better)
> 2. Is the idea to generate a self containing grammar artifact (wheel for python, crate for rust, and such) so you can import it, import the main `tree-sitter` library and use it?
> 3. Would those artifacts, for example Python's wheel, be uploaded to supply website, like PyPI for python? I think you should, since it would give best ease of use to the users (well, left to grammar repo maintainer of course).
> 
> Also good news, is that in Gentoo project, I help maintaining big amount of python packages packaged for Gentoo, and I also maintain some python projects, so for Python bindings I can help plan the best way. I can also try to add the CI for generating the wheels on release (since we want to do for multiple python targets, multiple CPU arches, glibc/musl). For issue where I'm unsure, I can ask for help from other strong Gentoo developers that handle Python.

1. Yes, try out any grammar, say Bash. First you'd build tree-sitter on this branch, then you'd run `tree-sitter generate` and it'll generate the extra bindings

2.  Exactly, there's no easy way for a user to just clone a grammar and use bindings beyond the built-in rust and node ones, this adds support for several more languages to easily build and integrate a parser into an application

3. Yes I think this is the way, once we've nailed down the API for this we can create binary packages/releases for more languages. The way Rust does it is it compiles the parser (and scanner, if any) at build time inside the app. I think that'd be good for Python too but Node users seem to want prebuilt binaries, your insights here would be nice

cli/src/generate/templates/setup.py
arthurzam: [Documentation](https://cffi.readthedocs.io/en/latest/cdef.html#ffibuilder-compile-etc-compiling-out-of-line-modules)

Looks like you can rename it manually after calling `compile`, or try the setuptools variant. In that same page search for `from setuptools import setup` line and see the example and one example below it. I **think** it should work well.

arthurzam: @amaanq I can confirm sadly that the python isn't working when installed into a virtual environment, it can't find the library (even after I changed to match my package_data extension). I suspect it isn't imported into the correct place. Let me try to fix it.

amaanq: Hm, yeah I'm not super familiar with Python's build process, running `python setup.py install --user` gives me this file structure:

```
$ fd -t f
build/lib/tree_sitter_capnp/__init__.py
dist/tree_sitter_capnp-0.0.1-py3.11.egg
setup.py
tree_sitter_capnp/__init__.py
tree_sitter_capnp/_tree_sitter_capnp.c
tree_sitter_capnp/_tree_sitter_capnp.cpython-311-x86_64-linux-gnu.so
tree_sitter_capnp/_tree_sitter_capnp.o
tree_sitter_capnp.egg-info/PKG-INFO
tree_sitter_capnp.egg-info/SOURCES.txt
tree_sitter_capnp.egg-info/dependency_links.txt
tree_sitter_capnp.egg-info/not-zip-safe
tree_sitter_capnp.egg-info/top_level.txt
```

amaanq: And then, once you install it in your venv, you can import a language:

```sh
Python 3.11.5 (main, Aug 28 2023, 20:02:58) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tree_sitter_capnp
>>> tree_sitter_capnp.language()
<cdata 'struct TSLanguage *' 0x7f741c5f5d00>
>>> 
```

Which is just the raw TSLanguage struct which is meant to be used in py-tree-sitter

amaanq: It seems like ffibuilder.compile returns the file name so...
```py
if __name__ == "__main__":
    file = ffibuilder.compile(verbose=True)
    setup(
        name='tree-sitter-capnp',
        version='0.0.1',
        description='Capnp grammar for tree-sitter',
        url='http://github.com/tree-sitter/tree-sitter-capnp',
        license='MIT',
        packages=find_packages(),
        package_data={
            'tree_sitter_capnp': [file.split('/')[-1]]
        },
        zip_safe=False
    )
```

This might work dynamically

arthurzam: So, yeah, a little bit of background for python, since a lot has changed since, because of PEP517.

1. Don't call directly `setup.py` with commands - this is deprecated
2. Use `build` package to build a wheel package from the directory. For example: `python3.12 -m build -w` (which by the way fails because of missing `setup_requires=["cffi>=1.0.0"],`
3. Install it into a venv to verify it working:

```
python3.12 -m venv /tmp/venv
source /tmp/venv/bin/activate
pip install dist/*.whl
python
```

amaanq: I see - I get the same error. It seems setup_requires is deprecated in favor of using a pyproject.toml to specify requirements, should this be the way to go? When I do that and specify cffi and setuptools as requirements, installation works just fine

arthurzam: @amaanq Yes, you should use pyproject.toml file.

Also, I'm sorry, but I don't think I'll have time for OSS work in the near half-week, so if my input is a must, but you want to merge it, feel free to continue and I can return after to python bindings.

amaanq: No worries! Your insight here was already very helpful for properly building a python package lately.

One last question if at all possible - The build commands builds a .whl file, is that fine? I do have to run pip install dist/tree_sitter_lang-ver-py3-none-any.whl to actually have the package available, I'm guessing that's normal

arthurzam: By some recommendations, I'll investigate the possibility of using meson-python for this. It should bring us the best interface and result.

> One last question if at all possible - The build commands builds a .whl file, is that fine? I do have to run pip install dist/tree_sitter_lang-ver-py3-none-any.whl to actually have the package available, I'm guessing that's normal

Yes, this is the "binary artifact" for python packages. "tar.gz" for source and "whl" for binary.

amaanq: My end goal with this is as such, at least for python:

tree-sitter-{LANG} will provide python packages, at least the upstream grammars, and with that will distribute the packages to PyPI, where a user can then run `pip install tree-sitter-c` to instantly fetch the compiled? (or should they compile it on install?) parser. The language() function will basically do what `Language()` does in `py-tree-sitter`

cli/src/generate/templates/Makefile
```diff
@@ -0,0 +1,113 @@
...
+	sed -e 's|@LIBDIR@|$(LIBDIR)|;s|@INCLUDEDIR@|$(INCLUDEDIR)|;s|@VERSION@|$(VERSION)|' \
+		-e 's|=$(PREFIX)|=$${prefix}|' \
+		-e 's|@PREFIX@|$(PREFIX)|' \
+		-e 's|@ADDITIONALLIBS@|$(ADDITIONALLIBS)|' \
+		-e 's|@LANGUAGENAME@|$(LANGUAGE_NAME)|' \
+		-e 's|@PARSERURL@|$(PARSER_URL)|' \
+		bindings/c/tree-sitter.pc.in > $@
+
+install: all
+	install -d '$(DESTDIR)$(LIBDIR)'
```
mavit, line 99: Is this the best place to install parser libraries, or would they better belong in a `tree-sitter` subdirectory?

Is my understanding correct that the parser libraries aren't stand-alone libraries that can be linked by any program, but instead have to be loaded by calling the appropriate function in `libtree-sitter`?  If so, I think that means they probably don't belong on in the default system library path.

ahlinc: They can be used as a standalone libraries that provide a parser data and a lexing logic for the parser core from `libtree-sitter.so` but if to install grammars to a system path as system libraries it's required to follow a lib names convention and name parser libs like `libtree-sitter-c.so`.

ObserverOfTime:
A few more bindings that could be added in the future if there's interest:

<details>
<summary><b>Haskell</b></summary><br/>

**Bindings**

- [haskell-tree-sitter](https://github.com/tree-sitter/haskell-tree-sitter) (official)
  - May need an update

**Files**

`bindings/haskell/tree-sitter-{language}.cabal`

Example: [tree-sitter-rust.cabal](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/tree-sitter-rust/tree-sitter-rust.cabal)

- Some stuff can be removed

`bindings/haskell/{Language}.hs`

Example: [Rust.hs](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/tree-sitter-rust/TreeSitter/Rust.hs)

- `getTestCorpusDir` can be removed
- Methods for query files should be added
- Contents should be [embedded at compile time](https://hackage.haskell.org/package/file-embed-0.0.16.0/docs/Data-FileEmbed.htm)

**Publishing**

[Hackage](https://hackage.haskell.org/)

</details>

<details>
<summary><b>Ruby</b></summary><br/>

**Bindings**

- [ruby-tree-sitter](https://github.com/tree-sitter/ruby-tree-sitter.old) (official)
  - Dead (should be archived)
- [ruby-tree-sitter](https://github.com/Faveod/ruby-tree-sitter)
  - Can't build the parser from source
- [ruby-tree-sitter-ffi](https://github.com/calicoday/ruby-tree-sitter-ffi)
  - Not production ready
  - May need an update

**Files**

_TBD_

**Publishing**

[RubyGems](https://rubygems.org/)

</details>

<details>
<summary><b>C#</b></summary><br/>

**Bindings**

- [csharp-tree-sitter](https://github.com/tree-sitter/csharp-tree-sitter) (official)
  - Work in progress
  - Windows-only
- [dotnet-tree-sitter](https://github.com/profMagija/dotnet-tree-sitter)
  - May need an update
  - Windows-only

**Files**

_TBD_

**Publishing**

[NuGet](https://www.nuget.org/)

</details>

<details>
<summary><b>Java</b></summary><br/>

**Bindings**

- [java-tree-sitter](https://github.com/serenadeai/java-tree-sitter)
  - Can't build the parser from source
  - May need an update
- [android-tree-sitter](https://github.com/AndroidIDEOfficial/android-tree-sitter)
  - Android only
- [tree-sitter-ng](https://github.com/bonede/tree-sitter-ng)

**Files**

_See [tree-sitter-javascript](https://github.com/bonede/tree-sitter-ng/tree/main/tree-sitter-javascript) (tree-sitter-ng)_

**Publishing**

[Maven central](https://central.sonatype.com/)

</details>

<details>
<summary><b>Kotlin</b></summary><br/>

**Bindings**

- [kotlintree](https://github.com/oxisto/kotlintree)
  - Old and incomplete
- _Might write my own_

**Files**

_TBD_

**Publishing**

[Maven central](https://central.sonatype.com/)

</details>

<details>
<summary><b>Lua</b></summary><br/>

**Bindings**

- [ltreesitter](https://github.com/euclidianAce/ltreesitter)
  - Can't build the parser from source
- [lua-tree-sitter](https://github.com/xcb-xwii/lua-tree-sitter)
  - Can't build the parser from source

**Files**

_TBD_

**Publishing**

[LuaRocks](https://luarocks.org/)

</details>

<details>
<summary><b>Perl</b></summary><br/>

**Bindings**

- [Text::Treesitter](https://metacpan.org/pod/Text::Treesitter)
  - Can't build the parser from source

**Files**

_TBD_

**Publishing**

[CPAN](https://www.cpan.org/)

</details>

<details>
<summary><b>R</b></summary><br/>

**Bindings**

- [rtreesitter](https://github.com/jimhester/rtreesitter)
  - Can't build the parser from source

**Files**

_TBD_

**Publishing**

[CRAN](https://cran.r-project.org/)

</details>

<details>
<summary><b>D</b></summary><br/>

**Bindings**

- [d-tree-sitter](https://github.com/aminya/d-tree-sitter)
  - May need an update

**Files**

_TBD_

**Publishing**

[DUB](https://code.dlang.org/)

</details>

---

[And more](https://github.com/search?q=tree-sitter%20bindings&type=repositories)

cli/src/generate/templates/binding_test.go
```diff
@@ -0,0 +1,14 @@
+package tree_sitter_PARSER_NAME_test
+
+import (
+	"testing"
+
```
ObserverOfTime, line 5:
```suggestion

    tree_sitter "github.com/smacker/go-tree-sitter"
```

cli/src/generate/templates/binding_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/tree-sitter/tree-sitter-PARSER_NAME"
+)
+
+func TestCanLoadGrammar(t *testing.T) {
+	parser := tree_sitter_PARSER_NAME.Language()
```
ObserverOfTime, line 10:
```suggestion
	language := tree_sitter_PARSER_NAME.Language()
	parser := tree_sitter.NewLanguage(language)
```

cli/src/generate/templates/build.rs
```diff
@@ -2,16 +2,12 @@ fn main() {
...
-        .flag_if_supported("-Wno-unused-but-set-variable")
-        .flag_if_supported("-Wno-trigraphs");
+    c_config.include(src_dir);
     let parser_path = src_dir.join("parser.c");
     c_config.file(&parser_path);
 
-    // If your language uses an external scanner written in C,
-    // then include this block of code:
+    // If your language uses an external scanner,
+    // then uncomment this block of code:
```
ObserverOfTime, line 10:
```suggestion
    // NOTE: if your language has an external scanner, uncomment this block:
```

cli/src/generate/templates/build.rs
```diff
@@ -21,20 +17,4 @@ fn main() {
 
     c_config.compile("parser");
```
ObserverOfTime, line 18:
```suggestion
    c_config.compile("PARSER_NAME");
```

amaanq: maybe tree-sitter-PARSER_NAME?

ObserverOfTime: It doesn't matter much since this just generates a static file.

cli/src/generate/templates/Makefile
```diff
@@ -0,0 +1,103 @@
...
+OBJ := $(addsuffix .o,$(basename $(SRC)))
+
+# ABI versioning
+SONAME_MAJOR := 0
+SONAME_MINOR := 0
+
+CFLAGS ?= -O3 -Wall -Wextra -I$(SRC_DIR)
+CXXFLAGS ?= -O3 -Wall -Wextra -I$(SRC_DIR)
+override CFLAGS += -std=gnu99 -fPIC
+override CXXFLAGS += -fPIC
```
ObserverOfTime, line 36: C++ is no longer used and tree-sitter is C11.
```suggestion
override CFLAGS += -std=gnu11 -fPIC
```

cli/src/generate/templates/Makefile
ObserverOfTime, line 1: We could also install queries under `/usr/share/tree-sitter/queries/PARSER_NAME`

cli/src/generate/templates/Makefile
ObserverOfTime: > Is this the best place to install parser libraries, or would they better belong in a `tree-sitter` subdirectory?

That's not a bad idea (GStreamer plugins for one also live in `/usr/lib/gstreamer-1.0`),
but distros like [Arch] and [Alpine] already have some parser libraries directly in `/usr/lib`.

[Alpine]: https://pkgs.alpinelinux.org/contents?branch=edge&name=tree-sitter-bash&arch=x86_64&repo=community
[Arch]: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tree-sitter-bash

cli/src/generate/templates/setup.py
```diff
@@ -0,0 +1,52 @@
...
+    package_data={
+        "tree_sitter_LOWER_PARSER_NAME": ["*.pyi", "py.typed"],
+        "tree_sitter_LOWER_PARSER_NAME.queries": ["*.scm"],
+    },
+    ext_package="tree_sitter_LOWER_PARSER_NAME",
+    ext_modules=[
+      Extension(
+        name="_binding",
+        sources=[
+            "bindings/python/binding.c",
```
arthurzam, line 34: I've tried to use this at commit a9a7b9a4689e5581d6a8743c9d94b6fbeb47cd25 over tree-sitter-bash, and it was failing for me to compile when building the wheel using `pip wheel .`

<details><summary>Output of command: pip wheel .</summary>
<p>

```
Processing /tmp/tree-sitter-bash
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: tree-sitter-bash
  Building wheel for tree-sitter-bash (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for tree-sitter-bash (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      running bdist_wheel
      running build
      copying queries/highlights.scm -> build/lib.linux-x86_64-cpython-311/tree_sitter_bash/queries
      running build_py
      copying bindings/python/tree_sitter_bash/__init__.py -> build/lib.linux-x86_64-cpython-311/tree_sitter_bash
      running egg_info
      writing bindings/python/tree_sitter_bash.egg-info/PKG-INFO
      writing dependency_links to bindings/python/tree_sitter_bash.egg-info/dependency_links.txt
      writing top-level names to bindings/python/tree_sitter_bash.egg-info/top_level.txt
      reading manifest file 'bindings/python/tree_sitter_bash.egg-info/SOURCES.txt'
      adding license file 'LICENSE'
      writing manifest file 'bindings/python/tree_sitter_bash.egg-info/SOURCES.txt'
      copying bindings/python/tree_sitter_bash/__init__.pyi -> build/lib.linux-x86_64-cpython-311/tree_sitter_bash
      copying bindings/python/tree_sitter_bash/binding.c -> build/lib.linux-x86_64-cpython-311/tree_sitter_bash
      copying bindings/python/tree_sitter_bash/py.typed -> build/lib.linux-x86_64-cpython-311/tree_sitter_bash
      running build_ext
      building '_binding' extension
      x86_64-pc-linux-gnu-gcc -Wsign-compare -DNDEBUG -march=native -mtune=native -O2 -pipe -fPIC -DPy_LIMITED_API=0x03080000 -DPY_SSIZE_T_CLEAN -Isrc -I/tmp/venv/include -I/usr/include/python3.11 -c bindings/python/binding.c -o build/temp.linux-x86_64-cpython-311/bindings/python/binding.o -std=c11
      cc1: fatal error: bindings/python/binding.c: No such file or directory
      compilation terminated.
      error: command '/usr/bin/x86_64-pc-linux-gnu-gcc' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tree-sitter-bash
Failed to build tree-sitter-bash                                                                                                                                                                                   
ERROR: Failed to build one or more wheels
```


</p>
</details> 

When I looked for that file, I found it under `bindings/python/tree_sitter_bash/binding.c`, so I guess we are missing here the directory. Adding it made it compile.

I then didn't forget to add `scanner.c`, and now it fully works after the change above:

```
$ pip wheel .
$ pip install tree_sitter_bash-0.0.1-cp38-abi3-linux_x86_64.whl
$ python3.11
>>> import tree_sitter_bash
>>> tree_sitter_bash.language()
140029515652384
```

ubolonton:
I'm late to this, but this is my opinion based on my experience with tree-sitter language bindings.

For context: At work, I built a Java binding, and an in-house DSL with bindings for NodeJS, Python and JVM (all use tree-sitter for parsing). I also built the Emacs Lisp (dynamic module) binding. 

IIUC, the approach here is generating additional glue code (and build files) for each language binding. This is similar to the approach taken by the NodeJS bindings.

My preferred approach has been different: building dynamic shared libs from each grammar's generated `src` (without any additional glue), letting the language binding load these, essentially treating them as resources. This way:
- Each grammar needs to be built only once.
- It's easier to support multiple platforms. Cross-building only the grammar's `src` (e.g. with `zig`) is simple. Cross-building the generated glue won't be.
- It's easier to integrate with other build setups, which only need to deal with `tree-sitter.{dylib,so.dll}` as resources.
- It's more robust against host language's tooling churn.

Note that it does not necessarily preclude being able to publish grammars as self-contained packages, as long as the language supports resource-only packages. At work, I use PyPI wheels and Maven jars that contain only pre-built shared libs. They work across platforms (Linux/Windows/macOS x AMD/ARM), and are painless, for both package providers and consumers. The NodeJS bindings, on the other hand, have been a huge source of pain.

I think it's great that we have both options. However, it seems like `py-tree-sitter` is going to also remove `Language.load()`, because of this. I think that would be a mistake.

ubolonton:
> @ubolonton Just to clarify, you're not proposing to remove the generated packaging files like `package.json`, `pyproject.toml`, `go.mod`, `Package.swift`, and `Cargo.toml`, right? But you think some of the code files like `js-binding.cc`, `index.js`, `setup.py` and `__init__.py` are unnecessary?
>
Yes, I think the generated code files are unnecessary. (Excluding code that is run at build time, like`setup.py`.)

> Which particular generated files would you remove, and how would that change the way that grammars are consumed?

I would do it like this:

### Python
- Remove `binding.c`, `py.typed`, `__init.__{py,pyi}`.
    ```bash
    tar -tzf dist/tree_sitter_python-0.21.0.tar.gz
    tree_sitter_python-0.21.0/LICENSE
    tree_sitter_python-0.21.0/README.md
    tree_sitter_python-0.21.0/pyproject.toml
    tree_sitter_python-0.21.0/src/grammar.json
    tree_sitter_python-0.21.0/src/node-types.json
    tree_sitter_python-0.21.0/src/parser.c
    tree_sitter_python-0.21.0/src/scanner.c
    tree_sitter_python-0.21.0/src/tree_sitter/alloc.h
    tree_sitter_python-0.21.0/src/tree_sitter/array.h
    tree_sitter_python-0.21.0/src/tree_sitter/parser.h
    tree_sitter_python-0.21.0/PKG-INFO
    ```
    ```bash
    tar -tzf dist/tree_sitter_python-0.21.0-cp38-abi3-macosx_12_0_arm64.whl
    tree_sitter_python/tree_sitter_python.dylib
    tree_sitter_python-0.21.0.dist-info/METADATA
    tree_sitter_python-0.21.0.dist-info/WHEEL
    tree_sitter_python-0.21.0.dist-info/licenses/LICENSE
    tree_sitter_python-0.21.0.dist-info/RECORD
    ```
- Publish a PEP517-compatible build backend. Use it in `pyproject.toml` and remove `setup.py`. The backend can shell out to either the platform's C compiler, or the `ziglang` package. (I would prefer the latter, which is more self-contained and reproducible, and enables easy cross-builds.)
    ```toml
    [build-system]
    requires = ["tree-sitter-build==0.1.0"]
    build-backend = "tree-sitter-build"
    ```
- Use `importlib.resources` to locate the dynamic shared lib at run time:
    ```python
    import importlib.resources

    from tree_sitter import Language
    import tree_sitter_python

    # This is for macOS. There would be some additional code to determine the suffix.
    language = Language(str(importlib.resources.files(tree_sitter_python) / 'tree_sitter_python.dylib'), 'python')

    # py-tree-sitter can also wrap the above so we can pass `tree_sitter_python` module directly:
    language = Language(tree_sitter_python)
    ```

### Java

```bash
tar -tzf jvm/target/tree-sitter-python-0.21.0.jar
META-INF/MANIFEST.MF
com/github/tree_sitter/python/jna/darwin-aarch64/tree_sitter_python.dylib
com/github/tree_sitter/python/jna/darwin-x86-64/tree_sitter_python.dylib
com/github/tree_sitter/python/jna/linux-aarch64/tree_sitter_python.so
com/github/tree_sitter/python/jna/linux-x86-64/tree_sitter_python.so
com/github/tree_sitter/python/jna/win32-x86-64/tree_sitter_python.dll
```

At run time, provide the namespace path, and the language name.
Underneath, the binding uses JNA to determine the right platform-specific suffix, then loads it with JNA.
```java
var language = Language.load("/com/github/tree_sitter/python/", "python");
```

### NodeJS
I'm not 100% sure it's doable, but the idea is similar to the Java case:
- Change `tree-sitter`'s NodeJS binding to be able to load a dylib at run time.
- Replace the grammars' binding files with a build script (which may just rely on a `tree-sitter-build` package).
- Use that build script to pre-build the dylibs for all supported platforms and put all of them in the tarball.

ObserverOfTime:
What about Rust, Go, and Swift bindings? What about query files and grammar.json (see Rust)? What about tests?

This would just make the whole process far more complicated with the sole benefit being the removal of a few files.

maxbrunsfeld:
For Node.js in particular, we generate classes for each node-type, to give fancier getter methods to certain nodes, so we need to generate a script that does that, unless we want to drop that feature.

maxbrunsfeld:
Overall, since we already need to have some generated config files for each binding language, I don’t think it matters that much if there are additional generated code files.

ubolonton:
> What about Rust

For Rust, the tree-sitter binding can already take a pointer to a language object, so a mechanism to load (not link) a dylib would be needed. Adding such mechanism was easy for my Emacs binding (which was built in Rust). Adding it to the Rust binding would have been even easier.

> Go, and Swift bindings?

I can't propose solutions for all languages. The core ideas are already clear enough, though.

> What about query files and node-types.json (see Rust)?

These are orthogonal to this discussion. Regardless of the approach taken, they can be included.

> What about tests?

I could be missing something, but tests also seem to be orthogonal to this.

> This would just make the whole process far more complicated with the sole benefit being the removal of a few files.

Like lines of code, number of files is not a good measurement of complexity. Of course it's more upfront work, which requires expertise in each language's packaging, and careful design. The gain is robustness against churns. As the approach in this PR gets used by more language grammars and system packagers, they will likely start experiencing toils. (Of which the NAPI transition for NodeJS bindings will be just a canary in the coal mine.) I have experienced that pain, so I know to avoid it in my own systems.

To make it clear, **I'm not advocating for undoing this change**. Its existence is fine, as long as the others have the option to make the dylib alternative work. It will only be a problem when the tree-sitter binding for a language relies on it as **the sole mechanism to consume compiled grammars**. The NodeJS binding was an example in that direction. Due to the pain with that binding, I'm close to the point of forking it, or writing an alternative. Therefore, the recent deprecation of `Language(path, ...)` in `py-tree-sitter` worried me. However, I've checked again, and looks like it's going to introduce `Language(ffi_pointer)`, so I'm relieved for now.

> For Node.js in particular, we generate classes for each node-type, to give fancier getter methods to certain nodes, so we need to generate a script that does that, unless we want to drop that feature.

I think this is a great argument to justify the tradeoff. Unfortunately for me, my uses cases don't need those.

nateabele:
Just FYI I had to roll back to 0.21 because this broke my project with an error about not having a `tree-sitter` config in `package.json`, and to refer to the documentation... despite there not really being any.

Is it strictly necessary to break the base-case configuration in order to implement this?

PR #2445, opened 2023-07-31
fix: do not allow eof to advance states if the state is the same state (token repetition)

### LINKED ISSUE #2040: Lexing forever with regex that matches `\0`
With the following `grammar.js`

```js
module.exports = grammar({
    name: 'hangs',

    rules: {
        // Match a string with only ASCII charecters
        source_file: $ => /"([[\x00-\x7f]&&[^"]])*"/,
    }
});
```

When run on `fail.txt`

```
"Hello
```

`tree-sitter parse ./fail.txt` loops forever, and produces no output.

`tree-sitter parse -d  ./fail.txt` outputs

```
new_parse
process version:0, version_count:1, state:1, row:0, col:0
lex_internal state:0, row:0, column:0
  consume character:'"'
  consume character:'H'
  consume character:'e'
  consume character:'l'
  consume character:'l'
  consume character:'o'
```

and then hangs

The generated parser can be viewed [here](https://github.com/painwacc/tree-sitter-hang/blob/trunk/src/parser.c). That repo can also be cloned to quickly reproduce this locally

amaanq:
If we have a regex or token repetition that includes the null character, there is a state that loops back to itself because it's meant to match as much as possible for one token. That state looks something like this (for a regex including \x01-\x7f)

```c
    case 4:
      ACCEPT_TOKEN(aux_sym_source_file_token1);
      if ((1 <= lookahead && lookahead <= 127)) ADVANCE(4);
      END_STATE();
```

This will eventually short-circuit and end once EOF is reached, but let's say we change the regex to include the null character (\x00-\x7f), the state looks like this
```c
    case 4:
      ACCEPT_TOKEN(aux_sym_source_file_token1);
      if ((0 <= lookahead && lookahead <= 127)) ADVANCE(4);
      END_STATE();
```

If the token doesn't end and we hit eof (which can happen in the middle of typing or with incomplete code), we loop indefinitely. A check should be added such that we're not at the end of the file, which this PR does by checking if the transition index and current state are equal, and if so, change the if statement to something like the following:
```c
    case 4:
      ACCEPT_TOKEN(aux_sym_source_file_token1);
      if (!eof && (0 <= lookahead && lookahead <= 127)) ADVANCE(4);
      END_STATE();
```

Closes #2040


ahlinc:
Please add tests to avoid regressions in the future.

cli/src/generate/render.rs
```diff
@@ -867,6 +878,7 @@ impl Generator {
         ranges: &[Range<char>],
         is_included: bool,
         indent_count: usize,
+        skip_eof: bool,
```
maxbrunsfeld, line 881: I think it would be correct to *always* add the `!eof &&` clause if the character range contains the null character. It seems like we shouldn't ever be "advancing" if we're at EOF.

maxbrunsfeld (commented):
Thanks so much for the fix @amaanq. Great improvement. I left one comment about an opportunity for simplification, which I think you mentioned too.

ahlinc (approved):

Issue #1156, opened 2021-06-10
Missing nodes for some token() constructs


mjambon:
All `token()` constructs should produce a node in the CST, but not all of them do. This is a similar problem as for patterns (see #1151) but it's trickier to deal with because:
1. Depending on their contents, the `token()` will or won't produce a node.
2. Such construct may contain precedence annotations, and moving it to form a named rule doesn't work (this is the workaround described for #1151).

Here's the grammar with annotations on which choice yields a node in the CST and which doesn't.
```javascript
module.exports = grammar({
  name: 'inline_token',
  rules: {
    program: $ => choice(
      token('hello'),              // ok
      token(seq('good', 'bye')),   // no node
      token(choice('a', 'b')),     // no node
      //token(repeat('c')),        // memory explosion on any input
      token(repeat1('d')),         // no node
      //token(optional('e')),      // memory explosion on any input
      token(prec.dynamic(0, 'f')), // ok
      token(prec(0, 'g')),         // ok
      token(prec.left(0, 'h')),    // ok
      token(prec.right(0, 'i')),   // ok
      token(token('k')),           // no node

      token.immediate('l'),                   // ok
      token.immediate(seq('m', 'm')),         // no node
      token.immediate(choice('n', 'o')),      // no node
      //token.immediate(repeat('p')),         // memory explosion on any input
      token.immediate(repeat1('q')),          // no node
      //token.immediate(optional('r')),       // memory explosion on any input
      token.immediate(prec.dynamic(0, 's')),  // ok
      token.immediate(prec(0, 't')),          // ok
      token.immediate(prec.left(0, 'u')),     // ok
      token.immediate(prec.right(0, 'v')),    // ok
      token.immediate(token.immediate('w')),  // no node

      token(token.immediate('x')),            // no node
      token.immediate(token('y')),            // no node
    ),
  }
});
```
Visualizing the tree obtained for various inputs is easy with @ahlinc's work in progress (https://github.com/ahlinc/tree-sitter/pull/4), which improves the output of `tree-sitter parse`. Here's how to see whether a node is created or not:
```
$ tree-sitter generate
$ tree-sitter-dev parse <(echo hello)
0:0  - 1:0    program
0:0  - 0:5      "hello"
$ tree-sitter-dev parse <(echo goodbye)
0:0  - 1:0    program
```

ahlinc:
@mjambon awesome investigation! :tada:
With all of these it would be good to address the #1111 issue too.

[...]

PR #2446, opened 2023-07-31
fix: disallow tokens that match the empty string

### LINKED ISSUE #98: Tokens that match the empty string can result in infinite loops during error recovery
This feels like a bug in tree-sitter, if I remove the stuff after `pkg demo =` then it at least stops crashing but I think a bug like that isnt good to have around.
Actually it seems to happen in a lot of scenarios, obviously my grammar is wrong but there should be error outputs then... not a stuck loop.

```
use bio
use std
use "test.use"

pkg demo = const sayhello : (-> void)
        ;;

```

```
{
   "name":"arithmetic",
   "extras":[
      {
         "type":"PATTERN",
         "value":"\\s"
      }
   ],
   "rules":{
      "translation_unit":{
         "type":"REPEAT",
         "content":{
            "type":"CHOICE",
            "members":[
               {
                  "type":"SYMBOL",
                  "name":"use"
               },
               {
                  "type":"SYMBOL",
                  "name":"pkg"
               }
            ]
         }
      },
      "pkg":{
         "type":"SEQ",
         "members":[
            {
               "type":"STRING",
               "value":"pkg"
            },
            {
               "type":"SYMBOL",
               "name":"system_lib_string"
            },
            {
               "type":"STRING",
               "value":"="
            },
            {
               "type":"SYMBOL",
               "name":"string_literal"
            }
         ]
      },
      "use":{
         "type":"SEQ",
         "members":[
            {
               "type":"STRING",
               "value":"use"
            },
            {
               "type":"CHOICE",
               "members":[
                  {
                     "type":"SYMBOL",
                     "name":"string_literal"
                  },
                  {
                     "type":"SYMBOL",
                     "name":"system_lib_string"
                  }
               ]
            }
         ]
      },
      "string_literal":{
         "type":"TOKEN",
         "content":{
            "type":"SEQ",
            "members":[
               {
                  "type":"STRING",
                  "value":"\""
               },
               {
                  "type":"REPEAT",
                  "content":{
                     "type":"CHOICE",
                     "members":[
                        {
                           "type":"PATTERN",
                           "value":"[^\"]"
                        },
                        {
                           "type":"STRING",
                           "value":"\\\""
                        }
                     ]
                  }
               },
               {
                  "type":"STRING",
                  "value":"\""
               }
            ]
         }
      },
      "system_lib_string":{
         "type":"TOKEN",
         "content":{
            "type":"SEQ",
            "members":[
               {
                  "type":"REPEAT",
                  "content":{
                     "type":"PATTERN",
                     "value":"\\w+"
                  }
               }
            ]
         }
      }
   }
}

```

### LINKED ISSUE #1111: Infinite loop with permanent memory consumption on empty token
**grammar.js**
```js
module.exports = grammar({
  name: 'tree_sitter_bugs',

  rules: {
    source_file: $ => '',
  }
});
```
**Commands for reproducing:**
```sh
tree-sitter parse -d <(echo -n '')  # Works as expected
tree-sitter parse -d <(echo -n 'a') # Infinite loop
```
**Debug log:**
```log
new_parse
process version:0, version_count:1, state:1, row:1, col:0
lex_internal state:0, row:1, column:0
lexed_lookahead sym:, size:0
shift state:2
process version:0, version_count:1, state:2, row:1, col:0
lex_internal state:0, row:1, column:0
lexed_lookahead sym:, size:0
detect_error
resume version:0
process version:0, version_count:1, state:0, row:1, col:0
skip_token symbol:
process version:0, version_count:1, state:0, row:1, col:0
skip_token symbol:
process version:0, version_count:1, state:0, row:1, col:0
skip_token symbol:
process version:0, version_count:1, state:0, row:1, col:0
skip_token symbol:
```

~~Similar issue #1110~~

May be such empty tokens should be prohibited on a `dsl.js` or a generator level, but having some fuse in the C library would be good, cause such behaviour in the parser infinitely consumes memory.

### LINKED ISSUE #1271: Tokens matching empty strings are allowed, and cause infinite loop
Grammar:
```js
module.exports = grammar({
  name: 'bugcase',

  rules: {
    source_file: $ => $.identifier,
    identifier: $ => token(repeat(/a/))
  }
});
```
Source file:
```
b
```

Full test case: https://gist.github.com/CyberShadow/fe5ce672cc68f065210a97ab116d69a7 (`npm install && node_modules/.bin/tree-sitter generate && node_modules/.bin/tree-sitter parse source.bugcase`).

Expected output (same as without `token(...)`):

```
The rule `identifier` matches the empty string.

Tree-sitter does not support syntactic rules that match the empty string
unless they are used only as the grammar's start rule.
```

Actual output:

The grammar compiles successfully, and upon attempting to parse the source file, `tree-sitter parse` loops infinitely until it consumes all available memory.

amaanq:
This keeps the behavior the same as non-token rules, and prevents an infinite loop bug

Closes #98
Closes #1111 
Closes #1271 
Fixes the memory explosion cases in #1156


Vicbiger21:
Mine

cli/src/generate/prepare_grammar/expand_tokens.rs
```diff
@@ -101,6 +113,16 @@ pub(crate) fn expand_tokens(mut grammar: ExtractedLexicalGrammar) -> Result<Lexi
 
     let mut variables = Vec::new();
     for (i, variable) in grammar.variables.into_iter().enumerate() {
+        if find_empty_rule(&variable.rule, i == 0) {
+            return Err(anyhow!(
+                "The rule `{}` matches the empty string.
+
+Tree-sitter does not support syntactic rules that match the empty string
+unless they are used only as the grammar's start rule.
```
ahlinc, line 121: Could you apply `indoc!` here?

ahlinc (changes requested):

cli/src/generate/prepare_grammar/expand_tokens.rs
amaanq: I copied it from `flatten_grammar.rs`

ahlinc: Ok, then somehow as a next step.

PR #2453, opened 2023-08-01
fix: suppress unused warnings for eof

amaanq:


XVilka (approved):

ahlinc (approved):

ahlinc:
Could you put into description how to reproduce the warning?

ahlinc:
I think you just broke tests the master by merging this without fixing the windows targets.

XVilka:
MSVC doesn't understand GNU style attributes, this is probably why.

PR #2454, opened 2023-08-01
fix: suppress unused warnings for eof 

amaanq:
Take 2, fixed for windows


lib/include/tree_sitter/parser.h
```diff
@@ -130,9 +130,16 @@ struct TSLanguage {
...
+#ifdef _MSC_VER
+#define UNUSED __pragma(warning(suppress : 4101))
+#else
+#define UNUSED __attribute__((unused))
+#endif
+
 #define START_LEXER()           \
   bool result = false;          \
   bool skip = false;            \
+  UNUSED_VAR                    \
```
ahlinc, line 142: May be `UNUSED` and not `UNUSED_VAR`?

ahlinc (changes requested):

lib/include/tree_sitter/parser.h
amaanq: yep..originally named it UNUSED_VAR but changed it in the def..

ahlinc (approved):

PR #2458, opened 2023-08-02
CLI: Re-use highlight cancellation flag for HTML highlights

the-mikedavis:
The `html` highlight function created its own cancellation flag which conflicts with the Ctrl-c handler set up in the CLI's `main` block for `tree-sitter highlight`. We can re-use the cancellation flag from that block to avoid a panic that happens when using `tree-sitter highlight -H <file>`

    thread 'main' panicked at 'Error setting Ctrl-C handler: MultipleHandlers', cli/src/util.rs:31:6

This change also aligns the parameters that `highlight::ansi` and `highlight::html` take.

### DIFF
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -1,4 +1,3 @@
-use super::util;
 use ansi_term::Color;
 use anyhow::Result;
 use lazy_static::lazy_static;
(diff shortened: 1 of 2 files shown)


amaanq:
Makes sense, thanks!

amaanq (approved):

maxbrunsfeld:
Nice work @the-mikedavis. Thanks!

PR #2460, opened 2023-08-02
Range: sort "ancestors" before "descendants"

helixbass:
I encountered an unexpected result when sorting a list of `Range`'s that included a "child" node whose `start_byte` coincided with its parent's `start_byte` - it seems maybe debatable but I'd expect the parent to be sorted _before_ the child, but I'm seeing the reverse

It looks like that is because `Range` derives `PartialOrd`/`Ord` which does "lexicographic ordering" so if the `start_byte` is the same it will sort the one with the smaller `end_byte` (aka the descendant) first

So this PR implements `PartialOrd`/`Ord` manually to achieve the expected ordering

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -5,7 +5,9 @@ mod util;
 use std::os::unix::io::AsRawFd;
 
 use std::{
-    char, error,
+    char,
+    cmp::Ordering,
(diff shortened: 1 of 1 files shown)


helixbass:
Sounds reasonable to me that "there is no canonical order for `Range`'s"

PR #2461, opened 2023-08-02
Remove `PartialOrd, Ord deriving from `Range` in Rust lib

ahlinc:
* Closes #2460 

> I think there shouldn't be PartialOrd, Ord on the Range at all because byte offsets and points are different things and the user can easily craft a situation when byte offsets stay the same but point rows changed significantly and vice versa by playing with space and new line symbols.

> I think users need to choice thoughtfully in the client code by what they want to compare by byte offsets or by points.


amaanq (approved):

ahlinc:
One another argument that ordering should be handled on client side:

Screenshot from 2023-08-02 19-21-23

On grammars where there may exist many parent nodes all such nodes may occupy the same _byte range_ and the same _points range_ so it's not meaningful to put such nodes in a list and try to sort to determine what is a parent for another.

PR #2464, opened 2023-08-03
feat: rework lifetimes, relax dependencies and tie only on implementations

ahlinc:
This PR:
* Adds meaningful names for lifetimes where it makes sense.
* Removes lifetime dependencies on structs and moves all dependencies on `impl`s.
* Removes redundant lifetime specifications and replaces with `'_` lifetime placeholder to let compiler deduce correct lifetime by itself in obvious situations.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -81,7 +81,7 @@ pub struct InputEdit {
 #[doc(alias = "TSNode")]
 #[derive(Clone, Copy)]
 #[repr(transparent)]
-pub struct Node<'a>(ffi::TSNode, PhantomData<&'a ()>);
+pub struct Node<'tree>(ffi::TSNode, PhantomData<&'tree ()>);
 
 /// A stateful object that this is used to produce a `Tree` based on some source code.
 #[doc(alias = "TSParser")]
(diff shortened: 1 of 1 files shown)


amaanq (approved):

smoelius:
I'm trying to test this code, as suggested in #2273.

How should I think of this parameter `I`, e.g., what intuitively does it iterate over? https://github.com/tree-sitter/tree-sitter/blob/c85f81f62515a915a2d26ef031c0fa013227d76d/lib/binding_rust/lib.rs#L202

ahlinc:
> How should I think of this type `I`, e.g., what intuitively does it iterate over?

As anything that can be dereferenced over `AsRef` trait to a `[u8]` slice of bytes, it could be anything like `&[u8]`, `Vec<u8>`, `String` or `str`. Anything from https://doc.rust-lang.org/nightly/std/convert/trait.AsRef.html for which `AsRef<[u8]>` or `AsRef<[T]>` implemented or any your own type with implemented `AsRef<[u8]>` trait.

smoelius:
Sorry for the noise. Changes that I thought were introduced by this PR were actually introduced long ago by https://github.com/tree-sitter/tree-sitter/pull/1130.

That PR's description gave me the additional background I needed.

And, yes, I believe all code I could not build previously now builds following these changes.

PR #2466, opened 2023-08-03
Web Bindings: Expose fieldNameForChild C function

tom95:
The function is currently not exposed. I believe the only way to get fields via the web bindings right now is to use the TreeCursor API instead.


amaanq (approved):

amaanq:
I'm not all that familiar w/ the node binding stuff but it looks okay

maxbrunsfeld (approved):

PR #2474, opened 2023-08-07
feat!: support the case-insensitive regex flag

### LINKED ISSUE #261: Support case-insensitive regex flag
Hello,
I follow the tutorial and change the identifier definition in this way:
`    identifier: $ => /[a-z_][a-z0-9_]*/i,`
but when I try to parse an identifier with uppercase character it raises an error.

Is it a bug?

amaanq:
Closes #261

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -139,10 +139,10 @@ pub(crate) fn expand_tokens(mut grammar: ExtractedLexicalGrammar) -> Result<Lexi
 impl NfaBuilder {
     fn expand_rule(&mut self, rule: &Rule, mut next_state_id: u32) -> Result<bool> {
         match rule {
-            Rule::Pattern(s) => {
+            Rule::Pattern(s, f) => {
                 let s = preprocess_regex(s);
                 let ast = parse::Parser::new().parse(&s)?;
-                self.expand_regex(&ast, next_state_id)
+                self.expand_regex(&ast, next_state_id, f.contains('i'))
             }
             Rule::String(s) => {
                 for c in s.chars().rev() {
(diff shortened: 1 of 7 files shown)


cli/src/generate/prepare_grammar/expand_tokens.rs
```diff
@@ -139,8 +141,11 @@ pub(crate) fn expand_tokens(mut grammar: ExtractedLexicalGrammar) -> Result<Lexi
 impl NfaBuilder {
     fn expand_rule(&mut self, rule: &Rule, mut next_state_id: u32) -> Result<bool> {
         match rule {
-            Rule::Pattern(s) => {
+            Rule::Pattern(s, f) => {
                 let s = preprocess_regex(s);
+                if f.contains('i') {
+                    self.case_insensitive = true;
+                }
```
ahlinc, line 148: May be it's better to pass flags directly to `expand_regex` and don't polute the self namespace?

cli/src/generate/prepare_grammar/expand_tokens.rs
```diff
@@ -229,6 +249,14 @@ impl NfaBuilder {
                     if class.negated {
                         chars = chars.negate();
                     }
+                    if self.case_insensitive {
+                        for char in chars.clone().chars() {
+                            let inverted = inverse_char(char);
+                            if char != inverted {
+                                chars = chars.add_char(inverted);
+                            }
+                        }
```
ahlinc, line 258: I'd wrap this to something like `with_inverse_char(&chars, |inverted| chars = chars.add_char(inverted))` and avoid repetition in other places.

ahlinc (changes requested):
Please look on my comments on the code.

cli/src/generate/prepare_grammar/expand_tokens.rs
amaanq: I thought about this before (and initially did it this way), it'd pollute expand_regex but that's fine

ahlinc: No, I meant it irrelevant for the self namespace because it's only used in expand regex so it's better to pass whole flags as additional parameter. It's not good idea to store everything in the self.

PR #2475, opened 2023-08-07
 Improve warning and errors  + doc improvements

### LINKED ISSUE #463: Document supported Javascript Regexp capabilities
I recently started playing around with tree-sitter. I've found it a pleasure to get started with, the documentation is easy to follow and I really like the integrated test system. Thank you!

One place where I found the documentation to be a bit lacking was when I tried to use some more advanced regex features (in particular Unicode property escapes like `\p{Ll}`). It's hard to figure out what the intended support of Javascript regexes and behaviour with certain given flags is. It would be nice to have some page listing the supported features (and differences in behaviour necessitated by the specific use-case and implementation of tree-sitter).

For [tree-sitter-beancount](https://github.com/yagebu/tree-sitter-beancount), I've resorted to using an additional build step using regexpu to compile it to an expression listing all characters separately. This makes it a bit harder to reason with in case of unicode-related grammar compilation errors as the build step itself might introduce more problems.

Regarding unicode escapes: it seems that the used regex_syntax crate has very complete [Unicode support](https://docs.rs/regex-syntax/0.6.12/regex_syntax/#crate-features), would it be hard to support them in tree-sitter as well?

### LINKED ISSUE #732: tree-sitter highlight has does not find the language.

### LINKED ISSUE #1159: Document that 'token($.foo)' is illegal
It looks like non-terminals (rule names) are not allowed in the body of `token()`. Adding a sentence to the documentation about what's allowed within `token()` would be great.

```
$ cat grammar.js
module.exports = grammar({
  name: "anon_pat",
  rules: {
    program: $ => seq(
      token($.x)
    ),
    x: $ => /x+/,
  }
});
```

```
$ tree-sitter generate

Error processing rule program_token1
Details:
  Grammar error: Unexpected rule Symbol(Symbol { kind: NonTerminal, index: 1 })

```

amaanq:
Closes #463 
Closes #732 (maybe, ticks the boxes in the last msg though)
Closes #1159


cli/loader/src/lib.rs
```diff
@@ -807,6 +810,13 @@ impl<'a> LanguageConfiguration<'a> {
                 path_ranges.push((path.clone(), prev_query_len..query.len()));
             }
         } else {
+            if default_path == "highlights.scm" {
```
ahlinc, line 813: Looks like a hack.

amaanq: Sort of, I didn't do it for other files like tags/injections, maybe we should warn about any defaults being used?

cli/src/util.rs
```diff
@@ -18,6 +22,33 @@ svg { width: 100%; }
 
 ";
 
+pub fn lang_not_found_for_path(path: &Path, loader_config: &LoaderConfig) -> String {
+    let path = path.display();
+    format!(
+        indoc! {"
+            No language found for path `{}`
+
+            If a language should be associated with this file extension, please ensure the path to `{}` is inside one of the following directories as specified by your 'config.json':\n\n{}\n
```
ruuda, line 31: I think this is incorrect for `tree-sitter highlight`. I built 2d652c90a7c60efe4b2820e257a7e79c1c5017d9 and it prints e.g.

```
$ tree-sitter highlight foo.bar
No language found for path `foo.bar`

If a language should be associated with this file extension, please ensure the path to `foo.bar` is inside one of the following directories as specified by your 'config.json':

  1. /home/ruud/repos
```

The file I’m trying to highlight should not need to be specified in the `config.json`.

amaanq: oh sorry, it should be extension of `foo.bar`, there must be a grammar dir where the filetype of `foo.bar` (bar) is in the repo's package.json

Issue #2478, opened 2023-08-08
implement first-line-regex


sigmaSd:
first-line-regex is documented here https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-detection

But in my testing it doesn't seem to work

Also looking at the source code it seems to be not implemented https://github.com/tree-sitter/tree-sitter/blob/82719cda01d52e098172e9f7c18b7dc8e04645d9/cli/loader/src/lib.rs#L573

PR #2479, opened 2023-08-08
feat: implement first-line-regex 

sigmaSd:
fix https://github.com/tree-sitter/tree-sitter/issues/2478

I added a PR for helix for the same thing https://github.com/helix-editor/helix/pull/7853

But there I went with a Vec<String> instead of a String, the example is strace output it can start with `["execve", ,"^[0-9:.]* *execve", "^__libc_start_main"]`

Obviously this is not a structured language by any mean (though I wrote a wip grammar for it) so it might be an invalid argument

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -85,7 +85,7 @@ const BUILD_TARGET: &str = env!("BUILD_TARGET");
 pub struct LanguageConfiguration<'a> {
     pub scope: Option<String>,
     pub content_regex: Option<Regex>,
-    pub _first_line_regex: Option<Regex>,
+    pub first_line_regex: Option<Regex>,
     pub injection_regex: Option<Regex>,
     pub file_types: Vec<String>,
     pub root_path: PathBuf,
@@ -107,6 +107,7 @@ pub struct Loader {
     language_configurations: Vec<LanguageConfiguration<'static>>,
     language_configuration_ids_by_file_type: HashMap<String, Vec<usize>>,
     language_configuration_in_current_path: Option<usize>,
+    language_configuration_ids_by_first_line_regex: HashMap<String, Vec<usize>>,
     highlight_names: Box<Mutex<Vec<String>>>,
     use_all_highlight_names: bool,
     debug_build: bool,
@@ -138,6 +139,7 @@ impl Loader {
             language_configurations: Vec::new(),
             language_configuration_ids_by_file_type: HashMap::new(),
             language_configuration_in_current_path: None,
+            language_configuration_ids_by_first_line_regex: HashMap::new(),
             highlight_names: Box::new(Mutex::new(Vec::new())),
             use_all_highlight_names: true,
             debug_build: false,
@@ -239,6 +241,26 @@ impl Loader {
                     .and_then(|extension| {
                         self.language_configuration_ids_by_file_type.get(extension)
                     })
+            })
+            .or_else(|| {
+                let Ok(file) = fs::File::open(path) else {
+                    return None;
+                };
+                let reader = BufReader::new(file);
+                let Some(Ok(first_line)) = std::io::BufRead::lines(reader).next() else {
+                    return None;
+                };
+
+                self.language_configuration_ids_by_first_line_regex
+                    .iter()
+                    .find(|(regex, _)| {
+                        if let Some(regex) = Self::regex(Some(regex)) {
+                            regex.is_match(&first_line)
+                        } else {
+                            false
+                        }
+                    })
+                    .map(|(_, ids)| ids)
             });
 
         if let Some(configuration_ids) = configuration_ids {
@@ -869,9 +891,9 @@ impl Loader {
                         scope: config_json.scope,
                         language_id,
                         file_types: config_json.file_types.unwrap_or(Vec::new()),
-                        content_regex: Self::regex(config_json.content_regex),
-                        _first_line_regex: Self::regex(config_json.first_line_regex),
-                        injection_regex: Self::regex(config_json.injection_regex),
+                        content_regex: Self::regex(config_json.content_regex.as_deref()),
+                        first_line_regex: Self::regex(config_json.first_line_regex.as_deref()),
+                        injection_regex: Self::regex(config_json.injection_regex.as_deref()),
                         injections_filenames: config_json.injections.into_vec(),
                         locals_filenames: config_json.locals.into_vec(),
                         tags_filenames: config_json.tags.into_vec(),
@@ -885,6 +907,12 @@ impl Loader {
                     for file_type in &configuration.file_types {
                         self.language_configuration_ids_by_file_type
                             .entry(file_type.to_string())
+                            .or_default()
+                            .push(self.language_configurations.len());
+                    }
+                    if let Some(first_line_regex) = &configuration.first_line_regex {
+                        self.language_configuration_ids_by_first_line_regex
+                            .entry(first_line_regex.to_string())
                             .or_default()
                             .push(self.language_configurations.len());
                     }
@@ -918,7 +946,7 @@ impl Loader {
                 file_types: Vec::new(),
                 scope: None,
                 content_regex: None,
-                _first_line_regex: None,
+                first_line_regex: None,
                 injection_regex: None,
                 injections_filenames: None,
                 locals_filenames: None,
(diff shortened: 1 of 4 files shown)


cli/loader/src/lib.rs
```diff
@@ -93,11 +93,26 @@ pub struct LanguageConfiguration<'a> {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_str() == other.0.as_str()
+    }
+}
+impl std::hash::Hash for Regex2 {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.as_str().hash(state)
+    }
+}
+
```
amaanq, line 109: It'd be better to do away with this and perhaps just use the string version of the regex for the hashmap entry, and then use Regex::from_str to convert back when using `find`

amaanq:
Also a test or two could be useful if you don't mind adding that

cli/loader/src/lib.rs
amaanq: Here's what I was thinking:

```patch
diff --git a/cli/loader/src/lib.rs b/cli/loader/src/lib.rs
index 0f9a33fe..88c47986 100644
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -8,6 +8,7 @@ use std::io::BufReader;
 use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::str::FromStr;
 use std::sync::Mutex;
 use std::time::SystemTime;
 use std::{env, fs, mem};
@@ -93,26 +94,12 @@ pub struct LanguageConfiguration<'a> {
     use_all_highlight_names: bool,
 }
 
-#[derive(Debug)]
-struct Regex2(Regex);
-impl Eq for Regex2 {}
-impl PartialEq for Regex2 {
-    fn eq(&self, other: &Self) -> bool {
-        self.0.as_str() == other.0.as_str()
-    }
-}
-impl std::hash::Hash for Regex2 {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.0.as_str().hash(state)
-    }
-}
-
 pub struct Loader {
     parser_lib_path: PathBuf,
     languages_by_id: Vec<(PathBuf, OnceCell<Language>)>,
     language_configurations: Vec<LanguageConfiguration<'static>>,
     language_configuration_ids_by_file_type: HashMap<String, Vec<usize>>,
-    language_configuration_ids_by_first_line_regex: HashMap<Regex2, Vec<usize>>,
+    language_configuration_ids_by_first_line_regex: HashMap<String, Vec<usize>>,
     highlight_names: Box<Mutex<Vec<String>>>,
     use_all_highlight_names: bool,
     debug_build: bool,
@@ -238,14 +225,22 @@ impl Loader {
                     })
             })
             .or_else(|| {
-                let Ok(file) = fs::File::open(path) else {return None};
+                let Ok(file) = fs::File::open(path) else {
+                    return None;
+                };
                 let reader = BufReader::new(file);
-                let Some(Ok(first_line)) = std::io::BufRead::lines(reader).next() else {return None};
+                let Some(Ok(first_line)) = std::io::BufRead::lines(reader).next() else {
+                    return None;
+                };
 
                 self.language_configuration_ids_by_first_line_regex
                     .iter()
-                    .find(|(regex,_)| dbg!(&regex.0).is_match(&first_line))
-                    .map(|(_,ids)| ids)
+                    .find(|(regex, _)| {
+                        Regex::from_str(regex)
+                            .and_then(|r| Ok(r.is_match(&first_line)))
+                            .unwrap_or(false)
+                    })
+                    .map(|(_, ids)| ids)
             });
 
         if let Some(configuration_ids) = configuration_ids {
@@ -616,7 +611,7 @@ impl Loader {
                     }
                     for first_line_regex in &configuration.first_line_regex {
                         self.language_configuration_ids_by_first_line_regex
-                            .entry(Regex2(first_line_regex.clone()))
+                            .entry(first_line_regex.to_string())
                             .or_insert(Vec::new())
                             .push(self.language_configurations.len());
                     }

```

sigmaSd: Should I just change https://github.com/tree-sitter/tree-sitter/blob/e854fc4731c20d56a6212569c36823a1aa84b166/cli/loader/src/lib.rs#L81 to String

Instead of the current Regex -> String -> Regex ?

sigmaSd: we loose the regex validation though

sigmaSd:
todo
- [ ] tests

cli/loader/src/lib.rs
amaanq: Up to you, but I don't really like the whole abstraction of a struct on top of regex..

sigmaSd:
I applied your suggestion and added a tes

cli/src/tests/helpers/dirs.rs
```diff
@@ -3,7 +3,7 @@ lazy_static! {
     static ref FIXTURES_DIR: PathBuf = ROOT_DIR.join("test").join("fixtures");
     static ref HEADER_DIR: PathBuf = ROOT_DIR.join("lib").join("include");
     static ref GRAMMARS_DIR: PathBuf = ROOT_DIR.join("test").join("fixtures").join("grammars");
-    static ref SCRATCH_DIR: PathBuf = {
+    pub static ref SCRATCH_DIR: PathBuf = {
```
sigmaSd, line 6: is this ok ?

amaanq: maybe `pub(crate)` to avoid accidental exports in the future?

amaanq: Instead - maybe we can use python and set the regex to check for `#!` and then python somewhere after.

amaanq: I've pushed a better change for this, if you're good w/ that do you mind squashing the commits down to 1? Or I can if you're wary about doing that.

sigmaSd: Yeah it's fine I leave that to you  I'm used to commiting random stuff and the maintainers squash and merge

cli/loader/src/lib.rs
```diff
@@ -78,7 +79,7 @@ const BUILD_TARGET: &'static str = env!("BUILD_TARGET");
 pub struct LanguageConfiguration<'a> {
     pub scope: Option<String>,
     pub content_regex: Option<Regex>,
-    pub _first_line_regex: Option<Regex>,
+    pub first_line_regex: Option<String>,
```
ahlinc, line 82: On this place the `Regex` type should be stored.

cli/loader/src/lib.rs
```diff
@@ -220,6 +223,24 @@ impl Loader {
...
+                };
+                let reader = BufReader::new(file);
+                let Some(Ok(first_line)) = std::io::BufRead::lines(reader).next() else {
+                    return None;
+                };
+
+                self.language_configuration_ids_by_first_line_regex
+                    .iter()
+                    .find(|(regex, _)| {
+                        Regex::from_str(regex)
```
ahlinc, line 239: See above comment, it's not a good idea to instantiate Regex instances here every time.

cli/loader/src/lib.rs
```diff
@@ -588,6 +609,12 @@ impl Loader {
                             .or_insert(Vec::new())
                             .push(self.language_configurations.len());
                     }
+                    for first_line_regex in &configuration.first_line_regex {
+                        self.language_configuration_ids_by_first_line_regex
+                            .entry(first_line_regex.clone())
```
ahlinc, line 614: Look on above loop and how a `Regex` converted to a `String` there.

cli/loader/src/lib.rs
```diff
@@ -220,6 +223,24 @@ impl Loader {
...
+                };
+
+                self.language_configuration_ids_by_first_line_regex
+                    .iter()
+                    .find(|(regex, _)| {
+                        Regex::from_str(regex)
+                            .and_then(|r| Ok(r.is_match(&first_line)))
+                            .unwrap_or(false)
+                    })
+                    .map(|(_, ids)| ids)
```
ahlinc, line 243: I think all this implementation should be improved by using a `RegexSet`: https://docs.rs/regex/1.9.3/regex/struct.RegexSet.html

ahlinc (changes requested):

cli/loader/src/lib.rs
amaanq: Does regexset implement Hash? That's the main issue

ahlinc: It's an additional note about implementation efficiency, it can be ignored cause requires more redesign, I can do it by myself later.

amaanq:
thanks, sorry for taking so long to get to this.

PR #2488, opened 2023-08-12
feat!: add an eof function

### LINKED ISSUE #160: Grammar construct to match EOF / end of file / document
In our project, we have a couple of grammar productions that basically say "do this until new-line or EOF". It's a mostly line-based grammar, where we don't like to error on the last line if it doesn't have a trailing newline.

Does that exist right now?

### LINKED ISSUE #1904: Identifying end-of-input
Searching the documentation for 'eof', 'eoi', 'eos' and 'end' finds no description of how a grammar can match the end-of-input.

For instance, a simple rule set:

```
const terminator = choice (";", /\r?\n/)
rules = {
  //...
  document: $ => repeat($.statement),
  statement: $ => seq("hello", terminator)
};
```

will fail to parse a document that does not have a newline on the last line.

I've looked at several other grammars and it seems they either just ignore this and produce end-user-confusing error messages or they resort to an external scanner, which seems a bit much for EOI.

amaanq:
Closes #160 

Instead of relying on the null char `'\0'` to match the end of a file (where that can also match any null char inside the file as well), one can now do this:

```js
module.exports = grammar({
    name: 'test',

    rules: {
        source_file: $ => seq('hello', eof()),
    }
});
```


lib/src/parser.c
```diff
@@ -482,7 +482,7 @@ static Subtree ts_parser__lex(
       current_position.extent.column
     );
     ts_lexer_start(&self->lexer);
-    bool found_token = self->language->lex_fn(&self->lexer.data, lex_mode.lex_state);
+    bool found_token = self->language->lex_fn(&self->lexer.data, lex_mode.lex_state, &self->lexer.accepted_eof);
```
ahlinc, line 485: I'm not sure that such approach for the implementation is correct, it looks like you created a parallel set of eof states to lex_states and their intersection in a generated parser.c. I thought implementation of the eof() would just generate additional lex states by using existing mechanism, so for some rules the would be generated aditional eof checks and on possitive values transitions to an  aditional state with the aux_sym_eof acceptance.

amaanq: Not exactly, another solution can be to use the field in the TSLexer instead of Lexer, but since that's exposed in the scanner it's maybe better not to.

The issue is we need to know if `aux_sym_eof` was already accepted once before (without advancing), and if so, then the builtin eof sym can be accepted

cli/src/generate/nfa.rs
```diff
@@ -445,27 +448,28 @@ impl<'a> NfaCursor<'a> {
         Self::group_transitions(self.raw_transitions())
     }
 
-    fn raw_transitions(&self) -> impl Iterator<Item = (&CharacterSet, bool, i32, u32)> {
+    fn raw_transitions(&self) -> impl Iterator<Item = (&CharacterSet, bool, i32, u32, bool)> {
```
maxbrunsfeld, line 451: I don't think `eof` should be modeled as part of an `Advance` action. I would think that it would just be a property of an `AcceptToken` action: "this token can only be accepted in this state if we are at EOF".

EOF seems different than a lookahead character set in that it's not something in the input stream that we *consume*; it's just a boolean property (are we or are we not at EOF) of a point in the input stream.

cli/npm/dsl.d.ts
```diff
@@ -344,6 +345,20 @@ declare const token: {
   immediate(rule: RuleOrLiteral): ImmediateTokenRule;
 };
 
+/**
+ * Creates an eof rule, matching the end of the input.
+ * This is useful in languages where the top level grammar is a repeated
+ * sequence of statements followed by a terminator. One might use newlines
+ * for a terminator, but the end of the file also should match. Previously,
+ * '\0' was used to match eof, but this can match any NUL byte in the input,
```
maxbrunsfeld, line 353: I'd remove the `Previously, '\0' was used` sentence from the docs. I'm assuming you mean some specific grammar used `\0` to detect EOF? Either way, I don't think the docs need to talk about some thing that previously existed.

lib/src/parser.c
maxbrunsfeld: I don't think it should matter whether or not the EOF token was accepted already. Conceptually, it seems like multiple layers of rules should be able to match on `EOF`. Is this intended for preventing infinite loops?

cli/src/generate/nfa.rs
amaanq: I was a bit unsure about how to model the eof action, yeah, there's no actual consumption of a character but we would still want to "advance" to the state when an eof token was used I'd think. Or, we can just accept in place in the zeroth state, is that what you were thinking? Here's my interpretation of it just being an `AcceptToken` action

```c
if (!*accepted_eof && eof) {
        *accepted_eof = true;
        ACCEPT_TOKEN(aux_sym_eof);
		END_STATE();
      }
```

So, instead of advancing to say state 4 and accepting `aux_sym_eof` there, just accept in place in state 0 instead

lib/src/parser.c
amaanq: Yeah it would endlessly accept `aux_sym_eof` without this check and the parsing would never end.

ItzCeeleeAdam:
i was completed

clason:
@amaanq is this still on the docket?

foxyseta:
> @amaanq is this still on the docket?

I was curious, too.

LostTime76:
Bump. Update on this?

I have tokens that have potential to be terminated by a line terminator or the end of the file. Being able to use eof() within the grammar.js would be perfect. Otherwise, we have sort of a hack that every text needs to have a line terminator at the end. That especially gets really annoying for single line comments that should be allowed to terminate using a line terminator or EOF.

amaanq:
I've reworked this to instead reuse the builtin end symbol, but I'm still not super confident on the implementation. It does seem to work okay, but it seems like it can increase state count a decent amount.

maxbrunsfeld:
I don’t think we should land this if it increases the state count much.

I think that we should never *shift* the EOF token onto the parse stack. EOF should only be able to be used at the end of a sequence, and we should handle it specially on the flattened rule, so that it if a rule in the grammar ends with EOF, then rather than becoming a part of the rule’s actual children, the rule must have EOF as a lookahead token in order to reduce.

WillLillis:
I rebased this and added a fix based on Max's feedback as a separate commit (to be squashed in if this is the right approach). 

- Productions ending in `eof()` are flattened with the end step stripped and a new `requires_eof_lookahead` flag set on the `Production`. The `Reduce` action is then emitted only under the EOF lookahead column instead of as a shift.

- `eof()` placed mid-sequence (i.e. inside an inlined `terminator = choice(..., eof())` used in a position where something must follow) is silently dropped at flatten time, since no input can follow EOF. If every alternative of a rule drops this way, generation fails with a clear `NoReachableProductions` error.
  - Two LR-table correctness fixes that turned out to be required:
    - The closure was propagating the parent's full FOLLOW set through requires_eof_lookahead productions, polluting inner reduce lookaheads with unrelated tokens. Restricting to {EOF} in the propagation eliminates a class of spurious conflicts.
    - ParseItem's Eq/Hash/Ord didn't include requires_eof_lookahead, so two productions differing only in that flag merged into one item and clobbered each other's production_id. Including it fixes correctness of reduce dispatch.


I tested with this change to tree-sitter-go:

```diff
diff --git a/grammar.js b/grammar.js
index 3648f02..b0315ea 100644
--- a/grammar.js
+++ b/grammar.js
@@ -26,7 +26,7 @@ const assignmentOperators = multiplicativeOperators.concat(additiveOperators).ma
 
 
 const newline = /\n/;
-const terminator = choice(newline, ';', '\0');
+const terminator = choice(newline, ';', eof());
 
 const hexDigit = /[0-9a-fA-F]/;
 const octalDigit = /[0-7]/;
@@ -115,15 +115,12 @@ module.exports = grammar({
   ],
 
   rules: {
-    source_file: $ => seq(
-      repeat(choice(
-        // Unlike a Go compiler, we accept statements at top-level to enable
-        // parsing of partial code snippets in documentation (see #63).
-        seq($._statement, terminator),
-        seq($._top_level_declaration, terminator),
-      )),
-      optional($._top_level_declaration),
-    ),
+    source_file: $ => repeat(seq(
+      // Unlike a Go compiler, we accept statements at top-level to enable
+      // parsing of partial code snippets in documentation (see #63).
+      choice($._statement, $._top_level_declaration),
+      terminator,
+    )),
 
     _top_level_declaration: $ => choice(
       $.package_clause,
```

All corpus tests still pass, and the state counts actually reduced a bit:
	- state count: 1442->1415
	- large state count: 29->28
	- symbol count: 214->213

WillLillis:
Slight wrinkle in the implementation, but shouldn't be blocking. I added an error for when a rule doesn't have any productions, as this can lead to silently broken parsers. The original intent behind this was for rules that dropped all of their eof productions and ended up empty. Unfortunately, tree-sitter-markdown actually uses an empty `choice` in its grammar today, which triggers the new error:

https://github.com/tree-sitter-grammars/tree-sitter-markdown/blob/c3570720f7f7bbad22fe96603f106276618e0cf5/tree-sitter-markdown/grammar.js#L27

Here's a minimal grammar demonstrating the problem:

```js
module.exports = grammar({
      name: 'empty_choice_demo',
      rules: {
          source_file: $ => seq('hello', $.dead, 'world'),
          dead: $ => choice(),
      }
  });
```

`tree-sitter generate` runs without any issues. However, if you try running `tree-sitter parse` on `hello world` you get a broken tree: `(ERROR [0, 0] - [0, 11])`.

I think this is worth fixing once I have a chance to understand tree-sitter-markdown's use case more, but shouldn't block this change. Instead (for now), we'll only emit an error if the rule _started_ with some choices and then flattening eliminated all of them.

WillLillis:
Rebased on top of latest master. Semantics should still be identical. Perf is neutral on walltime, at worst a 1% regression in peak rss for some grammars (e.g. razor).

WillLillis (approved):

clason:
This is one case where an error is actually appropriate ;) Emitting a broken parser would be the worse option.

Could you open an issue at tree-sitter-markdown about this for visibility? (Ideally, that kludge should be removed anyway.)

(Rebased the commit for good measure.)

clason (approved):

WillLillis:
Amended the commit and noting here: The commit is marked as breaking because it changes some public types for `tree-sitter-generate`. The original implementation on this branch was also breaking for the C lib, but this is no longer the case.

PR #2491, opened 2023-08-13
fix: replace & sanitize more characters

### LINKED ISSUE #2490: Escape null bytes in generated C code
https://github.com/tree-sitter/tree-sitter-go/commit/0308e24c57a379ea7e80da792882c8eb6b3e6fd5 allows `'\0'` in the grammar:

```
terminator = choice(newline, ';', '\0'),
```

However, the generated C code embeds the null character directly:

https://github.com/tree-sitter/tree-sitter-go/blob/bbaa67a180cfe0c943e50c55130918be8efb20bd/src/parser.c#L242

This generates a warning when compiling:

```
warning: vendored_parsers/tree-sitter-go-src/parser.c:242:17: warning: null character(s) preserved in literal
warning:   242 |   [anon_sym_] = " ",
warning:       |                 ^
```

I think null bytes should be escaped in the string literals generated for C code, so the C looks like this instead:

```
[anon_sym_] = "\0",

### PR DESCRIPTION
**Pull request by amaanq**
Closes #2490

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -1533,54 +1533,92 @@ impl Generator {
     fn sanitize_identifier(&self, name: &str) -> String {
         let mut result = String::with_capacity(name.len());
         for c in name.chars() {
-            if ('a' <= c && c <= 'z')
-                || ('A' <= c && c <= 'Z')
-                || ('0' <= c && c <= '9')
-                || c == '_'
-            {
+            if c.is_ascii_alphanumeric() || c == '_' {
                 result.push(c);
             } else {
-                let replacement = match c {
-                    '~' => "TILDE",
-                    '`' => "BQUOTE",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cli/src/generate/render.rs [hunk 1]
```diff
@@ -1593,10 +1625,18 @@ impl Generator {
...
+                '\u{0007}' => result += "\\a",
+                '\u{0008}' => result += "\\b",
+                '\u{000b}' => result += "\\v",
                 '\u{000c}' => result += "\\f",
                 '\n' => result += "\\n",
                 '\r' => result += "\\r",
                 '\t' => result += "\\t",
+                '\0' => result += "\\0",
+                '\u{0001}'..='\u{001f}' => {
+                    result += "\\x";
```
ahlinc, line 1637: Move this to the format and it's ok to merge.

ahlinc (approved):

PR #2494, opened 2023-08-13
chore(build-wasm): use spawn instead of output to give feedback

sigmaSd:
Currently when the user uses `build-wasm` he have no feedback on whats happening, and this command can take a while mainly the first time since its pulls the docker image. 

This PR spawn the command instead so the user can see whats happening live.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -677,12 +677,12 @@ impl Loader {
         }
 
         command.arg("parser.c");
-        let output = command.output().context("Failed to run emcc command")?;
-        if !output.status.success() {
-            return Err(anyhow!(
-                "emcc command failed - {}",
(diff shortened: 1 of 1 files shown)


Sonyashines (approved):

sigmaSd:
emcc shows a warning 
```
emcc: warning: EXPORTED_FUNCTIONS is not valid with LINKABLE set (normally due to SIDE_MODULE=1/MAIN_MODULE=1) since all functions are exported this mode.  To export only a subset use SIDE_MODULE=2/MAIN_MODULE=2 [-Wunused-command-line-argument]
```
Any idea how to disable/fix it ?

sigmaSd:
the ci seems to have no error ?

ahlinc:
I'll add more functionality to this PR, that's why I assigned it myself.

PR #2496, opened 2023-08-13
Revert "fix: musl compilation"

ahlinc:
Reverts tree-sitter/tree-sitter#2492

Atomic relaxed level doesn't provide any synchronization guaranties between multiple threads and I'm not sure that it's correct to change `SEQ_CST` used by `__sync_add_and_fetch` and `__sync_sub_and_fetch` to the `RELAXED` level.


amaanq:
will update `__atomic_*` to use SEQ_CST

PR #2500, opened 2023-08-14
feat: improve dsl definitions

amaanq:


cli/npm/dsl.d.ts
```diff
@@ -72,7 +73,10 @@ interface Grammar<
    * a *descending* ordering. Names listed earlier in one of these arrays
    * have higher precedence than any names listed later in the same array.
    */
-  precedences?: () => String[][],
+  precedences?: (
+    $: GrammarSymbols<RuleName | BaseGrammarRuleName>,
+    previous: RuleOrLiteral[][],
+  ) => RuleOrLiteral[][],
```
ahlinc, line 79: Why you decided that there possible to use references? This is not stated in the description too.

cli/npm/dsl.d.ts
```diff
@@ -42,6 +42,7 @@ type GrammarSymbols<RuleName extends string> = {
 
 type RuleBuilder<RuleName extends string> = (
   $: GrammarSymbols<RuleName>,
+  previous: RuleOrLiteral,
```
ahlinc, line 45: * Please remove this in favor of #2504

ahlinc (changes requested):

ahlinc (changes requested):
After an investigation that I made in #2504, I'm not sure that it's correct to use the `previous ` in all such arrays like `inline`, `supertypes` and so on, because the first `$` dollar parameter can be used to refer both rules from the previous and current grammars.

cli/npm/dsl.d.ts
```diff
@@ -101,8 +109,8 @@ interface Grammar<
    */
   externals?: (
     $: Record<string, SymbolRule<string>>,
-    previous: Rule[],
-  ) => (SymbolRule<string> | RegExp | string)[];
+    previous: RuleOrLiteral[],
+  ) => RuleOrLiteral[];
```
ahlinc, line 113: I think this return type `=> RuleOrLiteral[]` also should be just `=> Rule[]`.

ahlinc: And in other places too.

ahlinc (changes requested):
Please look on my comment about return types.

cli/npm/dsl.d.ts
ahlinc: It seems conversion to the `Rule` happens later.

ahlinc (approved):

PR #2503, opened 2023-08-14
feat: implement a cache for `get_column`

### LINKED ISSUE #1626: Improve performance of `get_column`

### LINKED ISSUE #3433: [perf] Bad performance when call get_column frequently
### Problem

When the get_column function is called frequently, the performance of get_column will be poor. Should we improve the performance of get_column?

example:

at first:

```
$ time pn tree-sitter parse ../1mb.data
pnpm tree-sitter parse ../1mb.data  4.18s user 0.13s system 100% cpu 4.285 total
```

change the code :
```c
usize start_pos = lexer->get_column(lexer);
if(lexer->get_column(lexer) == 0) {
}
```
to
```
usize start_pos = lexer->get_column(lexer);
if(start_pos == 0) {
}
```

```console

$ time pn tree-sitter parse ../1mb.data
pnpm tree-sitter parse ../1mb.data  3.01s user 0.15s system 101% cpu 3.122 total
```

> [!NOTE]
> the data should be 1MB but not 1mb.

### Expected behavior

_No response_

amaanq:
- _Supersedes_ #1637 
- Closes #3433
- Closes #1626

### DIFF
--- a/lib/src/lexer.c
+++ b/lib/src/lexer.c
@@ -37,6 +37,35 @@ static const TSRange DEFAULT_RANGE = {
   .end_byte = UINT32_MAX
 };
 
+/**
+ * Sets the column data to the given value and marks it valid.
+ * @param self The lexer state.
+ * @param val The new value of the column data.
+ */
+static void ts_lexer__set_column_data(Lexer *self, uint32_t val) {
+  self->column_data.valid = true;
+  self->column_data.value = val;
+}
+
+/**
+ * Increments the value of the column data; no-op if invalid.
+ * @param self The lexer state.
+ */
+static void ts_lexer__increment_column_data(Lexer *self) {
+  if (self->column_data.valid) {
+    self->column_data.value++;
+  }
+}
+
+/**
+ * Marks the column data as invalid.
+ * @param self The lexer state.
+ */
+static void ts_lexer__invalidate_column_data(Lexer *self) {
+  self->column_data.valid = false;
+  self->column_data.value = 0;
+}
+
 // Check if the lexer has reached EOF. This state is stored
 // by setting the lexer's `current_included_range_index` such that
 // it has consumed all of its available ranges.
@@ -104,6 +133,10 @@ static void ts_lexer__get_lookahead(Lexer *self) {
 }
 
 static void ts_lexer_goto(Lexer *self, Length position) {
+  if (position.bytes != self->current_position.bytes) {
+    ts_lexer__invalidate_column_data(self);
+  }
+
   self->current_position = position;
 
   // Move to the first valid position at or after the given position.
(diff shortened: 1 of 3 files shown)


amaanq:
@ahelwer I'd appreciate if you can try this fork out and still get the perf benefits like in your PR

ahlinc:
I've added the `postponed` tag to the original PR and asked to don't continue to push such PRs to master.

amaanq:
I think fuzzing scala + seeing the perf benefits as before would mean it's ok :slightly_smiling_face: but yeah get_column is a bit intricate so best to be careful

ahlinc:
> I think fuzzing scala + seeing the perf benefits as before would mean it's ok

That's not the reason, I asked to postpone it because I want to check anther solution and I don't want to make a battle of features and merge the feature that may disappear very fast if another solution would work better.

ahelwer:
There certainly are additional places that get_column could be cached, since the parser likes to reset its position (to well-known places) quite a bit.

dundargoc:
Does this close https://github.com/tree-sitter/tree-sitter/issues/1626?

ObserverOfTime:
This change is binary incompatible according to `abi-compliance-checker`.

<table class="ptable"><tbody><tr>
<th class="pn"></th>
<th class="chg">Change</th>
<th>Effect</th></tr><tr>
<th>1</th>
<td>Field <b>column_cache</b> has been added at the middle position of this structural type.</td>
<td>1) Size of the inclusive type has been changed.<br>2) Layout of structure fields has been changed and therefore fields at higher positions of the structure definition may be incorrectly accessed by applications.</td>
</tr>
</tbody></table>

maxbrunsfeld:
> This change is binary incompatible according to abi-compliance-checker.

Only changes to `api.h` affect binary compatibility. The rest of the headers are private.

amaanq:
I've leveraged the caching in one hotspot in the parser's main lex loop when it resets to the position before the external scanner was called and noticed decent performance improvements (~7-10% in my limited testing), so I think we can land this.

Issue #2505, opened 2023-08-15
clarify how to use dsl.d.ts


sigmaSd:
Following the tree sitter guide, I used `npm install --save-dev tree-sitter-cli` but the problem is the tsserver lsp don't know the types, I even created my own grammar.d.ts to fix this

But I now noticed that there is actually a dsl.d.ts, I' m adding this to grammar.js `/// <reference path="./node_modules/tree-sitter-cli/dsl.d.ts" />`

Is this how its intended to be used ?
Can we add this info to the tree siter web site in that case ?

sigmaSd:
Just adding this triple slash reference to the grammar.js example https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup here would be enough to have a better experience for newcomers

saud990:
> Following the tree sitter guide, I used `npm install --save-dev tree-sitter-cli` but the problem is the tsserver lsp don't know the types, I even created my own grammar.d.ts to fix this
> 
> But I now noticed that there is actually a dsl.d.ts, I' m adding this to grammar.js `/// <reference path="./node_modules/tree-sitter-cli/dsl.d.ts" />`
> 
> Is this how its intended to be used ? Can we add this info to the tree siter web site in that case ?

mmu094:
[Skip to content](https://github.com/tree-sitter/tree-sitter/issues/2505#start-of-content)
https://github.com/tree-sitter
/
[tree-sitter](https://github.com/tree-sitter/tree-sitter)
Public
[Code](https://github.com/tree-sitter/tree-sitter)
Issues
350
Pull requests
88
[Discussions](https://github.com/tree-sitter/tree-sitter/discussions)
[Actions](https://github.com/tree-sitter/tree-sitter/actions)
[Wiki](https://github.com/tree-sitter/tree-sitter/wiki)
[Security](https://github.com/tree-sitter/tree-sitter/security)
[Insights](https://github.com/tree-sitter/tree-sitter/pulse)
[Jump to bottom](https://github.com/tree-sitter/tree-sitter/issues/2505#issue-comment-box)
clarify how to use dsl.d.ts #2505
Open
[sigmaSd](https://github.com/sigmaSd) opened this issue yesterday · 2 comments
Labels
[documentation](https://github.com/tree-sitter/tree-sitter/labels/documentation)
Comments
[...]

PR #2523, opened 2023-08-17
feat(docs): mention how to use dsl.ts + node types

sigmaSd:
fix https://github.com/tree-sitter/tree-sitter/issues/2505

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -42,8 +42,8 @@ npm init
 # This installs a small module that lets your parser be used from Node
 npm install --save nan
 
-# This installs the Tree-sitter CLI itself
-npm install --save-dev tree-sitter-cli
+# This installs the Tree-sitter CLI itself, also installs node types
+npm install --save-dev tree-sitter-cli @types/node
 ```
 
 The last command will install the CLI into the `node_modules` folder in your working directory. An executable program called `tree-sitter` will be created inside of `node_modules/.bin/`. You may want to follow the Node.js convention of adding that folder to your `PATH` so that you can easily run this program when working in this directory.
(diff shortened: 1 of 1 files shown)


ahlinc:
As I said in  https://github.com/tree-sitter/tree-sitter/issues/2505#issuecomment-1679823244 this would be managed automatically for all new grammars.

sigmaSd:
Ah I see what you mean now

what about adding this part `npm install --save-dev tree-sitter-cli @types/node`  ?

sigmaSd:
Also that pr seems to be taking time, its not a bad idea to add improve the docs until it lands

ahlinc:
Could you tell for what purpose you want to install `@types/node` node types? We're moving to make Tree-sitter and its grammars having minimum dependencies.

amaanq:
Yeah you don't need that, just referencing the dsl works like so
```js
```

sigmaSd:
here is how it looks without @types/node
image

amaanq:
Weird, is that from tsserver? I forgot to mention adding ts-check below the reference gets you what you need

```js
// @ts-check
```

sigmaSd:
yes that's from tsserver, and adding ts-check doesn't change the error

amaanq:
Weird, I don't get the message you have, would be interesting if you have more context/info about your setup

sigmaSd:
I' m using helix with tsserver lsp 

I tried vscode, if I only mention module I get the same error 
image


But if I assign module to something the error disappear

If I'm going to start guessing ,maybe vscode is doing extra heuristics that the tsserver lsp can't do, I'm assuming you're using vscode ?

image

amaanq:
I'm using neovim w/ tsserver

sigmaSd:
I get the same behavior as vscode with neovim

So I guess I'll debug this on helix side, you can ignore this problem for now

sigmaSd:
in other words @types/node should be not needed indeed

sigmaSd:
helix seems to typecheck by default and in strict mode

neovim and vscode just ignore the error, and there is no typings for module, the only way to correctly fix this is to install node types

Either way its no big deal, and if something gets bothered I'm guess they will find this issue

PR #2526, opened 2023-08-18
Fix error recovery regression due to stack node_count change

maxbrunsfeld:
Fixes https://github.com/tree-sitter/tree-sitter/issues/2443

### DIFF
--- a/lib/src/stack.c
+++ b/lib/src/stack.c
@@ -120,6 +120,20 @@ recur:
   }
 }
 
+/// Get the number of nodes in the subtree, for the purpose of measuring
+/// how much progress has been made by a given version of the stack.
+static uint32_t stack__subtree_node_count(Subtree subtree) {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I wish this regression were caught by tests. I guess the behavior mattered more for some languages than others. And error recovery is somewhat difficult to test.

I'm gonna merge this for now. In the future, I'd like to better job testing in this area.

amaanq:
We could add the cases caught by scala or scheme, but that'd require adding those grammars to fixtures

maxbrunsfeld:
Yeah, I guess scala or PHP would be good options, since they are under the Tree-sitter org.

llllvvuu:
Noting that there's a performance regression in Neovim after this commit:

https://github.com/neovim/neovim/issues/25252

(I don't suspect this commit yet, but it is behaving differently somehow)

PR #2529, opened 2023-08-18
feat: add `injection.self` to inject a node with itself

amaanq:
# Motivation

In certain cases, a language can extend upon another, and in many cases, the language is backwards compatible (C/C++ *practically* are, bar a few features like `auto`, `register` and different number suffixes)

It makes sense for downstream to treat C queries as compatible with C++, and to only add C++ features to its queries (this is what Neovim does w/ its inherit directive at the start of a query file)

An issue arises though, when injections inject itself. Currently, preproc_arg in C is injected with C, but if that's inherited in C++, C parsing is applied instead of C++, thus removing the capability to inherit

# Solution

An `injection.self` property in the set directive of injections fixes this, where inheriting always ensures the injected language is that of the language itself.

This could be moved to just be a capture `@injection.self` similar to how `@injection.language` is used for heredoc delimiters

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -86,6 +86,7 @@ pub struct LanguageConfiguration<'a> {
     pub injections_filenames: Option<Vec<String>>,
     pub locals_filenames: Option<Vec<String>>,
     pub tags_filenames: Option<Vec<String>>,
+    pub language_name: String,
     language_id: usize,
     highlight_config: OnceCell<Option<HighlightConfiguration>>,
     tags_config: OnceCell<Option<TagsConfiguration>>,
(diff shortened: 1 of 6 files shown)


cli/loader/src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub struct LanguageConfiguration<'a> {
     pub injections_filenames: Option<Vec<String>>,
     pub locals_filenames: Option<Vec<String>>,
     pub tags_filenames: Option<Vec<String>>,
+    pub language_name: String,
```
ahlinc, line 89: That's nice.

ahlinc (approved):

PR #2530, opened 2023-08-18
Remove -Werror from default CFLAGS

### LINKED ISSUE #2531: Build issues on NixOS
After probably https://github.com/tree-sitter/tree-sitter/commit/f9e8802234cbd692b1979cff3bea90f5103d4187 I run into

```
warning: In file included from /nix/store/3w4yrspgcx1z9wi18rv0hbap4jjlhx4c-glibc-2.37-8-dev/include/bits/libc-header-start.h:33,
warning:                  from /nix/store/3w4yrspgcx1z9wi18rv0hbap4jjlhx4c-glibc-2.37-8-dev/include/stdlib.h:26,
warning:                  from include/tree_sitter/api.h:10,
warning:                  from src/./alloc.h:4,
warning:                  from src/./alloc.c:1,
warning:                  from src/lib.c:8:
warning: /nix/store/3w4yrspgcx1z9wi18rv0hbap4jjlhx4c-glibc-2.37-8-dev/include/features.h:413:4: error: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Werror=cpp]
warning:   413 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
warning:       |    ^~~~~~~
warning: cc1: all warnings being treated as errors
```

This is because NixOS enables it by default (`-O2 -D_FORTIFY_SOURCE=2`) and the build seems to override the level to 0: 

```
running: "gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "src" "-I" "include" "-Wall" "-Wextra" "-std=c99" "-fvisibility=hidden" "-Wshadow" "-Werror" "-o" "src/helix/target/re
lease/build/tree-sitter-0070f96918a4c153/out/src/lib.o" "-c" "src/lib.c"
```

This was hidden away before by `Werror` being disabled.

\cc @amaanq

jamessan:
Production builds shouldn't include -Werror by default since that could cause spurious build failures when there are toolchain updates.

CI uses -Werror to prevent warnings, so that should be sufficient.

Closes #2531

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -177,6 +177,7 @@ jobs:
         TARGET: ${{ matrix.target }}
         CC: ${{ matrix.cc }}
         AR: ${{ matrix.ar }}
+        IS_WINDOWS: ${{ contains(matrix.os, 'windows') }}
       run: |
         PATH="$PWD/.github/scripts:$PATH"
         echo "$PWD/.github/scripts" >> $GITHUB_PATH
@@ -191,6 +192,8 @@ jobs:
         [ -n "$CC" ] && echo "CC=$CC" >> $GITHUB_ENV
         [ -n "$AR" ] && echo "AR=$AR" >> $GITHUB_ENV
 
+        [ "$IS_WINDOWS" = "false" ] && echo "CFLAGS=-Werror" >> $GITHUB_ENV
+
         if [ "$USE_CROSS" == "true" ]; then
           echo "BUILD_CMD=cross" >> $GITHUB_ENV
           runner=$(BUILD_CMD=cross cross.sh bash -c "env | sed -nr '/^CARGO_TARGET_.*_RUNNER=/s///p'")
@@ -199,7 +202,7 @@ jobs:
 
     - name: Build C library
       if: ${{ !contains(matrix.os, 'windows') }} # Requires an additional adapted Makefile for `cl.exe` compiler
-      run: make.sh CFLAGS="-Werror" -j
+      run: make.sh -j
 
     - name: Build wasm library
       if: ${{ !matrix.cli-only && !matrix.use-cross }} # No sense to build on the same Github runner hosts many times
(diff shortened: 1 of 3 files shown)


amaanq:
can you elaborate on "spurious build failures when there are toolchain updates"? I don't see how that could happen w/o some complex build system at play

jamessan:
If a new gcc version is released that detects a new warning in tree-sitter, then existing tree-sitter releases will now fail to build because `-Werror` is present.  That's great in CI, but not so much when you're building elsewhere (e.g., rebuilding the package in a Linux distro).

amaanq:
That makes sense, thanks!

theHamsta:
`-Werror` makes sense for CI. If you apply `-Werror` for other people you're just penalizing everyone with a compiler which you didn't test.

amaanq (approved):

ahlinc:
@jamessan @theHamsta Thank you for describing clear reasons, we didn't have it until recently and I didn't support to add it but didn't have strong reasons proved by practice like in distros maintaining.

.github/workflows/build.yml
```diff
@@ -199,7 +202,7 @@ jobs:
 
     - name: Build C library
       if: ${{ !contains(matrix.os, 'windows') }} # Requires an additional adapted Makefile for `cl.exe` compiler
-      run: make.sh CFLAGS="-Werror" -j
```
amaanq, line 202: Just one note, can you bring this back here and remove "IS_WINDOWS"? currently make.sh only runs on Linux anyways

theHamsta:
> @jamessan @theHamsta Thank you for describing clear reasons, we didn't have it until recently and I didn't support to add it but didn't have strong reasons proved by practice like in distros maintaining.

Tree-sitter was never annoying to me in that regard as no warnings were triggered for me except one exception long time ago. No warning even with untested compilers are a good thing :slightly_smiling_face:

ahlinc (approved):

.github/workflows/build.yml
ahlinc: It's ok, in this way it works for build.rs too

PR #2532, opened 2023-08-18
feat!: properly handle predicates used on quantified captures

### LINKED ISSUE #2514: Implementation of predicates acting on quantified captures
Hi, while working on https://github.com/neovim/neovim/pull/24738, we came to think about the implementation of the predicates when applied upon a quantified match.

Take for example the following C code:
```c
// foobar
// baz
```

And the following query:
```
(
  (comment)+ @c
  (#eq? @c "// foobar")
)
```

In the current implementation of both the CLI and the web bindings, the predicate `#eq?` would match in the case because the __first__ captured node matches the predicate. We don't know if the current behavior was intended, and would like to make it clearer for consistency across the downstream implementors.

There are other possible versions of this:
- Match all node: obvious
- Match any node: similar as current implementation, but more flexible

Finally, I thought that giving proper names to predicates (just like `#not-eq?` for example) and prefixing them by e.g. `any` or `all` could allow to make the behavior explicit (having no prefix could default to e.g. the current implementation), then the predicates would look like the following examples: `any-eq?`, `all-match?`, ...

Pinging some people that took part in the discussion: @lewis6991 @clason @amaanq

amaanq:
Closes #2514


lib/binding_rust/lib.rs
```diff
@@ -1729,7 +1734,7 @@ impl Query {
                 // Build a predicate for each of the known predicate function names.
                 let operator_name = &string_values[p[0].value_id as usize];
                 match operator_name.as_str() {
-                    "eq?" | "not-eq?" => {
+                    "eq?" | "not-eq?" | "any-eq?" | "not-any-eq?" | "all-eq?" | "all-not-eq?" => {
```
ahlinc, line 1737: Could we use `not-all-eq?` or `any-not-eq?` for consistency?

ahlinc: I'm +1 for the second option

lib/binding_rust/lib.rs
```diff
@@ -1746,23 +1751,34 @@ impl Query {
...
                                 p[1].value_id,
                                 string_values[p[2].value_id as usize].clone(),
                                 is_positive,
+                                is_quantified,
                             )
                         });
                     }
 
-                    "match?" | "not-match?" => {
+                    "match?" | "not-match?" | "any-match?" | "not-any-match?" | "all-match?"
```
ahlinc, line 1780: The same here `any-not-match?` and where is `all-not-match?` ?

clason:
@vigoux

amaanq:
Added some docs/examples, any cool/useful predicates for cases I haven't covered are welcome to be added :)

docs/section-2-using-parsers.md
```diff
@@ -651,7 +666,79 @@ And this pattern would match key-value pairs where the `value` is an identifier
...
+```scheme
+((identifier) @constant
+  (#match? @constant "^[A-Z][A-Z_]+"))
+```
+
+Again, the prefix "any-" and "all-" are meant for use with quantified captures.
+
+Here's an example finding potential documentation comments in C
+
+```scm
```
ahlinc, line 700: No editors know the `scm` tag, use the `scheme` tag instead everywhere like you did above.

clason: Well, GitHub should create it -- after all, there's a tree-sitter parser for it ;)

cli/src/tests/query_test.rs
```diff
@@ -2843,7 +2843,7 @@ fn test_query_captures_with_text_conditions() {
              (#eq? @function.builtin "require"))
 
             ((identifier) @variable.builtin
-              (#any-of? @variable.builtin
+              (#eq? @variable.builtin
                         "arguments"
                         "module"
                         "console"
```
ahlinc, line 2849: This doesn't look clear, when I look on `(#eq? @node "foo" "bar" "baz")` it isn't clear what to expect, would be match over any of values, or it expects three nodes much to all values.

clason: Well, the latter doesn't really make sense, so the former is the only meaningful (and hopefully documented) Interpretation...

ahlinc: > the latter doesn't really make sense

why not? If it provides ordering and I for ex. wants to find all funcs with such parameter names in a row.

In any way I think it's not a good idea to add an additional sense on simple and clear things, the `#eq?` in it current form doesn't have sense when a pattern match captures more then a one capture by the same `@name` and such predicate should just filter out the whole pattern, to match in a more fine grained way there are other predicates.

I think predicate names should not be overloaded with many senses, it's better to have more precise predicates.

docs/section-2-using-parsers.md
ahlinc: We using the `scheme` tag not for Github, we use it for a renderer used in docs.

docs/section-2-using-parsers.md
```diff
@@ -629,18 +629,36 @@ The restrictions placed on a pattern by an anchor operator ignore anonymous node
...
+start with a _predicate name_ beginning with a `#` character. After that, they can
+contain an arbitrary number of `@`-prefixed capture names or strings.
 
-For example, this pattern would match identifier whose names is written in `SCREAMING_SNAKE_CASE`:
+Tree-Sitter's CLI supports the following predicates by default:
+
+##### eq?, not-eq?, any-eq?, any-not-eq?
+
+This family of predicates allows you to match against a single capture or string
+value.
```
ahlinc, line 642: It says about single capture or string, but in a bellow example there are two strings.

docs/section-2-using-parsers.md
```diff
@@ -629,18 +629,36 @@ The restrictions placed on a pattern by an anchor operator ignore anonymous node
...
-(
-  (identifier) @constant
-  (#match? @constant "^[A-Z][A-Z_]+")
-)
+((identifier) @variable.builtin
+  (#eq? @variable.builtin "self" "this"))
 ```
 
-And this pattern would match key-value pairs where the `value` is an identifier with the same name as the key:
+This pattern would match any identifier that is `self` or `this`.
```
clason, line 658: Wait, I'm confused. Didn't @ahlinc say that `#eq?` should _not_ behave like `#any-of?`?

ahlinc: I think it's time to draw a table somewhere with all predicates and their characteristics.

docs/section-2-using-parsers.md
amaanq: Fixed

docs/section-2-using-parsers.md
amaanq: I wrote a footer at the bottom that is basically a  recap

docs/section-2-using-parsers.md
```diff
@@ -629,18 +629,36 @@ The restrictions placed on a pattern by an anchor operator ignore anonymous node
...
-(
-  (identifier) @constant
-  (#match? @constant "^[A-Z][A-Z_]+")
-)
+((identifier) @variable.builtin
+  (#eq? @variable.builtin "self"))
 ```
 
-And this pattern would match key-value pairs where the `value` is an identifier with the same name as the key:
+This pattern would match any identifier that is `self` or `this`.
```
clason, line 658: Stray "or `this`".

docs/section-2-using-parsers.md
```diff
@@ -651,7 +669,87 @@ And this pattern would match key-value pairs where the `value` is an identifier
...
+an example finding a segment of empty comments
+
+```scheme
+((comment)+ @comment.empty
+  (#any-eq? @comment.empty "//"))
+```
+
+Note that "#any-eq?" will match a quantified capture if
+_any_ of the nodes match the predicate, while by default a quantified capture
+will only match if _all_ the nodes match the predicate.
```
clason, line 682: While/if you fix the above: Would be useful to also explicitly mention/give an example for the `#any-not-eq?` behavior.

Issue #1808, opened 2022-07-21
can't derive grammar from a base grammar which uses an external scanner


orzechowskid:
I'm attempting to create a grammar derived from tree-sitter's CSS grammar and, based on the suggestions in [this comment](https://github.com/tree-sitter/tree-sitter/issues/646#issuecomment-647781353), have come up with (basically) the following:

```
// tree-sitter-css/my_grammar/grammar.js

module.exports = grammar(require('../grammar'), {
  name: 'my_grammar",

  rules: {
    x: seq('foo', 'bar'),
    _block_item: ($, original) => choice(original, $.x)
  }
});
```
I'm able to `cd my_grammar` and run `tree-sitter generate` without errors; subsequent running of `npx node-gyp configure` and `npx node-gyp build` also finishes without errors.  but running `tree-sitter test` or `tree-sitter parse foo.txt` give me this error:

```
Error opening dynamic library "/home/orzechowskid/.tree-sitter/bin/my_grammar.so"
Details:
  /home/orzechowskid/.tree-sitter/bin/my_grammar.so: undefined symbol: tree_sitter_my_grammar_external_scanner_create
```

this derived grammar has no need for an external scanner, though the base grammar it derives from does.  is there a way to configure some combination of tree-sitter, node-gyp, and my grammar file to use tree-sitter-css' external scanner?

thanks!

amaanq:
currently you need to just copy the scanner over and rename the relevant tree_sitter_LANG_external_scanner_ACTION functions w/ your language name, but this should be done automatically if there's no scanner file maybe

ahlinc:
> "/home/orzechowskid/.tree-sitter/bin/my_grammar.so"

For now this looks like old Tree-sitter CLI, update your CLI to the latest.

> is there a way to configure some combination of tree-sitter, node-gyp, and my grammar file to use tree-sitter-css' external scanner?

It's possible to put most of the base external scanner impl into a separate file like `scanner_impl.c` except `tree_sitter_LANG_external_scanner_ACTION` functions. And in the `scanner.c` it would be possible to do:
```c
#include "scanner_impl.c"

void *tree_sitter_css_external_scanner_create() { return NULL; }
void tree_sitter_css_external_scanner_destroy(void *p) {}
void tree_sitter_css_external_scanner_reset(void *p) {}
unsigned tree_sitter_css_external_scanner_serialize(void *p, char *buffer) { return 0; }
[...]

Issue #1262, opened 2021-07-13
tree-sitter-cli doesn't support multiple scanner source files


jneem:
This is a follow-up from the discussion #1261, and seems related to #1246.

I'm trying to update a project from tree-sitter 0.13 to 0.19, and the scanner code has multiple source files (`src/scanner.cc`, `src/scanner_keywords.cc`, etc). It appears that tree-sitter-cli only compiles one of them; based on the discussion in #1246, this behavior is hard-coded.

Would you take a PR to make tree-sitter-cli take a list of source paths? It could be specified as a setting in `package.json`.

ahlinc:
I'm  :+1:  for a PR.
But there couple of questions:
- Do such settings in the `package.json` to be required or optional?
  - If to make it required than there should be added a logic that would update `package.json` automatically for all existing grammars.
    - But when to do this? On the `tree-sitter generate`, on the `tree-sitter parse`, or on both?
      - If only on the `tree-sitter generate` than the `tree-sitter parse` should silently behave as before and try to compile the `scanner.cc` or the `scanner.c` if one of such files exist and there no explicit `tree-sitter.scanner` setting in the `package.json` file?
- What settings structure needs to be supported in the `package.json` file?
  ```json
  {
    "tree-sitter": {
      "scanner": ???
    }
  }
  ```
  Does it make sense to support separate compilation flags per scanner file or some common flags for all files would be enough?

jneem:
I guess the easiest behavior to implement would be that the `package.json` supports the optional key
```
{
    "tree-sitter": {
        "scanner-sources": [ "src/scanner.cc", "src/scanner_extra.cc"]
    }
}
```

If this key is found, the files listed are the scanner source files. If this key is not found, the current behavior is used (`src/scanner.c` if it exists, or if not then `src/scanner.cc` if it exists).

I think it's fine to use common flags for all files; in fact, I was planning to just compile all files in the same compiler invocation. If the build system needs to get much fancier than that, it probably makes sense to start using an external build system.

By the way, [the docs](http://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners) say that you need to add the scanner sources to `binding.gyp`. Is this still the case?

ahlinc:
> By the way, [the docs](http://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners) say that you need to add the scanner sources to `binding.gyp`. Is this still the case?

[...]

PR #2533, opened 2023-08-19
feat: allow specifying an external scanner configuration

### LINKED ISSUE #1246: tree-sitter-cli does not allow C++11 language features in external scanners on macOS
See [this CI run](https://github.com/tlaplus-community/tree-sitter-tlaplus/actions/runs/1012002111) where the same grammar is built & tested in windows, ubuntu, and macOS. The build itself (via `npm install`) is fine; however, when `tree-sitter test` is run, it [fails on macOS](https://github.com/tlaplus-community/tree-sitter-tlaplus/runs/3020175648?check_suite_focus=true) with multiple errors of the following type:
```
/Users/runner/work/tree-sitter-tlaplus/tree-sitter-tlaplus/src/scanner.cc:144:37: error: non-aggregate type 'const std::vector<int32_t>' (aka 'const vector<int>') cannot be initialized with an initializer list
  static const std::vector<int32_t> THEN_TOKEN = {'T', 'H', 'E', 'N'};
```
Initializer lists are a C++11 feature that would be quite useful to have. Is the C++ standard version restricted on macOS somehow?

https://github.com/nvim-treesitter/nvim-treesitter/issues/1515 possibly related

amaanq:
Closes #1246 

Now one can specify external files, include paths (absolute, not meant to be portable, maybe this should be removed). The link section is an object where you specify windows, linux, and mac link args as they can be different from each other.

Just a side note, I think the loader can be improved in general, its code is getting a bit messy but for now this works great to have a more feature-rich external scanner setup

### DIFF
--- a/cli/benches/benchmark.rs
+++ b/cli/benches/benchmark.rs
@@ -209,7 +209,7 @@ fn parse(path: &Path, max_path_length: usize, mut action: impl FnMut(&[u8])) -> 
 fn get_language(path: &Path) -> Language {
     let src_dir = GRAMMARS_DIR.join(path).join("src");
     TEST_LOADER
-        .load_language_at_path(&src_dir, &src_dir)
+        .load_language_at_path(&src_dir, &src_dir, None)
         .with_context(|| format!("Failed to load language at path {:?}", src_dir))
         .unwrap()
 }
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -87,6 +87,7 @@ pub struct LanguageConfiguration<'a> {
     pub locals_filenames: Option<Vec<String>>,
     pub tags_filenames: Option<Vec<String>>,
     pub language_name: String,
+    pub external_scanner_config: Option<ExternalScannerConfig>,
     language_id: usize,
     highlight_config: OnceCell<Option<HighlightConfiguration>>,
     tags_config: OnceCell<Option<TagsConfiguration>>,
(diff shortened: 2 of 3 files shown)


ahlinc:
I think this PR adds too much functionality than I suggest in https://github.com/tree-sitter/tree-sitter/issues/1246#issuecomment-876536579 and in https://github.com/tree-sitter/tree-sitter/issues/1262#issuecomment-881111812.
We shouldn't encourage to link against additional external libraries from external scanners, this kills an idea of grammars portability, external scanners should serve for very limited purposes and be as much portable as Tree-sitter C core.

ahlinc:
By the way, a problem with multiple scanner files can be worked around by using simple `#include` statements if such files are of the same language type, like I've described in https://github.com/tree-sitter/tree-sitter/issues/1808#issuecomment-1684920433

amaanq:
Yeah, links are overkill I agree. 

I also agree that just including a file works, but this is nice in the case someone wants to have definitions in a header file, and then an implementation in a C file, and compile w/ that C file

ahlinc:
> I also agree that just including a file works, but this is nice in the case someone wants to have definitions in a header file, and then an implementation in a C file, and compile w/ that C file

Including non header files doesn't work if you have for ex. many C and C++ files, compiler frontend commands (gcc, clang) don't allow to mix different language files in a call (they allow it in a limited fashion) and to avoid mistakes it's better to compile all source files with a clear distinct compiler calls as a separate units into separate object files and later link all of them together by a separate linking command/

Original problems were of two sorts:
1. It's hard to combine multiple source files.
2. There is no way to pass additional compiler flags.

I think this should be implemented in a way in scanner configuration the end user would be able to define a common flags for all files and explicitly define every source code file with ability to override compilation flags per file.

amaanq:
After some discussion and thinking it over - a better idea (and the problem I *really* wanted to solve) - is to just add an external_files array to specify other files that should have their mtime monitored for compilation. This gives us some freedom in adding more files w/o having to worry about recompilation succeeding.

PR #2538, opened 2023-08-19
fix(safety): mark functions that potentially deref a raw pointer as unsafe

### LINKED ISSUE #2537: Unsound FFI functions in `*/c_lib`
Disclaimer: This is closer to a nitpick than a serious vulnerability.

Many of the functions in `*/c_lib.rs` are unsound. Since these are part of the public API, even on the Rust side, I believe they should be marked `unsafe` if calling them from safe Rust can lead to UB.

As an example [ts_tagger_delete](https://github.com/tree-sitter/tree-sitter/blob/70766f53275612e363410f71f58ae8ef1351dd3e/tags/src/c_lib.rs#L70C58-L70C58) can lead to double free or out of bounds access in safe rust. A snippet demonstrating this in the playground, which can be checked with Miri: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b34b49cac043bf4101077d5a1ab7d786.

I've verified the same example locally using the actual function found in the `tree-sitter-tags` crate.

amaanq:
Closes #2537 

@oskgo can I get your thoughts on this?

### DIFF
--- a/highlight/src/c_lib.rs
+++ b/highlight/src/c_lib.rs
@@ -47,11 +47,11 @@ pub unsafe extern "C" fn ts_highlighter_new(
     let highlight_names = slice::from_raw_parts(highlight_names, highlight_count as usize);
     let attribute_strings = slice::from_raw_parts(attribute_strings, highlight_count as usize);
     let highlight_names = highlight_names
-        .into_iter()
+        .iter()
         .map(|s| CStr::from_ptr(*s).to_string_lossy().to_string())
         .collect::<Vec<_>>();
     let attribute_strings = attribute_strings
-        .into_iter()
+        .iter()
         .map(|s| CStr::from_ptr(*s).to_bytes())
         .collect();
     let carriage_return_index = highlight_names.iter().position(|s| s == "carriage-return");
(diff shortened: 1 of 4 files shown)


ahlinc:
@amaanq You didn't marked all needed things as unsafe, I'll add a commit.

highlight/src/c_lib.rs
```diff
@@ -145,42 +152,60 @@ pub extern "C" fn ts_highlight_buffer_new() -> *mut TSHighlightBuffer {
     }))
 }
 
+/// Deleteis a [`TSHighlighter`] instance.
```
ahlinc, line 155: Misprint `Deleteis`

highlight/src/c_lib.rs
```diff
@@ -145,42 +152,60 @@ pub extern "C" fn ts_highlight_buffer_new() -> *mut TSHighlightBuffer {
...
+///
+/// # Safety
+///
+/// `this` must be non-null.
```
ahlinc, line 159: > `this` must be non-null.

This is not enough guaranty for safety, it also should be a valid pointer to a valid structure.

amaanq:
> @amaanq You didn't marked all needed things as unsafe, I'll add a commit.

I wasn't sure if all of them were needed to be marked, or just those derefing raw ptrs

highlight/src/c_lib.rs
```diff
@@ -145,42 +152,60 @@ pub extern "C" fn ts_highlight_buffer_new() -> *mut TSHighlightBuffer {
...
-    drop(unsafe { Box::from_raw(this) })
+pub unsafe extern "C" fn ts_highlighter_delete(this: *mut TSHighlighter) {
+    drop(Box::from_raw(this))
 }
 
+/// Deleteis a [`TSHighlightBuffer`] instance.
+///
+/// # Safety
+///
+/// `this` must be non-null.
```
ahlinc, line 169: If you don't have a tension to write a real docs for all this C API funcs right now I'd suggest to remove all such template doc messages because they don't provide useful meaning.

Every function that receives a raw pointer is unsafe by default.

ahlinc:
> I wasn't sure if all of them were needed to be marked, or just those derefing raw ptrs

Every function that receives raw pointers is unsafe by default because there is no guaranty that passed pointers are valid pointers and pointed structs are also valid.

ahlinc (approved):

PR #2540, opened 2023-08-20
feat: optionally generate Rust bindings during build process

ahlinc:
* Reimplements and closes #1524
---
# Motivation

This is required for a future work on making C lib to have an optional feature to have **zero dependencies** even from `libc` and  to resolve issues with `wasm32-unknown-unknown` target for a `Rust binding -> C lib` use case on the  `wasm32-unknown-unknown` target.

The `bindgen` dependency is optional and it can be used as:
```sh
cargo build -p tree-sitter -F bindgen
```

Because all optional dependencies automatically become features with the same names.


lib/binding_rust/build.rs
```diff
@@ -34,6 +37,44 @@ fn main() {
         .compile("tree-sitter");
 }
 
+#[cfg(feature = "bindgen")]
+fn generate_bindings() {
+    const HEADER_PATH: &'static str = "include/tree_sitter/api.h";
```
amaanq, line 42: Maybe move this outside the function but still guarded by the cfg? and 'static shouldn't be needed

lib/binding_rust/build.rs
```diff
@@ -34,6 +37,44 @@ fn main() {
...
+
+    let bindings = bindgen::Builder::default()
+        .header(HEADER_PATH)
+        .layout_tests(false)
+        .allowlist_type("^TS.*")
+        .allowlist_function("^ts_.*")
+        .allowlist_var("^TREE_SITTER.*")
+        .no_copy(no_copy.join("|"))
+        .generate()
+        .expect("Unable to generate bindings");
```
amaanq, line 68:
```suggestion
        .expect("Failed to generate bindings");
```

lib/binding_rust/build.rs
```diff
@@ -34,6 +37,44 @@ fn main() {
...
+
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let bindings_rs = out_dir.join("bindings.rs");
+
+    bindings.write_to_file(&bindings_rs).expect(&*format!(
+        "Unable to write bindings into path: {bindings_rs:?}"
```
amaanq, line 74:
```suggestion
        "Failed to write bindings into path: {bindings_rs:?}"
```

lib/binding_rust/bindings.rs
```diff
@@ -1,5 +1,7 @@
 /* automatically generated by rust-bindgen 0.66.1 */
 
+pub const TREE_SITTER_LANGUAGE_VERSION: u32 = 14;
+pub const TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION: u32 = 13;
```
amaanq, line 4: Why u32 :raised_eyebrow:

ahlinc: Because that's default type that came from `api.h` but I adapted it in https://github.com/tree-sitter/tree-sitter/pull/2540/files#diff-b3254ffc96186f12dd09ec7367f266556732e4bfc24dc84a39f966f1c73f21e3L29-R36.

Rust binding adapts `u32` to `usize` in many places but I'm not sure that it's a good approach.

lib/binding_rust/build.rs
ahlinc: :+1:

lib/binding_rust/build.rs
ahlinc: :+1:

lib/binding_rust/build.rs
ahlinc: It's not needed anywhere else so I'd keep it closer to the point of use. Thanks for the `'static` note, old habbit.

amaanq: Yeah I wasn't sure if you can remove static in a non top-level scope :sweat_smile:

lib/binding_rust/bindings.rs
amaanq: Makes sense

Issue #1279, opened 2021-07-19
cflags in binding.gyp do not affect scanner.c compilation.


m-kru:
I have added two flags to the `cflags_c` in the `binding.gyp` file.
Currently it looks like follows:
```
{
  "targets": [
    {
      "target_name": "tree_sitter_fbdl_binding",
      "include_dirs": [
        "<!(node -e \"require('nan')\")",
        "src"
      ],
      "sources": [
        "bindings/node/binding.cc",
        "src/parser.c",
        # If your language uses an external scanner, add it here.
        "src/scanner.c",
      ],
      "cflags_c": [
        "-std=c99",
        "-Wall",
        "-Werror",
      ]
    }
  ]
}
```

In one of the `scanner.c` functions I have written simple code that should generate warning treated as error:
```C
unsigned tree_sitter_fbdl_external_scanner_serialize(
	void *payload,
	char *buffer
) {
	int i = 5;
	char c = 3;

	c = i;

	return 0;
}
```
The `scanner.c` file is compiled when user runs `tree-sitter test`. However no warning/error resulting from `-Wall` and `-Werror` is reported.  Like the flags were not taken into account during `scanner.c` compilation.

ahelwer:
See https://github.com/tree-sitter/tree-sitter/issues/1246 and https://github.com/tree-sitter/tree-sitter/discussions/1184. `tree-sitter test` and `tree-sitter parse` do not use the binding.gyp file for compilation.

PR #2545, opened 2023-08-21
feat: allow specifying an external scanner's files

amaanq:
Supersedes #2533

This time, only an "external_files" array can be specified *only* meant to check the mtime.

The reason the old design was rejected was mainly due to complexity - it overcomplicated a simple problem and added too much that was not needed. Adding features such as controlling the linked-in libraries and compiler flags is not the goal of this PR, that adds unnecessary complexity and will cause platform incompatibilities down the line.

What *is* needed is knowing what files a `scanner.c` might include for compilation, so that the CLI may check its mtime for changes to know *when* it should recompile. This PR does exactly that by allowing one to specify an external-files array in package.json for additional files *beyond* scanner.c.

We want tree-sitter parsers to be as portable and reproducible on any system as possible - allowing one to specify flags and link args to the compiler execution inside the cli can break that if a user uses platform or compiler specific flags and links.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -89,7 +89,6 @@ pub struct LanguageConfiguration<'a> {
     pub locals_filenames: Option<Vec<String>>,
     pub tags_filenames: Option<Vec<String>>,
     pub language_name: String,
-    pub external_files: Option<Vec<String>>,
     language_id: usize,
     highlight_config: OnceCell<Option<HighlightConfiguration>>,
     tags_config: OnceCell<Option<TagsConfiguration>>,
@@ -99,7 +98,7 @@ pub struct LanguageConfiguration<'a> {
 
 pub struct Loader {
     parser_lib_path: PathBuf,
-    languages_by_id: Vec<(PathBuf, OnceCell<Language>, Option<Vec<String>>)>,
+    languages_by_id: Vec<(PathBuf, OnceCell<Language>, Option<Vec<PathBuf>>)>,
     language_configurations: Vec<LanguageConfiguration<'static>>,
     language_configuration_ids_by_file_type: HashMap<String, Vec<usize>>,
     language_configuration_in_current_path: Option<usize>,
@@ -322,7 +321,7 @@ impl Loader {
         &self,
         src_path: &Path,
         header_path: &Path,
-        external_files: &Option<Vec<String>>,
+        external_files: &Option<Vec<PathBuf>>,
     ) -> Result<Language> {
         let grammar_path = src_path.join("grammar.json");
         let parser_path = src_path.join("parser.c");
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
There’s something nice about all grammars using the same convention and naming here. What problem is solved by making it depend on the JSON file?

amaanq:
One problem that's solved is checking other files' modified times.

Grammars like typescript, [xml](https://github.com/ObserverOfTime/tree-sitter-xml), and others have more than 1 file for the scanner, so if a change is made in a different file included by the scanner, but not the scanner itself, recompilation won't occur as the cli only checks for the modified time of "scanner.c".

This is not *necessary*, it's only for specifying other files besides scanner.c for the cli to check the modified time of those files

cli/loader/src/lib.rs
```diff
@@ -929,6 +950,7 @@ fn needs_recompile(
     lib_path: &Path,
     parser_c_path: &Path,
     scanner_path: &Option<PathBuf>,
+    external_files_paths: &[PathBuf],
```
ahlinc, line 953: May be it's time to generalize this function and accept just a one parameter: `checked_paths: &[Path]`

cli/loader/src/lib.rs
```diff
@@ -304,16 +305,21 @@ impl Loader {
...
             })
             .map(|l| *l)
     }
 
-    pub fn load_language_at_path(&self, src_path: &Path, header_path: &Path) -> Result<Language> {
+    pub fn load_language_at_path(
+        &self,
+        src_path: &Path,
+        header_path: &Path,
+        external_files: &Option<Vec<String>>,
```
ahlinc, line 321: The `external_files` parameter looks not very semantically correct for this function, I think it would be better to replace it on a `force_recompiling: bool` parameter and make this decision on https://github.com/tree-sitter/tree-sitter/pull/2545/files#diff-eb2bc03777ae5bf9ec8c213b1be1a691a778c1d60594f46bed9a7f9831dafedaR308

cli/loader/src/lib.rs
```diff
@@ -338,11 +344,22 @@ impl Loader {
...
+            Vec::new()
+        };
+
         self.load_language_from_sources(
             &grammar_json.name,
-            &header_path,
+            header_path,
             &parser_path,
             &scanner_path,
+            &external_files,
```
ahlinc, line 362: I'd name it longer but more clear like `external_scanner_files`.

ahlinc:
I think this PR needs to have a clear explanation why _superseded design was rejected due to focus on portability_ and the next issues would be closed with explanation for the same reason:
* #1279
* #1246

The next one needs to be explained with examples and a reference on the xml grammar how to _include_ other files and closed also:
* #1262

ahlinc:
### Additional notes

Practice of migration of many grammars to pure C scanners for better portability showed that even complex scanners don't require too much of facilities and mostly require several things:
1. Be able to classify strings, `<wctype.h>`.
2. Manage stacks of strings and other types to track a context.

For all of the above we (@amaanq  and me) have a draft plan how to provide all this facilities in a generated `src/tree_sitter/scanner.h` file without dependencies from `libc` headers and by leveraging mechanisms that already exist in Tree-sitter.

amaanq:
superseded by #3031

PR #2566, opened 2023-08-25
fix: do not increment `current_included_range_index` past `included_range_count` in `__do_advance`

### LINKED ISSUE #2563: get_column segfaults when called while lexer->eof(lexer) == true
Calling `lexer->get_column(lexer)` will result in **Segmentation fault** if the condition of `lexer->eof(lexer)` is met.

In other words, when you try to get column when at the end of the file, it segfaults. Took me some time to figure that out, resolved the issue locally by creating a simple wrapper:

```c++
uint32_t get_column(TSLexer *lexer) {
        if (lexer->eof(lexer)) {
            return 0;
        }

        return lexer->get_column(lexer);

}
```

amaanq:
Closes #2563

### DIFF
--- a/.github/workflows/sanitize.yml
+++ b/.github/workflows/sanitize.yml
@@ -36,14 +36,14 @@ jobs:
       env:
         UBSAN_OPTIONS: halt_on_error=1
         CFLAGS: -fsanitize=undefined
-        RUSTFLAGS: -lubsan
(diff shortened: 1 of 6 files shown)


ahlinc (approved):
I assume I'm done with a correct hang test implementation. For now without a proc macro :joy: 

I've excluded such hang test for exotic arches and for ASAN target because don't want to spend more time on corner cases, on all important targets in CI and on local dev env such test would work well.

PR #2570, opened 2023-08-26
Add tree-sitter version to generated parser

MatthewGentoo:
Adds a comment to the top of generated parser.c files:

    // generated by tree-sitter 0.20.8 (6d41d99990cfc8527f424603cdfce1c9af2d78ef)
    or
    // generated by tree-sitter 0.20.8

This should help a bit with traceability, to make issues like https://github.com/tree-sitter/tree-sitter-bash/issues/199 a little easier to figure out.


cli/src/generate/render.rs
```diff
@@ -255,6 +258,17 @@ impl Generator {
             .count();
     }
 
+    fn add_generator_version(&mut self) {
+        let version = if let Some(build_sha) = BUILD_SHA {
+            format!("{} ({})", BUILD_VERSION, build_sha)
+        } else {
+            BUILD_VERSION.to_string()
+        };
```
MatthewGentoo, line 266: This is copied out of `main.rs` because I wasn't sure of a good place to put it...

cli/src/generate/render.rs
```diff
@@ -15,6 +15,8 @@ use std::{
     mem::swap,
 };
 
+const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
+const BUILD_SHA: Option<&'static str> = option_env!("BUILD_SHA");
```
amaanq, line 19: don't think 'static is needed

cli/src/generate/render.rs
amaanq: It's fine imo

cli/src/generate/render.rs
MatthewGentoo: Thanks, removed and amended commit :)

ahlinc:
I had a similar idea but I thought to add a more precise info form the output of `git describe` command:
```sh
# git describe --tags --match=v* --abbrev --dirty --broken --always 
v0.20.8-378-g6d41d999-dirty
# git describe --tags --match=v* --abbrev=40 --dirty --broken --always 
v0.20.8-378-g6d41d99990cfc8527f424603cdfce1c9af2d78ef-dirty
```

MatthewGentoo:
Would it be OK add a new dependency on the [git-version](https://lib.rs/crates/git-version) crate to handle that? It's a macro that calls `git describe`, and falls back to `CARGO_PKG_VERSION` on failure, which seems like just what we want.

ahlinc:
Yes, it would be good as a build time dependency, but its [default args](https://github.com/fusion-engineering/rust-git-version/blob/master/git-version-macro/src/lib.rs#L116) don't feet our needs and need to be adapted to args form a snippet that I posted previously.

amaanq:
I'm gonna close this for now because we have a plan in the very near future to bake in the parser's semver and the like, and expose it via the api for dynamic parser introspection

PR #2573, opened 2023-08-26
fix: makefile should install only `api.h` header

ahlinc:
This PR addresses an issue raised in https://github.com/tree-sitter/tree-sitter-bash/issues/199#issuecomment-1694416505
Because Gentoo distributive solely relies on Tree-sitter's Makefile and got an issue due to an issue in the Makefile.


amaanq (approved):

amaanq:
We shouldn't revert this, api.h is the only relevant file for users to interact with when installing the lib - parser.h is more internal and is meant for a grammar's parser/scanner to consume, not front-facing users

PR #2577, opened 2023-08-26
Intern/absorption edge case fix

amaanq:
# Problem/Background

Currently, the CLI and parser show weird and inconsistent behavior given a few edge cases that might come up when users are writing grammar. I initially thought it stemmed from the same problem, but there are actually two problems.

One, consider the following:

```js
module.exports = grammar({
    name: 'test',

    rules: {
        source: $ => choice($.a, $.b, $.c),
        
        a: $ => 'a',
        b: $ => seq('a'),
        c: $ => choice('c'),
});
```

At a glance, you'd assume `a`, `b`, and `c` to be similarly constructed in the parser, and the "under the hood" stuff to be roughly the same. This is *not* the case though when it comes to parsing and querying a grammar constructed like this. The problem lies in the usage of `seq` and `choice` on a single terminal, and the CLI not processing this the same way that it would given just one single terminal, in that it does not perform the same "absorbing" or "replacing" effect as it would given just the terminal itself.

When trying to query for a, we would need to use `(a)` since we know any named `a` nodes will *always* consist of the letter a, and so we cannot query for the anonymous literal node `"a"` because of this absorption effect. We know the same holds true for a sequence or choice of a singular terminal as well, given that it can only be that one item, so we *should* process these the same way as we do with just a single terminal.

However, for `b` and `c`, we can query for both the named nodes *and* their anonymous literal node *as well*, which is at best, unnoticed, and at worst, bizarrely inconsistent when one is writing queries with these edge cases in their grammar. 

Two, this is an even bigger problem when we consider hidden nodes: 

```js
module.exports = grammar({
    name: 'test',

    rules: {
        source: $ => choice($._a, $._b, $._c),
        
        _a: $ => 'a',
        _b: $ => seq('a'),
        _c: $ => choice('c'),
});
```

Hidden nodes that consist of a single terminal will get that terminal absorbed (in the above example, that would be `_a`), and then those that are a seq or choice of that single terminal can *query* for that literal, since as we know, a hidden node cannot be queried, but as we explained above, a single terminal wrapped in a seq or choice can be queried for. In this case, the hiding of a node provides extremely inconsistent behavior for a user, and can be a source of frustration for those that do not understand what's going on - this is the second problem, and while it does go hand-in-hand with the first one, it stems from another issue entirely.

Ideally, a user would not write rules that have these bizarre edge cases, but we should not have these inconsistencies to begin with as well.

# Solution

Fixing the first problem is not too hard - we should simply intern a sequence or choice of 1 rule the exact same way we intern a literal or pattern, and we should also print a warning to the user that this is a weird construct and is unnecessary.

The second problem is also not so hard, but it is a different problem entirely. When extracting terminal rules that have a single usage, we should not replace them if the rule encompassing them is a hidden rule, so that we can ensure these literals are query-able. In the case of visible nodes, the node itself should directly be used.

Tests were added for both cases that fail before the given changes, and illustrate the expected behavior of my changes :grin:

### DIFF
--- a/cli/src/generate/prepare_grammar/extract_tokens.rs
+++ b/cli/src/generate/prepare_grammar/extract_tokens.rs
@@ -53,10 +53,14 @@ pub(super) fn extract_tokens(
         {
             if i > 0 && extractor.extracted_usage_counts[index] == 1 {
                 let lexical_variable = &mut lexical_variables[index];
-                lexical_variable.kind = variable.kind;
-                lexical_variable.name = variable.name;
-                symbol_replacer.replacements.insert(i, index);
-                continue;
+                if lexical_variable.kind == VariableType::Auxiliary
+                    || variable.kind != VariableType::Hidden
+                {
+                    lexical_variable.kind = variable.kind;
+                    lexical_variable.name = variable.name;
+                    symbol_replacer.replacements.insert(i, index);
+                    continue;
+                }
             }
         }
         variables.push(variable);
@@ -488,6 +492,32 @@ mod test {
                 panic!("Expected an error but got no error");
             }
         }
+    }
+
+    #[test]
+    fn test_extraction_on_hidden_terminal() {
+        let (syntax_grammar, lexical_grammar) = extract_tokens(build_grammar(vec![
+            Variable::named("rule_0", Rule::non_terminal(1)),
+            Variable::hidden("_rule_1", Rule::string("a")),
+        ]))
+        .unwrap();
+
+        // The rule `_rule_1` should not "absorb" the
+        // terminal "a", since it is hidden,
+        // so we expect two variables still
+        assert_eq!(
+            syntax_grammar.variables,
+            vec![
+                Variable::named("rule_0", Rule::non_terminal(1)),
+                Variable::hidden("_rule_1", Rule::terminal(0)),
+            ]
+        );
+
+        // We should not have a hidden rule in our lexical grammar, only the terminal "a"
+        assert_eq!(
+            lexical_grammar.variables,
+            vec![Variable::anonymous("a", Rule::string("a"))]
+        );
     }
 
     fn build_grammar(variables: Vec<Variable>) -> InternedGrammar {
--- a/cli/src/generate/prepare_grammar/intern_symbols.rs
+++ b/cli/src/generate/prepare_grammar/intern_symbols.rs
@@ -18,13 +18,13 @@ pub(super) fn intern_symbols(grammar: &InputGrammar) -> Result<InternedGrammar> 
         variables.push(Variable {
             name: variable.name.clone(),
             kind: variable_type_for_name(&variable.name),
-            rule: interner.intern_rule(&variable.rule)?,
+            rule: interner.intern_rule(&variable.rule, Some(&variable.name))?,
         });
     }
 
     let mut external_tokens = Vec::with_capacity(grammar.external_tokens.len());
     for external_token in &grammar.external_tokens {
-        let rule = interner.intern_rule(external_token)?;
+        let rule = interner.intern_rule(external_token, None)?;
         let (name, kind) = if let Rule::NamedSymbol(name) = external_token {
             (name.clone(), variable_type_for_name(name))
         } else {
@@ -35,7 +35,7 @@ pub(super) fn intern_symbols(grammar: &InputGrammar) -> Result<InternedGrammar> 
 
     let mut extra_symbols = Vec::with_capacity(grammar.extra_symbols.len());
     for extra_token in &grammar.extra_symbols {
-        extra_symbols.push(interner.intern_rule(extra_token)?);
+        extra_symbols.push(interner.intern_rule(extra_token, None)?);
     }
 
     let mut supertype_symbols = Vec::with_capacity(grammar.supertype_symbols.len());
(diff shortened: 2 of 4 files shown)


ahlinc:
Good to see how fast you convert our conversations in the real code :rocket:

amaanq:
Nice, CI shows us C/C++ have a problem with this edge case :) fixing

cli/src/generate/prepare_grammar/extract_tokens.rs
```diff
@@ -51,12 +51,18 @@ pub(super) fn extract_tokens(
             index,
         }) = variable.rule
         {
-            if i > 0 && extractor.extracted_usage_counts[index] == 1 {
+            if i > 0 && extractor.extracted_usage_counts[index] == 1
+            //&& variable.kind != VariableType::Hidden
```
maxbrunsfeld, line 55: 🔥 ?

maxbrunsfeld (approved):
Great work, thanks for fixing these issues.

cli/src/generate/prepare_grammar/extract_tokens.rs
amaanq: whoops, extraneous comment I didn't mean to add

clason:
@amaanq I noticed some regressions from this: generating 

* tree-sitter-nim (of course) fails now with
```
Rule '_symbol_export_marker' cannot be used as both an external token and a non-terminal rule
```
* tree-sitter-fortran fails now with
```
Rule '_line_continuation' cannot be used as both an external token and a non-terminal rule
```

Do grammars have to be updated due to this PR?

amaanq:
Took a look, that is intentional (maybe should've marked this as breaking then, but none of the upstream grammars had any issues).

The problem is in both cases they're using a single terminal wrapped in a hidden node, which doesn't make much sense to me - the Fortran fix is pretty easy, but the Nim one is pretty convoluted (not exactly the cleanest grammar), they might have to just change the name and call it a day (unhide it)

clason:
rule of thumb: If it _can_ break something, it _will break something...

clason:
Here's a few more breakages where the generation succeeds but the (previously valid) queries no longer parse correctly:

```
hare (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 90:4. Invalid node type "":
  "@packed"
   ^
hurl (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 31:4. Invalid node type "status":
  "status"
   ^
liquid (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 62:4. Invalid node type "break":
  "break"
   ^
squirrel (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 65:4. Invalid node type "static":
  "static"
   ^
tlaplus (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 89:4. Invalid node type "skip":
  "skip"
   ^
vhs (highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 15:4. Invalid node type "Hide":
  "Hide"
   ^
```

maxbrunsfeld:
@amaanq the error message that we’re giving for the nim and Fortran grammars seems possibly wrong now? Is there actually a rule that is both an external token and non-terminal (having children)? Or are we falling into that code path incorrectly because of this transformation?

amaanq:
In both grammars, they do something that I find to be a bit odd (which this PR does affect)

e.g. in fortran, they have a rule `_line_continuation` which looks like this:

```js
_line_continuation: $ => '&',
``` 

which is also used as an external, likely for just checking if it's valid or not (not sure why that is, because it is also an extra, meaning that it is always valid), but it seems like they also actually lex this token in the scanner, defeating the purpose of the rule in the javascript.

My PR affects this because given a hidden rule (one that starts with `_`), we no longer absorb the terminal. If we "absorb" the terminal '&' to make `_line_continuation` the terminal node instead, this node cannot be queried for and nothing can be done to fetch this node from the parse tree or even query for it. But now, `_line_continuation` is no longer a terminal because it is a *hidden* node that consists of a terminal, so we consider '&' to be the terminal node, not `_line_continuation` itself, which allows us to query for this literal. (Nim is a pretty similar case to this btw)

So imo, the grammars are just wrong here, and probably should be adapted (I did PR a fix to Fortran). This could have been considered "UB" in tree-sitter land before, but now is a bit more rigidly defined after this PR

Side note, it'd be cool if we could do some kind of "crater run" (term that rust uses to test compilation of every single crate on crates.io to check for regressions) to test generating all grammars (like nvim-treesitter does), we don't need it in every PR, but a manual way to trigger it by a comment would be cool and pretty useful (we'd have to be a little selective - only grammars in a decently stable state can be considered)

clason:
What about the other grammars that just fail to parse?

>  it'd be cool if we could do some kind of "crater run" 

* `gh repo clone nvim-treesitter/nvim-treesitter -b main`
* `./scripts/install-parsers.lua --generate`
* `./scripts/check-queries.lua`
* profit! ;)

cli/src/generate/prepare_grammar/intern_symbols.rs
```diff
@@ -147,6 +151,21 @@ impl<'a> Interner<'a> {
 
         None
     }
+
+    // In the case of a seq or choice rule of 1 element in a hidden rule, weird
```
clason, line 155: This seems to be a trick grammars use to expose single anonymous nodes (see list in comment above), which now no longer works (the anonymous node is no longer exposed so queries are broken).

It also seems to behave unexpectedly with using a `const` in a simple (non-seq or choice) rule as in [clojure](https://github.com/sogaiu/tree-sitter-clojure/blob/f4236d4da8aa92bc105d9c118746474c608e6af7/grammar.js#L338-L339).

clason: (Definitely should have been marked breaking...)

sogaiu: For convenience, [here](https://github.com/sogaiu/tree-sitter-clojure/blob/f4236d4da8aa92bc105d9c118746474c608e6af7/grammar.js#L189-L194) is the definition of the `const` mentioned above to go along with the link to the rule in tree-sitter-clojure.

clason: I also think this should have been a separate PR and commit...

stadelmanma:
> In both grammars, they do something that I find to be a bit odd (which this PR does affect)
> 
> e.g. in fortran, they have a rule `_line_continuation` which looks like this:
> 
> ```js
> _line_continuation: $ => '&',
> ```
> 
> which is also used as an external, likely for just checking if it's valid or not (not sure why that is, because it is also an extra, meaning that it is always valid), but it seems like they also actually lex this token in the scanner, defeating the purpose of the rule in the javascript.
> 
> My PR affects this because given a hidden rule (one that starts with `_`), we no longer absorb the terminal. If we "absorb" the terminal '&' to make `_line_continuation` the terminal node instead, this node cannot be queried for and nothing can be done to fetch this node from the parse tree or even query for it. But now, `_line_continuation` is no longer a terminal because it is a _hidden non-terminal_ node that consists of a terminal, so we consider '&' to be the terminal node, not `_line_continuation` itself, which allows us to query for this literal. (Nim is a pretty similar case to this btw)
> 
> So imo, the grammars are just wrong here, and probably should be adapted (I did PR a fix to Fortran). This could have been considered "UB" in tree-sitter land before, but now is a bit more rigidly defined as erroneous after this PR
> 
> Side note, it'd be cool if we could do some kind of "crater run" (term that rust uses to test compilation of every single crate on crates.io to check for regressions) to test generating all grammars (like nvim-treesitter does), we don't need it in every PR, but a manual way to trigger it by a comment would be cool and pretty useful (we'd have to be a little selective - only grammars in a decently stable state can be considered)

I think in the Fortran case, in a much earlier version of tree-sitter you might have needed to name the same node in the grammar when using a scanner for external tokens? It's also equally likely I made a mistake since that was the first, and only, parser grammar I have created. Nonetheless, good find and thank you for taking the time to open a PR in the Fortran repo.

cli/src/generate/prepare_grammar/intern_symbols.rs
maxbrunsfeld: @amaanq seems like there is some negative fallout from this change. Some grammars explicitly wanted the old behavior, in which single-child seq was not special cased, but just treated like other sequences.

Should we consider reverting that part of this PR?

clason: I've already gone through the ecosystem and opened PRs to the affected parsers (I know of).

I've yet to hear from one that _wants_ to have a rule wrapping an anonymous node (which seems to me to just complicate parsers for little gain). 

However, the check _should_ be made smarter to not trigger the false positives reported in #3455.

(And, yes, some way of testing the side-effects of such changes _before_ landing them would be very useful -- tree-sitter has evolved quite a bit beyond the original handful of parsers under this org that are pulled in as fixtures.)

PR #2582, opened 2023-08-28
feat: support inline flags in regexes

soqb:
# Objective
- #2474 added support for our first regex flag (`i`), but did not handle inline flags.
- This PR implements the missing support for inline flags (`a(?i)b` matching `aB` but not `AB`).
- I also added support for negation of inline flags (`(?i)a(?-i)b` matching `Ab` but not `AB`).


amaanq:
Wow, this is great! I will take a look later today

dundargoc:
@soqb mind rebasing?

dundargoc:
@amaanq mind taking a look at this when you have time?

amaanq:
Actually, with the switch to Hir from Ast, this is no longer needed since the Hir handles translating flags for us. From the [docs](https://docs.rs/regex-syntax/latest/regex_syntax/hir/struct.Hir.html):

> The HIR will also “compile away” every flag present in the concrete syntax. For example, users of HIR expressions never need to worry about case folding; it is handled automatically by the translator (e.g., by translating (?i:A) to [aA]).

Thank you for the PR though!

PR #2584, opened 2023-08-28
doc: Add internal links and fix formatting of Rust docs

mgunyho:
In the documentation for the Rust bindings, there were several internal links that didn't have `` `backticks` ``, and several items with backticks but no links. With this PR, all internal links are changed to ``[`target`]``, or ``[`target`](StructName::target)``, which is the format used by the Rust standard library (and most other libraries). I also fixed a broken link.

Before:

docs-before-1

docs-before-2

After:

docs-after-1

docs-after-2


ahlinc (approved):
Thank you very much for all these function level doc improvements!

PR #2585, opened 2023-08-28
doc: Include README in top-level Rust module documentation

mgunyho:
The [README for the Rust bindings](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_rust/README.md) contains an introduction and some nice examples of using the bindings, but these examples are not found anywhere in the [online documentation](https://docs.rs/tree-sitter/latest/tree_sitter/). This PR adds the README to the front page of the docs.

Before:

docs-before

After:

docs-after


cli/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version.workspace = true
 [[bin]]
 name = "tree-sitter"
```
whytesandz89, line 15: Christopher Roll all apps

PR #2594, opened 2023-08-31
Add serialization support for queries, highlight configs

### LINKED ISSUE #1942: Caching or Serializing a `TSQuery`

SergioBenitez:
This is a continuation of #2374 which fully resolves #1942.

The approach here is to expose serializable versions of queries and highlight configurations instead of attempting to provide serialized versions directly. This enables the consumer to serialize the structures as they wish, and importantly, to do so in a way that respects things like byte order properly.

The above is exposed via new `serializable()` and `deserialize()` methods on `Query` and `HighlightConfiguration` and `Serializable{Query,HighlightConfiguration}` structures which implement `serde::{Serialize, Deserialize}` when the newly supported `serde` feature is enabled. I have successfully used this API to deserialize over 100 of highlight configurations in ~3 ms. This compares to ~3000ms to generate the configurations directly, or a 1000x improvement. The resulting structures can be used immediately without any additional preprocessing.

I should mention that there are a few concerns with the exact approach from #2374, which is replicated here:
  * There is no checking that the `const char *src` array's size is sufficient. Previously, a `size` argument was passed in but was unused. We should use the argument to check against the expected size.
  * Some sort of versioning needs to be added to (de)serialization. Otherwise, serialized bytes from a previous version can be deserialized by a new, incompatible version. While the bytes may physically fit, they may not semantically fit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -757,6 +757,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_regex"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a8136f1a4ea815d7eac4101cfd0b16dc0cb5e1fe1b8609dfd728058656b7badf"
+dependencies = [
+ "regex",
+ "serde",
+]
+
+[[package]]
 name = "serde_spanned"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -921,6 +931,8 @@ dependencies = [
  "bindgen",
  "cc",
  "regex",
+ "serde",
+ "serde_regex",
 ]
 
 [[package]]
@@ -982,6 +994,7 @@ version = "0.20.2"
 dependencies = [
  "lazy_static",
  "regex",
+ "serde",
  "thiserror",
  "tree-sitter",
 ]
--- a/cli/src/tests/helpers/query_helpers.rs
+++ b/cli/src/tests/helpers/query_helpers.rs
@@ -353,7 +353,7 @@ fn format_captures<'a>(
     captures
         .map(|capture| {
             (
-                query.capture_names()[capture.index as usize],
+                query.capture_names()[capture.index as usize].as_str(),
                 capture.node.utf8_text(source.as_bytes()).unwrap(),
             )
         })
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -2269,7 +2269,7 @@ fn test_query_captures_within_byte_range_assigned_after_iterating() {
         for (mat, capture_ix) in captures.by_ref().take(5) {
             let capture = mat.captures[capture_ix as usize];
             results.push((
-                query.capture_names()[capture.index as usize],
+                query.capture_names()[capture.index as usize].as_str(),
                 &source[capture.node.byte_range()],
             ));
         }
@@ -2292,7 +2292,7 @@ fn test_query_captures_within_byte_range_assigned_after_iterating() {
         for (mat, capture_ix) in captures {
             let capture = mat.captures[capture_ix as usize];
             results.push((
-                query.capture_names()[capture.index as usize],
+                query.capture_names()[capture.index as usize].as_str(),
                 &source[capture.node.byte_range()],
             ));
         }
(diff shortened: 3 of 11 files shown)


SergioBenitez:
Checking in on this. Is there a chance this will be merged soon? I'd like to release a library making use of this feature, so I'll need to publish a fork on crates.io to do so if this won't be merged any time soon, which I'd like to avoid.

ahlinc:
There is a lot of interest in queries (de)serialization and a more robust solution will be implemented that covers more use cases with better deserialization performance.

SergioBenitez:
> There is a lot of interest in queries (de)serialization and a more robust solution will be implemented that covers more use cases with better deserialization performance.

That sounds great.

I wonder if it would make sense to merge something like this PR in the interim, seeing as how my original issue is nearly a year old now. Even if it isn't particularly robust, doesn't cover every use-case (which are missing?), and could be improved performance-wise, it still provides functionality that dramatically changes the utility of tree-sitter by immediately making it usable in cases where it was previously prohibitive due to performance impacts.

Nevertheless, is there anywhere I can follow those efforts and potentially contribute? I'd love to avoid duplicating effort, and of course, maintaining a fork just to have something on crates.io.

mister-choo:
This is such a good implementation, which improved my projects performance 10 times. Yet I see no movement about query caching since. This is very strange

clason:
Thank you, but this PR has gone stale (sorry!) and diverged too far from the current HEAD to keep this alive.

We take query compilation performance _very_ seriously, but I'm not sure this is a route we want to go down for this. In any case, we are planning a significant rework of the query engine in the next development cycle (0.28), so it's better to hold off on adding such features until the refactor is done.

ofek:
@clason Where's the best place to learn about the refactor you mentioned and how it unlocks performance improvements?

WillLillis:
> @clason Where's the best place to learn about the refactor you mentioned and how it unlocks performance improvements?

The refactor Clason mentioned is still very early in the planning stages. We need to finish some other big efforts first before dedicating enough time to commit to any decisions.

ofek:
I see, thanks! Are we able to follow any issues/discussions or is the planning happening outside of GitHub?

WillLillis:
> I see, thanks! Are we able to follow any issues/discussions or is the planning happening outside of GitHub?

Nothing on GitHub right now, again it's _very_ early and there other features in the works. There aren't any plans to move things publicly until we have a validated design. It will be several months at the earliest.

PR #2597, opened 2023-09-01
refactor: remove `&Option<T>` where used

amaanq:


cli/loader/src/lib.rs
```diff
@@ -353,7 +353,7 @@ impl Loader {
         name: &str,
         header_path: &Path,
         parser_path: &Path,
-        scanner_path: &Option<PathBuf>,
+        scanner_path: Option<&PathBuf>,
```
ahlinc, line 356: Define it as `scanner_path: Option<&Path>`, please.

ahlinc (approved):

PR #2600, opened 2023-09-02
refactor: short circuit for known constraints in position lookups

lavafroth:
### Changes
- Read input backwards to find point corresponding to given offset
- Skip to first row that equals supplied position when finding offset

Additionally, all tests are passing.


ahlinc:
The second opt is not the opt actually:

image

PR #2602, opened 2023-09-02
chore(lib): fix fields naming in `QueryPattern` to singular

ahlinc:
This PR renames fields in `QueryPattern` struct because they aren't arrays.


maxbrunsfeld:
These fields had plural names because their values are slices: non-owning references into arrays.

ahlinc:
Thank you for explanation, I'll revert this.

PR #2607, opened 2023-09-03
fix: `dealloc` calls on zero pointers

ahlinc:


amaanq (approved):

leat41:
Wonderful

leat41:
Wonderful

maxbrunsfeld:
I think it's ok (and standard) to call `free` on possibly-null pointers without additional conditional logic [reference](https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c). I don't think these new conditional branches are an improvement. Am I missing something?

ahlinc:
Yes it's allowed to call the `free` many times but this makes hard to debug memory leaking situations in a client code when you trying precisely count memory allocations and related deal locations, so I'd suggest to be strict on this in the library code.

ahlinc:
Yes it's allowed to call the `free` many times but this makes hard to debug memory leaking situations in a client code when you trying precisely count memory allocations and related deal locations, so I'd suggest to be strict on this in the library code.

And doing such counting usually happens by overriding memory management functions.

ahlinc:
Yes it's allowed to call the free many times but this makes hard to debug memory leaking situations in a client code when you trying precisely count memory allocations and related deal locations, so I'd suggest to be strict on this in the library code.

And doing such counting usually happens by overriding memory management functions.

ahlinc:
_Sorry, may be you saw a mess up with my last reply, Gitnub client app on Android sometimes starts to duplicate sent messages many times._

PR #2614, opened 2023-09-09
feat: allow setting the output directory for the generated C code

JCWasmx86:
Automatically disables bindings as they are a bit more difficult to relocate into a single directory. Furthermore this makes integration into meson easier/possible(?) I'm not sure this is the right approach, but it works at least for my usecase.

### DIFF
--- a/cli/generate/src/lib.rs
+++ b/cli/generate/src/lib.rs
@@ -44,6 +44,7 @@ pub const ARRAY_HEADER: &str = include_str!("templates/array.h");
 
 pub fn generate_parser_in_directory(
     repo_path: &Path,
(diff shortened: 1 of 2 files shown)


ahlinc:
**TODO:**
* Needs to be coordinated with #2438

lvkunkun-bing (commented):
能不能组成GS数据

lvkunkun-bing:
能不能使用gs登录

clason:
Isn't that obsolete now with `tree-sitter build -o ...`?

amaanq:
> Isn't that obsolete now with `tree-sitter build -o ...`?

this is for parser.c, not parser.so

PR #2647, opened 2023-09-22
feat: use lockfiles to prevent duplicate recompilation in a multi-threaded scenario

amaanq:


cli/loader/Cargo.toml
```diff
@@ -15,6 +15,7 @@ rust-version.workspace = true
 anyhow = "1.0"
 cc = "^1.0.58"
 dirs = "3.0"
+fs2 = "0.4.3"
```
ahlinc, line 18: Why did you choice the `fs2` crate?

ahlinc:
I think this implementation may be improved to avoid double recompilation in a case if some another compilation process already is running.

cli/loader/Cargo.toml
amaanq: brings in a cross-platform trait to lock files easily

ahlinc: I meant it looks unmaintained and there are fs3, fs4 crates.

maxbrunsfeld:
Can we do this without any lockfiles, by just compiling to a temporary location, and then moving the compiled binaries into place atomically?

ahlinc:
I think yes, but instead of using a temporary place (if it exists on a separate fs it would have a cost to move a file), I would compile to a file in place with a temporary name like `grammar_name.<PID>.so` where the PID is a current process id or better a current thread id which called the loader to compile a library. This would give a unique name without any extra libs to generate a temp file name.

maxbrunsfeld:
That's a good idea. Maybe we could just use a randomly-generated UUID (u128).

ahlinc:
@maxbrunsfeld  I tried to recall what I thought about of possible fixes when I saw this issue at the first time.

If to use compiling in a temporary location and move to a destination location:
* There may be situation that many compilation of the same library would happen in parallel what would consume additional computing resources.
* The last compiling task wins and that is what is expected.
* This wouldn't work on Windows because opened files treated differently:
  * _Windows don't allow to delete or move an opened file_, it doesn't behave like the `unlink` syscall on Linux.
  * An opened file on Windows platform is in a locked state for other file manipulation operations.
  * Attempts of atomic replacements on Windows will lead to errors that after a successful compilation a process may fail on an attempt to place a file to a final destination.

As amount of unit tests would grow up this compiling issues would raise more often.

_This PR needs improvements in a more fine tuned locking process._

lib/binding_rust/lib.rs
```diff
@@ -113,7 +113,7 @@ pub struct TreeCursor<'cursor>(ffi::TSTreeCursor, PhantomData<&'cursor ()>);
 
 /// A set of patterns that match nodes in a syntax tree.
 #[doc(alias = "TSQuery")]
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
maxbrunsfeld, line 116: The `TSQuery` is not reference counted internally, so I don't think this derived `Clone` implementation will work. Queries are also immutable, so I don't think we should need to make them clone-able.

maxbrunsfeld (changes requested):
I'd like to **not** parallelize the `parse` or `query` commands. Let's leave them simple. They don't need maximal throughput. Users who need to do parallel parsing should probably use the library, not the CLI.

ahlinc:
> I'd like to **not** parallelize the `parse` or `query` commands. Let's leave them simple. They don't need maximal throughput. Users who need to do parallel parsing should probably use the library, not the CLI.

I also would like to add that PRs without any extra changes (and maybe with some doc or comment misprints fixes) are much easier for reviewers to be understand, this significantly helps to review changes and assume what points were the reasons for them.

lib/binding_rust/lib.rs
ahlinc: For the same reason I [removed](https://github.com/tree-sitter/tree-sitter/blob/016d8c24998cc268d6a6c5da471f2cc11a1c0573/script/generate-bindings#L12) [this](https://github.com/tree-sitter/tree-sitter/blob/016d8c24998cc268d6a6c5da471f2cc11a1c0573/lib/binding_rust/build.rs#L53) also from bindings generation.

amaanq:
Superseded by #2953, which is a simpler PR

Issue #1282, opened 2021-07-20
Getting line number in scanner.c.


m-kru:
Is it possible to get the current line number in `scanner.c`? I would like to use it to provide more robust error messages to the user in case of syntax errors. `TSLexer` has 

> uint32_t (*get_column)(TSLexer *) - (Experimental) A function for querying the current column position of the lexer. It returns the number of unicode code points (not bytes) since the start of the current line.

However, it looks like there is no method or attribute to get the current line number.

ahlinc:
No, it's not possible.
The external scanner is a very low level thing that intended for specific purposes.
How do you see that the `scanner.c` can be involved into error messages provisioning?

m-kru:
@ahlinc For example like this: `errx(EXIT_FAILURE, "Expected foo, found bar")`.
The context may be important here. I am defining grammar not for a text editor integration but for language prototyping.

ahlinc:
@m-kru Hmm, got it, I've used something similiar in a grammar development process to debug my external scanner.
I see it's enough for you to stop the parser right in a place of some error in the external scanner, but the same doesn't possible for any terminals defined in the `grammar.js` file. Tree-sitter was designed to recover from errors and parse to the end of input in any way. Do you plan to define all your terminals as external tokens?

m-kru:
@ahlinc Not all terminals are external tokens, but all external tokes are terminal. Is it even possible to parse non-terminal as external token?

ahelwer:
@m-kru kind of, but not really - you would have to parse it as something like `(external_marker_start) (nonterminal) (external_marker_end)` where `nonterminal` is defined in the grammar.

ahlinc:
> Is it even possible to parse non-terminal as external token?

No, external scanners are soleily a complement for tree-sitter's lexer and the production of non-terminals is the parser resbonsibility.

MeepTech:
I would also find this very useful for understanding whats going on in a custom scanner during debugging. Does the lexer already keep track of it somehow for `--stat` mode possibly?

As a workaround I was thinking of using an external `$._newline_` token in my extras and potentially counting them. Ill let yall know if it works.

amaanq:
Closing as not planned (see https://github.com/tree-sitter/tree-sitter/pull/2670#issuecomment-1791455992)

PR #2648, opened 2023-09-23
feat(lexer)!: `TSLexer.get_byte()`

llllvvuu:
Currently, since external scanners are called on-demand at arbitrary
points, they are limited in context/state.

An example of where this leads to redundant computation is when
determining the effective indent of Python comments, which depends not
only on the comment's indentation but also the indentation of the next
non-comment line, which is an expensive lookahead (O(comment_size)). If
there are many consecutive comments (e.g. as generated by an IDE when
commenting out a block of code), then this computation is redundant, and
in total quadratic.

`TSLexer.get_byte()` improves the statefulness of external scanners,
allowing to keep track of lookahead state, reducing redundancy.

see: https://github.com/tree-sitter/tree-sitter-python/pull/245

BREAKING CHANGE: This breaks the `TSLexer` ABI. To minimize disruption,
it can be bundled with other breaking changes.

### DIFF
--- a/lib/include/tree_sitter/parser.h
+++ b/lib/include/tree_sitter/parser.h
@@ -44,6 +44,7 @@ struct TSLexer {
   TSSymbol result_symbol;
   void (*advance)(TSLexer *, bool);
   void (*mark_end)(TSLexer *);
+  uint32_t (*get_byte)(TSLexer *);
   uint32_t (*get_column)(TSLexer *);
   bool (*is_at_included_range_start)(const TSLexer *);
   bool (*eof)(const TSLexer *);
--- a/lib/src/lexer.c
+++ b/lib/src/lexer.c
@@ -245,6 +245,10 @@ static void ts_lexer__mark_end(TSLexer *_self) {
   self->token_end_position = self->current_position;
 }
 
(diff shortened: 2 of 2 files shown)


ahlinc:
Could you explain why the `get_column` doesn't work for your case?

llllvvuu:
That was the first thing I tried, but it resets to 0 so there was no way to know whether we are still in a comment. If I have `get_row`, that would work as well

llllvvuu:
I also found someone else asked for `get_row` / `get_line`: https://github.com/tree-sitter/tree-sitter/issues/1282, though not for runtime reasons.

So maybe `get_line` would be more broadly useful than `get_byte`. If `current_position.extent.row` works for this I'd be happy to update the PR or make a competing one.

wkt740:
Man I glad u messaged me back. I did it as a file that was accessible. I
haven’t figured it out and got frustrated so stopped trying lol. How well
are you at coding? And have you tried to AiGitbot?

llllvvuu:
`get_line` alternative here: https://github.com/tree-sitter/tree-sitter/pull/2670

wkt740:
You’re the man!!!!!   Thanks a lot!!!! If I need help u care if I message
you

PR #2670, opened 2023-09-27
feat(lexer)!: `TSLexer.get_line()`

llllvvuu:
Closes: https://github.com/tree-sitter/tree-sitter/issues/1282
Related: https://github.com/nvim-treesitter/nvim-treesitter/issues/4839

Currently, since external scanners are called on-demand at arbitrary
points, they are limited in context/state.

An example of where this leads to redundant computation is when
determining the effective indent of Python comments, which depends not
only on the comment's indentation but also the indentation of the next
non-comment line, which is an expensive lookahead (O(comment_size)). If
there are many consecutive comments (e.g. as generated by an IDE when
commenting out a block of code), then this computation is redundant, and
in total quadratic.

`TSLexer.get_line()` improves the statefulness of external scanners,
allowing to keep track of lookahead state, reducing redundancy.

`TSLexer.get_column()` is insufficient in such cases where state needs
to be reused over lines, not columns.

`serialize/deserialize` is also insufficient in cases where state would
be invalidated by internally scanned text.

see: https://github.com/llllvvuu/tree-sitter-python/commit/cae9dbe58df56627ad4af76f9ba5ab262932c1c5

BREAKING CHANGE: This breaks the `TSLexer` ABI. To minimize disruption,
it can be bundled with other breaking changes.

### DIFF
--- a/lib/include/tree_sitter/parser.h
+++ b/lib/include/tree_sitter/parser.h
@@ -44,6 +44,7 @@ struct TSLexer {
   TSSymbol result_symbol;
   void (*advance)(TSLexer *, bool);
   void (*mark_end)(TSLexer *);
+  uint32_t (*get_line)(const TSLexer *);
   uint32_t (*get_column)(TSLexer *);
   bool (*is_at_included_range_start)(const TSLexer *);
   bool (*eof)(const TSLexer *);
--- a/lib/src/lexer.c
+++ b/lib/src/lexer.c
@@ -245,6 +245,10 @@ static void ts_lexer__mark_end(TSLexer *_self) {
   self->token_end_position = self->current_position;
 }
 
+static uint32_t ts_lexer__get_line(const TSLexer *_self) {
+  return ((Lexer *)_self)->current_position.extent.row;
+}
+
 static uint32_t ts_lexer__get_column(TSLexer *_self) {
   Lexer *self = (Lexer *)_self;
 
@@ -292,6 +296,7 @@ void ts_lexer_init(Lexer *self) {
       // library.
       .advance = ts_lexer__advance,
       .mark_end = ts_lexer__mark_end,
+      .get_line = ts_lexer__get_line,
       .get_column = ts_lexer__get_column,
       .is_at_included_range_start = ts_lexer__is_at_included_range_start,
       .eof = ts_lexer__eof,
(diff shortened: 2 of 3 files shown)


lib/include/tree_sitter/parser.h
```diff
@@ -44,6 +44,7 @@ struct TSLexer {
   TSSymbol result_symbol;
   void (*advance)(TSLexer *, bool);
   void (*mark_end)(TSLexer *);
+  uint32_t* line;
```
llllvvuu, line 47: Note: changing this to a getter could be worth the indirection if in the future this field might get a dynamic address or be computed on-demand

lib/src/lexer.c
```diff
@@ -310,6 +310,7 @@ void ts_lexer_init(Lexer *self) {
     .included_range_count = 0,
     .current_included_range_index = 0,
   };
+  self->data.line = &self->current_position.extent.row;
```
ahlinc, line 313: It's not a good idea to share a pointer every time, it's not better than to make a copy of the value directly. Tree-sitter's scanner API provides methods for most cases to be maximally lazy.

lib/include/tree_sitter/parser.h
ahlinc: There are two types of objects in scanner's API, method pointers and values, this change doesn't cost to introduce a third type, a pointer to a value. Also to minimize breaking changes it makes sense to implement the final best solution at once.

llllvvuu: Yup, I realized the direct pointer was a mistake too, just changed it to a getter

lib/src/lexer.c
llllvvuu: Removed

maxbrunsfeld:
If the scanner's output can depend on the line number where it was executed, then it is no longer valid to re-use that token when re-parsing incrementally if the content has above has been edited (because the token may occur on a different line number than before).

That's why the `get_column` API is designed the way it is. If that function is called, the scanner has to record that its output now depends on the start column, and is much less re-usable than it would otherwise be.

IMO, this isn't a good solution to the problem you're referring to in the Python parser, because it introduces an API that is somewhat incompatible with incremental parsing.

llllvvuu:
Gonna have another look at the Python parser when I get the chance then, thanks!

PR #2671, opened 2023-09-27
fix: split conditions into isolated iterations

lavafroth:
This should fix the [regression](https://github.com/tree-sitter/tree-sitter/pull/2600#issuecomment-1736672575) for `position_for_offset` introduced in #2600.

### Changes
- The function first counts the trailing characters for the last line.
- Once it finds a newline, it stops checking for anything but newlines and keeps incrementing the row count.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -401,15 +401,17 @@ fn offset_for_position(input: &[u8], position: Point) -> usize {
 
 fn position_for_offset(input: &[u8], offset: usize) -> Point {
     let mut result = Point { row: 0, column: 0 };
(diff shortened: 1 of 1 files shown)


ahlinc:
It's often situation that assumed optimization doesn't work on a real hardware, I'd suggest to benchmark all speedup optimizations before opening a PR.

This PR is also is not the optimization:

Screenshot from 2023-09-28 05-52-27

lavafroth:
Hi, sorry. Just to make sure we are on the same page, are we using `cargo bench`, `./script/benchmark` or something else entirely for benchmarking? I'd appreciate if you could kindly share your benchmark setup. Thanks!

PR #2672, opened 2023-09-28
feat: Better time and stat

daumantas-kavolis-sensmetry:
Since test and example files are usually small, parse times often ended up displayed as `0 ms` which was not very useful. Times reported by `--time` option and in benchmarks now also show decimals. Also added speed display to `--time` option, average speed to `--stat` option and improved benchmark speed calculation. Previously it used a resolution of 1 ms which was far too coarse for smaller files and resulted in skewed results.

### DIFF
--- a/cli/benches/benchmark.rs
+++ b/cli/benches/benchmark.rs
@@ -200,9 +200,12 @@ fn parse(path: &Path, max_path_length: usize, mut action: impl FnMut(&[u8])) -> 
         action(&source_code);
     }
     let duration = time.elapsed() / (*REPETITION_COUNT as u32);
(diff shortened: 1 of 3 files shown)


ahlinc:
What is good with ms resolution is that it's easy visually visible all 1ms outliers.

daumantas-kavolis-sensmetry:
Right aligned times achieve the same result.

amaanq:
Thanks @daumantas-kavolis-sensmetry! I rebased it and added a newline before printing stats just to be easier to distinguish from the data when parsing multiple files with the quiet flag

daumantas-kavolis-sensmetry:
No worries, thanks.

PR #2714, opened 2023-10-19
fix "test_point_range_captures not working"

MrPrezident:
Fix for https://github.com/tree-sitter/py-tree-sitter/issues/105

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -4048,9 +4048,20 @@ bool ts_query_cursor_next_capture(
         continue;
       }
 
-      // Skip captures that precede the cursor's start byte.
       TSNode node = captures->contents[state->consumed_capture_count].node;
-      if (ts_node_end_byte(node) <= self->start_byte) {
+
+      bool node_precedes_range = (
+        ts_node_end_byte(node) <= self->start_byte ||
+        point_lte(ts_node_end_point(node), self->start_point)
+      );
+      bool node_follows_range = (
+        ts_node_start_byte(node) >= self->end_byte ||
+        point_gte(ts_node_start_point(node), self->end_point)
+      );
+      bool node_outside_of_range = node_precedes_range || node_follows_range;
+
+      // Skip captures that are outside of the cursor's range.
+      if (node_outside_of_range) {
         state->consumed_capture_count++;
         continue;
       }
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -2106,6 +2106,73 @@ fn test_query_captures_within_byte_range() {
 }
 
 #[test]
+fn test_query_cursor_next_capture_with_byte_range() {
+    allocations::record(|| {
+        let language = get_language("python");
+        let query = Query::new(
+            language,
+            "(function_definition name: (identifier) @function)
+             (attribute attribute: (identifier) @property)
+             ((identifier) @variable)",
+        )
+        .unwrap();
+
+        let source = "def func():\n  foo.bar.baz()\n";
+        //            ^            ^    ^          ^
+        // byte_pos   0           12    17        27
+        // point_pos (0,0)      (1,0)  (1,5)    (1,15)
+
+        let mut parser = Parser::new();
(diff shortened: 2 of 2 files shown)


lib/src/query.c
```diff
@@ -4050,7 +4050,10 @@ bool ts_query_cursor_next_capture(
 
       // Skip captures that precede the cursor's start byte.
       TSNode node = captures->contents[state->consumed_capture_count].node;
-      if (ts_node_end_byte(node) <= self->start_byte) {
+      if (
+        ts_node_end_byte(node) <= self->start_byte ||
+        point_lte(ts_node_end_point(node), self->start_point)
```
maxbrunsfeld, line 4055: I think this should be an **and**, not an **or**. To be counted, a node must be within the supplied byte range **and** within any supplied point range.

MrPrezident: Thanks @maxbrunsfeld for the response! I think this should be **or** not **and** since this section of code is supposed to filter out nodes, not include them. When I run the debugger on a query that was specified with a point range, I see that the start/end bytes look like this:
```
(gdb) p self->start_byte
$29 = 0
(gdb) p self->end_byte
$30 = 4294967295
```
whereas the start/end points look like this:
```
(gdb) p self->start_point
$27 = {row = 9, column = 3}
(gdb) p self->end_point
$28 = {row = 9, column = 27}

```

So in this case, the node will always be inside the byte range, but not the point range, and so we need to **or** them together.

That being said, when I went over this again to see if I was missing anything, I realized that there are also some nodes that are not getting filtered out beyond the end of the range as well. So we probably need to do something like this. Let me know what you think. I'll test this out on my end and update the PR.

```python
      TSNode node = captures->contents[state->consumed_capture_count].node;
      
      bool node_precedes_range = (
        ts_node_end_byte(node) <= self->start_byte ||
        point_lte(ts_node_end_point(node), self->start_point)
      );
      bool node_follows_range = (
        ts_node_start_byte(node) >= self->end_byte ||
        point_gte(ts_node_start_point(node), self->end_point)
      );
      bool node_outside_of_range = node_precedes_range || node_follows_range;

      // Skip captures that are outside of the cursor's range.
      if (node_outside_of_range) {
        state->consumed_capture_count++;
        continue;
      }
```

MrPrezident: Alright tested out the new code to skip both preceding and following ranges. This works great! I have updated the PR branch.

MrPrezident: @maxbrunsfeld does this updated PR look ok to you?

maxbrunsfeld: Ah yes; you're quite right. Thanks for the explanation, and for making that code a bit clearer.

Would you be willing to add a test for this in `query_test.rs`?

MrPrezident: Sure. Will do.

amaanq (approved):

amaanq:
@MrPrezident run rustfmt on the files causing an error in CI

lib/src/query.c
MrPrezident: @maxbrunsfeld I added two unit tests. One for testing byte_range and one for testing point_range. Without my fix, the point_range test fails with extra captures both following and proceeding the point_range. The byte_range test fails with extra captures following the byte_range.

point_range fail:
> thread 'tests::query_test::test_query_cursor_next_capture_with_point_range' panicked at 'assertion failed: `(left == right)`
  left: `[("function", "func"), ("variable", "foo"), ("property", "bar"), ("property", "baz")]`,
 right: `[("variable", "foo")]`', cli/src/tests/query_test.rs:2168:9

byte_range fail:
> thread 'tests::query_test::test_query_cursor_next_capture_with_byte_range' panicked at 'assertion failed: `(left == right)`
  left: `[("variable", "foo"), ("property", "bar"), ("property", "baz")]`,
 right: `[("variable", "foo")]`', cli/src/tests/query_test.rs:2135:9

maxbrunsfeld:
Thank you!

PR #2724, opened 2023-10-24
cli - update to regex-syntax 0.8

plugwash:
Hi, I am currently investigating upgrading the regex crates in Debian, while we do have a mechanism for packaging multiple versions of a crate in paralell it is something we try to avoid where possible. Regex itself is not semver-breaking, but regex-syntax is. 

As a result I prepared a patch to make tree-sitter-cli build with the new regex-syntax.

regex-syntax has restructured the Ast enum a bit, rather than having a Class member it now has seperate members for different types of Class.

I get the same number of test failures with and without this change.

Any feedback would be appreciated.


dundargoc:
@amaanq the actual changes are very small if you ignore whitespace changes. Looks good to me tbh.

dundargoc (approved):

amaanq:
yeah its good

PR #2725, opened 2023-10-25
Fix `goto_previous_sibling` with padded first children

### LINKED ISSUE #2723: `goto_previous_sibling()` seems to create incorrect indices sometimes
I am having issues with `goto_previous_sibling()` (added in https://github.com/tree-sitter/tree-sitter/pull/2324) where it seems like using `utf8_text()` on the resulting node sometimes points to the wrong underlying bytes.

The easiest way to test this in a reproducible way for you was to fork and add a few "failing tests"
https://github.com/tree-sitter/tree-sitter/compare/master...DavisVaughan:tree-sitter:test/goto-previous-sibling

Snippet copied here for reference:

```rust
    let text = "
    // Hi there
    // This is fun!
    // Another one!
";
    let tree = parser
      .parse(text, None)
      .unwrap();

    let mut cursor = tree.walk();
    assert_eq!(cursor.node().kind(), "source_file");

    assert!(cursor.goto_last_child());
    assert_eq!(cursor.node().kind(), "line_comment");
    assert_eq!(cursor.node().utf8_text(text.as_bytes()).unwrap(), "// Another one!");

    assert!(cursor.goto_previous_sibling());
    assert_eq!(cursor.node().kind(), "line_comment");
    assert_eq!(cursor.node().utf8_text(text.as_bytes()).unwrap(), "// This is fun!");
```

If you run this, the last assertion fails with:

```
thread 'tests::tree_test::test_tree_cursor' panicked at 'assertion failed: `(left == right)`
  left: `"is is fun!\n\n   "`,
 right: `"// This is fun!"`', cli/src/tests/tree_test.rs:400:5
```

Notably I tried the same kind of thing with `node.prev_sibling()` and it seemed like it worked, so this is possibly a cursor only bug

daumantas-kavolis-sensmetry:
Fixes #2723

### DIFF
--- a/cli/src/tests/tree_test.rs
+++ b/cli/src/tests/tree_test.rs
@@ -393,15 +393,24 @@ fn test_tree_cursor_previous_sibling() {
 
     assert!(cursor.goto_last_child());
     assert_eq!(cursor.node().kind(), "line_comment");
-    assert_eq!(cursor.node().utf8_text(text.as_bytes()).unwrap(), "// Another one!");
+    assert_eq!(
+        cursor.node().utf8_text(text.as_bytes()).unwrap(),
+        "// Another one!"
+    );
 
     assert!(cursor.goto_previous_sibling());
     assert_eq!(cursor.node().kind(), "line_comment");
-    assert_eq!(cursor.node().utf8_text(text.as_bytes()).unwrap(), "// This is fun!");
+    assert_eq!(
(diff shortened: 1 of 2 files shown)


maxbrunsfeld (approved):

maxbrunsfeld:
Thanks for the fix @daumantas-kavolis-sensmetry

PR #2741, opened 2023-11-02
feat!: compare max dynamic precedence level first

### LINKED ISSUE #2733: reset child's dynamic precedence level
I'm making a tree-sitter parser for [Norg](https://github.com/nvim-neorg/norg-specs), and I'm having problem with conflict case described below.

I read official document several times, tried a lot of variations for months, but can't find a way to accomplish what I want.

# Problem

```
*/`word/*`
```

Syntactically this can be **bold containing italic with following punctuation** or **verbatim with preceding punctuations**.
But as verbatim has higher precedence level than bold/italic in Norg spec, this should be parsed as second one.

As this does not seem to be solvable in parser generation time, I give precedence level 2 to verbatim and 1 to both bold and italic.

But this doesn't work because **paragraph with bold**'s precedence level becomes 2, not 1 as it sums all descendant's precedence levels.

Simply giving more high precedence level to verbatim doesn't works because of following case

```
*/`word/ /word/ /word/*`
```

Now **paragraph with bold**'s precedence level becomes 4.
So theoretically dynamic precedence level can grow up endlessly.

# Solution

I suggest to have something like `prec.reset()` to allow parent node not contain child's precedence level.

So in cases above, only bold's precedence level will be compared to verbatim's, and italic's will be considered inside the bold (inside the `prec.reset()` function.)

---

I apologize if my explanation is unclear. I have bad English skill.

boltlessengineer:
Fixes #2733

When calculating dynamic precedence levels, tree-sitter sums up all child nodes' dynamic precedence level for parent node's. This leads issues described from #2733.

Dynamic precedences should not be summed to avoid this issue.

This PR changes parser behavior to give parent node a highest precedence level of child nodes, not sum of those.
And add extra logic to compare that _max dynamic precedence level_ first, and then compare the summed levels.

I tried my best to fix without changing previous behavior too much, but it eventually became breaking change.

```
failures:
    tests::corpus_test::test_corpus_for_c
    tests::corpus_test::test_corpus_for_cpp
    tests::corpus_test::test_corpus_for_go
    tests::corpus_test::test_corpus_for_javascript
    tests::corpus_test::test_corpus_for_python

test result: FAILED. 258 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out; finished in 19.58s
```

- tree-sitter-c
- tree-sitter-cpp
- tree-sitter-go
- tree-sitter-javascript
- tree-sitter-python

These 5 parsers are failing right now.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -917,7 +917,8 @@ static StackVersion ts_parser__reduce(
     } else {
       parent.ptr->parse_state = state;
     }
-    if (dynamic_precedence > parent.ptr->dynamic_precedence) {
+    if ((dynamic_precedence > parent.ptr->dynamic_precedence && dynamic_precedence != 0)
+    || (dynamic_precedence < 0 && parent.ptr->dynamic_precedence == 0)) {
       parent.ptr->dynamic_precedence = dynamic_precedence;
     }
     parent.ptr->dynamic_precedence_sum += dynamic_precedence;
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -745,6 +745,20 @@ static bool ts_parser__select_tree(TSParser *self, Subtree left, Subtree right) 
     return false;
   }
 
+  if (ts_subtree_dynamic_precedence_sum(right) > ts_subtree_dynamic_precedence_sum(left)) {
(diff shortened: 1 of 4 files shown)


ObserverOfTime:
Breaking (at least) 5 parsers to fix 1 is not really something we want.
It would be preferable to implement named dynamic precedence as mentioned in #2733.

PR #2167, opened 2023-04-04
CICD: new workflow

ahlinc:
_This is a full re-implementation for CI._

**New approach for releases:**
* Increment version in `cli/Cargo.toml`.
* Create a branch with prefix `release/v` like for ex. `release/v0.20.8`.
* Open a PR for the branch.
* If CI part would be completed then a `release.yml` workflow will create:
  * Create a tag for release.
  * A new [_Release_](https://github.com/tree-sitter/tree-sitter/releases) for the tag.
  * Merge the openned PR.

**TODO:**
- [ ] Relpace `.github/scripts` for `.github/actions`.
- [ ] Implement TODOs in `.github/workflows/publish.yml`.
- [x] Completely remove `appveyor.yml` stub file.

**Questions:**
- [ ] How to release changes in non CLI Rust crates or in the C library?

PR #2746, opened 2023-11-03
cicd: dynamic matrix construction + filter targets

ahlinc:
Workaround for an issue mentioned in: https://github.com/tree-sitter/tree-sitter/pull/1864#issuecomment-1792264724

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,18 +12,18 @@ on:
         default: ${{ github.ref }}
         type: string
       run-tests:
-        default: true
         type: boolean
+        default: true
   workflow_dispatch:
     inputs:
       run-tests:
         description: Run tests
-        default: true
         type: boolean
+        default: true
       rust-test-threads:
         description: Number of Rust test threads
-        default: ""
         type: string
+        default: ""
 
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


maxbrunsfeld:
@ahlinc Could you fill me on the overall design of the new CI system?

I see that back in https://github.com/tree-sitter/tree-sitter/pull/2167, we moved to a new releasing workflow, which is great. But since that PR landed, it seems like a fair amount of additional complexity has been added to the CI. I just want to make sure that the CI doesn't get more complex than it needs to be.

Why do we need to build for tier-2 and tier-3 targets on CI at all?

amaanq:
> Why do we need to build for tier-2 and tier-3 targets on CI at all?

I added most of the bizarre ones for coverage (re: that big endian query bug from a few months ago), they don't run the wasm tests, and don't produce binaries on release (I believe)

Just one tier 1 big endian target can suffice for coverage if mips is causing an issue

ahlinc:
@maxbrunsfeld I plan to write a `MAINTAINING.md` howto with describing all actions for the release process.

Briefly:
- I've introduced the `tier` parameter as an our testing tier level to easily mark targets that have different errors and separate them to different scopes instead of commenting out all related lines in case of issues.
- We don't plan to release CLI for all targets but different environments like Android, Raspberry Pi would be awesome to cover.
- Now all additional targets would be disabled by default but I saved a possibility to easily enable them by adding a `-all-tiers` suffix to a branch name or using a parameter in a manual action run window.

maxbrunsfeld:
Thanks for the context. I would support eliminating some of the targets, and leaving just one big-endian architecture; I don't have a strong opinion on which.

@ahlinc could you explain the different dispatch modes of the workflow? Could we eliminate some of the conditional logic in the YAML, and just have a fixed list of targets, or are there reasons we need to make it more dynamic?

ahlinc:
@maxbrunsfeld I thought to add dynamic construction of the targets list a long time ago to minimize an amount of parallel tasks that are trying to start because it easy to saturate a queue of running tasks (60 in Tree-sitter repo and 20 in forks) by updating several PRs. So having minimal amount of task produces less contention between tasks.

> could you explain the different dispatch modes of the workflow?

If understand you correctly, the `workflow_call` is used to make the build workflow callable from the main CICD entrypoint workflow, and the `workflow_dispatch` allows to run the build workflow manually, I use it from time to time if I don't want to push a branch to trigger CI run.

By tier 2 I marked all targets that I wouldn't like to run by default because from time to time a related QEMU emulators may stuck or show a poor performance. With the tier 3 I marked a target that have a hard issue so for now should be excluded. For what would go to release I thought to use only targets that don't have the tier label that what meens the 1. But we for example may introduce that tiers 1 and 2 would be for the release but on all PRs only the 1 tier would run by default to minimize and amount of work and time for a one workflow run.

I'd like to keep the full list to don't lost that details that needs to use corresponding targets, but in any way the info would be available in commits if we would remove some targets entirely because they are very similar to others.

@amaanq could you write a list what targets you think we could remove?

Also I plan to add a matrix of tests for different compilers like: tcc (Tiny C), mingw, cl.exe.

ahlinc:
Unfortunately also there are another unresolved issues that disturb CI stability:
* Grammar fixtures use master branches what randomly affects unrelated in progress PRs. I'd like to replace this approach to use latest commits and update them on a weekly basis in a separate CI task.
* Grammars loader doesn't properly [handle](https://github.com/tree-sitter/tree-sitter/pull/2647) concurrent compiler calls.

maxbrunsfeld:
> Add a matrix of tests for different compilers like: tcc (Tiny C), mingw, cl.exe.

I don't think we should add more builds for these compilers. Every job in our workflow has a pretty high cost, in my opinion, because they all fail from time to time for various reasons, and they slow down the feedback loop when working on features.

In my opinion, we should have:

* mac aarch64
* mac x86_64
* linux x86_64
* linux aarch64
* windows x86_64
* window x86

> Restore a CI cache behavior, it needs to replace a emsdk action with an own script because it doesn't maintained well and can't be properly used with the standard cache action

👍 This would be great.

> Grammar fixtures use master branches what randomly affects unrelated in progress PRs. I'd like to replace this approach to use latest commits and update them on a weekly basis in a separate CI task

👍  Love this idea.

> Grammars loader doesn't properly https://github.com/tree-sitter/tree-sitter/pull/2647 concurrent compiler calls

For some reason, I never see this come up as a problem when running tests. Has this caused many CI failures?

ahlinc:
> I don't think we should add more builds for these compilers.

Such compilers we thought to use only for C library because for the next benefits:
* It's cheap to compile Tree-sitter core C library and in parallel such tasks wouldn't be most long running tasks.
* `cl.exe` compiler is a standard compiler on Windows platform and when I tried to setup it CI it uncovers that there around 30-50 meaningful advises that makes sense to apply to the C code to make it more reliable. It's a benefit for Tree-sitter project to have it as an additional linting tool for C code in the CI flow.
* We have MinGW users so it makes sense to don't accidentally break a good experience for them from using Tree-sitter project.
* A Tiny C compiler is mentioned in Tree-sitter's C code and I suppose it may be used in some areas.
* Also it makes sense to introduce some guaranties about oldest supported compiler versions for the C code and check that in CI.

@maxbrunsfeld  I don't want to run all such checks in one giant CI flow that would took hours for execution.
My goal is to return a fun in development process so _I want to lower a time_ for commit checks in PRs branches _to 1-2 minutes_ and not more. For this I thought to introduce and test an approach with a dynamic CI matrix construction. Maybe with doing such matrix construction in clear Rust code instead of a cryptic `jq` / `yq` scripts.
Also I'm thinking about to introduce an [xtask](https://github.com/matklad/cargo-xtask) [approach](https://github.com/rust-lang/rust-analyzer/tree/master/xtask/src) to make all scripts in the `scripts/` folder fully portable and do many processes in parallel also it possible to smartly skip many compilation steps if to use git facilities to detect changes in code and build an own tool that would choice what parts of CI to run and what thing may be taken from CI cache.

At a current step I'd like to accumulate and don't fully drop things that we already have in CI and decide how to use all of that later.

_As for now I would like to merge this PR with small adjustments in targets that run by default and move forward with a smarter approach to CI flow._

ahlinc:
Another thought that about dynamic Github matrix construction is a possibility to a have a project adapted config of all actions and steps in clear and concise yaml file instead of inventing ways how to achieve needed goals with a current limited configuration possibilities of Github actions. Github actions provide a good environment for running CI tasks but configuration facilities aren't so good yet. Also it would allow to smarter run CI tasks with less duplication when a PR branch is pushed to the origin repo directly (Github runs a PR flow on an improvised merge commit to a trunk branch).

ahlinc:
My thoughts above look muddled because even the current solution in the CI is not what I would like to achieve, but a sufficiently good-quality one also needs a change in approaches.

ahlinc:
> Could you fill me on the overall design of the new CI system?

@maxbrunsfeld Let me try to describe by what thoughts I was guided and what I planned to do but not finished yet.

**Thoughts:**
1. A moment of appearance of _a release tag_ on the master branch is a very important thing because it's a very whose idea to do something with it even if CI would fail after the tag.
2. The best idea is to make an another release after a failed one but there is no guaranty that during a complex release process there everything would be fine without any new issues. What would lead to a one another broken tag.
3. If to left a broken tag there would be misunderstanding why there are no released packages in repos for it, what increases a mess up and misunderstanding.
4. Tree-setter has a huge amount of downstreams like various Linux distros and other big projects, all of them use different strategies and approaches for how they consume Tree-sitter source code and packages. If to play in a mess up game with a new releases after a one broken tag there is a possibility that some one will consume a code with a broken tag and wouldn't recheck that there appeared a one another version right after it.
5. As a **reliable approach** I see only a one way that the _tag appearance moment_ should be the guaranty that Tree-sitter release is fully good. What means that the tag for sure should appear only after all correctly passed CI checks. In this case it's not handy to manage the release tag manually and it should be an automatic result for all passed previous steps. After the tag appearance moment should happen only tasks that may be manually recovered like uploading artifacts to various package repos, but even this tasks may be placed before the tag if there is a possibility to rollback such actions or at least mark packages like yanked in the crates.io repository.

**Current CI approach** _(what is done)_:
1. It expects that releases would be managed by related PRs.
    - Release branches should be in a format like `release/v0.21.0` and the part after the `release/` will be the release tag name.
    - Release PR may be opened in a draft mode what allows to rebase it on top of the master or add more fixes until CI would be passed fine.
3. Parallel fast checks of _code compilation issues + Rust formatting checks_ to fail fast the whole CI flow.
4. Parallel "Build & Tests" flow for a broad list of architectures, to prove the state that Tree-sitter library is very portable and this is verified as much as we can.
5. Automatic release tag setting after passed of all tests.
6. Automatic Github release flow.
    - _**Not implemented** yet a posting of changes description._
7. Automatic release artifact packages construction and uploading in all repos where Tree-sitter is present.
   - _This is **not implemented** yet because a lot of questions appeared..._

**Problems with current unfinished approach:**
1. It's not flexible to manage a next release tag as a part of a release branch name because if there was a mistake with it it would require to reopen a new PR with a new branch name.
2. How to manage a description of release changes?
3. How to know for what changes parts of the project a release artifact packages should be constructed and uploaded to the repos? How to do this in an automatic CI/CD flow?
4. How to test more things that related to the project portability and don't increase a time for CI workflows for a day to day development?

For all of the above questions I have answers but that requires more time for development and to extend all of this automatisation on other project parts like the plaground and grammar repos to keep everything in sync and at a high quality level.

clason:
Don't let the perfect be the enemy of the good. Downstream is _really_ hurting currently from the lack of releases, even imperfect ones.

maxbrunsfeld:
@ahlinc Your plan of using a pull requests for initiating a new release sounds like  a good idea.

> It's cheap to compile Tree-sitter core C library and in parallel such tasks wouldn't be most long running tasks

Are you thinking of creating a separate workflow that just compiles the C library? I'm open to that, but since C compilers are so lenient, I worry that if we don't run the tests, we're not fully verifying that the library works with that compiler, so it may not be worth the additional CI complexity.

> `cl.exe` compiler is a standard compiler on Windows platform and when I tried to setup it CI it uncovers that there around 30-50 meaningful advises that makes sense to apply to the C code to make it more reliable.

Correct me if I'm wrong, but I believe `cl.exe` is what we're already using to compile the library on windows, as part of tree-sitter's `build.rs.` script, when building the rust crate. The other day, when working on https://github.com/tree-sitter/tree-sitter/pull/1864, I fixed some C errors that were only present on `cl.exe`.

maxbrunsfeld:
> I'd like to keep the full list to don't lost that details that needs to use corresponding targets, but in any way the info would be available in commits if we would remove some targets entirely because they are very similar to others.

> As for now I would like to merge this PR with small adjustments in targets that run by default and move forward with a smarter approach to CI flow

How about this compromise:

We completely remove any tier 3 targets. I don't think it's worth even mentioning tier 3 in our CI.

Instead of a numeric `tier` property on the matrix rows, we just have a **boolean** property called `ci`, which is `true` for the core targets, and `false` for the other ones. I think this would simplify our dynamic YAML-filtering logic.

During the main CI workflow, we would only run on the platforms where `ci: true`. You could keep the manual dispatch (in order to build for obscure targets), but it wouldn't take `tier` parameter. You could either:
* take a `platform` string, so that you could trigger a run for a particular obscure platform
* just run all platforms (including the obscure ones)

ahlinc:
> We completely remove any tier 3 targets. I don't think it's worth even mentioning tier 3 in our CI.

I did it as for now to move forward with this PR and move further with CI improvements.

> Instead of a numeric tier property on the matrix rows, we just have a boolean property called ci, which is true for the tier-1 targets, and false for the obscure targets. I think this would simplify the YAML-filtering logic that you added in this PR.

An idea with tier (or test levels) was in that that there are different sort of issues there are intermittent issues for some targets that are annoying but push-able to complete with CI restarts and there are hard issues for some targets that became broken at some point due to changes in third-party things. I thought at least to separate such things to different tiers to make it flexible of how to operate with different sorts of CI failures.

Another idea was to detect that a PR is in the draft mode and for such mode run a minimal subset of CI to make it complete faster and make the work with it more comfortable during development process and allow to postpone fixing any CI passing issues on the final stage when a PR would be marked as ready for review. For there tier like 0  for drafts and 1 for ready PRs would be useful to implement such logic.

ahlinc:
>  I really feel that all of this is too much.

I have the same feeling but I don't know simpler way except to make CI clearly specified but very smart enough to manage all details. With the same feeling I started the work on rewriting CI from previous implementation because on that time I found that I can't fix it at all, to really tests more things for sure and to automate more processes.

> for how much code we're maintaining

There is no enormous amount of code but even in this code there are a lot of tiny details that shouldn't be broken or changed without strict reasons, because there are a lot of downstream projects that expect stability and clear information about breaking changes. For similar reasons CMake support was rejected to don't get here even more details and complexity. I'd like to see all detail changes like version numbers in different places would be fully automated and controlled by CI so that at some moment to stop think about this at all and just know where to look on if something goes wrong.

> For the foreseeable future, we cannot be spending time worrying about tier 2 targets at all.

Agree, but I respect the work that @amaanq spent on this experiment for adding more target and we got a benefit from it by finding a subtle bugs related for the project portability and to be portable on big-endian target. Before adding possibility of how to build and test for such targets such issues were hidden for this project for years. I'm for example would like to see latest Tree-sitter CLI easily available for Android targets because from time to time I use it in Termux on my phone and there is a build of Tree-sitter in Termux packages repo but it isn't fresh and as Tree-sitter CLI isn't well tested for building for that environment that makes it not possible to simply build a fresh CLI with a one `cargo build --release` command or take a compatible version from a latest CI artifacts.

> I'm trying to make progress on features in small scraps of available time

The same thing, I don't like to manage CI at all, I want to see it fully automated that I can forget about it existence at all on daily basis and look on it just to identify mistakes in the projects code. I would better spend the time on new ideas for Tree-sitter but a set of boring issues that are mostly solved now didn't allowed to easily do the main development and experimenting. Also I would like just know that if I wan't to make a new release I just need to know to open a release PR, verify and correct a pre generated changelog skel and allow CI to fully complete and finish the release process automatically and in _a reliable way_.

> I'm getting bogged down because of the complexity of the CI

I also don't like how CI is organized there was a lot of time spent on thinking how to do a pretty simple things but with Github CI yaml files and facilities. Github CI is good thing but built with a thought in mind to make it common for everyone but at the same time it's hard to solve there a very project specific things. For example the dynamic matrix idea is cool but in our case it looks too complex and defined in a hacky way due to its restrictions. Another example there is a possibility to write actions for complex steps but many actions are with a poor quality inside or aren't well maintainable and it's a hard task to compose them with a project specific needs.

---

To finish automation of the release process I need to add more related logic in CI. I discovered many projects that do automation of the release and versioning process but don't found any that wouldn't break how that is doing here for years. Without dictated approaches of authors of such projects and just with easy adoption and with a help in automation in existing project specific flows.

For the automated release process in Tree-sitter I don't want to introduce an approach with a one common version for all components because this would lead to pointless version bumps for many components. I have an idea for this and will somehow describe it in separate proposal issue.

The most complex CI logic that would manage a lot of tiny project specific details I would like to write in Rust and I'm not kidding, the Rust language makes me more sure that I'm able to write a more reliable code with it than with Javascript or Typescript like it's doing in most of GH actions and with it I can easily parallelize everything what I thought to speed up CI processes.
I would organize such logic in a way that there is a lib of clear functions that do some steps and a separate file(s) where a high level CI logic described with such function in a clear and condensed way and with proper comments.
Like it was done [there](https://github.com/tree-sitter/tree-sitter/blob/3b0159d25559b603af566ade3c83d930bf466db1/cli/src/generate/binding_files.rs#L26-L39) and [there](https://github.com/amaanq/tree-sitter/blob/update-bindings/cli/src/generate/grammar_files.rs#L243-L279).

ahlinc:
The last [3 attempts](https://github.com/tree-sitter/tree-sitter/actions/runs/7000713143/job/19046563078?pr=2746#step:16:430) to complete CI for this PR suffer from the one issue related to #2647. I'll [re-implement](https://github.com/tree-sitter/tree-sitter/pull/2647#issuecomment-1793496915) that PR and merge a fix to push forward this one too.

dundargoc:
Unsolicited suggestion from passerby: I don't think the release workflow needs or should be tied to the main testing workflow at all. The release commit itself doesn't need to have code changes, and you should be fairly certain that the master branch passes CI since the changes should have passed the CI during a PR. I just don't think the benefit of testing a tag before setting it is worth the complexity it introduces.

If things do go south for some reason you can just create a new tag/release. It's not really a big deal.

PR #2790, opened 2023-11-28
fix: install only `include/tree_sitter/api.h` file

### LINKED ISSUE #2677: Makefile does not install `parser.h` but `npm install` and `cargo install` do
I was digging through the various install methods and the `Makefile` was recently changed so that it did not install `parser.h` but all the other install methods still include them in their various installs. So, for consistency those should probably be made to do the same.

ahlinc:
* Fixes #2677
* Related #2573
* Related https://github.com/tree-sitter/tree-sitter-bash/issues/199#issuecomment-1694416505

This PR moves the `parser.h` header from the `include` folder to the `src` folder:
```
lib/include/tree_sitter/parser.h -> lib/src/parser.h
```
The reasons for this:
* The `parser.h` file is actually not an API file for the Tree-sitter core library.
* There is a `lib/src/parser.c` and the `parser.h` header file is more related there among other header files.
* It's included in CLI as raw bytes and CLI puts it into right place in every grammar directory: `src/tree_sitter/parser.h`
* There was confusion from library users https://github.com/tree-sitter/tree-sitter-bash/issues/199#issuecomment-1694404449 and its definitely the `parser.h` header file shouldn't be shipped with Tree-sitter's core library by `make install` or in `tree-sitter` crate or `tree-sitter` npm package.

Also to avoid mistakes all external scanner files should use C include directive with double quotes to always refer the `parser.h` header file provided by CLI on the parser generation call:
```c
```


maxbrunsfeld (approved):
I agree, this is clearer.

aryx:
Can we revert this? Semgrep needs the tree-sitter.h to compile semgrep.
This prevents us from installing semgrep via homebrew here: https://github.com/semgrep/semgrep/issues/9693

PR #2794, opened 2023-11-28
ci: remove duplicate `cargo fmt`

dundargoc:


ahlinc:
Thanks, but the idea is to fail faster on the all running in parallel workflows.

dundargoc:
What parallel workflows? `full_rust_checks` depends on `fast_checks`.

ahlinc:
Sorry, I forget that I've made both of them dependent from the `fast_checks`. But the MSRV workflow is uncompleted for now and currently just driven by `cargo fmt` that you tried to remove. In any case it doesn't make sense for now to improve current CI implementation because current design doesn't work well and will be rewritten soon.

PR #2795, opened 2023-11-28
fix: Handle quantified predicates on `web-tree-sitter` properly

savetheclocktower:
#2532 added new `any-` versions of the `eq?` and `match?` predicates, plus their negations. But that PR didn't quite implement them properly on the web-tree-sitter side. (Test cases were added for the Rust bindings, but not for web-tree-sitter.)

I managed to untangle it and write some tests. I'll annotate below with some explanations of my choices.

### DIFF
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -842,78 +842,90 @@ class Language {
           const operator = steps[0].value;
           let isPositive = true;
           let matchAll = true;
+          let captureName;
           switch (operator) {
             case 'any-not-eq?':
-              isPositive = false;
-              matchAll = false;
-            case 'any-eq?':
-              matchAll = false;
             case 'not-eq?':
               isPositive = false;
+            case 'any-eq?':
             case 'eq?':
               if (steps.length !== 3) throw new Error(
-                `Wrong number of arguments to \`#eq?\` predicate. Expected 2, got ${steps.length - 1}`
+                `Wrong number of arguments to \`#${operator}\` predicate. Expected 2, got ${steps.length - 1}`
               );
               if (steps[1].type !== 'capture') throw new Error(
-                `First argument of \`#eq?\` predicate must be a capture. Got "${steps[1].value}"`
+                `First argument of \`#${operator}\` predicate must be a capture. Got "${steps[1].value}"`
               );
+              let eqMatchAll = !operator.startsWith('any-');
+              let eqIsPositive = isPositive;
               if (steps[2].type === 'capture') {
                 const captureName1 = steps[1].name;
                 const captureName2 = steps[2].name;
-                textPredicates[i].push(function(captures) {
+                textPredicates[i].push(function (captures) {
                   let nodes_1 = [];
                   let nodes_2 = [];
                   for (const c of captures) {
                     if (c.name === captureName1) nodes_1.push(c.node);
                     if (c.name === captureName2) nodes_2.push(c.node);
                   }
-                  return matchAll
-                    ? nodes_1.every(n1 => nodes_2.some(n2 => n1.text === n2.text)) === isPositive
-                    : nodes_1.some(n1 => nodes_2.some(n2 => n1.text === n2.text)) === isPositive;
+                  let compare = (n1, n2, positive) => {
+                    return positive ?
+                      n1.text === n2.text :
+                      n1.text !== n2.text;
+                  };
+                  return eqMatchAll
+                    ? nodes_1.every(n1 => nodes_2.some(n2 => compare(n1, n2, eqIsPositive)))
+                    : nodes_1.some(n1 => nodes_2.some(n2 => compare(n1, n2, eqIsPositive)));
                 });
               } else {
-                const captureName = steps[1].name;
+                captureName = steps[1].name;
                 const stringValue = steps[2].value;
-                textPredicates[i].push(function(captures) {
+                let matches = (n) => n.text === stringValue;
+                let doesNotMatch = (n) => n.text !== stringValue;
+                textPredicates[i].push(function (captures) {
                   let nodes = [];
                   for (const c of captures) {
                     if (c.name === captureName) nodes.push(c.node);
                   }
-                  return matchAll
-                    ? nodes.every(n => n.text === stringValue) === isPositive
-                    : nodes.some(n => n.text === stringValue) === isPositive;
+                  let test = eqIsPositive ? matches : doesNotMatch;
+                  return eqMatchAll
+                    ? nodes.every(test)
+                    : nodes.some(test);
                 });
               }
               break;
 
-            case 'not-any-match?':
-              isPositive = false;
-              matchAll = false;
-            case 'any-match?':
-              matchAll = false;
-            case 'not-match?':
-              isPositive = false;
-            case 'match?':
+            case "any-not-match?":
+            case "not-match?":
+                isPositive = false;
+            case "any-match?":
+            case "match?":
               if (steps.length !== 3) throw new Error(
-                `Wrong number of arguments to \`#match?\` predicate. Expected 2, got ${steps.length - 1}.`
+                `Wrong number of arguments to \`#${operator}\` predicate. Expected 2, got ${steps.length - 1}.`
               );
               if (steps[1].type !== 'capture') throw new Error(
-                `First argument of \`#match?\` predicate must be a capture. Got "${steps[1].value}".`
+                `First argument of \`#${operator}\` predicate must be a capture. Got "${steps[1].value}".`
               );
               if (steps[2].type !== 'string') throw new Error(
-                `Second argument of \`#match?\` predicate must be a string. Got @${steps[2].value}.`
+                `Second argument of \`#${operator}\` predicate must be a string. Got @${steps[2].value}.`
               );
-              const captureName = steps[1].name;
+              captureName = steps[1].name;
               const regex = new RegExp(steps[2].value);
-              textPredicates[i].push(function(captures) {
+              let matchIsPositive = isPositive;
+              let matchMatchAll = !operator.startsWith('any-');
+              textPredicates[i].push(function (captures) {
                 const nodes = [];
                 for (const c of captures) {
                   if (c.name === captureName) nodes.push(c.node.text);
                 }
-                if (nodes.length === 0) return !isPositive;
-                return matchAll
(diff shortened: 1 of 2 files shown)


lib/binding_web/binding.js
```diff
@@ -842,78 +842,90 @@ class Language {
           const operator = steps[0].value;
           let isPositive = true;
           let matchAll = true;
+          let captureName;
           switch (operator) {
             case 'any-not-eq?':
-              isPositive = false;
```
savetheclocktower, line 847: The `switch` statement fallthrough trick fails to work when there's a 2x2 matrix of flags — there's no way to structure them so that they don't get in each other's way. Instead, I check for the presence of `any-` at the start of `operator`.

lib/binding_web/binding.js
```diff
@@ -842,78 +842,90 @@ class Language {
...
               if (steps.length !== 3) throw new Error(
-                `Wrong number of arguments to \`#eq?\` predicate. Expected 2, got ${steps.length - 1}`
+                `Wrong number of arguments to \`#${operator}\` predicate. Expected 2, got ${steps.length - 1}`
               );
               if (steps[1].type !== 'capture') throw new Error(
-                `First argument of \`#eq?\` predicate must be a capture. Got "${steps[1].value}"`
+                `First argument of \`#${operator}\` predicate must be a capture. Got "${steps[1].value}"`
               );
+              let eqMatchAll = !operator.startsWith('any-');
+              let eqIsPositive = isPositive;
```
savetheclocktower, line 859: You may wonder what the point of this is. It makes sense in my head, but I haven't quite come up with the test case that would reveal why this is necessary:

`isPositive` is a variable whose value can change each time through this loop. But when we add to the `textPredicates` list, we're defining a closure that will consult `isPositive` at some time in the future — possibly after it has changed.

If we instead define a variable with `let` or `const` within an inner block, we'll get what we want: a snapshot of`isPositive`'s value at the time of evaluation.

I'm pretty sure I've still gotten this wrong, though, because a `case` consequence clause doesn't create its own implicit block. Bear with me.

Hard to think of cases that would illustrate this problem without being nonsensical, but this might be one:

```scm
(
  (comment) @foo
  (#match? @foo "^//")
  (#not-match? @foo "^///")
)
```

lib/binding_web/binding.js
```diff
@@ -842,78 +842,90 @@ class Language {
...
+                textPredicates[i].push(function (captures) {
                   let nodes_1 = [];
                   let nodes_2 = [];
                   for (const c of captures) {
                     if (c.name === captureName1) nodes_1.push(c.node);
                     if (c.name === captureName2) nodes_2.push(c.node);
                   }
-                  return matchAll
-                    ? nodes_1.every(n1 => nodes_2.some(n2 => n1.text === n2.text)) === isPositive
-                    : nodes_1.some(n1 => nodes_2.some(n2 => n1.text === n2.text)) === isPositive;
```
savetheclocktower, line 872: We can't do it this way because the logic inverts twice: first between `eq` and `not-eq`, then between both those predicates and their `any-` variants.

Hard to explain in words, but the `any-not-eq?` case falls down because it tries to assert that “some of these nodes match X” is `false`, when it needs to assert that “some of these nodes _don't_ match X” is `true`.

lib/binding_web/binding.js
```diff
@@ -842,78 +842,90 @@ class Language {
...
-                    : nodes.some(n => n.text === stringValue) === isPositive;
+                  let test = eqIsPositive ? matches : doesNotMatch;
+                  return eqMatchAll
+                    ? nodes.every(test)
+                    : nodes.some(test);
                 });
               }
               break;
 
-            case 'not-any-match?':
```
savetheclocktower, line 889: This should've been `any-not-match?`.

lib/binding_web/binding.js
savetheclocktower: Never mind — I think this can't happen because each new predicate has its own block scope from the `for` loop. I'll simplify this and update the PR.

RedCMD:
you could do something silly like...
image

savetheclocktower:
> you could do something silly like...

this is why I need an “oof” emoji reaction option ;-)

bitomukesh:
## PR Analysis

- **Main theme:** Enhancing predicate handling in web-tree-sitter bindings
- **PR summary:** This PR introduces fixes to the `any-` versions of the `eq?` and `match?` predicates in the web-tree-sitter bindings, ensuring correct behavior and adding tests to validate the changes.
- **Type of PR:** Bug fix
- **[Score ](https://docs.bito.ai)(0-100, higher is better):** 85
- **Relevant tests added:** Yes
- **[Estimated effort to review ](https://docs.bito.ai)(1-5, lower is better):** 3
The PR seems well-structured, with tests included for the new logic. However, due to the complexity of the predicate logic and the potential for subtle bugs, a thorough review is necessary.
## PR Feedback

- **General suggestions:** The PR successfully addresses the incorrect implementation of quantified predicates and adds the necessary tests to ensure the correctness of the implementation. Although the changes are quite extensive, the logic seems to be well-thought-out and structured. It's important to ensure that the added code does not introduce any performance regressions, especially in a critical component like query evaluation, which is likely to be called frequently. Also, consider abstracting some of the newly added complex logic into smaller, testable functions to improve maintainability and readability.
## Code feedback

## file: lib/binding_web/binding.js

- **Suggestions:**



 1. Consider refactoring the predicate functions into separate, named functions to improve readability and maintainability. This will also make it easier to add unit tests for these specific behaviors. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ textPredicates[i].push(function (captures) {
 ```


 2. The `compare` and `test` functions are redefined multiple times within different cases. Extract these as standalone functions to avoid duplication and potential errors if changes are required later. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ let compare = (n1, n2, positive) => {
 ```


 3. The `matchAll` variable is reassigned multiple times which could lead to errors. Consider restructuring the code to avoid reassignments or clarify the logic with more descriptive variable names. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ matchAll = !operator.startsWith('any-');
 ```


 4. Ensure that the error messages provide enough context for debugging. Including the actual number of arguments received in the error message would be beneficial. [medium] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ `Wrong number of arguments to \`#${operator}\` predicate. Expected 2, got ${steps.length - 1}`
 ```


 5. The repeated logic for pushing functions into `textPredicates` suggests that a higher-order function could be used to generate these predicate functions, reducing code duplication. [medium] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ textPredicates[i].push(function (captures) {
 ```


 6. The newly added tests in `query-test.js` should be reviewed to ensure they cover all edge cases, especially given the complexity of the predicates being tested. [medium] **Relevant line**:In lib/binding_web/test/query-test.js at line 259
 ```
+ it("handles quantified captures properly", () => {
 ```


 7. Consider adding more inline comments explaining the purpose and logic of the new predicate handling, especially for future developers who may not be familiar with the context of these changes. [medium] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ let captureName;
 ```

bitomukesh:
## PR Analysis

- **Main theme:** Enhancing predicate handling in web-tree-sitter bindings
- **PR summary:** This PR fixes and improves the implementation of quantified predicates in the web-tree-sitter's binding.js file, correcting the logic and adding new tests to ensure proper functionality.
- **Type of PR:** Bug fix
- **[Score ](https://docs.bito.ai)(0-100, higher is better):** 85
- **Relevant tests added:** Yes
- **[Estimated effort to review ](https://docs.bito.ai)(1-5, lower is better):** 2
The changes are quite isolated and well-defined, but require careful consideration of the logical flow and potential edge cases.
## PR Feedback

- **General suggestions:** The PR introduces significant improvements to the predicate handling in web-tree-sitter. It's important to ensure that the refactored code maintains readability and performance. Avoiding duplication of logic, such as the repeated definition of similar functions within different cases, could be beneficial. Leveraging higher-order functions or utility functions to encapsulate common behavior would enhance maintainability. Additionally, ensuring that the added tests cover all new branches and logic changes is crucial for preventing regressions.
## Code feedback

## file: lib/binding_web/binding.js

- **Suggestions:**



 1. Consider abstracting the repeated logic for text comparison into a higher-order function to reduce code duplication and improve maintainability. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ let compare = (n1, n2, positive) => {
 ```


 2. Refactor the repeated logic for regex matching into a single function outside of the switch cases to avoid duplication and potential errors in future maintenance. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ let test = (text, positive) => {
 ```


 3. Ensure that the new predicates are accurately reflected in the error messages, to provide clarity when an error is thrown. [medium] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ `Wrong number of arguments to \`#${operator}\` predicate. Expected 2, got ${steps.length - 1}`
 ```


 4. Validate that the changes made to the predicate logic do not introduce any performance regressions, especially in loops or heavily used code paths. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ return matchAll
 ```


 5. Consider adding more comprehensive tests to cover the new logic paths introduced by the changes in predicate handling. [important] **Relevant line**:In lib/binding_web/test/query-test.js at line 259
 ```
+ it("handles quantified captures properly", () => {
 ```


 6. Review the use of ternary operators for clarity, as complex expressions may reduce readability. Break down complex ternaries into if-else statements if needed. [medium] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ let compare = (n1, n2, positive) => {
 ```


 7. Ensure that the newly added code does not introduce any security vulnerabilities, particularly with regards to the handling of regular expressions and user-provided input. [important] **Relevant line**:In lib/binding_web/binding.js at line 842
 ```
+ const regex = new RegExp(steps[2].value);
 ```

amaanq:
Thanks for the PR @savetheclocktower, and sorry for the delay in response. I apologize for the issues I introduced here, js is not my forte :^)

amaanq (approved):

savetheclocktower:
Not a problem! Thanks for the review.

PR #2808, opened 2023-12-03
Grow wasm memory as needed when loading additional wasm language modules

maxbrunsfeld:
This PR changes the `TSWasmStore` to manage its linear memory more correctly as it loads more languages. It starts out with the minimum size linear memory allowed by emscripten (4MB), and then grows it as needed when loading each wasm language. It also makes the layout of the wasm stack, static data, and heap a bit clearer.


Lovea10:
****

Scutua (approved):

Issue #2814, opened 2023-12-08
feature request: support building wasm with podman in addition to docker 


SKalt:
**Context**: [`podman`](https://podman.io/) is a tool to run and manage containers. It's an alternative to docker that has most of the same command-line flags and arguments. I'd like [`tree-sitter build-wasm`](https://github.com/tree-sitter/tree-sitter/blob/660481dbf71413eba5a928b0b0ab8da50c1109e0/cli/loader/src/lib.rs#L571) to work with `podman` instead of complaining about the absence of docker.

I'd be happy to submit a PR for this!

maxbrunsfeld:
Sounds like a good feature. I'd be open to that PR.

PR #2815, opened 2023-12-09
podman support

SKalt:
see #2814; notes in commit messages.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -19,7 +19,7 @@ use tree_sitter_highlight::HighlightConfiguration;
 use tree_sitter_tags::{Error as TagsError, TagsConfiguration};
 use which::which;
 
-pub const EMSCRIPTEN_TAG: &'static str = concat!("emscripten/emsdk:", env!("EMSCRIPTEN_VERSION"));
+pub const EMSCRIPTEN_TAG: &str = concat!("docker.io/emscripten/emsdk:", env!("EMSCRIPTEN_VERSION"));
 
 #[derive(Default, Deserialize, Serialize)]
 pub struct Config {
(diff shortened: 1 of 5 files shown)


docs/section-6-contributing.md
```diff
@@ -18,7 +18,7 @@ To make changes to Tree-sitter, you should have:
 1. A C compiler, for compiling the core library and the generated parsers.
 2. A [Rust toolchain](https://rustup.rs/), for compiling the Rust bindings, the highlighting library, and the CLI.
```
Sonyashines, line 19:
```suggestion
 2. A [Rust toolchain](https://rustup.rs/), for compiling the Rust bindings, the highlighting library, and the CLI.
- [ ]

**Review by Sonyashines — APPROVED** [review]

File: docs/section-6-contributing.md (hunk 1 above)
Comment (SKalt) [reply docs/section-6-contributing.md:19]: Hi @Sonyashines, could you clarify what change you'd like me to make? I'm not able to tell what you're looking for based on the diff.

**Comment by SKalt** [comment]
The remaining test failures don't seem to relate to my changes. Instead, the failures smell like a timing-based intermittent failure:
```
... 
Caused by:
   /home/runner/work/tree-sitter/tree-sitter/target/scratch/x86_64-linux-unknown-gnu-little/bash.so: file too short
 
failures:
    tests::async_context_test::test_node_and_cursor_ref_in_fut
    tests::async_context_test::test_node_and_cursor_ref_in_fut_with_fut_fabrics
    tests::async_context_test::test_node_and_cursor_ref_in_fut_with_inner_spawns
```
See also
- https://github.com/tree-sitter/tree-sitter/actions/runs/7051392884/job/19197450237
- https://github.com/tree-sitter/tree-sitter/actions/runs/7040438109/job/19161367164

I'm going to push a dummy commit to see if re-running the pipeline fixes the failures.

**Comment by SKalt** [comment]
Confirmed heisenbug :( two of the three failures evaporated without any meaningful code changes.

```diff
  failures:
     tests::async_context_test::test_node_and_cursor_ref_in_fut
-    tests::async_context_test::test_node_and_cursor_ref_in_fut_with_fut_fabrics
-    tests::async_context_test::test_node_and_cursor_ref_in_fut_with_inner_spawns
```

cli/loader/src/lib.rs
```diff
@@ -532,45 +533,94 @@ impl Loader {
...
+                let emcc_bin = if cfg!(windows) { "emcc.bat" } else { "emcc" };
+                let emcc_path = which(emcc_bin)
+                    .ok()
+                    .and_then(|p| Command::new(&p).output().and(Ok(p)).ok());
+                emcc_path.map(|p| EmscriptenSource::Native(p))
+            } else {
+                None
+            };
+            emcc.or_else(|| {
+                if Command::new("docker").output().is_ok() {
```
maxbrunsfeld, line 553: I'm wondering if we should fall back to `podman` even if `docker` is installed, if the Docker server isn't running. To do that, we could run `docker images` instead of just `docker`, so that we'll get a failure if docker isn't running.

maxbrunsfeld:
Yeah, sorry for the spurious test failures. We'll fix those soon.

cli/loader/src/lib.rs
SKalt: I pushed this behavior in 61f7c2bc. For a future improvement, we might surface information about what executables are present on `$PATH` or what errors each executable raises.

alebastr:
This doesn't work if the `docker` command is a podman wrapper (`podman-docker` package in many Linux distributions).
```
$ /usr/bin/docker version
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Client:       Podman Engine
Version:      4.8.2
API Version:  4.8.2
Go Version:   go1.21.4
Built:        Mon Dec 11 05:23:04 2023
OS/Arch:      linux/amd64
```
`docker info` succeeds, so the code never takes the `podman` branch. Checking for `podman` first or testing `docker version`/`docker --version` for podman may help.

SKalt:
@alebastr good catch! I've pushed a patch to ensure `podman` aliased as `docker` always has user-namespacing mode `keep-id`. I decided to avoid the complexity of detecting if `docker` is an alias to `podman` since that seemed fragile and undesired by anyone who aliased `docker` to `podman` in the first place.

amaanq:
Hey @SKalt I've just rebased your PR on top of master and tweaked the code a bit - 

For one, I think we can just rely on linking with libc in a unix environment to directly use `getuid` instead of spawning a command to get the user's ID. I also just collapsed some code where possible. Let me know what you think, otherwise this seems good to me

SKalt:
Lgtm; thanks for the cleanup!

amaanq:
thanks!

PR #2824, opened 2023-12-16
fix(cli): cannot install via a HTTP tunnel proxy

stormyyd:
I found that the install script in `cli/npm` is not handling the HTTP proxy correctly, so I make a simple fix for that.

The proxy usage in former version is ok when the destination is a HTTP **(without the `S`)** website, but the install script will download tree-sitter binary file from GitHub by HTTPS protocol (which is good and I wouldn't to switching it to HTTP).

Using HTTP proxy as a tunnel (the HTTP `CONNECT` method) is the right way to do this. AFAIK all HTTP proxy implementation support the `CONNECT` method, this fix should be ok for all users.

### DIFF
--- a/cli/npm/install.js
+++ b/cli/npm/install.js
@@ -83,29 +83,42 @@ file.on('finish', () => {
 
 // Follow redirects.
 function get(url, callback) {
-  const requestUrl = new URL(url)
-  let request = https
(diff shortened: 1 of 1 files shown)


amaanq:
thank you!

PR #2838, opened 2023-12-25
Redundant redundant escapes

### LINKED ISSUE #2837: Regex preprocessing doesn't match any regex flavor
Tree-sitter currently does two preprocessing steps to regex:

- replace all `(^|[^\\pP])\{([^}]*[^0-9A-Fa-f,}][^}]*)\}` with `$1\{$2\}`[^curly-brace], and
- remove redundant escapes for `!'"/`[^redundant].

[^curly-brace]: This escapes the curly braces of any `{}` for which the contents do not look like a character class, repetition quantifier, or `\p`.
[^redundant]: This preprocessing is redundant now, as the Rust regex-syntax's [escape sequences](https://docs.rs/regex/latest/regex/#escape-sequences) now always permit escaping all ASCII except `[0-9A-Za-z<>]` (meaning to match the literal ASCII).

This doesn't actually implement any regex flavor. Worse than not implementing any well-known flavor, for each such flavor, there exists a diverging regex which is *valid* both in tree-sitter and that flavor, but which is interpreted differently.

- It obviously diverges from Rust flavored regex, since we're changing the regex before interpreting it as Rust flavored regex.
  - It doesn't even maintain that all valid Rust regexes are interpreted correctly, since Rust supports extended word boundary assertions which get mangled (`\b{start}`, `\b{end}`, `\b{start-half}`, `\b{end-half}`).
- It still doesn't match ECMAScript flavored regex:
  - `\0` is not recognized as meaning `\u{0}`.
  - It diverges in `[]]` being interpreted `\]` (ECMAScript flavor) versus `[\]]` (Rust flavor).
  - It diverges from Unicode-unaware mode, which doesn't have escapes `\a`, `\A`, `\z`, `\<`, `\>`, `\x{}`, `\u{}`, `\U{}`, `\p`, or `\P`, thus treats them as literal escapes.
    - Tree-sitter also still treats literal `{}` and unclosed `{` as an error.
    - [Unicode-aware mode] (the `u` or `v` flag) turns the [deprecated literal `{}` syntax][deprecated syntax] which is being emulated into a syntax error.
  - It diverges from [`u` mode][u], which doesn't have character class set operations, thus making `[a-y&&xyz]` mean `[a-z]` (instead of `[xy]`).
  - It diverges from [`v` mode][v], which doesn't have posix style character classes, thus making `[[:alpha:]]` mean `[:alph]` (instead of `[A-Za-z]`).
- It diverges from PCRE, which doesn't have `\<`, `\>`, or character class set operations.

[u]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode
[v]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicodeSets
[deprecated syntax]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features#regexp
[Unicode-aware mode]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#unicode-aware_mode

### Suggested resolution

Just stop preprocessing the regex. There's two possible justifications available for removing the curly brace preprocessing:

- State that tree-sitter uses Rust flavored regex expressions. It's even a flavor option on [regex101] nowadays.
- State that tree-sitter always applies the `u` or `v` flag to the ECMAScript flavored regex.

Reiterating: the regex preprocessing

- makes tree-sitter use an ad-hoc regex flavor
- in order to emulate an explicitly deprecated behavior of ECMAScript regex syntax
- and it doesn't even match ECMAScript behavior.

Given that Tree-sitter deliberately supports <code>\u{<i>hhhh</i>}</code> escapes[^unicode-class] despite being a `.js` syntax, which requires the use of the `u` or `v` flag, it imho just makes sense to stop ~~mangling~~ preprocessing the regex to support syntax which is an error given the use of the `u` or `v` flag. And *especially* now that Rust is a known flavor on [regex101], it's much more an option to just say that's what Tree-sitter is using.

[regex101]: https://regex101.com/
[^unicode-class]: If it didn't, and wanted to match ECMAScript behavior more closely, the curly brace preprocessing would probably match `[^[:digit:],}]` instead of `[^[:alnum:],}]`, to at least match the behavior of `\u{100F}`, although `\u{100}` would still be interpreted as `\u0100` (Rust flavor interpretation) instead of `u{100}` (deprecated Unicode-unaware ECMAScript flavor interpretation).

CAD97:
Fixes #2837; see that issue for further discussion.

The regex-syntax crate natively supports `\!`, `\'`, `\"`, and `\/` identity escapes now, so we don't need to preprocess the regex in order to support those.

Unescaped `{}` are only permitted in Unicode-unaware JS regexp, and that syntax is deprecated, only kept for web compatibility, and should not be relied upon. In Unicode-aware JS regexp (via the `u` or `v` flag), unescaped `{}` are syntax errors. Unicode-aware mode is required for support for `\u{hhhh}` or `\p{UnicodeProperty}` syntaxes, which Tree-sitter always supports.

The current behavior is sort of a "worst of both worlds" where Tree-sitter's regex flavor is neither Rust nor ECMAScript compatible. Removing the special curly brace preprocessing means Tree-sitter's regex now matches Rust flavor again and moves it closer to that of ECMAScript in `u` or `v` mode. (And the divergences, unlike unescaped `{}`, are much less likely to be hit unintentionally.) And even though this means any previously-preprocessed regex will change meaning, it will always[^almost] be changed to emit a clear error pointing out the `{` which now needs to be manually escaped.

[^almost]: Well, *almost* always. In the specific case of `\b{start}`, `\b{end}`, `\b{start-half}`, or `\b{end-half}`, the pre-preprocessing version is valid with a different meaning.

## Changelog

- Regexes are now always interpreted by Tree-sitter using Rust flavored regex syntax. The main impact is that `{` now need to be escaped outside of `[]`. If this occurs in your grammar, Tree-sitter will error, indicating the regex which needs to be edited.

### DIFF
--- a/cli/src/generate/prepare_grammar/expand_tokens.rs
+++ b/cli/src/generate/prepare_grammar/expand_tokens.rs
@@ -4,7 +4,6 @@ use crate::generate::nfa::{CharacterSet, Nfa, NfaState};
 use crate::generate::rules::{Precedence, Rule};
 use anyhow::{anyhow, Context, Result};
 use lazy_static::lazy_static;
-use regex::Regex;
 use regex_syntax::ast::{
     parse, Ast, ClassPerlKind, ClassSet, ClassSetBinaryOpKind, ClassSetItem, ClassUnicodeKind,
     RepetitionKind, RepetitionRange,
@@ -13,8 +12,6 @@ use std::collections::HashMap;
 use std::i32;
 
 lazy_static! {
-    static ref CURLY_BRACE_REGEX: Regex =
-        Regex::new(r"(^|[^\\pP])\{([^}]*[^0-9A-Fa-f,}][^}]*)\}").unwrap();
     static ref UNICODE_CATEGORIES: HashMap<&'static str, Vec<u32>> =
         serde_json::from_str(UNICODE_CATEGORIES_JSON).unwrap();
     static ref UNICODE_PROPERTIES: HashMap<&'static str, Vec<u32>> =
@@ -29,7 +26,6 @@ const UNICODE_CATEGORIES_JSON: &str = include_str!("./unicode-categories.json");
 const UNICODE_PROPERTIES_JSON: &str = include_str!("./unicode-properties.json");
 const UNICODE_CATEGORY_ALIASES_JSON: &str = include_str!("./unicode-category-aliases.json");
 const UNICODE_PROPERTY_ALIASES_JSON: &str = include_str!("./unicode-property-aliases.json");
-const ALLOWED_REDUNDANT_ESCAPED_CHARS: [char; 4] = ['!', '\'', '"', '/'];
 
 struct NfaBuilder {
     nfa: Nfa,
@@ -58,29 +54,6 @@ const fn get_completion_precedence(rule: &Rule) -> i32 {
         }
     }
     0
-}
-
-fn preprocess_regex(content: &str) -> String {
-    let content = CURLY_BRACE_REGEX.replace(content, "$1\\{$2\\}");
-    let mut result = String::with_capacity(content.len());
-    let mut is_escaped = false;
-    for c in content.chars() {
-        if is_escaped {
-            if !ALLOWED_REDUNDANT_ESCAPED_CHARS.contains(&c) {
-                result.push('\\');
-            }
-            result.push(c);
-            is_escaped = false;
-        } else if c == '\\' {
-            is_escaped = true;
-        } else {
-            result.push(c);
-        }
-    }
-    if is_escaped {
-        result.push('\\');
-    }
-    result
 }
 
 pub fn expand_tokens(mut grammar: ExtractedLexicalGrammar) -> Result<LexicalGrammar> {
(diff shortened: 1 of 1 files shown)


amaanq:
Yeah this makes a lot of sense - thanks a lot for the detailed issue and pr description as well as fixing this! I'll rebase this for you & merge

CAD97:
And thus, my pet peeve [from five years ago](https://github.com/tree-sitter/tree-sitter/pull/381) is resolved 🎉

I still have borderline impossible dreams for a [modern parser generator](https://matklad.github.io/2018/06/06/modern-parser-generator.html)[^1], but with this rough edge sanded down I have no qualms about suggesting/using tree-sitter anymore. Maybe it's finally time to revisit my toy language...

[^1]: (Not my post.) The main thing I want that tree-sitter doesn't have is a separately phased lexer instead of a lexerless design. However, I do know lexerless is in vogue and handles token splitting/gluing (e.g. `>>` vs `> >`) more directly, so I'll concede that. Other than that, it'd mostly be resolvable with tooling on top of the existing corpus testing support.

PR #2840, opened 2023-12-27
Introduce APIs for managing the lifetimes of languages, allow WASM languages to be deleted

maxbrunsfeld:
### Background

PR https://github.com/tree-sitter/tree-sitter/pull/1864 added initial support for loading languages from WASM files. Up until then, languages were always static data, so there were no APIs for managing their ownership or lifetime. So although you could load a language from WASM, there was no way to dynamically unload that language without leaking the memory.

### Change

This PR introduces two new C APIs for managing the lifetime of a `TSLanguage`: `ts_language_delete`, and `ts_language_copy`. For Tree-sitter's normal, statically-allocated language objects, these APIs are noops. But for languages loaded from WASM files, these APIs perform reference counting operations. When a language's reference count goes to zero, all of its memory is released.

This allows you to dynamically reload WASM languages at runtime. This could be used, for example, to create a Tree-sitter grammar playground that dynamically re-generates its parser, recompiles it using Emscripten, and re-loads it into the current process 🙀 !.

### Rust API

In Rust, `Language` is no longer `Copy`, and now has a `Drop`. Some functions (like `Parser::set_language` and `Query::new`) that take languages now take them by reference, so that the caller doesn't  need to call `.clone` (since the C APIs bump the reference count internally).


amaanq:
Hey @maxbrunsfeld, PHP had a rewrite splitting it into two grammars - you can rebase on top of master for the fix I added in #2878 that'll resolve the CI errors

maxbrunsfeld:
Nice, thanks @amaanq

PR #2841, opened 2023-12-28
Make Node.js language bindings context aware

segevfiner:
They don't have any dynamic global data, so all it takes is just declaring them as such

Contributed on behalf of [Swimm](https://swimm.io/)

### DIFF
--- a/cli/src/generate/templates/binding.cc
+++ b/cli/src/generate/templates/binding.cc
@@ -23,6 +23,6 @@ void Init(Local<Object> exports, Local<Object> module) {
   Nan::Set(module, Nan::New("exports").ToLocalChecked(), instance);
 }
 
-NODE_MODULE(tree_sitter_PARSER_NAME_binding, Init)
+NODE_MODULE_CONTEXT_AWARE(tree_sitter_PARSER_NAME_binding, Init)
 
 }  // namespace


verhovsky:
We'll have to submit a PR with this change to all 113 of these: https://tree-sitter.github.io/tree-sitter/#parsers

sogaiu:
Though possibly not all are active, I think there are [somewhat more than 113](https://github.com/sogaiu/ts-questions/blob/e9ae6c719dbd60b0d5556f29cb9966c0154953ea/ts-grammar-repositories.txt).

segevfiner:
Well if it comes down to it, we can always automate opening those PRs? Not sure how changes to the generation template or grammar generation were handled before. Not merging this and the accompanying PR to node-tree-sitter just means I'll have to fork and maintain patching everything by myself :(

clason:
> Though possibly not all are active, I think there are [somewhat more than 113](https://github.com/sogaiu/ts-questions/blob/e9ae6c719dbd60b0d5556f29cb9966c0154953ea/ts-grammar-repositories.txt).

More like [406](https://github.com/nvim-treesitter/nvim-treesitter/wiki/List-of-parsers).

maxbrunsfeld (approved):

PR #2844, opened 2024-01-02
docs: add css for inline code

sh-cho:
Recently I am building a parser with tree-sitter and while reading the docs, I thought it would be better if the inline code was more readable, so I changed the style.

What do you think? I'd love to hear your thoughts.

Here are some examples of the styling

| as-is | to-be |
| --- | --- |
| image | image |

ex) https://tree-sitter.github.io/tree-sitter/creating-parsers#keywords

| as-is | to-be |
| --- | --- |
| image | image |

ex) https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners

You can check other pages here: https://sh-cho.github.io/tree-sitter/

I only highlighted `<code>` inside `#main-content` except `<code>` inside `<a>` or `<pre>` tag.

### DIFF
--- a/docs/assets/css/style.scss
+++ b/docs/assets/css/style.scss
@@ -33,7 +33,7 @@ a[href^="http"]:after {
   padding: $padding 0;
 }
 
-#main-content code:not(pre code) {
+#main-content code:not(pre code, a code) {
   color: #c7254e;
   font-size: 0.9em;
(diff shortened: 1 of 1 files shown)


amaanq:
this is really nice, sorry for not getting to this sooner. thank you!

PR #2857, opened 2024-01-08
Add a `#replace!` predicate that can update a node

tgross35:
This PR is an attempt to add a `#replace! @capture "regexp(.*)" "rep$1" predicate that can be used to filter content. This is useful for things like extracting injection languages from shebangs or heredocs, and will provide an official equivalent of the `#gsub!` function supported by NeoVim.

However, I am not sure what the best way to do this is. It seems like this more or less requires instrumenting a `Node` in some way that `utf8_text` / `utf_16` text know to return data from a buffer rather than the input. I am going to try adding something to `Node` to allow this, but would like to know if there are other suggestions.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -714,6 +714,54 @@ fn test_query_matches_with_many_overlapping_results() {
 }
 
 #[test]
+fn test_query_predicate_replace() {
+    allocations::record(|| {
+        let language = get_language("ruby");
+        let query = Query::new(
+            language,
+            r#"(program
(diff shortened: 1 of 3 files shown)


tgross35:
I probably should look at how `#set!` works as reference

amaanq:
Closing as stale - I don't think this PR is ready as is, and I'm not entirely sure if we could support this kind of directive. I think it'd be better to have a discussion about this, or at the very least, come with a full-fledged implementation. No offense to you, I just want to declutter the open PRs.

PR #2875, opened 2024-01-24
Allow `--filter` to take regex

tgross35:
Replace `str::contains` with `Regex::is_match` to allow more flexible matching.


amaanq:
I think I'd like to replace filter with include/exclude in 0.21.0, and so we've added them in #2902 for 0.20.10. Thanks for the PR though!

PR #2879, opened 2024-01-25
fix: specify eof should be lf for windows cicd

### LINKED ISSUE #2756: Mess up with CRLF line endings between different platforms in `tree-sitter generate` command
# Background
It seems that the `git` tool in Github CI on Windows runners uses a `core.autocrlf = true` setting _(default value)_ what leads to that it checkouts all files with a platform native line ending what for Windows is CRLF `\r\n` even if a file stored in Git has just LF `\n` line endings.

All Rust's output macro in the std library [use only](https://doc.rust-lang.org/nightly/std/macro.writeln.html) LF `\n` for all platforms.

# Problem
Tree-sitter CLI on Windows platform produces generated files with different line endings. For example the `src/tree_sitter/parser.h` file would have a CRLF line ending while all other json files and the `src/parser.c` file would be with LF line ending.

It seems that the issue happens in CI when the `tree-sitter.exe` binary compiled for Windows platform, due to `git`'s default behavior where it first checkouts a static template file for the `src/tree_sitter/parser.h` file and replaces line endings to CRLF and after that this template file will then be included as is in a **binary form** in the Tree-sitter CLI executable.

We need to add the next config for the git in CI pipeline:
```
git config core.autocrlf false
git config core.eof lf
```

calebdw:
Closes #2756.

Thanks!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,5 @@
+* text=auto eol=lf
+
 /lib/src/unicode/*.h linguist-vendored
 /lib/src/unicode/LICENSE linguist-vendored
 


amaanq:
thank you!

PR #2880, opened 2024-01-25
0.20.9

amaanq:


Fapp3r3Lit3:
Was this for or about something I posted? Lmao idk why I'm getting notifications however, I do tend to,  text to talk a lot. And it prints out terrible.

amaanq:
> Publishing to Crates.io and NPM failed.. Is this something you are aware of and is being fixed? I would like to upgrade tree-sitter-cli to 0.20.9

Aware of it, will be fixed soon

PR #2884, opened 2024-01-27
docs: add `tree-sitter-magik` Magik parser

sebastiaanspeck:


verhovsky:
There's over 400 parsers, I don't think it would help anyone to link to this one on the front page.

https://github.com/nvim-treesitter/nvim-treesitter/wiki/List-of-parsers

verhovsky:
you don't see any difference between Magik and JavaScript?

sogaiu:
FWIW, the PR seems in line with similar requests for additions to the list of parsers.

The other mentioned list isn't curated by the same set of people IIUC, and the audiences of the two lists differ.

clason:
> The other mentioned list isn't curated by the same set of people IIUC

Well, there's some overlap... The main point is that if you aim to be comprehensive (which, given the merged PRs, I believe you are), having these lists on the front page is not sustainable. I'd recommend moving them to a separate page. (Ideally a Wiki to reduce noise on the main repo, but given the problems this project has with bot spam, that'd be unsafe.)

verhovsky:
It's very helpful to have a list of parsers in the docs, but it doesn't make sense to have it be a comprehensive list. It should be common languages people would expect to see, it's also an endorsement of the linked parser.

clason:
> It should be common languages people would expect to see

Your "common language" is another person's esolang. Who are you to be a gatekeeper here?

sogaiu:
> Well, there's some overlap... 

I agree.  I failed to communicate that nuance above.  My apologies.

---

If I'm not mistaken, motivation to get one's parser listed here in the past has included at least the following reasons (likely varying by individual):

1. Discoverability - letting other folks know that an effort of sufficient development exists
2. Fame / Pride

The list here is in some sense "official", but I think there has been some effort to not put just any parser.  The goals for the list here are not as explicit, but in practice I think one can see that there has been some curation with an eye toward including things that seemed to be functioning pretty well.

As for gatekeeping, this list is maintained by those who can merge PRs -- which in practice is only around three people?

The [page with the other list](https://github.com/nvim-treesitter/nvim-treesitter/wiki/List-of-parsers) has specific goals mentioned at the top of the page:

> This list contains all known tree-sitter parsers with publicly available repos. Checked parsers are already included in nvim-treesitter; if a parser is on the list but not included, please consider opening a PR (if you are willing to maintain them).

As I understand it, this list is a good place to look for whether any effort exists at all.  If there is nvim-treesitter support (represented by a check), I take that to mean it's likely to have been made to work with nvim-treesitter (so has some level of quality as well as integration work).

My two cents is that I think multiple lists serving somewhat different purposes seems fine, but may be it would be less work for there to only be one (TM)?  May be the tree-sitter maintainers have an opinion...

clason:
You are correct in your statements about the nvim-treesitter list (and no worries, I just wanted to mention that ;))

Of course there can be multiple lists _for different purposes_; I'd just recommend making the purpose of _this_ list explicit so there can be objective criteria for evaluating inclusion (as well as move it, if the criteria are inclusive enough for the list to become even longer...)

dundargoc:
Thank you for your contribution! We collect all typo fixes and similar small fixes into a single pull request and merge it once it gets big enough: https://github.com/tree-sitter/tree-sitter/pull/2957

dundargoc:
For what it's worth I'm really starting to dislike the parser link graveyard. They need to be constantly updated which is a drag. That said I don't think this is a discussion we should have in this PR since the author was just trying to be helpful by following the current trend (which is to link to as many parsers as possible). I think a wiki where people can update it themselves would work well for these.

@maxbrunsfeld allow wiki changes for the commoners please.

clason:
I don't think that's safe. If Neovim has to lock down their wiki, imagine what the cryptowolves would do to this repo...

dundargoc:
> I don't think that's safe. If Neovim has to lock down their wiki, imagine what the cryptowolves would do to this repo...

Yeah, fair. Maybe I'm being too optimistic.

> A central registry of parsers is useful but doesn't have to live on the front page of the tree-sitter documentation.

Yes, this is probably saner solution in hindsight. Although it doesn't solve the "churn" problem of needing to constantly update these. Still, an improvement is an improvement.

clason:
> Although it doesn't solve the "churn" problem of needing to constantly update these.

Yeah, I have no good answer to that. I just keep my eyes open and add any parser that I see mentioned somewhere.

PR #2896, opened 2024-02-02
ci: merge fast_checks into full_rust_checks

dundargoc:
It's not necessary to have two separate linting workflows as the full
lints themselves are very quick.


amaanq:
can you rebase please?

dundargoc:
Uh, I think the concurrency string in the workflow file is off since it looks like it cancelled itself or something lol

amaanq:
thanks!

PR #2900, opened 2024-02-03
ci: remove msrv.yml workflow as it's not really used

dundargoc:
If it does become relevant to actually implement this is can be done at
that time. Having a non-functional stub isn't useful.


amaanq:
thanks

PR #2901, opened 2024-02-03
cli: reworded test diff color key

buckynbrocko:
- Reworded the color key for diffed tests
- Added a `#[ignore]`ed test to make previewing any future changes trivial

image


cli/src/test.rs
```diff
@@ -916,4 +917,10 @@ code with ----
             }
         );
     }
+
+    #[ignore]
+    #[test]
+    fn test_print_diff_key() -> () {
+        print_diff_key()
+    }
```
amaanq, line 925: I don't think this is needed, we can just remove this - any future changes would be sparse and trivial to do something like this locally

Issue #2635, opened 2023-09-20
feature: add --exclude to tree-sitter-test


2ax3ax5:
It'd be nice if out of the box tree-sitter-test had a kind of exclude OPTION, a symmetric option to the already existing filter option.
```
$ tree-sitter test --help
tree-sitter-test 
Run a parser's tests

[...]

OPTIONS:
    -f, --filter <filter>       Only run corpus test cases whose name includes the given string
    -e, --exclude <filter>      Only run corpus test cases whose name does not include the given string
```

It'd be really cool if --filter and --exclude accepted a variable number of filters, instead of accepting only one.
```
    -f, --filter <filter>...    Only run corpus test cases whose name includes at least one of the given strings
    -e, --exclude <filter>...   Only run corpus test cases whose name does not include any of the given strings
```

ahlinc:
Then it makes sense to rename the `--filter` to the `--include` for symmetry.

sogaiu:
I prefer `--include` too as I think it's less ambiguous than `--filter`.  I think the latter could be interpreted as "filter to keep" or "filter to discard".

PR #2902, opened 2024-02-03
feat: non-breaking include/exclude regex options for subcommand test 

buckynbrocko:
cli `tree-sitter test`:
- Added `-i`/`--include` optional argument to run only tests whose names match the given regex string
- Added `-e`/`--exclude` optional argument to run only tests whose names _do not_ match the given regex string

This is notably different from #2875 [Allow _--filter_ to take regex](https://github.com/tree-sitter/tree-sitter/pull/2875) and #2635 [feature: add --exclude to tree-sitter-test](https://github.com/tree-sitter/tree-sitter/issues/2635)
in that it leaves the current `-f`/`--filter` option as-is to prevent a breaking change as well as allow it to be deprecated at a later time (if ever) after a suitable warning period. It may be a good idea to make `filter` conflict with `include` (or both) to prevent confusion/misuse, but that can be decided later.

Examples
```shell
tree-sitter test --include "(?i)statements?" # run only tests with `statement(s)` in their name, case insensitive
tree-sitter test --exclude "Identifier$" # run only tests whose name doesn't end with `Identifier`
tree-sitter test --include "(?i)statements?" --exclude "Identifier$" # include && exclude
```

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,10 +1,12 @@
 use anyhow::{anyhow, Context, Error, Result};
 use clap::{App, AppSettings, Arg, SubCommand};
 use glob::glob;
+use regex::Regex;
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::{env, fs, u64};
 use tree_sitter::{ffi, Parser, Point};
+use tree_sitter_cli::test::TestOptions;
 use tree_sitter_cli::{
     generate, highlight, logger,
     parse::{self, ParseFileOptions, ParseOutput},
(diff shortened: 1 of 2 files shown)


amaanq:
I think we can remove filter, and keep these two. It's definitely useful to just use a regex, and having too many flags gets annoying.

sogaiu:
Is it that bad to have at least one release where the old option coexists with the new ones?  Seems kinder from a breakage perspective.

Just my two cents.

amaanq:
Hm, maybe. Though 0.20.9 has filter, since it doesn't have these users might be bit by it. Let's keep it for 0.20.10 then remove it in 0.21.0

amaanq:
Ok, I also reworked it to not print out every group's name if it had no tests that would be run. I think this is cleaner overall

image

Before it'd be like this

image

Which feels annoying. Thanks for the PR!

PR #2915, opened 2024-02-04
fix: spaces in paths broke unquoted bash variables

buckynbrocko:


script/build-wasm
```diff
@@ -78,8 +78,9 @@ fi
 if [ -z "$emcc" ] && [ -n "$docker" ]; then
   export PODMAN_USERNS=keep-id
   emcc="$docker run                      \
+  --platform linux/amd64                 \
```
amaanq, line 81: are you sure about this? or was this just a leftover remnant from testing on your own machine

script/build-wasm
```diff
@@ -79,7 +79,7 @@ if [ -z "$emcc" ] && [ -n "$docker" ]; then
   export PODMAN_USERNS=keep-id
   emcc="$docker run                      \
     --rm                                 \
-    -v $(pwd):/src:Z                     \
+    -v \"$(pwd)\":/src:Z                 \
```
dundargoc, line 82: This gives

> docker: Error response from daemon: create ... includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.

Come on, at least test your PR :)

I think you should revert this completely, quotes shouldn't be needed here

script/build-wasm
```diff
@@ -79,7 +79,7 @@ if [ -z "$emcc" ] && [ -n "$docker" ]; then
   export PODMAN_USERNS=keep-id
   emcc="$docker run                      \
     --rm                                 \
-    -v $(pwd):/src:Z                     \
+    -v $(pwd):/src:Z                 \
```
dundargoc, line 82:
```suggestion
    -v $(pwd):/src:Z                     \
```

PR #2923, opened 2024-02-05
feat: allow the grammar to log to stdout

### LINKED ISSUE #703: console.log() in module causes some cryptic error

amaanq:
Closes #703

for anyone reading, you can now use console.log or any other stdout logging mechanism without having it break tree-sitter's generate command

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1239,7 +1239,6 @@ version = "0.20.9"
 dependencies = [
  "ansi_term",
  "anyhow",
- "atty",
  "clap",
  "ctor",
  "ctrlc",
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
Nice, thanks for fixing this @amaanq

PR #2924, opened 2024-02-05
ci: include cargo packages to dependabot

dundargoc:


dundargoc:
It's possible to e.g. specify to skip patch updates and only create a PR on minor version changes. Anything of interest?

amaanq:
yeah I think that'd be great to have, so we don't get a pr every week with unnecessary updates

dundargoc:
Done. It's possible to specify different strategies for different dependencies (e.g. never skip bumps for dependency A but skip patch updates for dependency B) but I can't really come up with a non-forced situation when this would be relevant tbh.

PR #2931, opened 2024-02-05
fix: align edits to utf-8 character boundaries if the input is valid utf-8

amaanq:
The php grammar's scanner recently switched to tracking and comparing heredoc delimiters with wchar_t, and as a result, the testing of its corpus with random edits sometimes would fail, albeit rarely. This is likely because it was editing inside the middle of a utf-8 character, inserting a byte where it really shouldn't - that is, if we're thinking in the context of editors where a user would edit the entire character rather than a specific byte inside it. 

You can observe the effects of this change with the following previously-failing seed, which succeeds now.

```sh
TREE_SITTER_SEED=14817424306918605657 cargo test tests::corpus_test::test_corpus_for_php -- --nocapture
```

### DIFF
--- a/cli/src/tests/helpers/edits.rs
+++ b/cli/src/tests/helpers/edits.rs
@@ -50,7 +50,7 @@ impl<'a> ReadRecorder<'a> {
     }
 }
 
-pub fn invert_edit(input: &Vec<u8>, edit: &Edit) -> Edit {
+pub fn invert_edit(input: &[u8], edit: &Edit) -> Edit {
     let position = edit.position;
     let removed_content = &input[position..(position + edit.deleted_length)];
     Edit {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I worry this is an actual bug in the PHP scanner (or Tree-sitter), because the incremental parsing **should** work even in the presence of invalid unicode.

amaanq:
Hm, the regression occurred **only** when the scanner switched from using an underlying `char*` for the string to `wchar_t*`, the exact commit is `https://github.com/tree-sitter/tree-sitter-php/commit/b9dc1624520971536e6564ac36390d1595a12f01` which you can test by using the command I wrote above and checking out to the commit before that

amaanq:
I can park this for now until we have a better insight as to what's really causing the issue

PR #2937, opened 2024-02-06
refactor!: expose the allocator family of functions for consumption in scanners

amaanq:
This PR allows all downstream scanners to use the user-supplied tree-sitter alloc family of functions when using them. The libc functions were also aliased via a macro to allow for a seamless transition if users upgrade their lib/cli without updating the name of the functions. I consider it breaking since it required the functions to actually be *visible* in the shared library when built/linked against it, so any upgraded grammars will need to use a tree-sitter version beyond this commit (if/when merged)


amaanq:
Superseded by #3063

PR #2700, opened 2023-10-10
feature: debug graph output on Windows

yehorb:
As I understand, dot graph output was essentially disabled on Windows, as file handles are not file descriptors.

Turns out, it is relatively easy to [get a file descriptor](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-osfhandle?view=msvc-170) from a file handle, which works seamlessly with existing C code, with few exceptions.

By creating a simple wrapper file to cover platform differences, slightly altering `ffi` bindings, and enabling `print_tree_graph` and `log_graphs` functions, I managed to get proper dot graphs on Windows.

I found that there were no tests for the dot graph output specifically, so I carried out a bit of manual testing.

I tested the changes on some of my projects - mostly Python and Rust files, mostly around 100 lines of code. All of the resulting `log.html` files differed between platforms, but only in specifics due to dot rendering being slightly different. All of the generated texts and graphs were equivalent.

To verify the changes on a bigger file, I used the `tree-sitter-c` parser to parse the `lib/src/parser.c` file. Both on Windows and on Linux the `tree-sitter parse -D lib/src/parser.c` command produced a 500mb `log.html` file, both of which crashed my browser.

I was able to obtain the unprocessed `log.dot` file using [some hacks](https://github.com/tree-sitter/tree-sitter/commit/8c32d2436aac4ed7646c4e3b3fbf8e4dfe2158e7). These files were easier to reason about. While not identical, they had the same number of lines, and, from what I can tell, were equivalent, only differing in specifics such as concrete dot node names.

Overall, this seems to be working pretty well.

PR #2938, opened 2024-02-06
feat: support dot graph output on windows

amaanq:
Supersedes #2700, effectively rebased and tidied up, without adding another header/c file


dundargoc:
Good to go?

PR #2947, opened 2024-02-07
refactor!: remove the apply-all-captures flag, make last-wins precedence the default

amaanq:
I'll leave this up for a while before merging - I will have to synchronize changing the queries for every upstream language at the same time to ensure highlight tests don't break, and to let downstream know this is coming soon.

Since no complaints were raised for several months since adding the flag, I'm going to assume it'll stay that way

Ref: #2412


amaanq:
7 months in the making! I still don't like the highlighting iteration code overall though but improvements to that can come at a later time

PR #2952, opened 2024-02-07
ci: introduce "response bot"

dundargoc:
It's basically a bot that closes issues or PRs if the author hasn't
responded after 30 days. This is only done on issues/PRs that has the
`more-information-needed` label and is meant to be used for issues that
has hit a dead-end and cannot progress without further information from
the author. If the author responds after an issue is closed and gives
more context the issue can simply be reopened.

### DIFF
--- a/.github/scripts/close_unresponsive.js
+++ b/.github/scripts/close_unresponsive.js
@@ -1,0 +1,58 @@
+function labeledEvent(data) {
+  return (
+    data.event === "labeled" && data.label.name === "more-information-needed"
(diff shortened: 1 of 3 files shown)


dundargoc:
30 days is what neovim goes with so I went with that, I don't mind making it longer.

If people dislike the idea we can scrap it, but this has been very useful for neovim at least.

amaanq:
Nice, LGTM

amaanq (approved):

PR #2953, opened 2024-02-07
feat: use lockfiles to dedup recompilation

amaanq:
This is a simplified rewrite of #2647, without adding a multithreaded option to the cli and without deriving clone unnecessarily, that was a bad mistake I made there but I did that to spark the conversation as to why that might be, I wasn't 100% sure back then.

I think testing grammars that parse large repos will benefit immensely anyways from having a `-j` flag that tells the cli how many threads to run parsing on, it does help a lot when it's in the order of magnitude of minutes. This is the first step in getting that potentially done, but is also great in that it fixes the spurious racy CI failures which happen when one test compiles a parser and the other attempts to load it when it's partially been written to the fs.

fs4 also seems to be actively maintained, unlike the fs2 crate in the previous PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -484,6 +484,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs4"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


amaanq:
Ok, macos-x86 failed 3x in a row on an async test which is super weird - I need to investigate that more, those tests might just be poorly written tbh.

Cintah888 (approved):

PR #2963, opened 2024-02-08
ci: remove reviewers when drafting or closing a PR

dundargoc:


dundargoc:
In hindsight I don't think this is that important (for now at least) due to the low number of maintainers currently.

PR #2965, opened 2024-02-08
build: enable -pedantic

dundargoc:


amaanq:
Noo we should keep -std=gnu99, tree-sitter is meant to be c99 compliant :grin:

dundargoc:
Oh? My bad. Lemme revert then?

But wait, why are we using c11 features in that case? That's the only reason I upgrades to gnu11: because we are in fact already using c11 features and therefore not c99 compliant.

amaanq:
Oh wait, yeah, I checked patrickt's prior issue, this is fine then actually since this would break on older compilers. If we want to go back to C99 we can let Max decide if we should name the unions or not.

Issue #2803, opened 2023-11-30
Build failure against emscripten 3.1.50


p-linnane:
Hello 👋 . I'm a maintainer for the [Homebrew](https://brew.sh) project. While doing regression testing for v3.1.50 of emscripten, we noticed tree-sitter fails to build against it. Here is the error we are receiving:

```
  ==> make AMALGAMATED=1
  ==> make install PREFIX=/opt/homebrew/Cellar/tree-sitter/0.20.8
  ==> npm install -ddd --build-from-source --cache=/Users/brew/Library/Caches/Home
  ==> script/build-wasm
  Last 150 lines from /Users/brew/actions-runner/_work/homebrew-core/homebrew-core/bottles/logs/tree-sitter/04.build-wasm:
  2023-11-30 20:08:33 +0000
  
  script/build-wasm
  
  shared:INFO: (Emscripten: Running sanity checks)
  cache:INFO: generating system headers: sysroot_install.stamp... (this will be cached in "/private/tmp/tree-sitter-20231130-7843-mdiej/tree-sitter-0.20.8/.brew_home/.emscripten_cache/sysroot_install.stamp" for subsequent builds)
  cache:INFO:  - ok
  cache:INFO: generating system asset: symbol_lists/9f47be1a9b71038911a47600338fb9a693fa7f82.json... (this will be cached in "/private/tmp/tree-sitter-20231130-7843-mdiej/tree-sitter-0.20.8/.brew_home/.emscripten_cache/symbol_lists/9f47be1a9b71038911a47600338fb9a693fa7f82.json" for subsequent builds)
  cache:INFO:  - ok
  cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/pic/libGL.a... (this will be cached in "/private/tmp/tree-sitter-20231130-7843-mdiej/tree-sitter-0.20.8/.brew_home/.emscripten_cache/sysroot/lib/wasm32-emscripten/pic/libGL.a" for subsequent builds)
  system_libs:INFO: compiled 3 inputs in 0.20s
  cache:INFO:  - ok
  cache:INFO: generating system library: sysroot/lib/wasm32-emscripten/pic/libal.a... (this will be cached in "/private/tmp/tree-sitter-20231130-7843-mdiej/tree-sitter-0.20.8/.brew_home/.emscripten_cache/sysroot/lib/wasm32-emscripten/pic/libal.a" for subsequent builds)
[...]

PR #2971, opened 2024-02-09
fix(wasm): remove C++ mangled symbols

amaanq:
Probably fixes https://github.com/tree-sitter/tree-sitter/issues/2803


amaanq:
Well that was a painful 3 hours of finding that ruby bug, now we can get this in

amaanq:
ah ok, thanks for the heads up

PR #2972, opened 2024-02-09
chore: document preferred language for scanner

calebdw:
Hello!

This documents that C is the preferred language for external scanners as mentioned in https://github.com/tree-sitter/tree-sitter/pull/2897#issuecomment-1928112073

Thanks!

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -626,6 +626,9 @@ grammar({
 
 Then, add another C or C++ source file to your project. Currently, its path must be `src/scanner.c` or `src/scanner.cc` for the CLI to recognize it. Be sure to add this file to the `sources` section of your `binding.gyp` file so that it will be included when your project is compiled by Node.js and uncomment the appropriate block in your `bindings/rust/build.rs` file so that it will be included in your Rust crate.
 
+> **Note**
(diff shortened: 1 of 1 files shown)


amaanq:
We might even consider removing c++ support, it's just not needed.

calebdw:
I'm fine with that, I can update the docs if you want.

sogaiu:
@amaanq I guess you'd know pretty well about how much C++ use there is in practice given how much rewriting you've done.

Should I check via ts-questions for any remaining `scanner.cc` files?

amaanq:
yeah sure I'll check as well and rewrite any that seem to have some sizable usage/relevance, I've just finished rewriting Ruby (last upstream one in c++ that I didn't bother to touch :laughing:)

sogaiu:
Some rough results of find | grep | sort | uniq | wc -l:

* `grammar.js`: 423
* `scanner.c`: 149
* `scanner.cc`: 31

Number of repositories: 406

amaanq:
A few of those I already sent PRs to rewrite and either the owner just didn't respond or didn't want it (astro, beancount) :shrug: 
Ikatyang also looks to be very inactive (yaml/vue) :grimacing:

amaanq:
Merging for now then I think we can remove C++ support in a month or two

PR #2973, opened 2024-02-09
ci: only create release for normal semver tags

dundargoc:
We don't need releases for pre-releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   push:
     tags:
-      - v*
+      - v[0-9]+.[0-9]+.[0-9]+
 
 jobs:
   build:


clason:
Do you even need pre-release tags? The new one broke `npm i -g tree-sitter-cli` (and with it, nvim-treesitter CI).

clason:
(failures are from @amaanq's -- very welcome -- switch of tree-sitter-ruby to a C scanner)

amaanq:
Yeah I'm looking into it, only the emulated platforms fail which is really weird

amaanq:
> Do you even need pre-release tags? The new one broke `npm i -g tree-sitter-cli` (and with it, nvim-treesitter CI).

No, idk what Max was trying to do w/ that

dundargoc:
> Do you even need pre-release tags?

No, but I haven't figured out a way to control people remotely yet :)

PR #2979, opened 2024-02-09
build: add useful development targets to makefile

dundargoc:


dundargoc:
There's probably a more rust-y way of doing this, but that's a headache for another day.

PR #2983, opened 2024-02-09
Extracted subcommands

buckynbrocko:
Broke out the subcommand logic in `run` to individual `.run` functions for each Command enum's relevant struct.

Because the subcommands, don't all make use of the all the common variables (`config`, `loader`, etc.) the arguments of each `.run` are specific to its needs for now. Might could instead bundle these up into some Common struct to make the calling conventions identical and instead create a `SubCommandRun` or some-such trait, and singularly call `command.run(common)` instead of using a match statement. TBD?

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -294,6 +294,18 @@ struct BuildWasm {
     pub path: Option<String>,
 }
 
+impl BuildWasm {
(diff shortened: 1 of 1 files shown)


amaanq:
I think you have disabled allowing maintainers pushing to your PR's branches, so I'll close this and open a new PR with you as a coauthor, thanks!

PR #3002, opened 2024-02-11
ci: simplify workflows

dundargoc:


dundargoc:
I have a lot on my plate at the moment and I don't think I'll be able to work much on this for a while. The changes so far are still an improvement IMO so I think we can merge this.

amaanq:
no worries, it's definitely an improvement as is, thanks :grin:

PR #3015, opened 2024-02-13
ci(windows): parallelize tests

amaanq:
after the lockfile implementation this might work now :slightly_smiling_face:


amaanq:
Closing as stale, this can be revisited once file locking is standardized in the Rust stdlib

PR #3031, opened 2024-02-16
feat: allow specifying an external scanner's files

amaanq:
rebasing #2545 was more painful than just rewriting it


dundargoc:
Walking down the chain of supersedes, the original PR claimed to close https://github.com/tree-sitter/tree-sitter/issues/1246. Is that still relevant?

amaanq:
nah cuz that wanted a method to customize compilation args, which we deemed unnecessary

PR #3037, opened 2024-02-17
fix: don't log NUL characters

### LINKED ISSUE #709: Graphviz generation sometimes fails with syntax error for dot
I think it has something to do with whitespace literals-- probably something needs to be escaped in the dot input. Here's a small example that causes failure:

Grammar:

```javascript
module.exports = grammar({

  name: 'memexwiki',

  extras: _$ => [],

  rules: {

    source_file: $ => repeat1(seq(
      $.paragraph,
      repeat(seq(
        '\n\n',
        $.paragraph,
      )),
    )),

    paragraph: $ => seq(
      $.plain_text,
      repeat(seq(
        $.space,
        $.plain_text,
      )),
    ),

    space: _$ => ' ',

    plain_text: $ => prec.right(seq(
      $.word,
      repeat(seq(
        $.space,
        $.word,
      )),
    )),

    word: _$ => token(prec(-1, /\S+/)),

  },

});
```

Test input (test.txt):

```
Lorem ipsum domini
```

Command:

```
$ tree-sitter parse test.txt -D
```

Output:

```
(source_file [0, 0] - [1, 0]
  (paragraph [0, 0] - [0, 18]
    (plain_text [0, 0] - [0, 18]
      (word [0, 0] - [0, 5])
      (space [0, 5] - [0, 6])
      (word [0, 6] - [0, 11])
      (space [0, 11] - [0, 12])
      (word [0, 12] - [0, 18])))
  (ERROR [0, 18] - [1, 0]
    (ERROR [0, 18] - [1, 0])))
test.txt	0 ms	(ERROR [0, 18] - [1, 0])
Error: <stdin>: syntax error in line 522 near '"'
Dot failed:
error Command failed with exit code 1.
```

amaanq:
Graphviz dot will fail with this character present

Closes #709

### DIFF
--- a/lib/src/subtree.c
+++ b/lib/src/subtree.c
@@ -997,7 +997,7 @@ void ts_subtree__print_dot_graph(const Subtree *self, uint32_t start_offset,
     ts_subtree_lookahead_bytes(*self)
   );
 
-  if (ts_subtree_is_error(*self) && ts_subtree_child_count(*self) == 0) {
+  if (ts_subtree_is_error(*self) && ts_subtree_child_count(*self) == 0 && self->ptr->lookahead_char != 0) {
     fprintf(f, "\ncharacter: '%c'", self->ptr->lookahead_char);
   }
 


lib/src/subtree.c
```diff
@@ -997,7 +997,7 @@ void ts_subtree__print_dot_graph(const Subtree *self, uint32_t start_offset,
     ts_subtree_lookahead_bytes(*self)
   );
 
-  if (ts_subtree_is_error(*self) && ts_subtree_child_count(*self) == 0) {
+  if (ts_subtree_is_error(*self) && ts_subtree_child_count(*self) == 0 && self->ptr->lookahead_char != 0) {
     fprintf(f, "\ncharacter: '%c'", self->ptr->lookahead_char);
   }
```
theHamsta, line 1002: Shouldn't we a print a replacement character here e.g.NUL or \0

amaanq: Ueah maybe, well \0 itself wouldn't work but escaping it like "\\0" would if that's what you mean. I initially just thought about the EOF case but there's also some languages where the nul byte is valid in certain spots as well.

PR #3045, opened 2024-02-19
ci: cache emsdk

dundargoc:


dundargoc:
Turns out it was already cached, emsdk just seems very thicc so even with caching it takes time.

PR #3046, opened 2024-02-19
test: fix html tests

dundargoc:
"fragment" was replaced with "document" in the html parser, so these
need to be changed in the tests as well.


dundargoc:
Superseded by https://github.com/tree-sitter/tree-sitter/pull/3047

PR #3047, opened 2024-02-19
chore: update html tests

amaanq:


dundargoc:
The conventional commit police says that the commit type should be `test`.

PR #3050, opened 2024-02-20
feat(cli): add optional config path argument

### LINKED ISSUE #3001: command line option to assign config path
### Problem

```sh
tree-sitter --config=/the/path/of/new/config.json highlight /the/path/of/a/file
```

will ignore `~/.config/tree-sitter/config.json`

### Expected behavior

_No response_

WillLillis:
I was interested in the alternate config option described in #3001 and took a shot at implementing. If this isn't a desired feature my apologies :)

Closes #3001

### DIFF
--- a/cli/config/src/lib.rs
+++ b/cli/config/src/lib.rs
@@ -60,15 +60,21 @@ impl Config {
     /// Locates and loads in the user's configuration file.  We search for the configuration file
     /// in the following locations, in order:
     ///
+    ///   - Location specified by the path paramter if provided
     ///   - `$TREE_SITTER_DIR/config.json`, if the `TREE_SITTER_DIR` environment variable is set
     ///   - `tree-sitter/config.json` in your default user configuration directory, as determined
     ///     by [`dirs::config_dir`](https://docs.rs/dirs/*/dirs/fn.config_dir.html)
     ///   - `$HOME/.tree-sitter/config.json` as a fallback from where tree-sitter _used_ to store
     ///     its configuration
-    pub fn load() -> Result<Self> {
-        let Some(location) = Self::find_config_file()? else {
-            return Self::initial();
+    pub fn load(path: Option<PathBuf>) -> Result<Self> {
+        let location = match (path, Self::find_config_file()?) {
+            (Some(alt_path), _) => alt_path,
+            (None, Some(default_path)) => default_path,
+            _ => {
+                return Self::initial();
+            }
         };
+
         let content = fs::read_to_string(&location)
             .with_context(|| format!("Failed to read {}", &location.to_string_lossy()))?;
         let config = serde_json::from_str(&content)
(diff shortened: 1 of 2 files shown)


cli/config/src/lib.rs
```diff
@@ -60,15 +60,21 @@ impl Config {
...
-            return Self::initial();
+    pub fn load(path: Option<PathBuf>) -> Result<Self> {
+        let location = match (path, Self::find_config_file()?) {
+            (Some(alt_path), _) => alt_path,
+            (None, Some(default_path)) => default_path,
+            _ => {
+                return Self::initial();
+            }
         };
+
```
amaanq, line 77: Just a nitpick, i find it a bit cleaner to maybe do the following instead of the interleaving match

```suggestion
    pub fn load(path: Option<PathBuf>) -> Result<Self> {
        let location = if let Some(path) = path {
            path
        } else if let Some(path) = Self::find_config_file()? {
            path
        } else {
            return Self::initial();
        };
```

cli/src/main.rs
```diff
@@ -150,6 +150,8 @@ struct Parse {
         help = "Open `log.html` in the default browser, if `--debug-graph` is supplied"
     )]
     pub open_log: bool,
+    #[arg(long, help = "The path to an alternate config.json file")]
```
amaanq, line 153: replace alternate with alternative in each help spot

amaanq:
> If this isn't a desired feature my apologies :)

it's totally welcome, just extremely low priority for me atm

cli/config/src/lib.rs
WillLillis: Makes sense, I went back and forth between the two myself :)

WillLillis:
Thanks for the feedback!

amaanq:
awesome, thanks

cli/config/src/lib.rs
amaanq: yeah I tend to flip flop too, it's a case-by-case basis for me ;)

sogaiu:
Late to the party but thanks a lot for this.

Regarding spelling out a use case (query about this [here](https://github.com/tree-sitter/tree-sitter/issues/3001#issuecomment-1937894676)), one scenario is when testing code that depends on making use of the `tree-sitter` cli tool programmatically [1] and one wants to be sure it is [not accidentally picking up existing settings](https://github.com/tree-sitter/tree-sitter/discussions/2049).  One might pull such code down into `/tmp` (or on to some other machine or account) and run some tests only to find out later that the `parser-directories` setting being used was pointing to inappropriate locations.

Using `--config-path` with a path that stays within one's repository can help with this situation because invocations of the cli can be in scripts within one's repository so testing operations can remain "contained" / "isolated".  There is then no necessity to change / check on one's `~/.config/tree-sitter/config.json` (or equivalent on some other platform).

In working with the `tree-sitter` cli, I have spent a little too much time being puzzled by its magical / automatic features (which can be nice when expectations align with them), so having a way to say "please use these settings" (e.g. when the `--no-bindings` option was added, it was much appreciated) seems like a good addition.

Now if there were some way to specify where the grammar dynamic libraries live...may be there is?

---

[1] For example, to use the `parse` subcommand to test against code samples.

PR #3052, opened 2024-02-20
ci: don't cancel jobs on master

dundargoc:
It can be useful to see the full results of all master commits when
trying to bisect a regression.


amaanq (approved):

Cicci82:
Project

PR #3068, opened 2024-02-23
Fix crash when attempting to load ancient languages via wasm

maxbrunsfeld:
This fixes a crash that we were seeing in Zed when trying to load the old, abandoned Tree-sitter-swift grammar via Wasm.

### DIFF
--- a/lib/src/language.h
+++ b/lib/src/language.h
@@ -10,6 +10,9 @@ extern "C" {
 
 #define ts_builtin_sym_error_repeat (ts_builtin_sym_error - 1)
 
+#define LANGUAGE_VERSION_WITH_PRIMARY_STATES 14
+#define LANGUAGE_VERSION_USABLE_VIA_WASM 13
+
 typedef struct {
   const TSParseAction *actions;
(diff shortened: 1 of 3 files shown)


lib/src/language.h
```diff
@@ -10,6 +10,9 @@ extern "C" {
 
 #define ts_builtin_sym_error_repeat (ts_builtin_sym_error - 1)
 
+#define LANGUAGE_VERSION_WITH_PRIMARY_STATES 14
+#define LANGUAGE_VERSION_USABLE_VIA_WASM 13
```
amaanq, line 14: Having a `NOTE`/`TODO` to drop this when we get to abi 15/1.0 might be useful to include (easy to find)

amaanq (approved):

lib/src/language.h
maxbrunsfeld: I think it would be ok to keep these around even if we change the API, unless we do so in a backward-incompatible change (and bump the MIN_COMPATIBLE_VERSION).

PR #3070, opened 2024-02-23
fix: wrap `||` comparison in parenthesis when `&&` is used

amaanq:
Subtle bug that probably didn't have any side effects, but better safe than sorry

I noticed this in tree-sitter-python when building the rust crate:

```
warning: tree-sitter-python@0.20.4: src/parser.c: In function ‘ts_lex’:
warning: tree-sitter-python@0.20.4: src/parser.c:8359:16: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
warning: tree-sitter-python@0.20.4:  8359 |       if (!eof && lookahead == 0 ||
warning: tree-sitter-python@0.20.4:       |           ~~~~~^~~~~~~~~~~~~~~~~
```

this code is probably not intended to be generated like this:
```
case 26:
      if (!eof && lookahead == 0 ||
          lookahead == '\n') ADVANCE(168);
```

It reads as `if ((!eof && lookahead == 0) || lookahead == '\n')` when it should be `if (!eof && (lookahead == 0 || lookahead == '\n'))`

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -871,6 +871,9 @@ impl Generator {
             line_break.push_str("  ");
         }
 
+        // parenthesis needed if we add the `!eof` condition to explicitly avoid confusion with
+        // precedence of `&&` and `||`
+        let (mut need_open_paren, mut need_close_paren) = (false, false);
         for (i, range) in ranges.iter().enumerate() {
             if is_included {
                 if i > 0 {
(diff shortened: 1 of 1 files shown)


slackner:
@amaanq I believe your change caused a regression. Please check https://github.com/tree-sitter/tree-sitter/pull/3274 with a suggested fix.

PR #3076, opened 2024-02-25
Use workspace dependencies for internal crates like Tree-sitter

maxbrunsfeld:
This way, the Rust version is just located in the top-level Cargo.toml.


amaanq:
Nice, this is great

PR #3077, opened 2024-02-25
refactor(bindings)!: convert node bindings to NAPI

segevfiner:
This is a breaking change that accompanys https://github.com/tree-sitter/node-tree-sitter/pull/190 and should only be merged along with it, with a proper strategy for how to update languages.

We need to consider how we are going to release such a thing as language authors will have to know to regenarate or modify those files...


ObserverOfTime:
Rust and Python will also have breaking changes soon, so it's fine to break Node as well.
Keep in mind that you will have to rebase once #2438 is merged since it also edits the bindings.

One question: how would `binding.cc` look like for multi-language parsers (e.g. [typescript](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/bindings/node/binding.cc)?

segevfiner:
Just `Object::New` and do the same, and just set those objects on `exports`, instead of directly treating exports as a language. `exports` is just a plain object and is there for convenience and you can really return anything.

cli/src/generate/templates/binding.cc
```diff
@@ -1,28 +1,18 @@
 #include "tree_sitter/parser.h"
-#include <node.h>
-#include "nan.h"
+#include <napi.h>
 
-using namespace v8;
+using namespace Napi;
```
verhovsky, line 4: I think you should remove this and just use `Napi::` everywhere

segevfiner: Can be done. The examples often do `use namespace Napi`, but you obviously don't have too.

cli/src/generate/templates/binding.cc
```diff
@@ -1,28 +1,16 @@
 #include "tree_sitter/parser.h"
-#include <node.h>
-#include "nan.h"
-
-using namespace v8;
+#include <napi.h>
 
 extern "C" TSLanguage * tree_sitter_PARSER_NAME();
 
 namespace {
```
verhovsky, line 6:
```suggestion
```

seems unnecessary and the examples don't have it https://github.com/nodejs/node-addon-examples/blob/main/src/1-getting-started/6_object_wrap/node-addon-api/addon.cc

cli/src/generate/templates/binding.cc
```diff
@@ -1,28 +1,16 @@
...
-  Nan::Set(module, Nan::New("exports").ToLocalChecked(), instance);
+Napi::Object Init(Napi::Env env, Napi::Object exports) {
+  exports["name"] = Napi::String::New(env, "PARSER_NAME");
+  exports["language"] = Napi::External<TSLanguage>::New(env, tree_sitter_PARSER_NAME());
+  return exports;
 }
 
-NODE_MODULE_CONTEXT_AWARE(tree_sitter_PARSER_NAME_binding, Init)
-
 }  // namespace
```
verhovsky, line 14:
```suggestion
```

cli/src/generate/templates/js-binding.cc
```diff
@@ -1,29 +1,12 @@
-#include "nan.h"
-#include <node.h>
+#include "tree_sitter/parser.h"
+#include <napi.h>
 
-using namespace v8;
+extern "C" TSLanguage * tree_sitter_PARSER_NAME();
```
verhovsky, line 4:
```suggestion
extern "C" TSLanguage *tree_sitter_PARSER_NAME();
```

verhovsky: last one, sorry

amaanq: the const qualifier will be discarded anyways at its call site, so its sorta useless since you'd need to cast it to `void ` where used

verhovsky (approved):

maxbrunsfeld (approved):

amaanq (approved):

amaanq:
thanks @segevfiner, I'll try and help out/review the PR in node-ts as well

segevfiner:
Just note that as it stands this will only work with node-tree-sitter with that PR merged as there is no backwards or forwards compatibility in place. So be sure to merge it before releasing this.

amaanq:
Yeah I know, we're not gonna have a release until then

ObserverOfTime:
> Just note that as it stands this will only work with node-tree-sitter with that PR merged as there is no backwards or forwards compatibility in place.

I did add forwards compatibility. The CLI will replace the old bindings with the new ones.

segevfiner:
I mean for importing the grammars in the node bindings. The non-napi node bindings can't import napi grammars and vice versa.

PR #3086, opened 2024-02-26
Fix invalid parse stack recursive merging with mismatched error cost

maxbrunsfeld:
Allowing this invalid merge caused an invariant to be violated later on during parsing, when handling a later error.


amaanq:
Oops, let me fix the macos build bug w/ the python parser

PR #3087, opened 2024-02-27
test: don't use TS_REUSE_ALLOCATOR on Darwin systems

amaanq:
For some reason, the linker seems to behave a bit differently with `-exported_symbols_list` on macOS **only** for tests vs other operating systems, so we'll disable this for now

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -30,9 +30,6 @@ fn main() {
         target_os = "dragonfly",
     ))]
     println!("cargo:rustc-link-arg=-Wl,--dynamic-list=cli/dynamic-symbols.txt");
-
-    #[cfg(any(target_os = "macos", target_os = "ios"))]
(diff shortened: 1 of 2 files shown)


cli/build.rs
```diff
@@ -30,9 +30,6 @@ fn main() {
         target_os = "dragonfly",
     ))]
     println!("cargo:rustc-link-arg=-Wl,--dynamic-list=cli/dynamic-symbols.txt");
-
-    #[cfg(any(target_os = "macos", target_os = "ios"))]
-    println!("cargo:rustc-link-arg=-Wl,-exported_symbols_list,cli/dynamic-symbols-darwin.txt");
```
maxbrunsfeld, line 35: do we still need the `dynamic-symbols-darwin.txt` file?

amaanq: not currently, but I sort of wanted to take a deeper dive and figure out why it doesn't work only for macOS, and see if I can fix it probably tomorrow.

maxbrunsfeld: Ok, sounds good.

PR #3088, opened 2024-02-27
build: use c11 everywhere

ObserverOfTime:
This should allow core to use the same header while disabling it by default in scanners, though I haven't tested it.

I changed the name to `TREE_SITTER_REUSE_ALLOCATOR` to be more consistent with `TREE_SITTER_FEATURE_WASM`.


maxbrunsfeld:
Do we need to require C11 mode? What newer C features are we using?

Package.swift
```diff
@@ -34,6 +35,9 @@ let package = Package(
                     "src/tree.c",
                     "src/query.c"
                 ],
+                cSettings: [
+                    .define("TREE_SITTER_REUSE_ALLOCATOR")
```
maxbrunsfeld, line 39: Why does this macro need to be defined when compiling the core library? I thought this was an optional macro you could define when compiling an external scanner.

build.zig
```diff
@@ -8,7 +8,10 @@ pub fn build(b: *std.Build) void {
     });
 
     lib.linkLibC();
-    lib.addCSourceFile(.{ .file = .{ .path = "lib/src/lib.c" }, .flags = &.{} });
+    lib.addCSourceFiles(&.{
+        "lib/src/lib.c",
+    }, &.{"-std=gnu11"});
+    lib.defineCMacroRaw("DTREE_SITTER_REUSE_ALLOCATOR");
```
maxbrunsfeld: Same question here - I think `TREE_SITTER_REUSSE_ALLOCATOR`  should only be needed when compiling an external scanner.

lib/binding_rust/build.rs
```diff
@@ -51,6 +51,7 @@ fn main() {
         .flag_if_supported("-fvisibility=hidden")
         .flag_if_supported("-Wshadow")
         .flag_if_supported("-Wno-unused-parameter")
+        .define("TREE_SITTER_REUSE_ALLOCATOR", "")
```
maxbrunsfeld, line 54: Same question here. I'd like to not require that this macro be defined.

lib/compile_flags.txt
```diff
@@ -1,4 +1,5 @@
 -std=c99
 -Isrc/wasm
 -Iinclude
--D TREE_SITTER_FEATURE_WASM
\ No newline at end of file
+-D TREE_SITTER_FEATURE_WASM
+-D TREE_SITTER_REUSE_ALLOCATOR
```
maxbrunsfeld, line 5: Same question about this.

script/generate-bindings
```diff
@@ -40,4 +40,5 @@ bindgen                                        \
   $header_path                                 \
   --                                           \
   -D TREE_SITTER_FEATURE_WASM                  \
+  -D TREE_SITTER_REUSE_ALLOCATOR               \
```
maxbrunsfeld, line 43: I don't think the bindgen call should need to set this macro either.

amaanq:
> Do we need to require C11 mode? What newer C features are we using?

anonymous unions, a bit more info [here](https://github.com/tree-sitter/tree-sitter/pull/837) and [here](https://github.com/tree-sitter/tree-sitter/pull/2965)

Package.swift
ObserverOfTime: So it should be used in the CLI and scanners but not in libraries and bindings?

amaanq: You're right, I suggested that to him and that was a mistake on my part. 

The code was deduped so there's effectively only 1 `alloc.h`. The core lib defining this will use the default `ts_current_*` functions which are just wrappers around the libc functions that abort on failure, but you're right in that the CLI is probably the only spot that should define this when compiling the lib. But otherwise, by default malloc/realloc/calloc/free would be used instead of `ts_malloc_default` and friends

ObserverOfTime: The `ts_*` memory allocation functions are now behind the same flag in core as in scanners, so these changes retain the previous behaviour (as far as I understood it) while eliminating the need to have 2 `alloc.h` files.

maxbrunsfeld: I think that we should retain the old behavior without requiring a new command line flag.

As for how we implement that, the `TREE_SITTER_REUSE_ALLOCATOR` macro (or some other macro that is checked in the `ifdef` could be defined in source code, prior to including `alloc.h`.

maxbrunsfeld:
> anonymous unions

I see, thanks. Do you have a sense of how widely-supported the C11 standard is? What's the difference between `gnu11` vs `c11`?

ObserverOfTime:
According to Wikipedia it should be supported in every non-EOL system:

> Some features of C11 are supported by the [GCC](https://en.wikipedia.org/wiki/GNU_Compiler_Collection) starting with version 4.6,[[16]](https://en.wikipedia.org/wiki/C11_(C_standard_revision)#cite_note-16) [Clang](https://en.wikipedia.org/wiki/Clang) starting with version 3.1,[[17]](https://en.wikipedia.org/wiki/C11_(C_standard_revision)#cite_note-17) [IBM XL C](https://en.wikipedia.org/wiki/IBM_XL_C) starting with version 12.1,[[18]](https://en.wikipedia.org/wiki/C11_(C_standard_revision)#cite_note-18) and [Microsoft Visual C++](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B) starting with VS 2019 (16.8)[[19]](https://en.wikipedia.org/wiki/C11_(C_standard_revision)#cite_note-19) in September 2020.

`gnu11` enables [GNU extensions](https://gcc.gnu.org/onlinedocs/gcc/C-Extensions.html).
The Makefile already used it so I added it everywhere in core.
Parsers/scanners likely won't need any extensions so I made them `c11`.

Package.swift
amaanq: so an idea I had for that is to define an internal flag for this as well, something like `TS_REUSE_ALLOCATOR` and the `TREE_SITTER` variant is the public one. but this would have a check for *either* flag being defined before it's included in any of the c source files, which might feel a little hacky to do this in both source files + the OR'd check in alloc.h. do you have any other ideas?

maxbrunsfeld: I think the simplest thing may be to just keep the duplicated header. There' not much code in there.

We don't need the `TS_PUBLIC` thing in the header, right? I would think those attributes only need to appear on the *definition* of the symbols (in alloc.c), not on their declarations?

amaanq (approved):

lib/compile_flags.txt
```diff
@@ -1,4 +1,5 @@
--std=c99
+-std=c11
 -Isrc/wasm
 -Iinclude
--D TREE_SITTER_FEATURE_WASM
\ No newline at end of file
+-D TREE_SITTER_FEATURE_WASM
+-D _POSIX_C_SOURCE=200112L
```
amaanq, line 5: probs not needed since the source files have them

PR #3098, opened 2024-02-27
test: add test case for parse stack merging with incorrect error cost bug

amaanq:
segfaults before #3086, but is fine now :grin:


maxbrunsfeld:
Nice, great idea.

PR #3099, opened 2024-02-27
refactor: remove extern/const where possible

ObserverOfTime:
`extern` is redundant in function prototypes.
`const` is discarded in most if not all binding functions.

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -263,7 +263,6 @@ impl Generator {
 
     fn add_pragmas(&mut self) {
         add_line!(self, "#if defined(__GNUC__) || defined(__clang__)");
-        add_line!(self, "#pragma GCC diagnostic push");
         add_line!(
             self,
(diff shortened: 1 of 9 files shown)


amaanq (approved):

cli/src/generate/render.rs
```diff
@@ -263,7 +263,6 @@ impl Generator {
 
     fn add_pragmas(&mut self) {
         add_line!(self, "#if defined(__GNUC__) || defined(__clang__)");
-        add_line!(self, "#pragma GCC diagnostic push");
```
maxbrunsfeld, line 266: why this change?

ObserverOfTime: No point in pushing if it's not popped. It will apply till the end of the file anyway.

maxbrunsfeld: oh ok, sounds good.

PR #3103, opened 2024-02-28
refactor(wasm)!: make `current*`, `is*`, and `has*` methods properties

### LINKED ISSUE #2195: TreeCursor.currentNode is a property in Node but a function in the browser
At least in the types. Compare Node:

https://github.com/tree-sitter/node-tree-sitter/blob/16f7319c993e07a4b28afad354b80624237c6660/tree-sitter.d.ts#L107

With binding_web:

https://github.com/tree-sitter/tree-sitter/blob/c6d663f3936aea4fb9a12e093fbaa25e762fa902/lib/binding_web/tree-sitter-web.d.ts#L118

verhovsky:
This is a breaking change to make the Wasm and Node bindings more interchangeable

Closes #2195

### DIFF
--- a/docs/assets/js/playground.js
+++ b/docs/assets/js/playground.js
@@ -175,7 +175,7 @@ let tree;
           const start = cursor.startPosition;
           const end = cursor.endPosition;
           const id = cursor.nodeId;
-          let fieldName = cursor.currentFieldName();
+          let fieldName = cursor.currentFieldName;
           if (fieldName) {
             fieldName += ': ';
           } else {
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -249,27 +249,27 @@ class Node {
     return C._ts_node_next_parse_state_wasm(this.tree[0]);
   }
 
-  isNamed() {
+  get isNamed() {
     marshalNode(this);
     return C._ts_node_is_named_wasm(this.tree[0]) === 1;
   }
 
-  hasError() {
+  get hasError() {
     marshalNode(this);
     return C._ts_node_has_error_wasm(this.tree[0]) === 1;
   }
 
-  hasChanges() {
+  get hasChanges() {
     marshalNode(this);
     return C._ts_node_has_changes_wasm(this.tree[0]) === 1;
   }
 
-  isError() {
+  get isError() {
     marshalNode(this);
     return C._ts_node_is_error_wasm(this.tree[0]) === 1;
   }
 
-  isMissing() {
+  get isMissing() {
     marshalNode(this);
     return C._ts_node_is_missing_wasm(this.tree[0]) === 1;
   }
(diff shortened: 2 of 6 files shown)


maxbrunsfeld:
@verhovsky Just curious - do you have thoughts on which interface makes more sense? We should definitely make the two interfaces consistent, but which interface should we change, the Node one, or this one? From an idiomatic JS perspective, when should we use getters and when not?

maxbrunsfeld:
I'm fine with making breaking changes, but I want to be open to making breaking changes in the other direction (to the Node binding) since we're already making breaking changes there due to the NAPI stuff. I just want to check in about which methods make sense to be getters.

One of the principles I was going for with the interface was to make the API feel somewhat like the most standard JS API for accessing trees, which is the web DOM APIs. I remember the DOM uses a lot of getters, but I don't remember the detailed shape of the APIs. Curious if you (or other folks who use this web binding) have thoughts.

verhovsky:
The way this PR is, we'll need breaking changes in Node as well, to `isMissing()` and `hasChanges()` and `hasError()`. Right now in the Node bindings we have

```ts
      isNamed: boolean; // <- not a function
      hasChanges(): boolean;
      hasError(): boolean;

      isMissing(): boolean;
      readonly currentNode: SyntaxNode;

      readonly currentFieldName: string;
```

and in Wasm

```ts
      isNamed(): boolean;
      hasChanges(): boolean;
      hasError(): boolean;
      isError(): boolean;
      isMissing(): boolean;
      currentNode(): SyntaxNode;
      currentFieldId(): number;
      currentFieldName(): string;
```

`isNamed`, `currentNode` and `currentFieldName` are incompatible and if we change `hasChanges`, etc. as in this PR we'll need to change it in Node too.

I think it makes sense for these to be properties because they don't change state or iterate the tree or perform any action really, just copy bytes from the given object from C into a JavaScript object and return it.

The DOM seems to consistently follow this too, things like [`Node.nextSibling`](https://developer.mozilla.org/en-US/docs/Web/API/Node/nextSibling) or [`Document.hidden`](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden) are properties. I don't see many functions with no arguments.

It doesn't make sense to me to make things that don't modify state in any meaningful way be functions. I could understand things that require traversing the tree being functions though. It's a minor annoyance to have to type `()` after `currentNode`, especially when it doesn't change any state. If tree-sitter was a pure JavaScript library then `.currentNode` would definitely just be the property it uses internally for tracking the current node and `isError` would probably just be a property of a node.

ObserverOfTime:
I think `has*` (maybe `is*` as well?) make more sense as methods.

maxbrunsfeld:
Yeah, I agree with `currentNode`, `currentFieldName`, and `currentFieldId` being properties. For the rest, I could go either way. I'm ok with either methods or properties. It'll be good to have it consistent between Node and Warm.

verhovsky:
People seemed opposed to changing the other methods to properties on Discord so I undid that change so this PR can be merged. We will still have to change `isNamed` in the Node bindings to be a function to be consistent with the Wasm bindings.

But I'm still convinced that these things should be properties. They are just reading a value that already exists in memory, wrapping it in a JavaScript object and returning it. No change of state happens. The only function that does any sort of computation, `hasError` barely does any (an if statement and checks if a number exceeds a threshold). Semantically, that's a property. These are already properties in the Python bindings and everything else around these things in the JavaScript code is also already `get` properties, which kind of suggests this was just an oversight. There's no reason to make people type `()` to read a piece of data on an object.

That being said, more breaking changes will be really annoying, and since it's not as necessary as changing `.currentNode` and friends, for the sake of backwards compatibility I'm okay with not doing it if people are against it.

maxbrunsfeld:
Ok, I'm ok with either direction of change, to make the two bindings consistent.

PR #3106, opened 2024-02-28
Add typings for the node bindings

### LINKED ISSUE #3072: Add type definitions to language bindings
### Problem

Add TypeScript type definitions to generated language bindings, so that TypeScript won't complain when importing them.

They can also include actual typing, so that TypeScript will only allow passing a proper language object to `setLangauge` and types for the exported `nodeTypeInfo`. Though that may be difficult, as TypeScript does structural typing and doesn't have a proper concept of a newtype, and emulating it using `unique symbol` will require importing the symbol from a shared place, which requires some new dependency.

### Expected behavior

For language grammars in Node to have TypeScript typings.

segevfiner:
One annoying thing here is that since we use `export =` we can't also export the types themselves with names from the typing file... e.g. To make it possible to declare variables using those types by name rather than `typeof`. CJS/ESM is fun...

P.S. It is possible to reference `dsl.d.ts` by `tree-sitter-cli/dsl`, see https://github.com/tree-sitter/tree-sitter-javascript/blob/51aed7263d33bee46b0a8a5e165ffcb690d5aca8/grammar.js#L10, I think we might want to point there instead of generating `types/dsl.d.ts` with the grammar. @amaanq

Fixes #3072


ObserverOfTime:
Did you test them?

segevfiner:
In one of my own projects by coping and pasting into a local .d.ts file.

cli/src/generate/grammar_files.rs
```diff
@@ -97,16 +98,21 @@ pub fn generate_grammar_files(
...
+            }) || package_json_types.is_none();
             if package_json_needs_update {
                 eprintln!("Updating package.json with new binding path");
                 package_json.insert(
                     "main".to_string(),
                     Value::String("bindings/node".to_string()),
                 );
+                package_json.insert(
+                    "types".to_string(),
+                    Value::String("bindings/node/index.d.ts".to_string()),
```
ObserverOfTime, line 114:
```suggestion
                    Value::String("bindings/node".to_string()),
```

cli/src/generate/templates/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "CAMEL_PARSER_NAME grammar for tree-sitter",
   "repository": "github:tree-sitter/tree-sitter-PARSER_NAME",
   "license": "MIT",
+  "types": "bindings/node/index.d.ts",
```
ObserverOfTime, line 7:
```suggestion
  "types": "bindings/node",
```
`index.d.ts` is the default.

cli/src/generate/grammar_files.rs
```diff
@@ -97,16 +98,21 @@ pub fn generate_grammar_files(
             let mut package_json = serde_json::from_str::<Map<String, Value>>(&package_json_str)
                 .with_context(|| "Failed to parse package.json")?;
             let package_json_main = package_json.get("main");
+            let package_json_types = package_json.get("types");
             let package_json_needs_update = package_json_main.map_or(true, |v| {
```
ObserverOfTime, line 102: I think this old check should be removed. Instead we should add checks for the other newly-added or changed fields. That's out of scope for this PR but if you want to handle it you can.

segevfiner: I might handle it in a followup PR that I'm creating that adds prebuild, assuming this one get's merged.

ObserverOfTime: Personally, I'd prefer `prebuildify` over `prebuild` (ideally without `node-gyp-build`). Most parser binaries are not particularly large so bundling them would be fine. Worst case, we can limit the prebuilds to the major platforms (win-x64, linux-x64, mac-x64, mac-arm64). The [workflow](https://github.com/tree-sitter/workflows/blob/main/.github/workflows/package-npm.yml) can be adapted accordingly.

amaanq: > I think this old `main` check should be removed. Instead we should add checks for `files` and `dependencies` (`nan` -> `node-addon-api`). That's out of scope for this PR but if you want to handle it you can.

Just did this

amaanq:
thanks @segevfiner

PR #3109, opened 2024-02-28
Type tag the language in the node bindings

segevfiner:
This will help make sure we won't unwrap unrelated external objects if they are passed and possibly seg fault. To be used by https://github.com/segevfiner/node-tree-sitter/pull/1

### DIFF
--- a/cli/src/generate/templates/js-binding.cc
+++ b/cli/src/generate/templates/js-binding.cc
@@ -4,11 +4,7 @@ typedef struct TSLanguage TSLanguage;
 
 extern "C" TSLanguage *tree_sitter_PARSER_NAME();
 
-// tstag() {
-//   b2sum -l64 <(printf tree-sitter) <(printf "$1") | \
-//   awk '{printf "0x" toupper($1) (NR == 1 ? ", " : "\n")}'
-// }
-// tstag language # => 0x8AF2E5212AD58ABF, 0xD5006CAD83ABBA16
+// "tree-sitter", "language" hashed with BLAKE2
 const napi_type_tag LANGUAGE_TYPE_TAG = {
   0x8AF2E5212AD58ABF, 0xD5006CAD83ABBA16
 };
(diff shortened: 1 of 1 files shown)


cli/src/generate/templates/js-binding.cc
```diff
@@ -4,9 +4,15 @@ typedef struct TSLanguage TSLanguage;
 
 extern "C" TSLanguage *tree_sitter_PARSER_NAME();
 
+const napi_type_tag LANGUAGE_TYPE_TAG = {
+  0x95840BEBF71E4E90, 0x9DC9419B874C0271
```
ObserverOfTime, line 8: But what are these magic numbers?

ObserverOfTime: Does it have to be a UUID? It'd be nice if it could represent tree-sitter:
```bash
b2sum -l64 <(printf tree) <(printf sitter) | \
  awk '{printf "0x" toupper($1) (NR == 1 ? ", " : "\n")}'
# 0x7FA28BFC1966AC2D, 0xA24BD16E50AD8E52
```
In any case, there should be a small explanation in a comment.

segevfiner: Dunno. They specified a UUID... But I don't think anything stops you from using any magic number. You can probably even hex encode tree-sitter or something...

cli/src/generate/templates/js-binding.cc
```diff
@@ -4,9 +4,20 @@ typedef struct TSLanguage TSLanguage;
 
 extern "C" TSLanguage *tree_sitter_PARSER_NAME();
 
+// tstag() {
+//   b2sum -l64 <(printf tree-sitter) <(printf "$1") | \
+//   awk '{printf "0x" toupper($1) (NR == 1 ? ", " : "\n")}'
+// }
+// tstag language # => 0x8AF2E5212AD58ABF, 0xD5006CAD83ABBA16
+const napi_type_tag LANGUAGE_TYPE_TAG = {
```
ObserverOfTime, line 12: A simple comment is enough here.
```suggestion
// "tree-sitter", "language" hashed with BLAKE2
const napi_type_tag LANGUAGE_TYPE_TAG = {
```

ObserverOfTime (approved):

PR #3111, opened 2024-02-29
Fix regression in `subtree_compare`

daumantas-kavolis-sensmetry:
Fixes regression introduced by #2788, which changed comparison to last-one wins instead of first-one. We saw a regression in conditional (ternary) expressions where nested operator expressions in else branch would bind to the whole conditional expression instead of the else branch.


maxbrunsfeld:
Nice debugging @daumantas-kavolis-sensmetry , thanks so much.

PR #3114, opened 2024-02-29
docs: Add `Ohm` language parser

novusnota:
## Brief description

Adding the complete parser of the [Ohm](https://ohmjs.org/) language.
It has highlights, locals and tags generic queries, as well as editor-specific ones.
Besides, the parser itself, tag and highlighting queries are test-covered.

## About the language

Ohm is a parsing toolkit consisting of a library and a domain-specific language. One can use it to parse custom file formats or quickly build parsers, interpreters, and compilers for programming languages.

The Ohm language is based on [parsing expression grammars](http://en.wikipedia.org/wiki/Parsing_expression_grammar) (PEGs), which are a formal way of describing syntax, similar to regular expressions and context-free grammars. The Ohm library provides a JavaScript interface for creating parsers, interpreters, and more from the grammars you write.


maxbrunsfeld:
Thanks!

PR #3120, opened 2024-03-03
remove repetitive code

geekvest:


sogaiu:
Changes look ok -- they appear to be replacing "the the" instances with "the" in comments.

PR #3123, opened 2024-03-03
fix(cli): only output the sources with `--no-bindings`

ObserverOfTime:
In a multi-language repo, running `tree-sitter generate --no-bindings` in a language subdirectory will still generate `.editorconfig`, `.gitignore`, `.gitattributes`, `types/dsl.d.ts` and modify `package.json`, which is undesirable.
This is a stopgap solution until (a) the commands are separated and (b) multi-language generation is improved.

### DIFF
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -113,11 +113,11 @@ pub fn generate_grammar_files(
                 fs::read_to_string(path).with_context(|| "Failed to read package.json")?;
             let mut package_json = serde_json::from_str::<Map<String, Value>>(&package_json_str)
                 .with_context(|| "Failed to parse package.json")?;
-            let package_json_types = package_json.get("types");
-            let package_json_needs_update = package_json
-                .get("dependencies")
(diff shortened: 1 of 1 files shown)


cli/src/generate/mod.rs
```diff
@@ -74,7 +74,7 @@ pub fn generate_parser_in_directory(
                 .strip_prefix("tree-sitter-")
                 .or_else(|| Some(dir_name.as_ref()))
             {
-                grammar_files::generate_grammar_files(&repo_path, language_name, false)?;
+                grammar_files::generate_grammar_files(&repo_path, language_name)?;
```
ObserverOfTime, line 77: Should this be removed instead?

amaanq: it should only generate grammar.js i think, but I'm not sure that makes sense anymore since it'd just use the default template to generate from. we could remove it and have the user create it instead, but it's nice to have a template created

ObserverOfTime: It will also generate `package.json`.

DyroxOne: Ja

PR #3126, opened 2024-03-04
feat(bindings): add peerDependencies for npm

ObserverOfTime:
Also add the "files" array if missing.

Requires #3109 and the node-tree-sitter update.

### DIFF
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -1,16 +1,12 @@
 use super::write_file;
 use anyhow::{anyhow, Context, Result};
-use filetime::FileTime;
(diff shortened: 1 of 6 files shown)


DyroxOne (approved):

ObserverOfTime:
I'm afraid I don't know enough rust to fix this.

```
error[E0499]: cannot borrow `package_json` as mutable more than once at a time
   --> cli/src/generate/grammar_files.rs:143:40
    |
121 |                 let dependencies = package_json
    |                                    ------------ first mutable borrow occurs here
...
143 |                 let dev_dependencies = package_json
    |                                        ^^^^^^^^^^^^ second mutable borrow occurs here
...
150 |                     dependencies.insert(
    |                     ------------ first borrow later used here
```

amaanq (approved):

PR #3131, opened 2024-03-05
feat: add to_sexp_pretty to rust_binding

rhapsodyn:
Just adding a helper for ease of debugging.


amaanq:
well, the code for this already exists with edge cases tackled in `test.rs` (format_sexp) - duplicating it would be fine since it's for different use cases.

PR #3132, opened 2024-03-05
Fix windows

amaanq:


cli/src/generate/templates/alloc.h
```diff
@@ -9,16 +9,10 @@ extern "C" {
...
-#define TS_PUBLIC
-#else
-#define TS_PUBLIC __attribute__((visibility("default")))
-#endif
-
-TS_PUBLIC extern void *(*ts_current_malloc)(size_t);
-TS_PUBLIC extern void *(*ts_current_calloc)(size_t, size_t);
-TS_PUBLIC extern void *(*ts_current_realloc)(void *, size_t);
-TS_PUBLIC extern void (*ts_current_free)(void *);
+extern void *(*ts_current_malloc)(size_t);
```
ObserverOfTime, line 12: If `TS_PUBLIC` is no longer defined here then the `#undef` in `render.rs` is unnecessary.

Also you could probably move these within the `#ifdef` since they're not needed otherwise.

PR #3159, opened 2024-03-11
fix(make): Allow to override CFLAGS on the commandline

cryptomilk:
If you package tree-sitter for a distribution, the distribution is setting the CFLAGS which should be used for all packages.  Allow those flags to be used.


amaanq:
makes sense - sorry about that

Issue #2669, opened 2023-09-26
Please consider lowering MSRV


Wilfred:
In difftastic, I use tree-sitter extensively, but I'm trying to support the last 12 months of Rust versions (1.64 was released 22nd September 2022) as well as Debian stable (currently [1.63](https://tracker.debian.org/pkg/rustc)).

I see the MSRV was specified in #2167 by @ahlinc, which was released in tree-sitter 0.20.10. This means that I can't upgrade beyond 0.20.9.

Would you be willing to lower the MSRV to e.g. 1.60? As far as I can tell tree-sitter doesn't depend on any newer Rust features.

ahlinc:
I'm very open to make the MSRV much lower and for the _tree-sitter_ crate lower it maximally. At the end of this week there will be a new tree-sitter release so I'll include this request in it.

dundargoc:
Actually from quick testing it looks like treesitter depends on rust version 1.70.0+. Closing as this is not possible.

Wilfred:
@dundargoc are you sure? Cargo.toml has rust 1.65 as the MSRV, I don't see anything that obviously requires 1.70. The CLI crate doesn't seem to override the MSRV either.

Limiting all tree-sitter binaries to rust 1.70 or later would cause issues for difftastic, as it would prevent me building on distros that upgrade rustc less frequently.

Would you be opposed to PRs that lower the MSRV? This is only a concern for the library, the CLI is absolutely fine :)

amaanq:
Using cargo-msrv, it does show that 1.70.0 is the minimum requirement because of `which`, but if you only need the lib to have a lower msrv that would be totally doable

dundargoc:
> as it would prevent me building on distros that upgrade rustc less frequently.

That is generally the price of choosing to use a super "stable" distro. Humanity wouldn't have left caves if we all did what debian do.

> Would you be opposed to PRs that lower the MSRV? This is only a concern for the library, the CLI is absolutely fine :)

Tentative yes, although I'll defer to my more experienced colleagues.

dundargoc:
> is this an issue with Ubuntu/Debian?

Debian stable specifically. Debian unstable has rust 1.70.0.

This feels like a conflicting set of requirements, because you want to use an old compiler due to its purported stability but also want to use the latest and greatest packages for some reason. Using rustup is a simple solution and is according to a debian maintainer friend a recommended way for debian stable users to build new software. I don't see the problem of using rustup.

[...]

PR #3169, opened 2024-03-12
Lower the lib's MSRV

the-mikedavis:
Generally libraries should specify their MSRV as the minimum that is required to build. Keeping the MSRV below 1.70 is crucial for us as we follow [Firefox's MSRV policy](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html). Like @Wilfred mentioned in https://github.com/tree-sitter/tree-sitter/issues/2669, some distros lag with their Rust versions and keeping a low MSRV is the only way we can be packaged in those distros.

I've only made this change for the lib since that's the important part for downstream consumers.


maxbrunsfeld:
Thanks for the PR @the-mikedavis. This totally makes sense. I wish we could lower it for the whole workspace, but we currently cannot because of `clap`, the command-line arg parsing library.

PR #3171, opened 2024-03-12
feat: add `--no-position` flag to `Parse`

mkatychev:
Added the ability to omit the byte range when calling `tree-sitter parse` to help with seeding corpus tests:
```
$ tree-sitter parse my.test --no-position
(source_file
  (test
    name: (text)
    (input)
    (output
      (node
        (identifier)
        (node
          (identifier)))))
  (comment)
  (test
    name: (text)
    (input)
    (output
      (node
        (identifier)
        (node
          (identifier))))))
```

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -147,6 +147,7 @@ struct Parse {
     )]
     pub position: bool,
     #[arg(
+        long,
         num_args = 1..,
         help = "Apply edits in the format: \"row, col delcount insert_text\""
     )]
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
Just use `tree-sitter test -u`?

mkatychev:
Would there be value if we could produce the same output parsing an file?

mkatychev:
Unlikely that this will be reviewed

PR #3184, opened 2024-03-17
feat(lib): add the language name

ObserverOfTime:
## Problem

Bindings have no way of retrieving the name of the language.
The only exception is Node which wraps it in a N-API object.

## Solution

Add a `name` field to the generated struct and a `ts_language_name` function to retrieve it.
This should be backwards compatible since it will just return `NULL` if unset.


clason:
Actually, I would recommend turning this into a full `ts_language_info` object that can also take the version, possible aliases, ... (future proof).

ObserverOfTime:
Looks like this actually segfaults if the compiled parser is older.
I can make it check the ABI version and only work on 15.

maxbrunsfeld (approved):

ObserverOfTime:
(Should not be merged before ABI 15 is a thing.)

PR #3189, opened 2024-03-17
Update index.md to include the NGINX implementation repo

JonCoole:
Added the NGINX tree-sitter implementation to the index.md repository list.


amaanq:
There's nothing in that repo though? At least *try* to write a decent grammar before requesting to add it to the docs.

PR #3191, opened 2024-03-18
Fix some bugs with ts nodes

### LINKED ISSUE #1057: Missing matches with some queries

### LINKED ISSUE #2580: Nodes of an aliased rule that contains aliases have weird behavior and sexp output
Let's take tree-sitter-markdown and try setting the root node to the 2nd heading of this code example:

```md
# H1

Text

## H2

More text
```

If we run this rust code to view that node, it shows some very weird output:

```rs
use tree_sitter::{Parser, Query, QueryCapture, QueryCursor, QueryMatch};

pub static CODE: &str = include_str!("../a.md");

pub fn collect_matches<'a>(
    matches: impl Iterator<Item = QueryMatch<'a, 'a>>,
    query: &'a Query,
    source: &'a str,
) -> Vec<(usize, Vec<(&'a str, &'a str)>)> {
    matches
        .map(|m| {
            (
                m.pattern_index,
                format_captures(m.captures.iter().cloned(), query, source),
            )
        })
        .collect()
}

fn format_captures<'a>(
    captures: impl Iterator<Item = QueryCapture<'a>>,
    query: &'a Query,
    source: &'a str,
) -> Vec<(&'a str, &'a str)> {
    captures
        .map(|capture| {
            (
                query.capture_names()[capture.index as usize].as_str(),
                capture.node.utf8_text(source.as_bytes()).unwrap(),
            )
        })
        .collect()
}

fn main() {
    parser.set_language(tree_sitter_md::language()).unwrap();

    let tree = parser.parse(CODE, None).unwrap();

    let node = tree.root_node().child(0).unwrap().child(2).unwrap(); // .child(2).unwrap();

    println!("node: {}\n{}", node.utf8_text(CODE.as_bytes()).unwrap(), node.to_sexp());

    let query = Query::new(tree_sitter_md::language(), "(section) @context").unwrap();

    let matches = cursor
        // .set_byte_range(node.start_byte() - 1..node.end_byte())
        .matches(&query, node, CODE.as_bytes());

    let matches = collect_matches(matches, &query, CODE);

    println!("matches: {:?}", matches);
}
```

Relevant output:
```
("_section2")(atx_heading (atx_h2_marker) heading_content: (inline))("_section2_repeat1")("_blank_line")("_blank_line_start")("_newline")("_line_ending")("_block_not_section")(paragraph (inline))
```

And, because of this weird node, a query involving sections fails.

This seems to be because of aliases where the aliased rule has aliases itself, I made a minimal reproducer here that can be tested w/ `cargo test -- --nocapture` in the bug1 dir:

https://github.com/amaanq/tree-sitter-bug

This was first reported by @lewis6991 here: https://github.com/tree-sitter/tree-sitter/issues/1057#issuecomment-1694674420

### LINKED ISSUE #3185: `ts_node_parent` sometimes returns the wrong node
### Problem

`ts_node_parent` sometimes returns the preceding sibling node instead of the parent node.

This was originally reported at https://github.com/emacs-tree-sitter/elisp-tree-sitter/issues/274.

I could reproduce it with the Python bindings.

### Steps to reproduce

Build the Python grammar
```bash
git clone https://github.com/tree-sitter/tree-sitter-python
cd tree-sitter-python
git reset --hard b8a4c641
tree-sitter generate
tree-sitter build -o tree-sitter-python.dylib
```

Install the Python bindings. Current version is `0.21.0`.
```python
from tree_sitter import *
py = Language("tree-sitter-python.dylib", "python")

p = Parser()
p.set_language(py)

tree = p.parse("def dupa(foo):".encode('utf-8')); tree
# <tree_sitter.Tree at 0x10691a050>

tree.root_node.sexp()
# '(module (function_definition name: (identifier) parameters: (parameters (identifier)) body: (block)))'

fd = tree.root_node.child(0); fd
# <Node type=function_definition, start_point=(0, 0), end_point=(0, 14)>

blk = fd.child(4); blk
# <Node type=block, start_point=(0, 14), end_point=(0, 14)>

blk.parent
# <Node type=":", start_point=(0, 13), end_point=(0, 14)>
```

### Expected behavior

`blk.parent` should be the same as `fd`.

0.22.2

### Operating system/version

macOS 12.6.3

amaanq:
- Closes #1057 (last comment that's related to #2580)
- Closes #2580
- Closes #3185

It seems like inlined nodes weren't taken into account when going to a node's parent, and getting the string representation of a node didn't take into account if the root node wasn't the actual root node, but rather a child node that is either a) aliased, or b) has children (both cases had bugs). This issue with aliased root nodes was also present when resetting a tree cursor to one of these nodes, thus making queries that are run on a node where the root node is an alias and queries for this root node invalid. This PR fixes these problems (after several hours of debugging... :sleeping:)

### DIFF
--- a/cli/src/tests/node_test.rs
+++ b/cli/src/tests/node_test.rs
@@ -249,6 +249,24 @@ fn test_node_parent_of_child_by_field_name() {
 }
 
 #[test]
+fn test_parent_of_zero_width_node() {
+    let code = "def dupa(foo):";
+
+    let mut parser = Parser::new();
+    parser.set_language(&get_language("python")).unwrap();
+
+    let tree = parser.parse(code, None).unwrap();
+    let root = tree.root_node();
+    let function_definition = root.child(0).unwrap();
+    let block = function_definition.child(4).unwrap();
+    let block_parent = block.parent().unwrap();
+
+    assert_eq!(block.to_string(), "(block)");
+    assert_eq!(block_parent.kind(), "function_definition");
+    assert_eq!(block_parent.to_string(), "(function_definition name: (identifier) parameters: (parameters (identifier)) body: (block))");
+}
+
+#[test]
 fn test_node_field_name_for_child() {
     let mut parser = Parser::new();
     parser.set_language(&get_language("c")).unwrap();
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -513,7 +513,7 @@ TSNode ts_node_parent(TSNode self) {
         ts_node_start_byte(child) > ts_node_start_byte(self) ||
         child.id == self.id
       ) break;
-      if (iterator.position.bytes >= end_byte) {
+      if (iterator.position.bytes >= end_byte && ts_node_child_count(child) > 0) {
         node = child;
         if (ts_node__is_relevant(child, true)) {
           last_visible_node = node;
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -439,7 +439,14 @@ const char *ts_node_grammar_type(TSNode self) {
 }
 
 char *ts_node_string(TSNode self) {
-  return ts_subtree_string(ts_node__subtree(self), self.tree->language, false);
+  TSSymbol alias_symbol = ts_node__alias(&self);
+  return ts_subtree_string(
+    ts_node__subtree(self),
+    alias_symbol,
+    ts_language_symbol_metadata(self.tree->language, alias_symbol).visible,
+    self.tree->language,
+    false
+  );
 }
 
 bool ts_node_eq(TSNode self, TSNode other) {
--- a/lib/src/subtree.c
+++ b/lib/src/subtree.c
@@ -890,9 +890,15 @@ static size_t ts_subtree__write_to_string(
       }
     }
   } else if (is_root) {
-    TSSymbol symbol = ts_subtree_symbol(self);
+    TSSymbol symbol = alias_symbol ? alias_symbol : ts_subtree_symbol(self);
     const char *symbol_name = ts_language_symbol_name(language, symbol);
-    cursor += snprintf(*writer, limit, "(\"%s\")", symbol_name);
+    if (ts_subtree_child_count(self) > 0) {
+      cursor += snprintf(*writer, limit, "(%s", symbol_name);
+    } else if (ts_subtree_named(self)) {
+      cursor += snprintf(*writer, limit, "(%s)", symbol_name);
+    } else {
+      cursor += snprintf(*writer, limit, "(\"%s\")", symbol_name);
+    }
   }
 
   if (ts_subtree_child_count(self)) {
@@ -947,6 +953,8 @@ static size_t ts_subtree__write_to_string(
 
 char *ts_subtree_string(
   Subtree self,
+  TSSymbol alias_symbol,
+  bool alias_is_named,
   const TSLanguage *language,
   bool include_all
 ) {
@@ -954,13 +962,13 @@ char *ts_subtree_string(
   size_t size = ts_subtree__write_to_string(
     self, scratch_string, 1,
     language, include_all,
-    0, false, ROOT_FIELD
+    alias_symbol, alias_is_named, ROOT_FIELD
   ) + 1;
   char *result = ts_malloc(size * sizeof(char));
   ts_subtree__write_to_string(
     self, result, size,
     language, include_all,
-    0, false, ROOT_FIELD
+    alias_symbol, alias_is_named, ROOT_FIELD
   );
   return result;
 }
--- a/lib/src/tree_cursor.c
+++ b/lib/src/tree_cursor.c
@@ -151,7 +151,7 @@ static inline bool ts_tree_cursor_child_iterator_previous(
 // TSTreeCursor - lifecycle
 
 TSTreeCursor ts_tree_cursor_new(TSNode node) {
-  TSTreeCursor self = {NULL, NULL, {0, 0}};
+  TSTreeCursor self = {NULL, NULL, {0, 0, 0}};
   ts_tree_cursor_init((TreeCursor *)&self, node);
   return self;
 }
@@ -162,6 +162,7 @@ void ts_tree_cursor_reset(TSTreeCursor *_self, TSNode node) {
 
 void ts_tree_cursor_init(TreeCursor *self, TSNode node) {
   self->tree = node.tree;
+  self->root_alias_symbol = node.context[3];
   array_clear(&self->stack);
   array_push(&self->stack, ((TreeCursorEntry) {
     .subtree = (const Subtree *)node.id,
@@ -474,7 +475,7 @@ uint32_t ts_tree_cursor_current_descendant_index(const TSTreeCursor *_self) {
 TSNode ts_tree_cursor_current_node(const TSTreeCursor *_self) {
   const TreeCursor *self = (const TreeCursor *)_self;
   TreeCursorEntry *last_entry = array_back(&self->stack);
-  TSSymbol alias_symbol = 0;
+  TSSymbol alias_symbol = self->root_alias_symbol;
   if (self->stack.size > 1 && !ts_subtree_extra(*last_entry->subtree)) {
     TreeCursorEntry *parent_entry = &self->stack.contents[self->stack.size - 2];
     alias_symbol = ts_language_alias_at(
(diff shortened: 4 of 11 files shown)


lib/src/node.c
```diff
@@ -513,7 +518,7 @@ TSNode ts_node_parent(TSNode self) {
         ts_node_start_byte(child) > ts_node_start_byte(self) ||
         child.id == self.id
       ) break;
-      if (iterator.position.bytes >= end_byte) {
+      if (iterator.position.bytes >= end_byte && !ts_node__subtree(child).data.is_inline) {
```
maxbrunsfeld, line 521: I don't think the `is_inline` bit is the right criteria for this. The `is_inline` thing is just an optimization where certain leaf nodes are stored w/o heap allocations. Leaf nodes could be inlined or not based on the numerical values of their `size` and `padding` (e.g. the size needs to be small enough to fit in a certain number of bits).

Maybe we should be checking if `child` has children here (`ts_subtree_child_count(node) > 0`)? Or is that we need to skip empty nodes (where `ts_subtree_size(node).bytes > 0`)?

lib/src/subtree.c
```diff
@@ -940,27 +946,28 @@ static size_t ts_subtree__write_to_string(
...
 
-  if (is_visible) cursor += snprintf(*writer, limit, ")");
+  if (is_visible || (is_root && ts_subtree_child_count(self) > 0)) cursor += snprintf(*writer, limit, ")");
 
   return cursor - string;
 }
 
 char *ts_subtree_string(
   Subtree self,
+  TSSymbol alias_symbol,
```
maxbrunsfeld, line 956: 👍  It makes total sense that we'd need to pass this in. Could you add a Rust test for a calling `to_sexp()` on a node that is aliased?

lib/src/tree_cursor.c
```diff
@@ -172,6 +172,7 @@ void ts_tree_cursor_init(TreeCursor *self, TSNode node) {
     .child_index = 0,
     .structural_child_index = 0,
     .descendant_index = 0,
+    .root_alias = node.context[3],
```
maxbrunsfeld, line 175: I don't understand why we need this `root_alias` field. Nodes can only be aliased by their parent node having an `alias_sequence` (which indicate the alias of every child), and we already account for that. Can you explain why the `root_alias` is there?

maxbrunsfeld: Actually, re-reading your comment about calling `ts_tree_cursor_init` about a non-root, aliased node, I get the intent now. I don't think `root_alias` should be on *every* node of the stack though. It's just the alias for the node at the bottom of the stack right?

lib/src/subtree.c
```diff
@@ -940,27 +946,28 @@ static size_t ts_subtree__write_to_string(
...
   const TSLanguage *language,
   bool include_all
 ) {
   char scratch_string[1];
   size_t size = ts_subtree__write_to_string(
     self, scratch_string, 1,
     language, include_all,
-    0, false, ROOT_FIELD
+    alias_symbol, false, ROOT_FIELD
```
maxbrunsfeld, line 964: It looks like, in addition to passing in the `alias_symbol` here, we need to look up whether the alias symbol represents a "named symbol", similarly to what we do in the recursive calls to this function

```c
        bool subtree_alias_is_named = subtree_alias_symbol
          ? ts_language_symbol_metadata(language, subtree_alias_symbol).named
          : false;
```

We need to pass in that boolean, instead of always passing in `false`. I think that if you did that, then the `is_visible` flag would always be correct, and you wouldn't need a lot of the other logic that you added to the `ts_subtree__write_to_string` function, in order to work around `is_visible` being wrong.

lib/src/subtree.c
```diff
@@ -890,9 +890,15 @@ static size_t ts_subtree__write_to_string(
...
+    TSSymbol symbol = alias_symbol ? alias_symbol : ts_subtree_symbol(self);
     const char *symbol_name = ts_language_symbol_name(language, symbol);
-    cursor += snprintf(*writer, limit, "(\"%s\")", symbol_name);
+    if (ts_subtree_child_count(self) > 0) {
+      cursor += snprintf(*writer, limit, "(%s", symbol_name);
+    } else if (ts_subtree_named(self)) {
+      cursor += snprintf(*writer, limit, "(%s)", symbol_name);
+    } else {
+      cursor += snprintf(*writer, limit, "(\"%s\")", symbol_name);
+    }
```
maxbrunsfeld, line 901: I think that if we're in this `else` block (`is_visible` is false), we're on a node that wouldn't normally even be included in the S-expression, so we don't need this logic. I made a comment below about how `is_visible` might not be accounting for aliases, which should be fixed upstream of this code.

lib/src/tree_cursor.c
amaanq: yep. it's *only* used when the stack size is just 1, sorry if the else if bit wasn't clear about that

lib/src/subtree.c
amaanq: Yeah in python's suite a single `_newline` is aliased to `block`, that is tested in the rust test

lib/src/node.c
amaanq: Yeah `ts_subtree_child_count(node) > 0` is better actually, `ts_subtree_size(node).bytes > 0` doesn't work since the child is 0 bytes as well as the inlined parent

cli/src/tests/node_test.rs
```diff
@@ -248,6 +248,24 @@ fn test_node_parent_of_child_by_field_name() {
     );
 }
 
+#[test]
+fn test_node_parent_with_inlined_parent() {
```
maxbrunsfeld, line 252: I think this test could use a different name, since `inlined` isn't really part of the issue. Maybe `parent_with_zero_length_node`? I think the issue is that there are zero-length nodes inside of the `suite`,  right?

lib/src/tree_cursor.h
```diff
@@ -9,6 +9,7 @@ typedef struct {
   uint32_t child_index;
   uint32_t structural_child_index;
   uint32_t descendant_index;
+  TSSymbol root_alias;
```
maxbrunsfeld, line 12: What do you think of putting this on the `TreeCursor` instead of on each node of the stack? I think it would make it clearer that it is only meaningful for the root node.

cli/src/tests/node_test.rs
amaanq: yep you're right

lib/src/tree_cursor.h
amaanq: Yeah this makes a lot of sense, it's only relevant to whatever node the cursor was set with

cli/src/tests/node_test.rs
maxbrunsfeld: Yeah, I think the test should be called `test_parent_of_zero_width_node`. The `block` in this case (which is what you're testing the parent of) is a zero-width node, and that's why the problem was happening before.

The issue here doesn't have to do with inline subtrees. Inline subtrees are just a slightly more compact way of representing certain child-less subtrees.

I think your fix (avoiding descending into nodes without children) helps in this case (and most cases), but it's still possible that `ts_node_parent` could return the wrong value in the presence of zero-width nodes.

Fundamentally `ts_node_parent(node)` works by walking down the tree, starting from the root, trying to find the smallest possible node that contains the `node`'s byte range (and is not `node` itself). But if the node's byte range is empty, there are some edge cases.

I still think this fix improves the situation, and I'm glad you figured it out. I just think we should change the title of the test and 🚢 .

amaanq: Ahh gotcha, makes sense. Thanks for clarifying

maxbrunsfeld (commented):
Thanks for debugging all of this stuff @amaanq . The TreeCursor changes look 💎 .

maxbrunsfeld (approved):

PR #3197, opened 2024-03-19
docs: add typescript directives to grammar.js snippet in creating-parsers.md

michaelfortunato:
Adds typescript comment directives to the grammar.js boilerplate, allowing  parser authors to get type inference if their editor has typescript lsp. These directives are mostly present in the major tree-sitter grammars. 

See [here](1) and [here](2).

[1]: https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-
[2]: https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -56,6 +56,12 @@ export PATH=$PATH:./node_modules/.bin
 Once you have the CLI installed, create a file called `grammar.js` with the following contents:
 
 ```js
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I don't think I want the docs to have this. I'm trying to show a very minimal example here. I don't even think that these comment directives are the best way of setting up Typescript on a JS file.

Also, the links that you provided don't seem relevant to this topic.

PR #3203, opened 2024-03-23
fix: default output directory for `build --wasm` should use current_dir

### LINKED ISSUE #3202: Can't run playground
### Problem

Since tree-sitter 0.22.1 (99b93d83a11a84ad9a23f5248e126fe45c1a61d3), I can't run playground because it can't find the .wasm file, because `build --wasm` outputs the .wasm in `parser_lib_path` instead of `current_dir`

### Steps to reproduce

- Create the ["Getting Started" parser](https://tree-sitter.github.io/tree-sitter/creating-parsers)
- `tree-sitter playground` will complain:
```
Failed to read tree-sitter-YOUR_LANGUAGE_NAME.wasm. Run `tree-sitter build --wasm` first.

Caused by:
    No such file or directory (os error 2)
```
- `tree-sitter build --wasm`
- `tree-sitter playground` will still complain the same complaint

### Expected behavior

Playground running

tree-sitter 0.22.2 (b7fcf9878e1144a30e71ae94f951f163258770d9)

### Operating system/version

macOS 13.6.1

rooney:
Fix #3202

This PR has two commits:
3492a487e6733ec58d7e0f0ed325f2bfb91a21e9 is the minimal fix that fixes the problem.
452bac66179e47c6d6e2076eed5517e81d3d146f is the proper fix.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -28,6 +28,7 @@ enum Commands {
     InitConfig(InitConfig),
     Generate(Generate),
     Build(Build),
+    BuildWasm(BuildWasm),
     Parse(Parse),
     Test(Test),
     Query(Query),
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -498,6 +508,18 @@ fn run() -> Result<()> {
             }
         }
 
+        Commands::BuildWasm(wasm_options) => {
+            eprintln!("`build-wasm` is deprecated and will be removed in the next version, you should use `build --wasm` instead");
```
ObserverOfTime, line 512:
```suggestion
            eprintln!("`build-wasm` is deprecated and will be removed in v0.24.0, you should use `build --wasm` instead");
```

PR #3204, opened 2024-03-23
feat: allow external scanner to use the logger

rooney:
Allow external scanners to use Tree-sitter's logger via `lexer->log(lexer, "message", ...)` so they don't have to reinvent e.g. wasm compatibility.


amaanq:
ping @maxbrunsfeld

PR #3207, opened 2024-03-24
docs: creating parser: setup instruction: need to add …

rooney:
… "tree-sitter" entry to package.json

Otherwise, when you do`tree-sitter generate`, it will fail with the following message:

```
Failed to locate a package.json file that has a "tree-sitter" section, please ensure you have one, and if you don't then consult the docs
```


rooney:
Fixes #3208

PR #3210, opened 2024-03-24
feat(bindings): update npm scripts

ObserverOfTime:
* `npm run build` is removed
* `npm run parse` is removed
* `npm run prebuildify` is removed
  - use `npm x -- prebuildify ...` instead
* `npm start` will run `ts playground` (after `ts build --wasm`)
  - the build can be skipped with `npm start --ignore-scripts`
* `npm test` will run `node --test` instead of `ts t`
  - requires #3178
  
  tree-sitter commands can be run with `npm x -- tree-sitter ...`

### DIFF
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -126,12 +126,12 @@ pub fn generate_grammar_files(
                     .unwrap();
                 if dependencies.remove("nan").is_some() {
                     eprintln!("Replacing nan dependency with node-addon-api in package.json");
-                    dependencies.insert("node-addon-api".to_string(), "^7.1.0".into());
+                    dependencies.insert("node-addon-api".to_string(), "^8.0.0".into());
                     updated = true;
                 }
                 if !dependencies.contains_key("node-gyp-build") {
                     eprintln!("Adding node-gyp-build dependency to package.json");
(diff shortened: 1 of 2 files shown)


clason:
I would prefer to keep the management commands local to `npm` (or ideally `tree-sitter`) and leave the `Makefile` as a simple build harness (the sole usecase in my view is to be able to easily compile the parser from a release tarball without having `tree-sitter` CLI installed).

cli/src/generate/grammar_files.rs
```diff
@@ -142,33 +142,28 @@ pub fn generate_grammar_files(
...
+                    scripts.clear();
+                    scripts.append(
+                        json!({
+                            "install": "node-gyp-build",
+                            "prestart": "tree-sitter build --wasm",
+                            "start": "tree-sitter playground",
+                            "test": node_test
+                        })
+                        .as_object_mut()
+                        .unwrap(),
```
amaanq, line 166: we can avoid the overhead of clearing

```suggestion
                    *scripts = Map::from_iter([
                        ("install".to_string(), "node-gyp-build".into()),
                        ("prestart".to_string(), "tree-sitter build --wasm".into()),
                        ("start".to_string(), "tree-sitter playground".into()),
                        ("test".to_string(), node_test.into()),
                    ]);
```

PR #3211, opened 2024-03-24
Include header files in Python bindings package source distributions

geigerzaehler:
Including the headers allows the Python package to be build from the source distribution.

Before, the following would fail when using [`build`](https://build.pypa.io/en/stable/index.html) for Python.

```bash
mkdir tree-sitter-foo
cd tree-sitter-foo
tree-sitter generate
python -m build
```


ObserverOfTime:
Using `data_files` is discouraged and wheels are enough.

PR #3218, opened 2024-03-26
fix: editing subtree that depends-on-column: ...

### LINKED ISSUE #3217: Parsing-After-Editing Test Case is incorrect (and passing)
### Problem

The test [`test_parsing_after_editing_tree_that_depends_on_column_values`](https://github.com/tree-sitter/tree-sitter/blob/0fc92c9a7d0ddb417bd74bf7f533bb8f3042dbe3/cli/src/tests/parser_test.rs#L429) is incorrect.

It starts with parsing the following source code:
```
a = b
c = do d
       e + f
       g
h + i
```
and asserting that the parse tree should be:
```
"(block ",
"(binary_expression (identifier) (identifier)) ",
"(binary_expression (identifier) (do_expression (block (identifier) (binary_expression (identifier) (identifier)) (identifier)))) ",
"(binary_expression (identifier) (identifier)))",
```

Then, it `perform_edit` to the source code, to become:
```
a = b
c1234 = do d
       e + f
       g
h + i
```
(so far so good)

The problem is, it then asserts that the parse tree of the edited source code should become:
```
 "(block ",
 "(binary_expression (identifier) (identifier)) ",
 "(binary_expression (identifier) (do_expression (block (identifier)))) ",
 "(binary_expression (identifier) (identifier)) ",
 "(identifier) ",
 "(binary_expression (identifier) (identifier)))",
```

Which doesn't seem to be correct because _what the `perform_edit` did is just renaming the identifier `c` to `c1234` -- it should not result in any change to the tree structure at all._

But yes, the test passes. Which means, there's some bug in the tree-editing implementation.

### Steps to reproduce

- open `cli/src/tests/parser_test.rs`
- [Run the test suite](https://tree-sitter.github.io/tree-sitter/contributing#testing)
- All tests pass

### Expected behavior

`test_parsing_after_editing_tree_that_depends_on_column_values` shall not pass.

Or, if it were to, then the edited parse tree should be equal to the original.

tree-sitter 0.22.2 (b7fcf9878e1144a30e71ae94f951f163258770d9)

### Operating system/version

macOS 13.6.1

rooney:
... invalidation should be based on the edit's end-column.

Fix #3217


maxbrunsfeld:
Hi, I don't know what behavior this new code would implement, but the existing behavior is what we want, I believe.

The existing behavior is that if the external scanner's `get_column` method is called when scanning a token, then any edit that occurs *before* that token on a line (thus shifting the token's start column) will invalidate the token.

maxbrunsfeld:
I'm going to close this out because I believe the existing test is correct, and your change to the test would break the desired behavior. If I'm  missing some more subtle reason why this invalidation logic should be changed, then feel free to explain further, and we can reopen.

rooney:
Hi @maxbrunsfeld 

Alright, I have reworked my PR so it's much simpler now -- only adding two checks and nothing else (no test change, and all the tests still pass).

https://github.com/tree-sitter/tree-sitter/compare/master...rooney:tree-sitter:col-dependent-fix
(I assume it's pretty clear since it's only a two-liner, but please let me know if I should explain)

The change is needed because there are some cases where a token-that-depends-on-column should be invalidated but isn't. Let me demo it using a simple-as-possible example:

```js
// grammar.js

  rules: {
    source_file: ($) => seq(/[ \t\r\n]+/, choice($.odd, $.even), $.x),
    x: () => "x",
  },
  externals: ($) => [$.odd, $.even],
```
Here, our grammar will: 1. consume leading whitespace(s), 2. consult the external scanner to see if the column is at odd/even position, and 3. match "x".

(the scanner is stateless):
```c
// scanner.c

enum TokenType { ODD, EVEN };

bool tree_sitter_coldemo_external_scanner_scan(void *payload, TSLexer *lexer,
                                               const bool *valid_symbols) {
  lexer->result_symbol = lexer->get_column(lexer) % 2 ? ODD : EVEN;
  return true;
}
```

Now, here's how the playground goes on `master`:
MOV to GIF
See how it stays "odd" even though the positions are correctly updating.

And here's how the playground goes after the two-line patch:
Mov to Gif Conversion
See how it correctly updates the odd/even now.

[tree-sitter-coldemo.zip](https://github.com/tree-sitter/tree-sitter/files/14848556/tree-sitter-coldemo.zip) (contains only `grammar.js` and `src/scanner.c`)

To be extra sure, I also have checked this new patch against the complete [tree-sitter-haskell](https://github.com/tree-sitter/tree-sitter-haskell) -- all tests there still pass running this patch.

maxbrunsfeld:
Thanks for the response, and nice work. As part of your branch, can you add your new example scanner as a test to the Tree-sitter test suite?

maxbrunsfeld:
Reopening.

rooney:
> Reopening.

Thanks @maxbrunsfeld 

- OK test added, made sure it doesn't pass without the patch.

Also, added an optimization for when the edits don't actually move the columns. (The test for this optimization checks that there's fewer bytes being read by the recorder). 
https://github.com/tree-sitter/tree-sitter/compare/master...rooney:tree-sitter:col-dependent-fix

Also have re-run [tree-sitter-haskell](https://github.com/tree-sitter/tree-sitter-haskell) `test` with the optimization, all good there.

amaanq:
I don't think it can't be reopened, can you open another PR @rooney?

maxbrunsfeld:
Oh sorry, thought that I reopened it. The GitHub UI mislead me a bit there.

rooney:
sure, np, created another PR https://github.com/tree-sitter/tree-sitter/pull/3257

PR #3223, opened 2024-03-27
fix(windows): add `/utf-8` flag for parsers using unicode symbols

### LINKED ISSUE #3222: Cannot use unicode in queries on windows in rust in tree-sitter 0.22.2
### Problem

On tree-sitter versions 0.21.0 and higher, attempting to compile a query containing a unicode character fails on windows using the rust bindings. The same code succeeds on linux and macOS, and succeeds across all platforms on 0.20.x versions:

```rust
use tree_sitter::{Parser, Query, QueryCursor};

fn main() {
    parser.set_language(&tree_sitter_test::language()).expect("Error loading grammar");
    let source_code = "op == expr op ≜ expr";
    let tree = parser.parse(source_code, None).unwrap();
    println!("{}", tree.root_node().to_sexp());

    let query = Query::new(&tree_sitter_test::language(), "(def_eq \"≜\" @def_eq)").unwrap();
    for capture in cursor.matches(&query, tree.root_node(), "".as_bytes()) {
        println!("{:?}", capture);
    }
}
```

### Steps to reproduce

1. Clone this branch containing a minimal tree-sitter grammar and rust program: https://github.com/ahelwer/tree-sitter-test/tree/windows-unicode
2. `cd` into the `rust` directory and run `cargo run`

On Windows, this will produce the following error value:
```
called `Result::unwrap()` on an `Err` value: QueryError { row: 0, column: 9, offset: 9, message: "", kind: NodeType }
```
On Linux and macOS, it will succeed.

Alternatively, you can see a less minimal cross-platform reproduction of it in this CI run: https://github.com/tlaplus-community/tlauc/actions/runs/8440383121

### Expected behavior

I expected to be able to continue using unicode characters in queries on all supported platforms. This behavior worked across all platforms when generating and consuming grammars with tree-sitter 0.20.x.

0.22.2, 0.21.0 or higher

### Operating system/version

N/A

amaanq:
Fixes #3222

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -591,6 +591,7 @@ impl Loader {
                 command.arg(scanner_path);
             }
             command
+                .arg("/utf-8")
                 .arg("/link")
                 .arg(format!("/out:{}", output_path.to_str().unwrap()));
         } else {
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -1,6 +1,7 @@
 use super::write_file;
 use anyhow::{anyhow, Context, Result};
 use heck::{ToKebabCase, ToShoutySnakeCase, ToSnakeCase, ToUpperCamelCase};
+use indoc::indoc;
 use serde::Deserialize;
 use serde_json::{json, Map, Value};
 use std::fs::File;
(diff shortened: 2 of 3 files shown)


clason:
This also fixes issues with purescript, Haskell, and (possibly) Perl utf-8 queries in nvim-treesitter.

cli/loader/src/lib.rs
```diff
@@ -592,6 +592,7 @@ impl Loader {
             }
             command
                 .arg("/link")
+                .arg("/utf-8")
```
ObserverOfTime, line 595: It should be before `/link` since it's not a linker flag.

cli/src/generate/grammar_files.rs
```diff
@@ -261,9 +261,19 @@ pub fn generate_grammar_files(
...
-        })?;
+        missing_path_else(
+            path.join("build.rs"),
+            |path| generate_file(path, BUILD_RS_TEMPLATE, language_name),
+            |path| {
+                let build_rs =
+                    fs::read_to_string(path).with_context(|| "Failed to read build.rs")?;
+                if !build_rs.contains("/utf-8") {
+                    eprintln!("Updating build.rs with the /utf-8 flag for Windows");
+                    generate_file(path, BUILD_RS_TEMPLATE, language_name)?;
```
ObserverOfTime, line 272: It might be better to print a warning here instead of replacing the file.

cli/src/generate/templates/build.rs
```diff
@@ -8,6 +8,9 @@ fn main() {
     c_config.file(&parser_path);
     println!("cargo:rerun-if-changed={}", parser_path.to_str().unwrap());
 
+    #[cfg(target_os = "windows")]
+    c_config.flag("/utf-8");
```
ObserverOfTime, line 12:
```suggestion
    #[cfg(target_env = "msvc")]
    c_config.flag("-utf-8");
```

The flag should not be added in mingw and cc-rs uses dashes.

cli/src/generate/grammar_files.rs
amaanq: it'd be annoying for users that don't care about the rust grammar, but have this warning printed every time they run generate

cli/src/generate/templates/build.rs
ObserverOfTime: `msvc` not `windows`

cli/src/generate/grammar_files.rs
ObserverOfTime: But users that do care about the rust bindings may have to edit them again.

cli/src/generate/templates/build.rs
amaanq: oops, sorry

cli/src/generate/grammar_files.rs
amaanq: what if we just place it after `println!("cargo:rerun-if-changed={}", parser_path.to_str().unwrap());` instead of generating from the template?

PR #3224, opened 2024-03-27
fix: insert "tree-sitter" section in current directory's package.json…

### LINKED ISSUE #3170: Documentation: "Creating parsers" missing information
### Problem

The [Project setup](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) section does not indicate that `package.json` must be updated with a *tree-sitter* section. As result, the given steps do not work and the message:

*Failed to locate a package.json file that has a "tree-sitter" section, please ensure you have one, and if you don't then consult the docs*

is issued.

I found a fix by adding in `package.json` the text:

```
  "tree-sitter": [
    {"file-types": ["YOUR_LANGUAGE_NAME"]}
  ] '
```

### Steps to reproduce

Follow the steps of the  [Project setup](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) section until

`tree-sitter generate`

### Expected behavior

Generation of a grammar for the given example

tree-sitter 0.22.1

### Operating system/version

Windows 11

rooney:
… if none of the parents have it

The section will only be inserted if:
- There's no package.json in any of the parent directories (incl. grandparents etc.) containing "tree-sitter" section.
- There is a package.json in current directory, but it does not have a "tree-sitter" section.

Fix #3170

### DIFF
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -213,6 +213,19 @@ pub fn generate_grammar_files(
                     updated = true;
                 }
 
+                // insert `tree-sitter` at the end
+                if !package_json.contains_key("tree-sitter") {
+                    eprintln!("Adding a `tree-sitter` section to package.json");
+                    package_json.insert(
+                        "tree-sitter".to_string(),
+                        json!([{
+                            "scope": format!("source.{language_name}"),
+                            "injection-regex": format!("^{language_name}$"),
+                        }]),
+                    );
+                    updated = true;
+                }
+
                 if updated {
                     let mut package_json_str = serde_json::to_string_pretty(&package_json)?;
                     package_json_str.push('\n');
@@ -425,8 +438,7 @@ fn lookup_package_json_for_path(path: &Path) -> Result<(PathBuf, PackageJSON)> {
             .then(|| -> Result<PackageJSON> {
                 let file =
                     File::open(pathbuf.as_path()).with_context(|| "Failed to open package.json")?;
-                let package_json: PackageJSON = serde_json::from_reader(BufReader::new(file))?;
-                Ok(package_json)
+                Ok(serde_json::from_reader(BufReader::new(file))?)
             })
             .transpose()?;
         if let Some(package_json) = package_json {


cli/src/generate/grammar_files.rs
```diff
@@ -213,6 +213,18 @@ pub fn generate_grammar_files(
                     updated = true;
                 }
 
+                // insert `tree-sitter` right after `files`
+                if !package_json.contains_key("tree-sitter") {
+                    eprintln!("Adding tree-sitter to package.json");
```
amaanq, line 218:
```suggestion
                    eprintln!("Adding a `tree-sitter` section to package.json");
```

cli/src/generate/grammar_files.rs
```diff
@@ -213,6 +213,18 @@ pub fn generate_grammar_files(
...
+                    package_json = insert_after(
+                        package_json,
+                        "files",
+                        "tree-sitter",
+                        json!([{"file-types": language_name}]),
```
amaanq, line 223: `file-types` should be an array

```suggestion
                        json!([{"file-types": [language_name]]),
```

also, since this should always be at the end, we could just use `package_json.insert` in this case

```rs
package_json.insert(
    "tree-sitter".to_string(),
    json!([{"file-types": [language_name]}]),
);
```

cli/src/generate/grammar_files.rs
```diff
@@ -436,6 +449,10 @@ fn lookup_package_json_for_path(path: &Path) -> Result<(PathBuf, PackageJSON)> {
         }
         pathbuf.pop(); // package.json
         if !pathbuf.pop() {
+            if path.exists() {
```
amaanq, line 452: I disagree with this part, I'd almost say the `tree-sitter` section is necessary for a lot of stuff, especially for the cli. Language configurations can't be looked up without this, and now having written an entry with `file-types` beforehand should alleviate the need for this blob now

rooney: This is just a check that we are using a `package.json` that already exists in the current directory (if the file does not already exist, we will not create it). Or is it the comment in the following line that's confusing? The comment is supposed to mean "If the package.json is in the current directory (rather than in parent directories), it doesn't need to have a `tree-sitter` section because in that case we will create the section." (if the package.json is in parent dirs, it needs to have a `tree-sitter` section/we will not create the section and will just exit with the error message) -- If that's the case I'll reword the comments.

amaanq: But wasn't the section written to the file *before* this lookup is done though? It *will* exist due to the code you added where the `tree-sitter` section is added to the end of `package.json`, so this shouldn't be needed I think? I might be missing something though

cli/src/generate/grammar_files.rs
```diff
@@ -213,6 +213,20 @@ pub fn generate_grammar_files(
...
 
+                // insert `tree-sitter` at the end
+                if !package_json.contains_key("tree-sitter") {
+                    eprintln!("Adding a `tree-sitter` section to package.json");
+                    package_json.insert(
+                        "tree-sitter".to_string(),
+                        json!([{
+                            "scope": format!("source.{language_name}"),
+                            "injection-regex": language_name,
+                            "file-types": [language_name],
```
ObserverOfTime, line 224:
```suggestion
                            "injection-regex": format!("^{language_name}$")
```
Extensions do not always match the language name.

PR #3230, opened 2024-03-28
 fix(wasm): add iswpunct to stdlib 

### LINKED ISSUE #3228: `getenv()` should probably be a blessed stdlib-symbol for WASM builds?
### Problem

i.e. maybe include it in here https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/wasm/stdlib-symbols.txt

Since external scanners are now intended to do something like `getenv("TREE_SITTER_DEBUG")` to check for debug mode, it seems like `getenv()` may be reasonable to include as a function for WASM builds?

### Steps to reproduce

Use `getenv()` in an external scanner and then compile to wasm, it will fail.

### Expected behavior

Expected to be able to use `getenv()` in an external scanner for detecting tree-sitter debug mode

tree-sitter 0.21.0

### Operating system/version

macOS 13.6.5

ObserverOfTime:
This also added `getenv` at first (for #3228) but that breaks the wasm tests.


amaanq:
Is `iswpunct` needed? I don't think any scanners use this

ObserverOfTime:
[Vim](https://github.com/tree-sitter-grammars/tree-sitter-vim/blob/master/src/scanner.c#L327) uses it.

maxbrunsfeld:
Thanks for explaining what the use case was. I'm fine adding it, since it's in a similar category of `wctype` functions that we're already including.

amaanq:
think I'll close this since a catch-all punctuation function doesn't make sense (and the vim usage needs to be refactored, it has a comment questioning its usage :grin:)

clason:
[It's not the only parser](https://github.com/search?q=iswpunct+path%3A**%2Fsrc%2Fscanner.c&type=code) but admittedly niche. That promised refactoring would be very welcome.

PR #3231, opened 2024-03-28
refactor(scripts): clean up bash scripts

ObserverOfTime:
This also fixes `script/build-fuzzers` which didn't work before.


ObserverOfTime:
Windows has the cmd scripts for some of them (which I'm not touching).

PR #3234, opened 2024-03-30
Generate simpler code for matching large character sets

maxbrunsfeld:
To match large character sets, Tree-sitter previously generated code with an explicit binary search of a set of ranges. This took a long time when compiling to wasm. Now, we generate simple static arrays, containing ordered lists of ranges, and match these arrays using a fixed binary search routine. This improves compile time for parsers that use regexes with large character classes.

In addition, I have changed the way that we extract large character sets, so that we don't generate as many mostly-duplicated-but-slightly different character sets.

/cc @fdncred, I think this will help the Nu grammar somewhat. I also am going to submit a PR to the grammar itself with a couple of minor changes that speed up compiles and shrink the size of the parser.


fdncred:
Thanks @maxbrunsfeld. Appreciate you tagging me, and I look forward to all this working together better.

maxbrunsfeld:
@fdncred With these changes, plus a couple of small grammar changes, the Nu parser compiles to wasm in about 45 seconds on my machine. Still slow, but there should no longer be any issue publishing the Zed extension.

fdncred:
Thanks @maxbrunsfeld. I really appreciate all your hard work here. I've been watching the commits. Looking forward to seeing that PR proceed. 45 seconds seems like a crazy amount of time to compile the nushell grammar though.

PR #3237, opened 2024-03-31
feat: allow inlined rules to match empty

### LINKED ISSUE #3220: Is there a particular reason why inlining happens _after_ checking?
### Problem

It's more a question than a FR.

Let's say I'm translating the EBNF grammar and what to keep the structure as close as possible. So I end up with a non-starter rule that can match empty input, so tree-sitter complains about it. That's actually super easily solvable, KUDOs for using a programming language to write a grammar, I can just turn these rules into functions.

But there is also this `inline` feature, that looks appealing, so I though I should be able to just inline these rules. Unfortunately that doesn't work: the check for rules matching an empty input is done _before_ inlining, so tree-sitter still complains.

So, my question is: is there a particular reason for this ordering? Could we do inlining first, and then probably some bad rules will go away already?

### Expected behavior

_No response_

rooney:
Fix #3220

### DIFF
--- a/cli/src/generate/prepare_grammar/flatten_grammar.rs
+++ b/cli/src/generate/prepare_grammar/flatten_grammar.rs
@@ -192,7 +192,13 @@ pub(super) fn flatten_grammar(grammar: ExtractedSyntaxGrammar) -> Result<SyntaxG
     }
     for (i, variable) in variables.iter().enumerate() {
         for production in &variable.productions {
-            if production.steps.is_empty() && symbol_is_used(&variables, Symbol::non_terminal(i)) {
+            if production.steps.is_empty()
+                && symbol_is_used(&variables, Symbol::non_terminal(i))
+                && !grammar
+                    .variables_to_inline
+                    .iter()
+                    .any(|symbol| symbol.index == i)
+            {
                 return Err(anyhow!(
                     "The rule `{}` matches the empty string.
 


amaanq:
Tokens (terminals) can't be inlined - the linked issue doesn't have an example, is this for the case where a rule is an optional (thus can match nothing) but is inlined? So wouldn't these rules containing this inline need to be checked *as well* then? If rule a is comprised of a single rule b, and rule b is an inlined rule matching an optional, then rule a also becomes invalid since it can match an empty string.

maxbrunsfeld:
> So wouldn't these rules containing this inline need to be checked as well then? If rule a is comprised of a single rule b, and rule b is an inlined rule matching an optional, then rule an also becomes invalid since it can match an empty string.

Yeah, I think you are right @amaanq. @rooney if we allow `inline` rules to match the empty string, we'll need additional logic to correctly check that other rules (which *use* inline rules) don't match the empty string due to inlined rules. Is the complexity worth it, or is the existing behavior fine?

rooney:
> Tokens (terminals) can't be inlined - the linked issue doesn't have an example, is this for the case where a rule is an optional (thus can match nothing) but is inlined?

Yes -- more specifically (in real-world scenario): some sequence of optionals that ultimately can match empty, that in turn be used as a common subrule by other (non empty-matching) rules.

> > So wouldn't these rules containing this inline need to be checked as well then? If rule a is comprised of a single rule b, and rule b is an inlined rule matching an optional, then rule an also becomes invalid since it can match an empty string.
> 
> Yeah, I think you are right @amaanq. @rooney if we allow `inline` rules to match the empty string, we'll need additional logic to correctly check that other rules (which _use_ inline rules) don't match the empty string due to inlined rules. Is the complexity worth it, or is the existing behavior fine?

Right, ok, I'll try to find a way that handles the transitivity in a non-complex way.

Btw, current behavior (before this patch) can actually match empty string just fine, e.g:

```
  rules: {
    source_file: ($) => seq("foo", $.blank, "bar"),
    blank: () => "",
  },
```
image

(not sure if this is intentional or not)

amaanq:
Closing this as unplanned, feel free to reopen once you'll take it up again, if so

PR #3252, opened 2024-04-03
Use two macros for reduce, to restore positional arguments

maxbrunsfeld:
https://github.com/tree-sitter/tree-sitter/pull/3229

/cc  @ObserverOfTime @amaanq @DavisVaughan

I don't want to have the designated initializers for `.symbol` and `.child_count` in every `REDUCE` invocation in the parse table, because it makes those entries pretty verbose, and there are a lot of them.

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -1296,11 +1296,12 @@ impl Generator {
                         production_id,
                         ..
                     } => {
-                        add!(
-                            self,
-                            "REDUCE(.symbol = {}, .child_count = {child_count}",
-                            self.symbol_ids[&symbol]
-                        );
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
Would it be fine to make the other two arguments positional as well?

maxbrunsfeld:
That's a good point @ObserverOfTime. I guess I made them use designated initializers initially, since it was less obvious what they were (`production_id`, `dynamic_precedence`), but we could just make them all positional. Then we could go back to a single variadic macro.

ObserverOfTime:
It doesn't have to be variadic. You can just have four named arguments and keep the designated initializers.

PR #3255, opened 2024-04-04
Fix signedness in select_higher_precedence log message

tek:
It appears that the previous fix only caught one of the two occurrences in these messages.


amaanq:
thanks

PR #3257, opened 2024-04-04
fix: handle more cases of editing subtrees that depend on column values

rooney:
Reopen https://github.com/tree-sitter/tree-sitter/pull/3218

### DIFF
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -502,6 +502,67 @@ h + i
 }
 
 #[test]
+fn test_parsing_after_editing_tree_that_depends_on_column_position() {
+    let dir = fixtures_dir()
+        .join("test_grammars")
+        .join("depends_on_column");
+
+    let grammar_json = load_grammar_file(&dir.join("grammar.js"), None).unwrap();
+    let (grammar_name, parser_code) = generate_parser_for_grammar(grammar_json.as_str()).unwrap();
+
+    let mut parser = Parser::new();
+    parser
+        .set_language(&get_test_language(&grammar_name, &parser_code, Some(&dir)))
+        .unwrap();
+
+    let mut code = b"\n x".to_vec();
+    let mut tree = parser.parse(&code, None).unwrap();
+    assert_eq!(tree.root_node().to_sexp(), "(x_is_at (odd_column))");
+
+    perform_edit(
+        &mut tree,
+        &mut code,
+        &Edit {
+            position: 1,
+            deleted_length: 0,
+            inserted_text: b" ".to_vec(),
+        },
+    )
+    .unwrap();
+
+    assert_eq!(code, b"\n  x");
+
+    let mut recorder = ReadRecorder::new(&code);
+    let mut tree = parser
+        .parse_with(&mut |i, _| recorder.read(i), Some(&tree))
+        .unwrap();
+
+    assert_eq!(tree.root_node().to_sexp(), "(x_is_at (even_column))",);
+    assert_eq!(recorder.strings_read(), vec!["\n  x"]);
+
+    perform_edit(
+        &mut tree,
+        &mut code,
+        &Edit {
+            position: 1,
+            deleted_length: 0,
+            inserted_text: b"\n".to_vec(),
+        },
+    )
+    .unwrap();
+
+    assert_eq!(code, b"\n\n  x");
+
+    let mut recorder = ReadRecorder::new(&code);
+    let tree = parser
+        .parse_with(&mut |i, _| recorder.read(i), Some(&tree))
+        .unwrap();
+
+    assert_eq!(tree.root_node().to_sexp(), "(x_is_at (even_column))",);
+    assert_eq!(recorder.strings_read(), vec!["\n\n  x"]);
+}
+
+#[test]
 fn test_parsing_after_detecting_error_in_the_middle_of_a_string_token() {
     let mut parser = Parser::new();
     parser.set_language(&get_language("python")).unwrap();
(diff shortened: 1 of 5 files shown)


test/fixtures/test_grammars/depends_on_column/scanner.c
```diff
@@ -0,0 +1,20 @@
...
+// STATELESS
+void *tree_sitter_depends_on_column_external_scanner_create() { return NULL; }
+void tree_sitter_depends_on_column_external_scanner_destroy(void *payload) {}
+unsigned tree_sitter_depends_on_column_external_scanner_serialize(void *payload,
+                                                                  char *buffer) {
+  return 0;
+}
+void tree_sitter_depends_on_column_external_scanner_deserialize(void *payload,
+	                                                            const char *buffer,
+																unsigned length) {}
```
maxbrunsfeld, line 12: The formatting is a bit crazy here.

test/fixtures/test_grammars/depends_on_column/grammar.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "depends_on_column",
```
maxbrunsfeld, line 2: These test grammars can now be written in JS instead of JSON, and I've been gradually converting them all to JS, since it's so much more readable. Since this grammar is a new addition, would you mind converting it to JS?

lib/src/subtree.c
```diff
@@ -773,6 +774,10 @@ Subtree ts_subtree_edit(Subtree self, const TSInputEdit *input_edit, SubtreePool
       ) && (
         !invalidate_first_row ||
         child_left.extent.row > entry.tree->ptr->padding.extent.row
+      ) && (
+      	!ts_subtree_depends_on_column(*child) ||
+        !column_shifted ||
+        child_left.extent.row > edit.old_end.extent.row
```
maxbrunsfeld, line 780: This new condition makes good sense. Did you check if you can actually *remove* the previous condition? I would think that you'd only need to check `depends_on_column` in one spot. Do any tests fail if you remove the second part of this 3-part `&&` expression?

cli/src/tests/parser_test.rs
```diff
@@ -502,6 +502,80 @@ h + i
...
+
+    let mut parser = Parser::new();
+    parser
+        .set_language(&get_test_language(&grammar_name, &parser_code, Some(&dir)))
+        .unwrap();
+
+    let mut code = b"
+ x
+    "
+    .to_vec();
```
maxbrunsfeld, line 521: What do you think about writing this declaration, and the assertions about `code` using single line string literals? I think it would be easier to read. And I don't think the trailing whitespace is needed (correct me if I'm wrong there).

I know that you insert a newline in the final step. I kinda think it would be fine to write the final code assertion with a newline escape sequence, like:

```rs
assert_eq!(code, "\n  x");
```

maxbrunsfeld (commented):
Thanks for the fix @rooney. I left some minor suggestions.

test/fixtures/test_grammars/depends_on_column/scanner.c
rooney: > The formatting is a bit crazy here.

ok prettified

test/fixtures/test_grammars/depends_on_column/grammar.json
rooney: > These test grammars can now be written in JS instead of JSON, and I've been gradually converting them all to JS, since it's so much more readable. Since this grammar is a new addition, would you mind converting it to JS?

np, converted to grammar.js

cli/src/tests/parser_test.rs
rooney: > What do you think about writing this declaration, and the assertions about `code` using single line string literals? I think it would be easier to read. And I don't think the trailing whitespace is needed (correct me if I'm wrong there).
> 
> I know that you insert a newline in the final step. I kinda think it would be fine to write the final code assertion with a newline escape sequence, like:
> 
> ```rust
> assert_eq!(code, "\n  x");
> ```

ok all assertions are now single-line

lib/src/subtree.c
rooney: > Did you check if you can actually _remove_ the previous condition?

Yea I did, and I can't: it would make the Haskell-like indentation test fail -- I originally thought that it's the test that were wrong.

> I would think that you'd only need to check `depends_on_column` in one spot.

That would be the case if trees that depend-on-column make their supertree depend-on-column too. But that's not the case because currently the condition for setting the flag is:
```c
// subtree.c line 394
    if (
      self.ptr->size.extent.row == 0 &&
      ts_subtree_depends_on_column(child)
    ) {
      self.ptr->depends_on_column = true;
    }
```
So it's possible for a tree to NOT depend on column, but it contains subtree that DOES. (I find that counterintuitive, but maybe there's a reason for that?).
That's why we need to check `depends_on_column` on two spots: one when iterating the bisecting-stack, and another when delving into the children -- the condition needs to be re-checked for each subtree.

Now, I did try changing the flag-setting code to:
```c
// subtree.c line 394
    if (
      ts_subtree_depends_on_column(child)
    ) {
      self.ptr->depends_on_column = true;
    }
```
- All the tests still pass after this change.
- It's more intuitive: if a tree depends-on-column, then its supertree depends-on-column too.
- On subtree-edit, we just need to check the flag at one spot only.

But is it OK? If so, I'll make the change.

> Do any tests fail if you remove the second part of this 3-part `&&` expression?

No, all tests still pass without the second condition -- it's an optimization for when the column positions does not change e.g. if the length of `inserted_text` == `deleted_length`. Do you think it's incorrect/premature? Or is it ok but I should add a test for that (in the form of a recorder reading fewer bytes)?

maxbrunsfeld: No, then almost the entire tree would depend on its column.

A tree only depends on its start column of any of its children **that start on its first line** depend on their column. For all subsequent children, the tree's start column does not change their column.

PR #3260, opened 2024-04-06
fix: Decouple error and time parse prints

### LINKED ISSUE #3232: Parse always add time info if grammar error
### Problem

`tree-sitter parse` and a file with a grammar error always adds a line at the end that shows the same information as if the `--time` flag had been activated.

This behavior is caused by:

### tree-sitter/cli/src/parse.rs
```
...
if first_error.is_some() || opts.print_time {
...
```
Why show execution time when an error occurs, even if the corresponding flag `--time` has not been specified?

### Expected behavior

I think it is not a good idea to mix both situations in the same piece of code, or at least redirect this kind of error info via `stderr` instead of `stdout`.

WillLillis:
#3232 points out that if a parse operations fails, the CLI tool will always print the operation's time information as if the `--time` had been supplied. This PR separates the two into separate if blocks, so that time info only prints if the `--time` flag was used. If this isn't the intended behavior then I'm fine closing this down :)

Closes #3232

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -317,7 +317,7 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
             }
         }
 
-        if first_error.is_some() || opts.print_time {
+        if opts.print_time {
             write!(
                 &mut stdout,
                 "{:width$}\t{duration_ms:>7.2} ms\t{:>6} bytes/ms",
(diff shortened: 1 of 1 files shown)


amaanq:
The reason it shows the "time" even though `-t` wasn't specified is to show the stats info is to show *where* parsing went wrong; it shows the first error node's position which makes it easy to debug. It'd also be a bit awkward to *just* remove the time aspect here, and only show it *if* `-t` was specified as well imo.

WillLillis:
That makes sense, thanks so much for the explanation! :) I'll go ahead and close this PR.

amaanq:
No problem :)

LuisMorenoSerrano:
> The reason it shows the parse time even though `-t` wasn't specified is to show the stats info is to show _where_ parsing went wrong; it shows the first error node's position which makes it easy to debug. It'd also be a bit awkward to _just_ remove the time aspect here, and only show it _if_ `-t` was specified as well imo.
> 
> I think you can see why this doesn't look great as is...
> 
> image

I disagree: if I want to see time info, I can always repeat the command with flag active. IMHO for me is awkward to see time info if I have not requested it, but it's your decision.

Regards

PR #3262, opened 2024-04-06
Added debug messages for missing/misplaced test directories

buckynbrocko:
Using the `test --debug` now prints if a corpus/highlight/tags directory is not found. Also displays that top-level `corpus/` directories are no longer supported if `test/corpus/` is not found, but a top-level `corpus/` is.


amaanq:
I would've merged this at the time of the PR, but I think now after several months pretty much every active repo has adapted to no longer use a top-level `corpus` dir

PR #3263, opened 2024-04-07
Change cursor to treat child's end point exclusive

blmarket:
...when running goto_first_child_for_{byte or point}.

My problem: I have a following markdown document

```
{Node document (0, 0) - (11, 0)} document 91
  {Node section (0, 0) - (8, 0)} section 101
    // .. ignorable details
  {Node section (8, 0) - (11, 0)} section 101
    // .. ignorable details
```

(indent describes child)

When I run `goto_first_child_for_point` with (8, 0), I expected to get the second section which spans (8, 0) to (11, 0). Instead I got the first section which spans (0, 0) to (8, 0).

My suggestion: Given that the range is start-inclusive and end-exclusive, I'm wondering tree-sitter can change the walk behavior to follow the same semantics, by excluding the end position of the node when considering the child.

Code change itself is trivial. Affected test cases were updated which I believe for a good reason. For example, even though `;` is located at Point (6, 5), the test code previously had to call `goto_first_child_for_point(6,6)` to get to the `;` node.

### DIFF
--- a/cli/src/tests/tree_test.rs
+++ b/cli/src/tests/tree_test.rs
@@ -475,12 +475,13 @@ fn test_tree_cursor_child_for_point() {
     assert_eq!(c.node().kind(), "program");
 
     // descend to expression statement
-    assert_eq!(c.goto_first_child_for_point(Point::new(6, 6)), Some(0));
+    assert_eq!(c.goto_first_child_for_point(Point::new(6, 5)), Some(0));
     assert_eq!(c.node().kind(), "expression_statement");
 
     // step into ';' and back up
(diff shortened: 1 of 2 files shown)


blmarket:
I found the previous chance was intentional - https://github.com/tree-sitter/tree-sitter/pull/1640

I don't understand... when I want to goto the child located at (6, 5) Point, I feel it's more intuitive to call the method with (6, 5) as parameter, instead of weird (6, 6). I can just add 1 as a workaround for my use cases but anyway please consider this PR.

amaanq:
I'm really for this - this has bitten me too, and now that we *can* walk back to previous siblings, does it make sense to go back on #1640? @maxbrunsfeld

PR #3254, opened 2024-04-04
fix: Add lifetime to matches function

WillLillis:
This fixes an issue where a `QueryCursor` can be dropped but still have `.matches()` called on it, leading to an attempt to index into a zero-length array in the C library. [My comment](https://github.com/tree-sitter/tree-sitter/issues/3253#issuecomment-2036006041) in the issue reporting the bug has a more in-depth explanation.

Closes #3253

PR #3266, opened 2024-04-09
fix: Add lifetime to QueryCursor::captures()

WillLillis:
Realized `QueryCursor::captures()` has the same issue as `QueryCursor::matches()` that I fixed in #3254.


amaanq:
thanks!

PR #3274, opened 2024-04-10
fix: fix mismatched parenthesis when && is used

slackner:
@amaanq This fixes an issue introduced by the recent changes in https://github.com/tree-sitter/tree-sitter/pull/3070

Tree-sitter 0.22.2 generates the following incorrect code:

```c
      if (lookahead == '-' ||
          ('A' <= lookahead && lookahead <= 'Z') ||
          lookahead == '_' ||
          ('a' <= lookahead && lookahead <= 'z')) ADVANCE(244);
      if (!eof && (0 <= lookahead && lookahead <= 8) ||
          (14 <= lookahead && lookahead <= 31) ||
          (lookahead == '"' ||
          lookahead == '\'' ||
          lookahead == '\\' ||
          (127 <= lookahead && lookahead <= 159)) ADVANCE(381);
      if (lookahead != 0 &&
          (lookahead < '\t' || ' ' < lookahead) &&
          lookahead != '(' &&
          lookahead != ')') ADVANCE(245);
```

Note that for the middle `if`, the parenthesis are not balanced and compilation fails.
With this change, it looks good again:

```c
      if (lookahead == '-' ||
          ('A' <= lookahead && lookahead <= 'Z') ||
          lookahead == '_' ||
          ('a' <= lookahead && lookahead <= 'z')) ADVANCE(244);
      if ((!eof && (0 <= lookahead && lookahead <= 8)) ||
          (14 <= lookahead && lookahead <= 31) ||
          lookahead == '"' ||
          lookahead == '\'' ||
          lookahead == '\\' ||
          (127 <= lookahead && lookahead <= 159)) ADVANCE(381);
      if (lookahead != 0 &&
          (lookahead < '\t' || ' ' < lookahead) &&
          lookahead != '(' &&
          lookahead != ')') ADVANCE(245);
```

Feel free to adjust the PR / create your own one, if a different approach is preferred.

### DIFF
--- a/cli/src/generate/render.rs
+++ b/cli/src/generate/render.rs
@@ -870,39 +870,40 @@ impl Generator {
             line_break.push_str("  ");
         }
 
-        // parenthesis needed if we add the `!eof` condition to explicitly avoid confusion with
-        // precedence of `&&` and `||`
-        let (mut need_open_paren, mut need_close_paren) = (false, false);
         for (i, range) in ranges.iter().enumerate() {
             if is_included {
                 if i > 0 {
                     add!(self, " ||{line_break}");
                 }
+                // parenthesis needed if we add the `!eof` condition to explicitly avoid confusion with
+                // precedence of `&&` and `||`
+                let mut close_paren = false;
                 if range.start == '\0' {
-                    add!(self, "!eof && ");
-                    (need_open_paren, need_close_paren) = (true, true);
+                    add!(self, "(!eof && ");
+                    close_paren = true;
                 }
                 if range.end == range.start {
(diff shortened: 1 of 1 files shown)


slackner:
Note that it would probably also be valid to generate code like this, if preferred: `!eof && (... all conditions ...)`. `!eof` is only relevant for the check `lookahead == 0` if I understand the code correctly, but it also doesn't hurt for the other range checks.

amaanq:
yeah my initial intention was to wrap every condition, but you're right that only the condition that's checking if the lookahead is 0 is needed anyways, thanks

PR #3284, opened 2024-04-12
style(bindings): fix indent & line endings

ObserverOfTime:
This makes those files consistent with the core codebase and the generated `parser.c` which use 2 spaces.

However, most scanners tend to use 4 spaces from what I can tell.
Anyone who wants 4 spaces can add this to the end of `.editorconfig`:
```ini
[src/parser.c]
indent_size = 2
```

_Can be squashed._

### DIFF
--- a/cli/src/generate/templates/.editorconfig
+++ b/cli/src/generate/templates/.editorconfig
@@ -2,9 +2,6 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
-insert_final_newline = true
-trim_trailing_whitespace = true
 
 [*.{json,toml,yml,gyp}]
 indent_style = space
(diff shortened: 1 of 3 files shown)


cli/src/generate/templates/.editorconfig
```diff
@@ -37,3 +34,6 @@ indent_size = 8
 [Makefile]
 indent_style = tab
 indent_size = 8
+
+[parser.c]
+indent_size = 2
```
fwcd, line 39: I think we need to add all of the generated headers too (`tree_sitter/{alloc,array,parser}.h`). Since almost all of the generated C code and most of the external scanners I have seen in the wild use 2-space indent, maybe it would make sense to just use `indent_size = 2` for all C/C++ here?

YazanHabib:
ممكن تسجلي دخول عالفيس الخاص بي وتكون معي

PR #3288, opened 2024-04-12
fix: CRLF line endings in test titles

### LINKED ISSUE #3287: Test runner can't parse CRLF in some cases
### Problem

If the test has the `:error` attribute and the test file uses CRLF line endings, the test will report failure incorrectly.  Other attributes are also likely affected by the bug.  The programmer of the tests may not figure out the line separators are the problem for some time.  This appears to come from the following loop in `cli/src/test.rs`, maybe just tweak the closure:

```rs
for line in test_name_and_markers
    .split(|&c| c == b'\n')
    .filter(|s| !s.is_empty())
{
// ...
}
```

### Steps to reproduce

Add `:error` to a unit test that has an error deliberately introduced.  Change line separators to CRLF.  Run `tree-sitter test`.

### Expected behavior

Expect test runner to be able to parse either standard line separator.

tree-sitter 0.22.2

### Operating system/version

Windows 11

WillLillis:
This fixes an issue with CRLF line endings in test headers by filtering out '\r'. 

Closes #3287

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -518,9 +518,14 @@ fn parse_test_content(name: String, content: &str, file_path: Option<PathBuf>) -
         let (mut skip, mut platform, mut fail_fast, mut error, mut languages) =
             (false, None, false, false, vec![]);
 
-        let test_name_and_markers = c
-            .name("test_name_and_markers")
-            .map_or("".as_bytes(), |m| m.as_bytes());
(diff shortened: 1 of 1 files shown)


amaanq:
I have a different fix for this that I was just about to submit a pr for ;)

WillLillis:
> I have a different fix for this that I was just about to submit a pr for ;)

Sounds good! :)

PR #3292, opened 2024-04-13
Added F# to parsers

Thorium:
This is the most well known community project to implement F# tree-sitter parser.


amaanq:

PR #3295, opened 2024-04-14
feat(grammar): add "inherits" field if available

### LINKED ISSUE #3294: Add parent language name to grammar.json
### Problem

There's no easy way to find the parent of a child grammar in tooling.

### Expected behavior

`tree-sitter generate` should add an "inherits" field to `grammar.json`.

ObserverOfTime:
The field can be excluded entirely by changing `null` to `undefined` if we want that.

Closes #3294.

### DIFF
--- a/cli/src/generate/dsl.js
+++ b/cli/src/generate/dsl.js
@@ -223,6 +223,8 @@ function RuleBuilder(ruleMap) {
 }
 
 function grammar(baseGrammar, options) {
+  let inherits = null;
+
   if (!options) {
     options = baseGrammar;
     baseGrammar = {
@@ -237,6 +239,7 @@ function grammar(baseGrammar, options) {
     };
   } else {
     baseGrammar = baseGrammar.grammar;
+    inherits = baseGrammar.name;
   }
 
   let externals = baseGrammar.externals;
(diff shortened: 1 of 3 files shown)


cli/src/generate/dsl.js
```diff
@@ -279,6 +282,14 @@ function grammar(baseGrammar, options) {
     throw new Error("Grammar's 'name' property must not start with a digit and cannot contain non-word characters.");
   }
 
+  if (inherits && typeof inherits !== "string") {
+    throw new Error("Grammar's 'inherits' property must be a string.");
```
amaanq, line 286:
```suggestion
    throw new Error("The inheriting grammar's 'name' property must be a string.");
```

cli/src/generate/dsl.js
```diff
@@ -279,6 +282,14 @@ function grammar(baseGrammar, options) {
...
+  }
+
+  if (inherits && !/^[a-zA-Z_]\w*$/.test(name)) {
+    throw new Error("Grammar's 'inherits' property must not start with a digit and cannot contain non-word characters.");
```
amaanq, line 290:
```suggestion
    throw new Error("The inheriting grammar's 'name' property must not start with a digit and cannot contain non-word characters.");
```

Also allowing the name to start with digits might be okay

ObserverOfTime: It would not be a valid identifier (not sure if it matters here but better safe than sorry).

amaanq: yeah that's fine, if it is ever an issue we can adjust then

PR #3301, opened 2024-04-16
Fix `childrenFromFieldXXX` methods

friggeri:
These do not expect a `cursor` parameter, it's unclear to me why they're in that signature, see: 
https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/binding.js#L366


amaanq:
thanks, sorry about that!

PR #3304, opened 2024-04-17
fix(rust): use unix/wasi instead of not windows for fd

liamrosenfeld:
`not(windows)` includes non-wasi web assembly which causes issues when using the rust bindings from web non-wasi assembly.

In lib.rs, `AsRawFd` does not exist for wasm so it was causing the build to fail.

I also changed ffi.rs `_ts_dup` for consistency even though that does not cause a build issue.

This does lead to the situation where if you are not on either of those platforms, print_dot_graphs has no parameter to take a file. While odd, I think that does successfully convey that the current platform is not supported to use that function.

### DIFF
--- a/lib/binding_rust/ffi.rs
+++ b/lib/binding_rust/ffi.rs
@@ -8,7 +8,7 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 #[cfg(not(feature = "bindgen"))]
 include!("./bindings.rs");
 
-#[cfg(not(windows))]
+#[cfg(any(unix, target_os = "wasi"))]
 extern "C" {
     pub(crate) fn _ts_dup(fd: std::os::raw::c_int) -> std::os::raw::c_int;
 }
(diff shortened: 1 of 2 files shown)


clason:
What about macos/darwin?

liamrosenfeld:
Darwin is included under Unix for Rust. What I did just realize might cause an issue though is this is essentially undoing #3293 which was a fix for WASI. I wonder if there's a way to have it be for either Unix or WASI but exclude platforms such as wasm32-unknown-unknown?

amaanq:
> What I did just realize might cause an issue though is this is essentially undoing https://github.com/tree-sitter/tree-sitter/pull/3293

Yeah I was going to bring that up, but I didn't realize your case was for wasm32-unknown-unknown. You could definitely combine the cfgs for unix + wasi to exclude wasm32 then

maxbrunsfeld:
Yeah, the Rust standard library uses this attribute for the `AsRawFd` trait:

```
#[cfg(any(unix, target_os = "wasi"))]
```

https://doc.rust-lang.org/std/os/fd/trait.AsRawFd.html

liamrosenfeld:
Just pushed with that change

amaanq:
run `cargo fmt --all` @liamrosenfeld

amaanq:
thanks!

PR #3305, opened 2024-04-17
fix: leaf nodes produced by an external scanner should not be reused

amaanq:
This is the real fix for the issue described in PR #2931 

The problem is the `heredoc_start` node, given the right conditions, is being reused, which is *not* what we want to happen, because it's an external token and the edit that takes place affects this token. You can observe the bug here - https://github.com/amaanq/tree-sitter-bug/tree/master/bug6-php-external-reused, and if you patch the tree-sitter dependency to use a local path with the change in this PR, you'll observe that it's fixed. What really helped to visualize and understand the issue was printing every node's contents inside the nowdoc node + dumping the parser logging; I had also printed out every invocation of advance + what token got advanced and where, and noticed that the edited portion of the heredoc start was *not* consumed at all whatsoever, which led me to spotting one log message "reuse_node symbol:heredoc_start", which did not sound right at all.

/cc @maxbrunsfeld


maxbrunsfeld:
@amaanq I don't think we want to avoid reusing external tokens' nodes completely. There's a lot of logic around *when* we can reuse them:
* the external scanner state needs to match
* all of the text that was read by the scanner needs to be un-edited
* if the scanner called `get_column`, then the previous content on that line needs to be un-edited

Can you tell why it is that we're getting the wrong parse results here? Why exactly can't the `heredoc_start` node be reused in this case?

amaanq:
Yeah I tried to see if there was a better condition to check for but couldn't really come up with one, I'll try to summarize the issue

Given this php corpus test:

```php
<?php
$读写;

<<<漢漢字
  Heredoc detected.
漢漢字;

<<<'中文'
  Nowdoc detected..
中文;
```

We perform an edit that changes the heredoc start of the nowdoc (中文) which messes with the utf8 codepoint, thus destroying the nowdoc. When we un-edit this, the heredoc_start *should* be reparsed, but it is *not* being reparsed when we pass in the 2nd tree (tree2) to `parse`. If we pass in tree1 or None, it parses fine because the heredoc is intact or is parsed from scratch. But given that we un-edit tree2 to restore the heredoc start portion, it should be reparsed yet it's not actually being reparsed. I'm not sure what else we could do here to fix this

maxbrunsfeld:
Huh, I would think that if that node's text changed, it would be marked as edited and not reused. Maybe there's an edge case with edits that occur within code points?

maxbrunsfeld:
@amaanq i still think the problem here is specific to invalid Unicode code points in the source, and isn’t really specific to external tokens per se. In particular, I worry that we’re not extending a node’s lookahead bytes far enough, if the lexer stops on an invalid code point. We probably want to record that the node’s lookahead depends on the content up to 4 bytes past the start of the invalid code point, since characters can take up to 4 bytes.

I still haven’t debugged this locally, but that’s my sense, from the way you’ve described the bug after your detailed investigation. Let me know what you think

maxbrunsfeld:
Yes, exactly. I think the problem is something like that. The relevant code is around here:

https://github.com/tree-sitter/tree-sitter/blob/946acfd70fbbd5c1d2190f8b288cf804c955894e/lib/src/lexer.c#L363-L369

Maybe the 1 byte increment that we're doing is not enough in the general case.

maxbrunsfeld:
Awesome!

Issue #3296, opened 2024-04-15
*** stack smashing detected ***: terminated Fatal error 6: Aborted


sahirzm:
### Problem

Hi,
I updated tree-sitter on ArchLinux from 0.22.2-1 -> 0.22.4-1. After the upgrade, emacs has started to crash whenever I load any tsx file with the following backtrace:
```
*** stack smashing detected ***: terminated
Fatal error 6: Aborted
Backtrace:
emacs(+0x18a436)[0x5de6a212b436]
emacs(+0x24df3)[0x5de6a1fc5df3]
emacs(+0x25d01)[0x5de6a1fc6d01]
emacs(+0x2cf70d)[0x5de6a227070d]
/usr/lib/libc.so.6(+0x3c770)[0x727ab68fd770]
/usr/lib/libc.so.6(+0x8d32c)[0x727ab694e32c]
/usr/lib/libc.so.6(gsignal+0x18)[0x727ab68fd6c8]
/usr/lib/libc.so.6(abort+0xd7)[0x727ab68e54b8]
/usr/lib/libc.so.6(+0x25395)[0x727ab68e6395]
/usr/lib/libc.so.6(+0x11473b)[0x727ab69d573b]
/usr/lib/libc.so.6(+0x115a56)[0x727ab69d6a56]
emacs(+0x299883)[0x5de6a223a883]
/usr/lib/emacs/29.3/native-lisp/29.3-561b282f/treesit-37439c61-97df641d.eln(F747265657369742d666f6e742d6c6f636b2d666f6e746966792d726567696f6e_treesit_font_lock_fontify_region_0+0x1f2)[0x727a8eb6e832]
emacs(+0x21052e)[0x5de6a21b152e]
/usr/bin/../lib/emacs/29.3/native-lisp/29.3-561b282f/preloaded/font-lock-895216f6-1f3b244f.eln(F666f6e742d6c6f636b2d666f6e746966792d73796e746163746963616c6c792d726567696f6e_font_lock_fontify_syntactically_region_0+0x62)[0x727ab1e38f42]
emacs(+0x21052e)[0x5de6a21b152e]
/usr/bin/../lib/emacs/29.3/native-lisp/29.3-561b282f/preloaded/font-lock-895216f6-1f3b244f.eln(F666f6e742d6c6f636b2d64656661756c742d666f6e746966792d726567696f6e_font_lock_default_fontify_region_0+0x4af)[0x727ab1e36bef]
[...]

PR #3308, opened 2024-04-18
Fix soname versions

ObserverOfTime:
## Library

The soname will now include the minor version.
We _could_ also include the patch but that would be overkill.

Alternatively, we can use a manually bumped arbitrary number for the minor.

## Parsers

The soname will include the language version as the minor, matching [Gentoo](https://github.com/gentoo/gentoo/blob/master/eclass/tree-sitter-grammar.eclass#L164).

Different majors likely mean incompatible queries so we need to keep that as is.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,14 +34,14 @@ SONAME_MINOR := $(word 2,$(subst ., ,$(VERSION)))
 # OS-specific bits
 ifeq ($(shell uname),Darwin)
 	SOEXT = dylib
-	SOEXTVER_MAJOR = $(SONAME_MAJOR).dylib
-	SOEXTVER = $(SONAME_MAJOR).$(SONAME_MINOR).dylib
-	LINKSHARED += -dynamiclib -Wl,-install_name,$(LIBDIR)/libtree-sitter.$(SONAME_MAJOR).dylib
+	SOEXTVER_MAJOR = $(SONAME_MAJOR).$(SOEXT)
+	SOEXTVER = $(SONAME_MAJOR).$(SONAME_MINOR).$(SOEXT)
+	LINKSHARED += -dynamiclib -Wl,-install_name,$(LIBDIR)/libtree-sitter.$(SOEXTVER)
 else
 	SOEXT = so
-	SOEXTVER_MAJOR = so.$(SONAME_MAJOR)
-	SOEXTVER = so.$(SONAME_MAJOR).$(SONAME_MINOR)
-	LINKSHARED += -shared -Wl,-soname,libtree-sitter.so.$(SONAME_MAJOR)
+	SOEXTVER_MAJOR = $(SOEXT).$(SONAME_MAJOR)
+	SOEXTVER = $(SOEXT).$(SONAME_MAJOR).$(SONAME_MINOR)
+	LINKSHARED += -shared -Wl,-soname,libtree-sitter.$(SOEXTVER)
 endif
 ifneq ($(filter $(shell uname),FreeBSD NetBSD DragonFly),)
 	PCLIBDIR := $(PREFIX)/libdata/pkgconfig
--- a/cli/src/generate/templates/makefile
+++ b/cli/src/generate/templates/makefile
@@ -21,10 +21,6 @@ endif
 
 TS ?= tree-sitter
 
-# ABI versioning
-SONAME_MAJOR := $(word 1,$(subst ., ,$(VERSION)))
-SONAME_MINOR := $(word 2,$(subst ., ,$(VERSION)))
-
 # install directory layout
 PREFIX ?= /usr/local
 INCLUDEDIR ?= $(PREFIX)/include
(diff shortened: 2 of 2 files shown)


ObserverOfTime:
Automatically fixing parser makefiles would be cumbersome but we could print a warning. :thinking:

dundargoc (approved):

amaanq (approved):

cryptomilk:
I just stumbled upon this as the SO version changed. You actually broke the ABI in the past. So the SO major version should be 0 in the meantime.

However you might want to take a look at:

https://abimap.readthedocs.io/en/latest/readme.html

clason:
Tree-sitter is technically zerover...

cryptomilk:
Shared object versioning != release versioning

I'm the maintainer of https://libssh.org/ and we just released version 0.11. However the SO version is already libssh.so.4. I don't know what the problem is correctly bumping the SO version when needed :-)

cryptomilk:
https://www.akkadia.org/drepper/dsohowto.pdf

sogaiu:
Perhaps starting from around [here](https://www.akkadia.org/drepper/dsohowto.pdf#page=34) is relevant?

ObserverOfTime:
> You actually broke the ABI in the past.

Which is what sparked this (long overdue) change. What's the problem here?

dundargoc:
@cryptomilk I suspect you might know more about ABI versioning because I don't understand what the problem is nor what the correct solution is. I understand *something* broke but I don't know what. Did the soname version change in this PR, and if so from which version and to which version? I assumed it was just a refactor (didn't take very close look at it I admit). I don't mind reading on how to properly ABI version but I don't even know what the problem is to begin with.

cryptomilk:
Between 0.22.2 and 0.22.3 the ABI changed. To be precise, the structure `TSTreeCursor` changed.

Here is the abigail output:

```
[C] 'function void ts_query_cursor_delete(TSQueryCursor*)' at query.c:2986:1 has some indirect sub-type changes:
    parameter 1 of type 'TSQueryCursor*' has sub-type changes:
      in pointed to type 'typedef TSQueryCursor' at api.h:48:1:
        underlying type 'struct TSQueryCursor' at query.c:301:1 changed:
          type size changed from 1152 to 1216 (in bits)
          15 data member changes:
            type of 'TSTreeCursor cursor' changed:
              underlying type 'struct TSTreeCursor' at api.h:105:1 changed:
                type size changed from 192 to 256 (in bits)
                1 data member change:
                  type of 'uint32_t context[2]' changed:
                    type name changed from 'uint32_t[2]' to 'uint32_t[3]'
                    array type size changed from 64 to 96
                    array type subrange 1 changed length from 2 to 3
            'struct {QueryState* contents; uint32_t size; uint32_t capacity;} states' offset changed from 256 to 320 (in bits) (by +64 bits)
            'struct {QueryState* contents; uint32_t size; uint32_t capacity;} finished_states' offset changed from 384 to 448 (in bits) (by +64 bits)
            'CaptureListPool capture_list_pool' offset changed from 512 to 576 (in bits) (by +64 bits)
            'uint32_t depth' offset changed from 832 to 896 (in bits) (by +64 bits)
            'uint32_t max_start_depth' offset changed from 864 to 928 (in bits) (by +64 bits)
            'uint32_t start_byte' offset changed from 896 to 960 (in bits) (by +64 bits)
            'uint32_t end_byte' offset changed from 928 to 992 (in bits) (by +64 bits)
            'TSPoint start_point' offset changed from 960 to 1024 (in bits) (by +64 bits)
            'TSPoint end_point' offset changed from 1024 to 1088 (in bits) (by +64 bits)
            'uint32_t next_state_id' offset changed from 1088 to 1152 (in bits) (by +64 bits)
            'bool on_visible_node' offset changed from 1120 to 1184 (in bits) (by +64 bits)
            'bool ascending' offset changed from 1128 to 1192 (in bits) (by +64 bits)
            'bool halted' offset changed from 1136 to 1200 (in bits) (by +64 bits)
            'bool did_exceed_match_limit' offset changed from 1144 to 1208 (in bits) (by +64 bits)
```

`TSQueryCursor` has a `struct TSTreeCursor`. The array `context` was increased from 2 to 3. So the size of `struct TSQueryCursor` changed. The size increase was in the middle of the structure creating and incompatible interface change.

If TSQueryCursor could have use a pointer for TSTreeCursor there wouldn't have been a problem. It has the array at the end and the size change wouldn't be a problem.

This broke the API. From the libssh CMakeLists.txt:

```
# SOVERSION scheme: CURRENT.AGE.REVISION
#   If there was an incompatible interface change:
#     Increment CURRENT. Set AGE and REVISION to 0
#   If there was a compatible interface change:
#     Increment AGE. Set REVISION to 0
#   If the source code was changed, but there were no interface changes:
#     Increment REVISION.
```

As this was an incompatible interface change, the SO version between 0.22.2 and 0.22.3 should have change from 0.x.x to 1.0.0.

https://github.com/tree-sitter/tree-sitter/commit/09d2b23a640c60449b2b55ecae47d6483da82c95 is the commit.

Version scripts make it easier to maintain a stable API and even have old interfaces only available for old versions. abimap helps you with that.

cryptomilk:
See https://www.man7.org/linux/man-pages/man1/abidiff.1.html how to figure out ABI changes.

clason:
This broke ABI, not API.  And I am not aware of any versioning scheme that requires such a change to jump to 1.0. 

But the explanation of what changed is appreciated.

dundargoc:
> https://github.com/tree-sitter/tree-sitter/commit/09d2b23a640c60449b2b55ecae47d6483da82c95 is the commit.

Derp, so it's not even related to the changes in this PR then? OK, that explains some things at least.

dundargoc:
Is every application expected to keep track of their ABI versioning? I'm guessing this should be done via a CI job in that case? Do you have an example CI workflow where people test this?

I doubt many people is actually willing to suffer that 5000 page pdf file you linked so this is either a relatively simple check or projects simply don't do this check in general.

ObserverOfTime:
> Between 0.22.2 and 0.22.3 the ABI changed. To be precise, the structure `TSTreeCursor` changed.

You're _very_ late to [the party](https://github.com/tree-sitter/tree-sitter/issues/3296#issuecomment-2054409905).

> As this was an incompatible interface change, the SO version between 0.22.2 and 0.22.3 should have change from 0.x.x to 1.0.0.

The SO version was always 0 before. It's 0.x now.

> Version scripts make it easier to maintain a stable API and even have old interfaces only available for old versions. abimap helps you with that.

Cool, and who's going to maintain those?

---

> Do you have an example CI workflow where people test this?

Nobody wanted #3344.

cryptomilk:
> > Between 0.22.2 and 0.22.3 the ABI changed. To be precise, the structure `TSTreeCursor` changed.
> 
> You're _very_ late to [the party](https://github.com/tree-sitter/tree-sitter/issues/3296#issuecomment-2054409905).
> 
> > As this was an incompatible interface change, the SO version between 0.22.2 and 0.22.3 should have change from 0.x.x to 1.0.0.
> 
> The SO version was always 0 before. It's 0.x now.

Either you keep it at 0 and add a big warning sign, that you have to expect that the API/ABI breaks with every time, or you maintain it.

What do you think what would happen if libc would break the API/ABI with each (minor) release?

> > Version scripts make it easier to maintain a stable API and even have old interfaces only available for old versions. abimap helps you with that.
> 
> Cool, and who's going to maintain those?

The following image isn't up2date, but I would say quite good.

image

> > Do you have an example CI workflow where people test this?
> 
> Nobody wanted #3344.

:-(

clason:
> Either you keep it at 0 and add a big warning sign, that you have to expect that the API/ABI breaks with every time, or you maintain it.

Again, why are you complaining _here_ about something that happened _in the past_ when it _was_ at 0, like you suggest? What is your complaint about _this PR_?

dundargoc:
> Nobody wanted https://github.com/tree-sitter/tree-sitter/pull/3344.

Yes but that PR was insane and was anything but a small, simple check. You're mixing up what is being tested contra how it is tested.

clason:
> Nobody wanted https://github.com/tree-sitter/tree-sitter/pull/3344.

Nobody wanted automated comment spam; there was no objection against an ABI check. But now the situation is different: We _can_ fail on breaking ABI changes _if_ the soname version isn't bumped in the same commit.

ObserverOfTime:
> We _can_ fail on breaking ABI changes _if_ the soname version isn't bumped in the same commit.

What if there are multiple ABI breaking commits within a release?

clason:
Then this should not be a check that runs on every PR; only on the release/version bump PR.

ObserverOfTime:
No such thing. :)

clason:
Well, then make it a thing. Do you even project management smh.

Issue #3309, opened 2024-04-19
Parsing string literals in predicates isn't accounting for `\s`


DavisVaughan:
### Problem

Consider this example from the docs of finding doc comments in C using a match query

```
((comment)+ @comment.documentation
  (#match? @comment.documentation "^///\s+.*"))
```

This actually doesn't work!

It looks like the `\s` isn't accounted for as an allowed special case in `ts_query__parse_string_literal()` and instead the `\\` gets swallowed and we just end up with an `s` to try and match against.

Here's an example of this in the playground where we should be getting a match, but aren't

https://github.com/tree-sitter/tree-sitter/assets/19150088/8f1c7d07-1c67-4c7f-bf52-2fac2a65b078

Here's some weird stuff that _does_ match, because it's looking for a literal `s`

<img width="731" alt="Screenshot 2024-04-19 at 1 54 47 PM" src="https://github.com/tree-sitter/tree-sitter/assets/19150088/904f1fa0-bd33-4612-9db2-141475051ca6">

Looks like it is going through the `default` case here
https://github.com/tree-sitter/tree-sitter/blob/4cd23ff6b0b74f35996cfec873c3ca6b0bfac2f7/lib/src/query.c#L2023-L2038

### Steps to reproduce

-

### Expected behavior

-

### Tree-sitter version (tree-sitter --version)

0.21.0

### Operating system/version

macOS 13.6.5

DavisVaughan:
Ah I think you're right! Swapping to `\\s` does fix the playground too.

So I think that may just be a small typo on the docs page where it needs to be `\\s` rather than `\s`, and then it makes sense to me that `\n`, `\t`, `\r`, and `\0` are the only other ones handled specially.

maxbrunsfeld:
Yes, because we support escape sequences in those strings, you need to escape the backslash if you want to write a literal backslash (which you want in this case, so that the regex engine sees `\s`). Want to open a PR for the docs fix?

DavisVaughan:
Sure I'll do one monday!

PR #3310, opened 2024-04-20
Swap `\s` for `\\s` in documentation example

DavisVaughan:
As discussed in https://github.com/tree-sitter/tree-sitter/issues/3309#issuecomment-2067197293


maxbrunsfeld:
Thank you!

PR #3312, opened 2024-04-21
Build the project with Bazel.

zadlg:
Build the project with Bazel.

This commit introduces a Bazel-based build system for the core library
of `tree-sitter`.

In order not to "pollute" the existing source directories, an overlay directory
has been created at `bazel/`.

People can build the `tree-sitter` C API by doing:

```
$ cd bazel/
$ bazel build @tree-sitter
```

The following two flags are available to enable or disable specific features:

 - `--@tree-sitter//build_config:wasm=<true|false>`: enable WASM. **not yet supported.**
 - `--@tree-sitter//build_config:hide_symbols=<true|false>`: if enabled, do not expose tree-sitter symbols.

Bazel users can import `tree-sitter` into their existing Bazel workspace by
adding the following to their `WORKSPACE` file:

```starlark
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

# The git commit.
_COMMIT = ""

# SHA-256 digest of the archive.
_SHA256SUM = ""
http_archive(
    name = "tree-sitter-raw",
    sha256 = _SHA256SUM,
    urls = [
        "https://github.com/tree-sitter/tree-sitter/archive/{commit}.tar.gz".format(
            commit = _COMMIT,
        ),
    ],
    strip_prefix = "tree-sitter-{commit}".format(
        commit = _COMMIT,
    ),
)

load("@tree-sitter-raw//bazel:configure.bzl", "tree_sitter_configure")

# Configures the repository `tree-sitter`.
tree_sitter_configure(name = "tree-sitter")

# Fetches `tree-sitter` Bazel dependencies.
load("@tree-sitter//:repositories.bzl", "tree_sitter_repositories")

tree_sitter_repositories()

# Loads `tree-sitter` Bazel dependencies.
load("@tree-sitter//:deps.bzl", "tree_sitter_dependencies")

tree_sitter_dependencies()
```

Then, users can link their libraries/binaries against tree-sitter:

```starlark
cc_library(
    name = "my-parser",
    srcs = ["my-parser.cc"],
    deps = ["@tree-sitter"],
)
```


ObserverOfTime:
Several CMake PRs were rejected for being too much. Now this, this is *way* too much.

zadlg:
> Several CMake PRs were rejected for being too much. Now this, this is _way_ too much.

Hi @ObserverOfTime, thanks for your reply.
Could you elaborate on why you think this is too much?

ObserverOfTime:
16 files for a build system that very few people use.

zadlg:
> 16 files for a build system that very few people use.

I'm not going to debate Bazel's popularity, because it's pretty hard to get numbers.

I understand your feeling about having 16 files exclusively for a build system. However, since the project is quite ""simple"" (it doesn't have tons of external dependencies), I think it would be easy to maintain it.

I was inspired by [what was done for LLVM's monorepo](https://github.com/llvm/llvm-project/tree/a4bacb0f42ee8bd9d09baaae0fc0ff4650fbef2b/utils/bazel). I made a directory to self-contain everything related to Bazel (`/bazel`), so that it does not ""pollute"" the source tree.
About maintaining it, I suggest the following (again, inspired by what was done for LLVM):
 - the Bazel based build system can be broken at anytime, meaning that people who contribute to `tree-sitter` don't have to care about Bazel.
 - from time to time, someone updates/fixes the Bazel build system.

I've added new GitHub jobs, I'll make them non-blocking to ensure that non-Bazel-aware contributors won't be blocked.

What do you think?

ObserverOfTime:
In my opinion, if we were to add another (optional) build system it should be CMake (just one or two files).

zadlg:
It seems that [we cannot flag a job](https://github.com/actions/runner/issues/2347) as non-blocking, we can only use [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error), which does not make sense here.

However, an admin can flag these jobs as [non-required](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule).

zadlg:
> In my opinion, if we were to add another (optional) build system it should be CMake (just one or two files).

I can also add an optional CMake-based build system for tree-sitter if you want.
The reason I added Bazel in the first place is that I use it for my personal projects and for a few projects in the company I work for.
In other words, I'm not trying to add Bazel because it's popular (or not).

ObserverOfTime:
Downstream projects can use whatever build system they want. It doesn't need to be added _here_.

maxbrunsfeld:
Hi, this is great work but I don’t want to maintain this in this repository.

zadlg:
> Hi, this is great work but I don’t want to maintain this in this repository.

I understand. Thanks anyway for your reply.

maxbrunsfeld:
Awesome, thank you @zadlg !

PR #3315, opened 2024-04-24
fix(zig): make usable as a zig dependency

bfredl:
Make this usable as a dependency of a zig 0.12 project:

- build.zig.zon must exist to be used as a zig dependency
- public headers need to be installed as part of the artifact

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,0 +1,10 @@
+.{
+    .name = "tree-sitter",
+    .version = "0.22.5",
+    .paths = .{
+        "build.zig",
+        "build.zig.zon",
+        "lib/src",
+        "lib/include",
+    },
+}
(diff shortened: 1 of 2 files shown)


build.zig.zon
```diff
@@ -0,0 +1,10 @@
+.{
+    .name = "tree-sitter",
+    .version = "0.22.5",
```
clason, line 3: Note: another manifest file that needs to be bumped on release; should be added to whatever workflow you have for that @amaanq

softprops:
Can you share an example of a what a zig build consuming this would look like?

bfredl:
@softprops https://github.com/neovim/neovim/pull/28344/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR57 and usages of this `treesitter` package below ( ie `nvim_exe.linkLibrary(treesitter.artifact("tree-sitter"));` )

softprops:
Very cool! Is it possible to use something like that for libraries vs executables? Something like `treesitter.module("tree-sitter")`. I wasn't able to get that to work but I'm hopeful there is a way.

PR #3318, opened 2024-04-25
Improve handling of serialization buffer overflows

maxbrunsfeld:
Refs https://github.com/tree-sitter/tree-sitter-julia/issues/128

This is a bug in the Julia grammar's external scanner, where the scanner will write past the size of the serialization buffer. Previously, there was no explicit guard against this happening, it would just corrupt other state, and cause confusing crashes downstream.

Now, it is handled better in two different ways:
* When running a parser natively, we can at least crash with a more informative error location if the scanner indicates that it has written beyond the end of the buffer:

    ```
    Assertion failed: (length <= TREE_SITTER_SERIALIZATION_BUFFER_SIZE), function ts_parser__external_scanner_serialize, file parser.c, line 404.
    ```
* When running a parser via WASM (e.g. [in Zed](https://zed.dev/blog/language-extensions-part-1)), because the parser is only writing to the wasm memory, we can completely prevent a crash, and prevent it from affecting anything, except for failing the current parse.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -397,10 +397,12 @@ static unsigned ts_parser__external_scanner_serialize(
       self->lexer.debug_buffer
     );
   } else {
-    return self->language->external_scanner.serialize(
+    uint32_t length = self->language->external_scanner.serialize(
       self->external_scanner_payload,
       self->lexer.debug_buffer
     );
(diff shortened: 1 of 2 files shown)


Romek-cloud (commented):
Ok

PR #3323, opened 2024-04-27
fix: don't error out on missing/malformed tree-sitter fields when `--no-bindings` is passed in

amaanq:
If `--no-bindings` is passed in, a default tree-sitter field doesn't get inserted, which can lead to build issues only in *this* case

### DIFF
--- a/cli/src/generate/grammar_files.rs
+++ b/cli/src/generate/grammar_files.rs
@@ -243,7 +243,11 @@ pub fn generate_grammar_files(
         },
     )?;
 
-    let (_, package_json) = lookup_package_json_for_path(package_json_path_state.as_path())?;
+    let package_json = match lookup_package_json_for_path(package_json_path_state.as_path()) {
+        Ok((_, p)) => p,
+        Err(e) if generate_bindings => return Err(e),
+        _ => return Ok(()),
+    };
 
     // Do not create a grammar.js file in a repo with multiple language configs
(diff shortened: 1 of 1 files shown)


clason:
That's not quite the fix I had in mind; I wanted `--no-bindings` to ignore all malformed manifest files that are not necessary for generating a `parser.c`.

(Not that other improvements can't be made still, mind you.)

cli/src/generate/grammar_files.rs
```diff
@@ -243,7 +243,8 @@ pub fn generate_grammar_files(
         },
     )?;
 
-    let (_, package_json) = lookup_package_json_for_path(package_json_path_state.as_path())?;
+    let (_, package_json) = lookup_package_json_for_path(package_json_path_state.as_path())
+        .map_err(|e| generate_bindings.then(|| Err(e)).unwrap_or(Ok(())))?;
 
     // Do not create a grammar.js file in a repo with multiple language configs
     if !package_json.has_multiple_language_configs() {
```
clason, line 250: wouldn't that trigger the error here? you probably want to pull the `!generate_bindings` early return earlier?

clason: (Drive-by comment; feel free to disregard.)

clason:
Can confirm this fixes the related nvim-treesitter issues!

PR #3339, opened 2024-05-05
Support C99 (again, but for real this time!)

DavisVaughan:
In https://github.com/tree-sitter/tree-sitter/pull/3088, there was a move to C11 over (officially) supporting C99.

I'd like to test the waters to see if there is some flexibility there, and make a request to more officially support C99 instead. If this is not something you are willing to consider, that's totally fine, but it didn't seem _too_ complicated to make it so that tree-sitter can compile with C99.

According to https://github.com/tree-sitter/tree-sitter/pull/2965#issuecomment-1934843618, there was some desire to "keep" tree-sitter C99 compliant, but in reality it hasn't been for awhile due to the usage of anonymous structs and unions, which was surfaced with the usage of `-pedantic` there.

This PR first has 3 commits that _name_ these anonymous structs and unions:
- A `state` union inside `ExternalScannerState`, which seemed to make sense as a name
- A `children_state` struct within a union, which I do think actually makes things more clear at its calls sites, because it is often guarded by something like `if (child_count > 0)`
- A `data` union inside `SubtreeHeapData`, which is actually what `children_state` lives in. This name goes along with the fact that you use `data` as a name in the inline case, i.e. typically something like `if (inlined) self->data else self->ptr->data`.

I will be the first to admit, having to name these structs and unions does make the code a little uglier because of the added indirection. But it might be worth this little bit of code style pain for C99 support.

The commits after those:
- Update everything from `std=c11` to `std=c99` everywhere I could find
- Removes some usages of `#define _POSIX_C_SOURCE 200112L`, added in https://github.com/tree-sitter/tree-sitter/pull/3088. Note that was using the 2001 standard, not 2011, so I'm not entirely sure if that commit was necessary in that PR. There is one other call to this in `lib.c` related to https://github.com/tree-sitter/tree-sitter/pull/295, which I left in place. (Update: Turns out this is required for `fdopen()` support in non-amalgamated builds, presumably where you didn't go through `lib.c` so you don't have `_POSIX_C_SOURCE` defined yet. I guess in theory you _only_ need to put `_POSIX_C_SOURCE` in the individual files where it is required rather than in `lib.c` itself, but I didn't touch that.)

---

My motivation for this is that I am working on wrapping tree-sitter in an R package, and R's history with C99 is a big gnarly. R is currently at R 4.4.0 (released a month ago). Only as of R 4.3.0 (last year) did package authors gain a way to specify the C std version with `USE_Cxx` here:
https://github.com/wch/r-source/blob/bacd56452465bd7fd95839c3e5dd1adf21be4de6/doc/NEWS.Rd#L1572-L1604

Additionally at that time, there was a change in R on Windows to allow the default C compiler to be used (typically C17), rather than always defaulting to C99. For Mac and Linux, it was already doing this:
https://github.com/wch/r-source/blob/bacd56452465bd7fd95839c3e5dd1adf21be4de6/doc/NEWS.Rd#L1545-L1548

Unfortunately, we typically try to support 5 minor versions of R, back to R 4.0.0 right now, so C99 support is actually still pretty important for us 😞 (R users are typically pretty slow to update their R version), which is what leads me here to see if tree-sitter can be tweaked for C99 support!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ OBJ := $(SRC:.c=.o)
 # define default flags, and override to append mandatory flags
 ARFLAGS := rcs
 CFLAGS ?= -O3 -Wall -Wextra -Wshadow -pedantic
-override CFLAGS += -std=c11 -fPIC -fvisibility=hidden
+override CFLAGS += -std=c99 -fPIC -fvisibility=hidden
 override CFLAGS += -Ilib/src -Ilib/src/wasm -Ilib/include
 
 # ABI versioning
--- a/Package.swift
+++ b/Package.swift
@@ -37,5 +37,5 @@ let package = Package(
                 ],
                 sources: ["src/lib.c"]),
     ],
-    cLanguageStandard: .c11
+    cLanguageStandard: .c99
 )
--- a/build.zig
+++ b/build.zig
@@ -8,7 +8,7 @@ pub fn build(b: *std.Build) void {
     });
 
     lib.linkLibC();
-    lib.addCSourceFile(.{ .file = .{ .path = "lib/src/lib.c" }, .flags = &.{"-std=c11"} });
+    lib.addCSourceFile(.{ .file = .{ .path = "lib/src/lib.c" }, .flags = &.{"-std=c99"} });
     lib.addIncludePath(.{ .path = "lib/include" });
     lib.addIncludePath(.{ .path = "lib/src" });
 
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -576,7 +576,7 @@ impl Loader {
                 cc_config.cpp(true);
                 eprintln!("Warning: Using a C++ scanner is now deprecated. Please migrate your scanner code to C, as C++ support will be removed in the near future.");
             } else {
-                cc_config.std("c11");
+                cc_config.std("c99");
             }
             cc_config.file(scanner_path);
         }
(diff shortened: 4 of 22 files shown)


clason:
There is not, since wasmtime requires C11.

DavisVaughan:
Not everyone uses the wasm features though!

ObserverOfTime:
But everyone uses the C11 features. Unfortunately, you'll just have to drop support for versions that default to C99.

clason:
But tree-sitter does.

DavisVaughan:
@clason I don't understand what you are trying to say. Building the C library of tree-sitter without `TREE_SITTER_FEATURE_WASM` defined does not require wasmtime IIUC

DavisVaughan:
> But everyone uses the C11 features

Sorry, I don't think I understand. Who is using them?

ObserverOfTime:
> @clason I don't understand what you are trying to say. Building the C library of tree-sitter without `TREE_SITTER_FEATURE_WASM` defined does not require wasmtime IIUC

But building it *with* `TREE_SITTER_FEATURE_WASM` defined requires wasmtime and consequently C11.

> Sorry, I don't think I understand. Who is using them?

Tree-sitter itself and all the parsers.

clason:
> @clason I don't understand what you are trying to say. Building the C library of tree-sitter without `TREE_SITTER_FEATURE_WASM` defined does not require wasmtime IIUC

Yes, but you can't use "optional standards". And Max considers wasmtime a required feature (for Zed).

DavisVaughan:
Alright, it was worth a shot. Thanks!

PR #3343, opened 2024-05-06
ci: add ABI compatibility workflow

ObserverOfTime:
Temporarily includes #3184 for testing.


ObserverOfTime:
GitHub refuses to set the correct permissions. I'll open a new one.

PR #3344, opened 2024-05-06
 ci: add ABI compatibility workflow 

ObserverOfTime:
~~Temporarily includes https://github.com/tree-sitter/tree-sitter/pull/3184 for testing.~~

### DIFF
--- a/.github/workflows/abicc.yml
+++ b/.github/workflows/abicc.yml
@@ -1,0 +1,88 @@
+name: Check ABI changes
+
+on:
+  pull_request_target:
+    branches: [master]
+    paths:
+      - lib/include/tree_sitter/api.h
+
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: true
+
+jobs:
+  dump:
+    name: Dump ${{matrix.version}} ABI data
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        version: [old, new]
+        include:
+          - { version: old, ref: "${{github.event.pull_request.base.sha}}" }
+          - { version: new, ref: "${{github.event.pull_request.head.sha}}" }
+    steps:
+      - name: Checkout ${{matrix.version}} commit
+        uses: actions/checkout@v4
+        with:
+          ref: ${{matrix.ref}}
+          sparse-checkout: |
+            lib/src/
+            lib/include/
+      - name: Compile library
+        run: make libtree-sitter.so
+        env:
+          CFLAGS: -Og -g -fno-omit-frame-pointer
+      - name: Run ABI dumper
+        id: abi-dumper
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
Apparently PRs from forks [can't change token permissions](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#restrictions-on-repository-forks). Thanks GitHub.

ObserverOfTime:
I changed it to `pull_request_target` but it can no longer be previewed here.
See https://github.com/ObserverOfTime/tree-sitter/pull/3#issuecomment-2096527160 (non-breaking change).

ObserverOfTime:
Reverting 09d2b23a640c60449b2b55ecae47d6483da82c95 is only reported as breaking if the private ABI is also checked. See https://github.com/ObserverOfTime/tree-sitter/pull/4#issuecomment-2096536144.

ObserverOfTime:
Should a warning annotation be issued if incompatible changes are detected?

dundargoc:
> I changed it to `pull_request_target` but it can no longer be previewed here. See [ObserverOfTime#3 (comment)](https://github.com/ObserverOfTime/tree-sitter/pull/3#issuecomment-2096527160) (non-breaking change).

That comment is incredibly intrusive. It looks like this workflow is only run if `lib/include/tree_sitter/api.h` is changed so at the very least not every PR will have this comment. I'd still argue it sets a bad precedent to start conveying CI information via comments as it disrupts discussion. This practice taken to its extreme would be https://github.com/microsoft/winget-pkgs/pull/151704, where any discussion would be borderline impossible amidst the bots bickering with each other.

More importantly, this PR combines `pull_request_target` with checking out the HEAD of the PR which is very dangerous as that allows the pull requester to make changes to the tree-sitter repository. See e.g. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/:

> ***TL;DR**: Combining `pull_request_target` workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.*

I know you've only enabled pull request in `permissions`, but the problem is that future contributors will need to be on guard when they make any changes to the workflow. Worse, they might not even be aware of this problem (it's surprisingly not well-known) which makes this super dangerous, and in my opinion, a ticking time bomb. I think using `pull_request` and skipping commenting altogether is a more ideal solution. If the workflow fails they can just read the CI logs to find more information.

ObserverOfTime:
> I think using `pull_request` and skipping commenting altogether is a more ideal solution.

I could make the action parse the XML report instead and turn it into annotations but it would be considerably difficult.

dundargoc:
I think I might have missed something. Is simply failing the CI job not an option?

ObserverOfTime:
The CI should not fail. It should, at most, warn.

ObserverOfTime:
Alternatively I could reduce the comment to just the "problems" tables.

PR #3352, opened 2024-05-08
fix(rust): access `CARGO_MANIFEST_DIR` at runtime in build script

redsun82:
The `CARGO_MANIFEST_DIR` environment variable should be accessed by `build.rs` at run time rather than compile time. This was for example causing issues when importing `tree-sitter` via
[`rules_rust`](https://github.com/bazelbuild/rules_rust) in bazel, where compilation and running happen in separate environments.

### DIFF
--- a/lib/binding_rust/build.rs
+++ b/lib/binding_rust/build.rs
@@ -1,6 +1,6 @@
 use std::{
     env, fs,
-    path::{Path, PathBuf},
+    path::PathBuf,
 };
 
 fn main() {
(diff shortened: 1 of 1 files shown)


hendrikvanantwerpen (approved):

dcreager (approved):
This looks right to me, too, but I haven't touched this part of the code in awhile.  @maxbrunsfeld and/or @amaanq does this look :+1: to you?

amaanq:
yeah this is great! thanks for the reminder @dcreager

amaanq (approved):

PR #3355, opened 2024-05-08
feat(dsl)!: support other JS runtimes

ObserverOfTime:
## Problem

The `generate` command currently only supports Node out of the box and Bun via a shim that calls `bun -`.

## Solution

* Replace the Node-specific `global` with the universal `globalThis`.
* Replace the Node-specific `require()` with the universal `await import()`.
  - This also adds support for ES modules but no longer works in Node out of the box.
* Adapt the environment & stdout functions depending on the runtime.
* Set the required arguments for known runtimes (Node, Bun, Deno).
  - The user can still specify a different command or the full path for custom runtimes.
 
 ## Notes
 
 - In Deno, the DSL only needs `--allow-env --allow-read` but we use `--allow-all` since grammars may need more.
   - The other runtimes are not restrictive anyway.
 - QuickJS does not support reading input from stdin at the moment.
 - MozJS does not support reading input from stdin *as a module* at the moment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,9 +101,9 @@ jobs:
         echo "image = \"$image\""                              >> Cross.toml
         echo "CROSS_CONFIG=$PWD/Cross.toml"                    >> $GITHUB_ENV
 
-        echo "FROM ghcr.io/cross-rs/$target:edge"              >> Dockerfile
-        echo "ENV DEBIAN_FRONTEND=noninteractive"              >> Dockerfile
-        echo "RUN apt-get update && apt-get install -y nodejs" >> Dockerfile
+        echo "FROM ghcr.io/cross-rs/$target:edge"                            >> Dockerfile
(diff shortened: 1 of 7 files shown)


ObserverOfTime:
We could also support QuickJS & MozJS by writing a temporary file, but loading node modules (if needed) would be tricky.

ObserverOfTime:
Cross does not seem to support `--input-type`. Is it using a version of Node older than 12.0?

Windows is, as usual, weird:

```
 TypeError [ERR_INVALID_MODULE_SPECIFIER]: Invalid module "\\?\D:\a\tree-sitter\tree-sitter\test\fixtures\test_grammars\uses_current_column\grammar.js" is not a valid package name imported from D:\a\tree-sitter\tree-sitter\cli\[eval1]
 ```

ObserverOfTime:
> Is it using a version of Node older than 12.0?

Right, [of course it is](https://github.com/cross-rs/cross/blob/764d258e8b88323df7e88b1b6f178d09f951966a/docker/Dockerfile.cross#L9)…

PR #3356, opened 2024-05-08
feat: node kinds iterator

andrewgazelka:


ObserverOfTime:
I don't think this needs to be in the library.

andrewgazelka:
> I don't think this needs to be in the library.

I'm curious why do you think this is? Just because it adds unnecessary bloat and would not be used often?

ObserverOfTime:
Pretty much.

PR #3358, opened 2024-05-09
Move cli::generate into its own crate

### LINKED ISSUE #3357: Split tree_sitter_cli::generate off into its own crate
### Problem

The tree-sitter-cli `generate` module is useful for tools like `rust-sitter`, but because it is part of tree-sitter-cli, it brings along lots of unnecessary dependencies like `webbrowser`.

### Expected behavior

It would be nice if there was a tree-sitter-generate package with more minimal dependencies.

taktoa:
Fixes #3357.

The new `tree-sitter-generate` crate has much fewer dependencies than `tree-sitter-cli`, while being enough for tools like `rust-sitter-tool`.


ObserverOfTime:
It might be better to wait for the command refactoring in 0.24 where `generate` will no longer be responsible for bindings.

adrijana-belgrade (approved):

PR #3385, opened 2024-05-25
feat: add `fuzz` subcommand

amaanq:
# Motivation

It is a common occurrence where users write a complex parser or have a scanner that is quite involved for their parser, and this is a very common pitfall for fuzzing bugs. It is especially prevalent when dealing with properly structured source code vs using something like libFuzzer and feeding in random bytes (which is still useful too!), triggering parsing paths in the core library or scanner that are uncommon but can be the source for some nasty bugs. 

This is typically really hard to reason about and debug reproducibly, and relies on luck or a user submitting a perfect test case that exemplifies these bugs. A good example is the Rust crash bug Max, Marshall, and I fixed a few months ago, and the still-existing Javascript infinite loop bug given certain input, and the Typescript one I fixed a few months back.

# Solution

Luckily, our testing infrastructure has most of the groundwork laid out for us here to edit source code at random, effectively fuzzing the parser with *real* input. It does rely on the tests that grammar authors write for the grammar, so it's really important to write more than one or two tests to see if any edge cases resulting in crashes, infinite loops, or inconsistent parses come up.

I moved a lot of the code from the test code to fuzz, and moved any needed dependencies out of `dev-dependencies` to `dependencies`, and have the cli build a "sanitize" build for fuzzing which we can have link in ASAN or UBSAN as well for certain targets if desired - though when trying this out ASAN was not playing nicely for me. We also have a check for memory leaks  with `allocations::record` which can help users quickly spot if their scanner leaks memory. I don't think this is *fully* ready, I still want to move some of the configurable env vars to be flags/args instead, but those are nice to have anyways when running and debugging our own tests.

I've been using this for a few weeks now and it works pretty well


maxbrunsfeld (approved):
Awesome, this will be super useful.

amaanq:
thanks :see_no_evil: @maxbrunsfeld can I poke you to look at #2577?

PR #3389, opened 2024-05-26
build(deps): bump wasmtime to v21

amaanq:
It seems like wasmtime got rid of `DEP_WASMTIME_C_API_WASM_INCLUDE` and changed `index` to `__private`, I'm not sure exactly why (couldn't really find a reason in the PR that contained the field name change), but it seems to be okay on my end. Maybe you might know more @maxbrunsfeld


maxbrunsfeld (approved):
Thanks!

It's nice that there aren't the two separate include paths anymore.

PR #3395, opened 2024-05-30
feat(lib): make lib no_std

### LINKED ISSUE #3391: Support nostd
### Problem

Have you considered adding support for nostd? Or will similar PRs be allowed to be merged?

### Expected behavior

_No response_

cathaysia:
close #3391

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -26,10 +26,13 @@ include = [
 ]
 
 [features]
+default = ["std"]
+std = ["regex/std", "regex/perf", "regex-syntax/unicode"]
 wasm = ["wasmtime-c-api"]
 
 [dependencies]
-regex.workspace = true
+regex = { version = "1.10.4", default-features = false, features = ["unicode"] }
+regex-syntax = { version = "0.8.4", default-features = false }
 tree-sitter-language = { version = "0.1", path = "language" }
 
 [dependencies.wasmtime-c-api]
--- a/lib/binding_rust/ffi.rs
+++ b/lib/binding_rust/ffi.rs
@@ -9,6 +9,7 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 include!("./bindings.rs");
 
 #[cfg(any(unix, target_os = "wasi"))]
+#[cfg(feature = "std")]
 extern "C" {
     pub(crate) fn _ts_dup(fd: std::os::raw::c_int) -> std::os::raw::c_int;
 }
@@ -18,7 +19,7 @@ extern "C" {
     pub(crate) fn _ts_dup(handle: *mut std::os::raw::c_void) -> std::os::raw::c_int;
 }
 
-use std::{marker::PhantomData, mem::ManuallyDrop, ptr::NonNull, str};
+use core::{marker::PhantomData, mem::ManuallyDrop, ptr::NonNull, str};
 
 use crate::{
     Language, LookaheadIterator, Node, Parser, Query, QueryCursor, QueryError, Tree, TreeCursor,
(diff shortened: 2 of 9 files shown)


lib/binding_rust/ffi.rs
```diff
@@ -9,6 +9,7 @@ include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 include!("./bindings.rs");
 
 #[cfg(any(unix, target_os = "wasi"))]
+#[cfg(feature = "std")]
```
ObserverOfTime, line 12: Should be merged with the above condition and also added to the windows version.

lib/binding_rust/lib.rs
```diff
@@ -2025,22 +2026,25 @@ impl Query {
...
-                        let regex = &string_values[p[2].value_id as usize];
-                        text_predicates.push(TextPredicateCapture::MatchString(
-                            p[1].value_id,
-                            regex::bytes::Regex::new(regex).map_err(|_| {
-                                predicate_error(row, format!("Invalid regex '{regex}'"))
-                            })?,
-                            is_positive,
-                            match_all,
-                        ));
+                        #[cfg(feature = "std")]
```
ObserverOfTime, line 2029: This should apply to the entire match case.

lib/binding_rust/lib.rs
```diff
@@ -1,22 +1,32 @@
 #![doc = include_str!("./README.md")]
-
+#![cfg_attr(not(feature = "std"), no_std)]
 pub mod ffi;
 mod util;
 
+#[cfg(not(feature = "std"))]
+extern crate alloc;
```
ObserverOfTime, line 7: Is this required?

cathaysia: yes, alloc crate not included by default, we need include it by hand.

ref: https://docs.rust-embedded.org/book/collections/index.html#using-alloc

> [!NOTE]
> rust seems has no `#![feature(alloc)]` at this time.

ObserverOfTime:
It seems that regex [does](https://docs.rs/regex/latest/regex/index.html#ecosystem-features) support `no_std`, albeit with some performance loss.

cathaysia:
yes, I add regex to no_std. But as a trade-off, regex itself is removed from the workspace. Because `default-features = false ` need to be set in the workspace. This will cause changes to other crates

ObserverOfTime:
There are more features that need to be toggled, and I also see [this](https://github.com/rust-lang/regex/blob/master/Cargo.toml#L40-L41) note now, so maybe regex should just be disabled in `no_std` after all.

cathaysia:
that note is outdated. I test this pr for --target=thumbv7m-none-eabi, and this passed.

lib/Cargo.toml
```diff
@@ -26,10 +26,12 @@ include = [
 ]
 
 [features]
+default = ["std"]
+std = ["regex/std"]
 wasm = ["wasmtime-c-api"]
 
 [dependencies]
-regex.workspace = true
+regex = { version = "1.10.4", default-features = false, features = ["unicode"] }
```
ObserverOfTime, line 34:
```suggestion
regex = { version = "1.10.4", default-features = false, features = ["perf", "unicode", "regex-syntax/unicode"] }
```

lib/Cargo.toml
```diff
@@ -26,10 +26,12 @@ include = [
 ]
 
 [features]
+default = ["std"]
+std = ["regex/std"]
```
ObserverOfTime, line 30:
```suggestion
std = ["regex/std", "regex-syntax/std"]
```

cathaysia:
this pr should ready be merged.

cathaysia:
@amaanq anything block this?

PR #3397, opened 2024-05-31
chore: Hoist out common subexpressions in satisfies_text_predicates

osiewicz:
This commit stores the result of text predicates evaluation in a separate variable to ensure that they're computed just once. As is, it is possible for e.g. #match predicates to match node against a regex twice.


osiewicz:
Hm, I'm not sure why that change breaks WASM tests specifically..

amaanq:
Sorry about that @osiewicz - I updated the js grammar w/ a bug fix for rare infinite loops, which caused the failure here but I've fixed that now. Can you rebase?

osiewicz:
No worries at all @amaanq; I was just confused because I wouldn't have expected this change to break anything specifically for WASM.

amaanq:
yeah for sure - I wouldn't expect that to break wasm either. thanks for the perf improvement!

PR #3404, opened 2024-06-05
cli: Use Hir as underlying regex representation 

osiewicz:
This picks up work from #383 . This PR is a bit simpler, as it doesn't include any changes for `CharacterSet` (which have already been applied in the meantime).
I've benchmarked resulting grammars and found that the performance of parsers is identical. In fact, `parser.c` for e.g. Rust is bit-for-bit identical.


ObserverOfTime:
Presumably related error:
```
 thread 'main' panicked at cli/src/generate/render.rs:259:63:
no entry found for key
```

amaanq:
Superseded by #3831

PR #3408, opened 2024-06-08
build(zig): Git ignore updated Zig cache directory

mochalins:
Recent versions of Zig have renamed the previously used `zig-cache` directory to `.zig-cache`. While `zig-out` and `zig-cache` were covered by the previous `.gitignore`, the new `.zig-cache` also should be handled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ log*.html
 *.xcodeproj
 .vscode
 .cache
+.zig-cache
 
 fuzz-results
 


.gitignore
```diff
@@ -4,6 +4,7 @@ log*.html
 *.xcodeproj
 .vscode
 .cache
+.zig-cache
```
amaanq, line 7:
```suggestion
zig-cache
```

Pretty sure it's just `zig-cache`, no dot.

mochalins: Please refer to the PR description; `zig-cache` is already covered in the `.gitignore` by `zig-*`. This is an addition to handle the new rename to `.zig-cache`.

https://github.com/ziglang/zig/pull/20115

amaanq: sorry, did not read the PR desc, my bad

PR #3414, opened 2024-06-12
fix(cli): installation via authenticated proxy

tuananhvga:
The fix provides the ability to run installation script behind an authenticated proxy


amaanq:
thank you!

PR #3417, opened 2024-06-13
Use tabs only in recipes so as to avoid confusing 'make'

mjambon:
Some version of `make` that we run on Windows (Cygwin) chokes on a tab on line 32 with the following error message:
```
Makefile:32: *** recipe commences before first target.  Stop.
```
I don't know why this problem is showing up only now. Also, we don't have this error on other platforms (Linux, MacOS). We're in the process of upgrading tree-sitter from 0.20 to 0.22.6. I didn't spot a change in the makefile that would suddenly cause this error.

I apologize for not taking the time to reproduce the error outside of our private CI system. However, I think it's good hygiene to not use tabs outside of recipes anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,12 @@ PCLIBDIR ?= $(LIBDIR)/pkgconfig
 
 # collect sources
 ifneq ($(AMALGAMATED),1)
-	SRC := $(wildcard lib/src/*.c)
-	# do not double-include amalgamation
-	SRC := $(filter-out lib/src/lib.c,$(SRC))
+  SRC := $(wildcard lib/src/*.c)
+  # do not double-include amalgamation
+  SRC := $(filter-out lib/src/lib.c,$(SRC))
 else
-	# use amalgamated build
-	SRC := lib/src/lib.c
+  # use amalgamated build
+  SRC := lib/src/lib.c
 endif
 OBJ := $(SRC:.c=.o)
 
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
It *was* using tabs.

mjambon:
> Windows is not supported by the Makefile

Has it changed since 0.20.6? I know 0.20.6 is old but it was working fine for us.

This [commit](https://github.com/tree-sitter/tree-sitter/commit/799833f9cf10b535978443b95604c89cce66e32d) introduces:
```
ifeq ($(OS),Windows_NT)
	$(error "Windows is not supported")
```

mjambon:
Great news: I found out that building tree-sitter worked for us on Windows (after removing the `ifeq ($(OS),Windows_NT)`).  We run the following commands:
```
        make PREFIX="$prefix" CFLAGS="-O3 -Wall -Wextra"
        make PREFIX="$prefix" install
```
(copy-pasted from our CI job)

It uses `CC=x86_64-w64-mingw32-gcc` for the compiler.

I also see this comment in our CI:
```
        # tree-sitter fails to compile without an ar, you can use
        # CC=x86_64-w64-mingw32-gcc but there is no AR=x86_64-w64-mingw32-ar
        which ar
        ar --version
```

ObserverOfTime:
It works in shells that emulate Unix and builds Unix-like libraries.
It does not (normally) work in PowerShell or CMD and can't build DLLs.
It's not easy to support Windows in Makefiles, and CMake was rejected in the past.

> Are there instructions on how to build tree-sitter on Windows/Cygwin?

You can build a static library via zig. A shared library can also be built by adapting `build.zig` (PR welcome).

mjambon:
> > Are there instructions on how to build tree-sitter on Windows/Cygwin?
>
> You can build a static library via zig. A shared library can also be built by adapting build.zig (PR welcome).

As reported earlier, it works on Cygwin. All we need is a Windows check that excludes Cygwin.

The following should work. I can test it and make a PR:

```
ifneq ($(OSTYPE),cygwin)
  ...
endif
```

ObserverOfTime:
> All we need is a Windows check that excludes Cygwin.

What about MSYS, MinGW, or Git Bash?

mjambon:
I wasn't able to modify the makefile so that it only produces an error for a non-Cygwin Windows environment. I don't have access to a Windows machine, only to a Github Actions job that was set up by someone else. As usual, this comes with no shell access. The approaches I tried are:

* using Bash to make it print the value of the OSTYPE variable (`OSTYPE := $(shell bash -c 'echo "$$OSTYPE"')`), expecting `cygwin`.
* using `uname -s` and check if its output starts with `CYGWIN`.

:man_shrugging: 

Meanwhile we're using a patched makefile that removes the check for Windows.

ObserverOfTime:
Both of those approaches will fail on PowerShell/CMD.
The Makefile can't properly support Windows without significant hacks.
We would have to either adapt build.zig and/or add a CMakeLists.txt file.

aryx:
@maxbrunsfeld  @amaanq can we merge this? (not sure what's going on with the CI failure).
Also looks like Martin and I are not part of the tree-sitter org anymore. When was this changed?

Issue #3426, opened 2024-06-21
Replace `sprintf()` with `snprintf()`


DavisVaughan:
### Problem

I noticed that you use `sprintf()` in one place for the logger, but the slightly safer `snprintf()` in a different place. Would it be possible to consistently use `snprintf()`?

https://github.com/tree-sitter/tree-sitter/blob/3da7deedd8e0a3decd2db32b76ea51513876914e/lib/src/parser.c#L26-L36

CRAN (for R packages) grumbles about the fact that `sprintf()` is used, since it "is not safe", requiring a manual patch on my end.

### Steps to reproduce

NA

### Expected behavior

Use `snprintf()` instead

### Tree-sitter version (tree-sitter --version)

main

### Operating system/version

macOS 13.6.5

ObserverOfTime:
PR welcome.

PR #3430, opened 2024-06-24
Swap `sprintf()` for `snprintf()`

DavisVaughan:
Closes https://github.com/tree-sitter/tree-sitter/issues/3426

(Matched the style of the other `snprintf()` call below this one)


maxbrunsfeld:
Thanks!

PR #3438, opened 2024-06-26
doc(index): Added alternative VHDL link

jpt13653903:
This adds a VHDL link to the new (and maintained) VHDL parser and highlight queries, currently in alpha-testing.

Ref: https://github.com/nvim-treesitter/nvim-treesitter/discussions/6701


jpt13653903:
By the way - I'm also very happy to add the new link as "an alternative" instead of a direct update - let me know and I'll change it.

I just don't want users to be directed to a repo with broken highlights queries and that is not maintained anymore, and then think that that is the only option.

amaanq:

Issue #2869, opened 2024-01-19
Query `(_ . (name) @firstChild)` captures every `name` node


RedCMD:
The query `(_ . (_) @firstChild)` will correctly capture the first child of every named node
returns: `function_definition` and the first `identifier` ✅

```
(source_file
  (function_definition
    (identifier)
    (identifier)
    (identifier)))
```

however if you specify a node name inside of `(_)`, the query will instead capture every single node of that name
regardless of if its the first named node or not
`(_ . (identifier) @everyChild)`
returns: `function_definition` and **every single** `identifier` ❌

specifying the parent name does work correctly: `(function_definition . (identifier) @firstChild)`
returns: the first `identifier` ✅

https://github.com/tree-sitter/tree-sitter/blob/master/docs/section-2-using-parsers.md#anchors
https://github.com/tree-sitter/tree-sitter/blob/master/docs/section-2-using-parsers.md#wildcard-node

amaanq:
this isn't actionable without a grammar to go along with your query reproducer

RedCMD:
go to https://tree-sitter.github.io/tree-sitter/playground
select `JavaScript`
paste code `function name(one, two, three) { }`
paste query `(formal_parameters . (identifier) @firstChild)`
notice how only the first `identifier` is highlighted ✅

change `formal_parameters` to `_`
`(_ . (identifier) @firstChild)`
notice how every single `identifier` is now highlighted ❌

amaanq:
Fixed by #3440

PR #3440, opened 2024-06-27
fix(lib):  #2869 restrict pattern_map optimization

quentinLeDilavrec:
Fixing #2869 by restricting pattern_map optimization in `ts_query_new()`.

[`pattern_map`](https://github.com/tree-sitter/tree-sitter/blob/9d1ac2139221b2e19389ca620767c537607a2f63/lib/src/query.c#L287
) is used to efficiently spawn query states given the current node symbol.
As an additional optimization, [`it`](https://github.com/tree-sitter/tree-sitter/blob/9d1ac2139221b2e19389ca620767c537607a2f63/lib/src/query.c#L2743) also handles patterns with a wildcard root given that the first child is not a wildcard itself.
But for a pattern with a wildcard root, it is actually unsound to select a second step which is immediate,
because it will not be checked during the matching process [here](https://github.com/tree-sitter/tree-sitter/blob/9d1ac2139221b2e19389ca620767c537607a2f63/lib/src/query.c#L3624).

Moreover, it would be difficult to solve the unsoundness in `advance()`, and restricting the optimization should be a good enough trade-off.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -5171,3 +5171,29 @@ fn test_query_compiler_oob_access() {
     // UBSAN should not report any OOB access
     assert!(Query::new(&language, "(package_declaration _ (_) @name _)").is_ok());
 }
+
+// test for https://github.com/tree-sitter/tree-sitter/issues/2869
+#[test]
+fn test_query_first_immediate_pattern_map() {
+    let language = get_language("javascript");
+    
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
Could you add a test?

amaanq (approved):

amaanq:
nice, thank you @quentinLeDilavrec!

PR #3443, opened 2024-06-30
docs: clean up binding & parser lists

ObserverOfTime:
- Separate official and third-party bindings
- Remove links to outdated bindings
- Move parser list to the wiki

---

We can more or less copy the list from [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/wiki/List-of-parsers).


sogaiu:
For some of the PRs that requested addition (e.g. [prolog, etc.](https://github.com/tree-sitter/tree-sitter/pull/3435/files)), should these be mentioned at matrix / discord for addition to the nvim-treesitter list (those that aren't there already)?

ObserverOfTime:
If we do move forward with this then all the open PRs will be added to the wiki here.
The nvim-treesitter page will likely be removed.

clason:
Feel free to steal more info from the nvim-treesitter wiki (like the expanded links for parser writing resources).

clason:
You should then also add a note how someone would get their shiny new parser into that list (ask on Discord?)

ObserverOfTime:
> Feel free to steal more info from the nvim-treesitter wiki (like the expanded links for parser writing resources).

Those can be added to the [useful external resources](https://github.com/tree-sitter/tree-sitter/wiki/Useful-external-resources) page.
_Should we link to it somewhere in the docs?_

> You should then also add a note how someone would get their shiny new parser into that list (ask on Discord?)

Yes that will be in the wiki page.

PR #3450, opened 2024-07-05
fix(lib): an empty root node should not precede an empty range

### LINKED ISSUE #3448: Empty files do not match queries
### Problem

A query for the root node does not match if the file is empty, but it does match if there's anything in the file.

### Steps to reproduce

1. Ensure that you have `tree-sitter` and some language grammar installed. (For this example, I'm using `tree-sitter-javascript`).

2. Create a file with a query for the root node:
```bash
echo "(program)" > queries
```

3. Query an empty file:
```bash
touch empty.js
tree-sitter query queries empty.js
# output:
# empty.js
```

4. Query a file that contains a single whitespace character:
```bash
echo > whitespace.js
tree-sitter query queries whitespace.js
# output:
# newline.js
#   pattern: 0
```

### Expected behavior

I'd expect both queries to match on the root node.

Note that when parsing these files, both return the root node:
```bash
> tree-sitter parse empty.js
(program [0, 0] - [0, 0])
> tree-sitter parse newline.js
(program [1, 0] - [1, 0])
```

tree-sitter 0.22.6

### Operating system/version

macOS 14.4.1

amaanq:
The problem is, given an empty file, the root node of this file spans 0 bytes. As such, the logic for determining whether or not the node precedes the range fails, and is true when it should be false. Later on, the logic for determining whether a node intersects the range fails because of this edge case, when it shouldn't, and a state is not added to the query cursor.

By checking if the node is empty (end byte is the same as the start byte) in the logic for determining whether a node precedes a range, we can avoid this issue.

Closes #3448

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -3541,6 +3541,13 @@ static inline bool ts_query_cursor__advance(
       // Get the properties of the current node.
       TSNode node = ts_tree_cursor_current_node(&self->cursor);
       TSNode parent_node = ts_tree_cursor_parent_node(&self->cursor);
+
+      uint32_t start_byte = ts_node_start_byte(node);
+      uint32_t end_byte = ts_node_end_byte(node);
+      TSPoint start_point = ts_node_start_point(node);
+      TSPoint end_point = ts_node_end_point(node);
+      bool is_empty = start_byte == end_byte;
+
       bool parent_precedes_range = !ts_node_is_null(parent_node) && (
         ts_node_end_byte(parent_node) <= self->start_byte ||
         point_lte(ts_node_end_point(parent_node), self->start_point)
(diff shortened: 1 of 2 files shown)


lib/src/query.c
```diff
@@ -3541,8 +3542,8 @@ static inline bool ts_query_cursor__advance(
         point_gte(ts_node_start_point(parent_node), self->end_point)
       );
       bool node_precedes_range = parent_precedes_range || (
-        ts_node_end_byte(node) <= self->start_byte ||
-        point_lte(ts_node_end_point(node), self->start_point)
+        (!is_empty && ts_node_end_byte(node) <= self->start_byte) ||
+        (!is_empty && point_lte(ts_node_end_point(node), self->start_point))
```
NinoScript, line 3546: Maybe it's clearer if you factor `is_empty` out of those conditions
```suggestion
        !is_empty && (
          ts_node_end_byte(node) <= self->start_byte ||
          point_lte(ts_node_end_point(node), self->start_point))
```

amaanq: totally agree, thanks

lib/src/query.c
```diff
@@ -3541,8 +3542,10 @@ static inline bool ts_query_cursor__advance(
         point_gte(ts_node_start_point(parent_node), self->end_point)
       );
       bool node_precedes_range = parent_precedes_range || (
-        ts_node_end_byte(node) <= self->start_byte ||
-        point_lte(ts_node_end_point(node), self->start_point)
+        !is_empty && (
+          ts_node_end_byte(node) <= self->start_byte ||
+          point_lte(ts_node_end_point(node), self->start_point)
+        )
```
maxbrunsfeld, line 3548: I think this would falsely *include* the node if it is empty, but disjoint from the range. For example, say we had an empty node at 0, but were looking for matches in the range `5 - 10`. I think that instead of special-casing byte offset 0, we may want to have a special-case where the node is empty and touches the range boundary.

What about something like this?

```c
uint32_t start_byte = ts_node_start_byte(node);
uint32_t end_byte = ts_node_end_byte(node);
TSPoint start_point = ts_node_start_point(node);
TSPoint end_point = ts_node_end_point(node);
bool is_empty = start_byte == end_byte;

// ...

bool node_precedes_range =
  parent_precedes_range ||
  end_byte < self->start_byte ||
  point_lt(end_point, self->start_point) ||
  (!is_empty && end_byte == self->start_byte) ||
  (!is_empty && point_eq(end_point, self->start_point));
```

amaanq: This makes total sense, and is definitely a lot clearer too

PR #3451, opened 2024-07-05
feat(lib): add `ts_query_end_byte_for_pattern`

amaanq:
Adds a function to the api to fetch the end byte for a query pattern, as requested in https://github.com/tree-sitter/tree-sitter/discussions/3431

cc @ziontee113, since there's no issue associated w/ this PR


maxbrunsfeld:
Nice work on this @amaanq

PR #3453, opened 2024-07-06
Extend CLI with a method to generate the Swift equivalent of parser.c

daspoon:
The motivation for the change is a package (https://github.com/daspoon/tree-sitter-kit) aiming to provide seamless integration of tree-sitter in Swift applications. There a grammar is expressed as a Swift data structure and is transformed to a TSLanguage definition by a macro. Swift code generation and a callable interface are required because Swift packages don't support mixed language targets and Swift macros can't invoke subprocesses.
    
The changes (on branch swiftgen) are limited to cli/src/generate and are summarized as:
  - the code formating functionality of Generator is factored out as the RenderTarget trait with implementations RenderTargetC and RenderTargetSwift;
  - all mutation of Generator is now limited to its init method
  - the RenderContext trait is added as the interface by which RenderTargets query Generator
  - mod.rs now has a C-compatible method which takes a JSON grammar spec and returns Swift source

I am new to Rust and this is my first open source contribution, so I will be grateful for any feedback you provide. I'd also like to thank all who have made tree-sitter a reality.


ObserverOfTime:
The generated bindings support [Swift Tree Sitter](https://github.com/viktorstrate/swift-tree-sitter).
TreeSitterKit  is an alternative DSL (similar to [Rust Sitter](https://github.com/hydro-project/rust-sitter)) and as such will not be supported here.

PR #3456, opened 2024-07-07
feat: improve warning report for unnecessary seq/choice wrapping

### LINKED ISSUE #3455: Warning mismatch?  rule _ is just a `seq` or `choice` rule with a single element.
### Problem

Since 252e2a4bc09c14ef5289fad89acd2a92f02aa66a, `tree-sitter generate --no-bindings` produces the following warning for [tree-sitter-clojure](https://github.com/sogaiu/tree-sitter-clojure/commit/f4236d4da8aa92bc105d9c118746474c608e6af7):

```
Warning: rule char_lit is just a `seq` or `choice` rule with a single element. This is unnecessary.
```

See [this comment](https://github.com/tree-sitter/tree-sitter/pull/2577#discussion_r1652328901) for related discussion.

AFAIU, the warning text does not apply in tree-sitter-clojure's case, as [`char_lit`'s definition](https://github.com/sogaiu/tree-sitter-clojure/blob/f4236d4da8aa92bc105d9c118746474c608e6af7/grammar.js#L338-L339) is:

```js
    char_lit: $ =>
    CHARACTER,
```

and [`CHARACTER` is defined earlier](https://github.com/sogaiu/tree-sitter-clojure/blob/f4236d4da8aa92bc105d9c118746474c608e6af7/grammar.js#L189-L194) as:

```js
const CHARACTER =
      token(seq("\\",
                choice(OCTAL_CHAR,
                       NAMED_CHAR,
                       UNICODE,
                       ANY_CHAR)));
```

Here the `seq` takes two arguments and the `choice` takes four.  Neither take a single argument.

### Steps to reproduce

0. Ensure appropriate version of tree-sitter cli is installed.  Likely commits from (and including) 252e2a4bc through 3095fbe0 (most recent at time of writing) will work.

1. Clone and checkout appropriate tree-sitter-clojure

    ```
    git clone https://github.com/sogaiu/tree-sitter-clojure
    cd tree-sitter-clojure
    git checkout f4236d4da8aa92bc105d9c118746474c608e6af7
    ```

2. Use the `generate` subcommand

    ```
    tree-sitter generate --no-bindings
    ```

3. Observe warning:

    ```
    Warning: rule char_lit is just a `seq` or `choice` rule with a single element. This is unnecessary.
    ```

### Expected behavior

No warning or if there is something incorrect, a different warning message.

tree-sitter 0.22.6 (3095fbe07bbcc605910b7ad8bbd6313e0a07e678)

### Operating system/version

Void Linux

amaanq:
## Background

This PR doesn't exactly fix #3455, but definitely improves on the warning emitted when a rule contains an unnecessary sequence or choice wrapping a single terminal (stemming from the changes in #2577). It also does not produce the warning if this seq/choice is within a `token`, since it does not alter runtime behavior in this case (only when a terminal is "absorbed" does it actually impact behavior).

## Impact

Given the Clojure grammar, I've taken it as is and only wrapped a single string (that being `"~@"`) in `unquote_splicing_lit` with a `seq`  to demonstrate the before and after. This is because the other rule (`char_lit`) being marked as having an unnecessary seq is part of a `token` construct, and so the warning isn't necessary in this case. After the PR, we can see there being only one warning that *should* exist, and it also attaches useful information about where the problem lies, in case the rule is large or built with constants defined all over the place (thus making it harder to track down)

Output before the PR:

image

```
Warning: rule char_lit is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule unquote_splicing_lit is just a `seq` or `choice` rule with a single element. This is unnecessary.
```

After:

image

```
Warning: rule `unquote_splicing_lit` contains a `seq` or `choice` rule with a single element. Wrapping the element with this function is unnecessary.
seq(repeat(), field('marker', seq('~@')), alias($._form, 'value'), repeat(), field('value', $._form))
                              ^^^^^^^^^
```

The "reconstruction" of the JavaScript code by implementing `ToString` for `Rule` will never be 100% accurate, but a majority of the time it will be, and it'd still be useful enough to aid grammar authors in debugging even if it doesn't match the original source code 1:1, especially when they're reusing constants defined elsewhere.

### DIFF
--- a/cli/src/generate/rules.rs
+++ b/cli/src/generate/rules.rs
@@ -41,8 +41,6 @@ pub struct MetadataParams {
     pub dynamic_precedence: i32,
     pub associativity: Option<Associativity>,
     pub is_token: bool,
-    pub is_string: bool,
-    pub is_active: bool,
     pub is_main_token: bool,
     pub alias: Option<Alias>,
     pub field_name: Option<String>,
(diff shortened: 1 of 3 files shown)


clason:
I think the biggest issue for me is the phrasing "This is unnecessary", which implies that it is harmless and can be ignored, while in fact it does alter the produced parser in an unexpected way.

It would be better to replace this with a clear description of how the generated parser differs from expectation, e.g. "the `'@~'` node will be absorbed and not be visible to queries".

maxbrunsfeld:
Personally, I would rather not have warnings, at least in potentially valid cases like this.

It's not invalid or wrong to have a single-element sequence: it just gives you a different syntax tree, which you may want, for some reason (maybe for consistency with some other rule).

cli/src/generate/rules.rs
```diff
@@ -70,6 +68,98 @@ pub enum Rule {
     Seq(Vec<Rule>),
 }
 
+impl fmt::Display for Rule {
```
maxbrunsfeld, line 71: I don't really think we should we should have this code for converting a rule back into JavaScript.

IMO, we should remove the warning.

clason:
But there should be _some_ way of notifying users of this breaking change in parser generation. It's ok to make that change, but not silently -- you should come up with some sort of deprecation strategy for such breaking changes.

clason:
> It's not invalid or wrong to have a single-element sequence: it just gives you a different syntax tree, which you may want, for some reason (maybe for consistency with some other rule).

@maxbrunsfeld Maybe, but that consistency was broken by the earlier PR -- it _was_ used like that in a number of parsers, but the changed syntax tree means it no longer works. Let me give a concrete example: assume you have the grammar
```js
$.foo = choice("a",b"),
$.bar = choice("c"),
```
Before the earlier PR, you could do
```lisp
[
  "a"
  "b"
  "c"
] @quux
```
After the earlier PR, you can't: `"c"` is no longer visible -- you **must** use `(bar)` in the query, which is a breaking change, whether you warn about it or not.

maxbrunsfeld:
@amaanq, I worry the new behavior is actually **less** intuitive, since `choice("a")` behaves completely differently than `choice("b", "c")`. And the non-obvious nature of the change is forcing us to add a warning that we wouldn't otherwise need.

@clason @amaanq I don't think it's too late to simply revert that behavior change. The grammar changes that were made in order to accommodate the new behavior are still going to be valid, even with the old rule behavior, right?

I think it's super important to keep the set of rules that grammar authors need to know as simple as possible, rather than having a more complex set of rules, and adding warnings to explain them. Now, having seen the practical use cases, I'm really not sold that special casing the one-child version of `choice` and `seq` was the right move. Where are you at on this?

clason:
>  The grammar changes that were made in order to accommodate the new behavior are still going to be valid, even with the old rule behavior, right?

Yes (and possibly still welcome as inlining singleton `choice`/`seq` rules simplifies the grammar). In this respect, the current warning _is_ useful -- just like the warning about unnecessary conflicts.

(To be clear, my concern was with the combination of "soft warning" and "behaviour change" only.)

Maybe these "are you sure you want to do that? you don't need to, you know" warnings could be gated behind a `--verbose` or `--warn` or `--lint` switch? (As writing grammars is such an arcane art, having such -- and more -- builtin hints seems desirable to me.)

clason:
@amaanq since a release seems to be coming up, can we
1. keep the (improved) warning;
2. revert the (silent) behavior change

at least for 0.23?

amaanq:
After some thinking, I think the behavior *should* be reverted, but the warning can stay. It's just too intrusive, although imo, from a querying perspective, having a rule `foo: $ => 'foo'` and `foo: $ => seq('foo')` behave differently is *very* unintuitive. I'll put up a PR to revert shortly

PR #3483, opened 2024-07-23
docs: add Hare grammar

esawady:
there're a couple tree-sitter-hares floating around, this is the one [endorsed by upstream](https://harelang.org/documentation/install/editors.html#tree-sitter)


amaanq:

PR #3490, opened 2024-07-27
script/build-wasm: support big endian

juru1234:
With this fix, I'm able to use `bash-language-server` with `neovim` on a big endian s390x Linux machine.  `bash-language-server` uses `web-tree-sitter`.


juru1234:
Sorry, will open it again.

PR #3492, opened 2024-07-28
fix: support big endian machines

juru1234:
With this fix, I'm able to use `bash-language-server` with `neovim` on a big endian s390x Linux machine. `bash-language-server` uses `web-tree-sitter`.

### DIFF
--- a/script/build-wasm
+++ b/script/build-wasm
@@ -119,6 +119,7 @@ $emcc                                            \
   -s WASM=1                                      \
   -s INITIAL_MEMORY=33554432                     \
   -s ALLOW_MEMORY_GROWTH=1                       \
+  -s SUPPORT_BIG_ENDIAN=1                        \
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
Can you check if this has a performance impact on LE machines?

juru1234:
@ObserverOfTime I think the use of DataView's was heavily improves so this shouldn't have a performance impact: https://v8.dev/blog/dataview

ObserverOfTime:
Looks like we need to update emscripten to 3.1.58+ for this to work (https://github.com/emscripten-core/emscripten/issues/21751).

juru1234:
@ObserverOfTime Oh yes, I tested with a local emscripten.... Should I push again and update it to exact this version? Or a newer one?

piero1100:
gh pr checkout 1

PR #3495, opened 2024-07-29
Reset language when resetting wasm store

CGamesPlay:
This is necessary, otherwise [ts_wasm_store_start](https://github.com/tree-sitter/tree-sitter/blob/2512f3ab179867de282b89221a62c55ff6feffa3/lib/src/parser.c#L1909) will never have been called.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -2144,11 +2144,19 @@ TSTree *ts_parser_parse_string_encoding(
 }
 
 void ts_parser_set_wasm_store(TSParser *self, TSWasmStore *store) {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Thanks for catching this. Could you add a test to [this file](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/tests/wasm_language_test.rs) that sets the wasm store after the language? I feel like it should be possible to leave the language set (but using the new store), but maybe I'm missing something.

CGamesPlay:
Done for both.

PR #3501, opened 2024-07-29
Added `--overview-only` to `test` subcommand

buckynbrocko:
`tree-sitter test --overview-only`

Adds the option to only show the pass-fail tree - something I frequently find myself wanting when I'm only concerned with the pass-fail status and don't want to scroll back up through a sea of color-diffs.


amaanq:
I think this is a great change - do you mind rebasing?

elliotfontaine:
Thanks, that's so great !

PR #3502, opened 2024-07-30
feat(cli): Add `--show-fields` flag to cli `test` subcommand

buckynbrocko:
eg `tree-sitter test --show-fields`

Forces the test diff to output fields, regardless of their lack of presence in the test file's expected output section.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -252,7 +252,7 @@ struct Test {
     pub open_log: bool,
     #[arg(long, help = "The path to an alternative config.json file")]
     pub config_path: Option<PathBuf>,
-    #[arg(long, help = "force showing fields in test diffs")]
+    #[arg(long, help = "Force showing fields in test diffs")]
     pub show_fields: bool,
 }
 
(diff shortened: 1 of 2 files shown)


cli/src/test.rs
```diff
@@ -388,7 +389,9 @@ fn run_tests(
                     }
                 } else {
                     let mut actual = tree.root_node().to_sexp();
-                    if !has_fields {
+                    let show_fields = opts.show_fields || has_fields;
+
+                    if !show_fields {
```
ObserverOfTime, line 394:
```suggestion
                    if !(opts.show_fields || has_fields) {
```

piero1100: > 
```suggestion
                    let show_fields = opts.show_fields || has_fields;

                    if !show_fields _{_better thinks divide fields#
```
Examp 🥇 former framer

cli/src/main.rs
```diff
@@ -252,6 +252,8 @@ struct Test {
     pub open_log: bool,
     #[arg(long, help = "The path to an alternative config.json file")]
     pub config_path: Option<PathBuf>,
+    #[arg(long, help = "force showing fields in test diffs")]
```
ObserverOfTime, line 255:
```suggestion
    #[arg(long, help = "Force showing fields in test diffs")]
```
    
```suggestion
    #[arg(long, help = "Force showing fields in test diffs")]
```

Issue #3503, opened 2024-07-30
collecting match results into a Vec in Rust returns incorrect nodes


pr0me:
### Problem

Collecting the results of a match query into a `Vec<QueryMatches>` returns an unexpected (and incorrect) tree node:
```rust
let matches: Vec<tree_sitter::QueryMatch> = query_cursor.matches(&ts_func_query, root_node, CODE.as_bytes()).collect();
```

### Steps to reproduce

You can run the following minimal example:
```rust
use tree_sitter::{Query, QueryCursor};


fn main() {
    let mut ts_parser = tree_sitter::Parser::new();
    ts_parser
        .set_language(&tree_sitter_c::language())
        .expect("Error loading C grammar");

    let tree = ts_parser
        .parse(&CODE, None)
        .unwrap();

    let raw_query = format!("(function_declarator declarator: (identifier) @name (#eq? @name \"tzparse\")) @func");
    let ts_func_query =
        Query::new(
        &tree_sitter_c::language(),
        &raw_query)
        .unwrap();

    let root_node = tree.root_node();
    let mut query_cursor = QueryCursor::new();
    let matches: Vec<tree_sitter::QueryMatch> = query_cursor.matches(&ts_func_query, root_node, CODE.as_bytes()).collect();
    
    // let mut query_cursor = QueryCursor::new();
    let mut matches_x = query_cursor.matches(&ts_func_query, root_node, CODE.as_bytes());

    println!("{:?}", matches);
    println!("{:?}", matches_x.next());
}

static CODE: &str = r#"
static int
tzparse(
  const char    *       name,
  struct state  * const sp,
  const int             lastditch
  )
{
  return 0;
}

void
tzset(void)
{
  settzname();
}"#;
```

### Expected behavior

expected output:
```
Some(QueryMatch { id: 0, pattern_index: 0, captures: [QueryCapture { node: {Node function_declarator (2, 0) - (6, 3)}, index: 1 }, QueryCapture { node: {Node identifier (2, 0) - (2, 7)}, index: 0 }] })
[QueryMatch { id: 0, pattern_index: 0, captures: [QueryCapture { node: {Node function_declarator (2, 0) - (6, 3)}, index: 1 }, QueryCapture { node: {Node identifier (2, 0) - (2, 7)}, index: 0 }] }]
```

i.e., match the `tzparse` function_declaration node.

Actual output:
```
[...]

Issue #2847, opened 2024-01-05
probably UB: `QueryMatches::next` in Rust library clobbers previously returned values.


umanwizard:
Consider `test.c`:

```
void
do_stuff()
{
  int x = 42;
  char const *y = "Hello, world!";
  float z = 3.14159;
}
```

and `main.rs`:

```
use tree_sitter::{Parser, Query, QueryCursor};

fn main() {
    let mut parser = Parser::new();
    let l = tree_sitter_c::language();
    parser.set_language(l).unwrap();

    let contents = std::fs::read("/home/brennan/test.c").unwrap();
    let tree = parser.parse(&contents, None).unwrap();
    let q = Query::new(l, "(declaration type: (primitive_type) @t)").unwrap();
    let mut qc = QueryCursor::new();

    let matches: Vec<_> = qc
        .matches(&q, tree.root_node(), contents.as_slice())
        .collect();
    for m in matches {
        println!(
            "{}",
            m.captures[0].node.utf8_text(contents.as_slice()).unwrap()
        );
    }
}
```

This prints:

```
float
float
float
```

rather than the expected

```
int
char
float
```

The issue is that each call to `QueryMatches::next` constructs a `QueryMatch` object that points to data returned by `ts_query_cursor_next_match`. However, this data references buffers that become invalidated on each new call to `ts_query_cursor_next_match`. So it's in general not okay to keep the old `QueryMatch` objects alive when constructing new ones.

ObserverOfTime:
Closing as a duplicate of #2843.

Issue #2843, opened 2023-12-31
Discrepancy between query with Rust bindings and CLI tool


WillLillis:
Hi! I'm having some trouble using the [tree-sitter-asm](https://github.com/rush-rs/tree-sitter-asm/) grammar (the latest commit on the main branch) with the tree-sitter Rust bindings. This is the query I'm using:
```
(label (ident (reg (word) @label)))
(meta kind: (meta_ident)
(
        (int)
        ("," (int))*
    )
) @data
```

The results from the CLI tool and Rust bindings match when used on the following input file:

```asm
data_items:
    .long 1, 2, 3
```

The playground for this file looks like:

![image](https://github.com/tree-sitter/tree-sitter/assets/87077023/423d6a08-6db0-437d-9079-30f06dc4b0b8)

The tree looks like this:

![image](https://github.com/tree-sitter/tree-sitter/assets/87077023/a0a25b0c-1a5e-47d0-a2a1-ee63a49fb7f9)

Using the Rust bindings with the above query, I get these matches in response:

```
Matches: [
QueryMatch { id: 0, pattern_index: 0, captures: [QueryCapture { node: {Node meta (1, 4) - (1, 17)}, index: 1 }] }, 
QueryMatch { id: 1, pattern_index: 1, captures: [QueryCapture { node: {Node meta (1, 4) - (1, 17)}, index: 1 }] }
]
```

The CLI tool produces this output:

```
working.asm
  pattern: 0
    capture: 0 - label, start: (0, 0), end: (0, 10), text: `data_items`
  pattern: 1
    capture: 1 - data, start: (1, 4), end: (1, 17), text: `.long 1, 2, 3`
```

However the CLI tool and Rust bindings return different results on the next (very similar) input: 

```asm
data_items:
    .long 1, 2, 3

main:
    movq %rbp, %rsi
```

The playground for this file looks like this:

![image](https://github.com/tree-sitter/tree-sitter/assets/87077023/974e0849-a53c-4ea5-bd3f-153e5476ed37)

The tree looks like this:

![image](https://github.com/tree-sitter/tree-sitter/assets/87077023/ebac7b6a-941e-4b5b-a35a-63b085d01093)

Currently with the rust bindings,  using the above query and printing out the result from calling ``.matches()`` yields:

```
Matches: [
QueryMatch { id: 0, pattern_index: 0, captures: [QueryCapture { node: {Node word (3, 0) - (3, 4)}, index: 0 }] }, 
[...]

Issue #2265, opened 2023-05-16
`tree_sitter::QueryMatches` mutates `captures` field during iteration


Jakobeha:
Related to https://github.com/tree-sitter/tree-sitter/issues/1656 and https://github.com/tree-sitter/tree-sitter/issues/608

If you `collect` a `QueryMatches`, then the earlier match's captures will be overwritten by later ones. The issue seems to be that `QueryMatches` is actually a streaming iterator, and once you call `next` the old `QueryMatches` should no longer be used.

I think this should be addressed, it violates Rust's safety guarantees since the match captures are behind an immutable reference. Also, GATs are now stable and that seemed to be a blocker.

Some possibilities:

- Make QueryMatches implement [https://docs.rs/streaming-iterator/latest/streaming_iterator/] or a similar trait, which will maintain some of the portability and convenience of being an iterator.
- Make `QueryMatches` own its captures (`Vec<QueryCapture<'tree>>`). This means we get to keep `QueryMatches` an iterator, at the cost of a slight performance hit
- Do both: create a new trait `StreamingQueryMatches` which is a streaming iterator, and make `QueryMatches` own its captures

hadronized:
[kak-tree-sitter](https://github.com/hadronized/kak-tree-sitter) author here. I have faced that exact problem lately while adding support for advanced semantics selections modification. It took me a whole day to eventually enter a discussion with @the-mikedavis (from Helix) to realize that collecting iterators is unsound with `tree-sitter`.

I’m joining @Jakobeha on this one.

ObserverOfTime:
Closing as duplicate of #608. (PR very welcome.)

PR #3504, opened 2024-07-30
StreamingIterator impl for QueryMatches in Rust Bindings

pr0me:
This PR addresses issues with `QueryMatches`' existing iterator.
**Problem**: Using `.next()` and related methods such as `.collect()` would potentially yield incorrect results / nodes.

**Root cause**: Each call to .next() on the iterator updates the state of underlying object in the C library.
This has the problem that if you collect a `QueryMatches`, the earlier match's captures will be overwritten by later ones, leading to Undefined Behavior.

This problem was discussed in the following issues:
https://github.com/tree-sitter/tree-sitter/issues/608
https://github.com/tree-sitter/tree-sitter/issues/2265
https://github.com/tree-sitter/tree-sitter/issues/2843
https://github.com/tree-sitter/tree-sitter/issues/2847
https://github.com/tree-sitter/tree-sitter/issues/3503

One repeatedly proposed solution to this problem was the use of a streaming iterator, correctly yielding references to `QueryMatch`es instead of copies with undefined internal states.

The old, unsafe implementation of `Iterator` for `QueryMatches` was removed.

The new `StreamingIterator` can, e.g., be used as follows:
```rust
use tree_sitter::{Node, QueryMatch};
use streaming_iterator::StreamingIterator;

let mut query_matches = query_cursor.matches(&ts_func_query, root_node, code.as_bytes());
let mut matches: Vec<Node> = Vec::new();
while let Some(m) = query_matches.next() {
    matches.push(m.captures[0].node);
    println!("{:?}", m);
}
```
Things to note:
- `for` loops do not work with streaming iterators afaik
- we cannot simply collect all `QueryMatch`es into a vector as that would require `Clone`. Instead, we can collect those parts of a Match we are actually interested in


pr0me:
fixing tests

pr0me:
fixed issues and ran tests locally.
except for these three tests, which I'm not sure correlate to my changes, everything completed successfully:
```
failures:
    tests::corpus_test::test_feature_corpus_files
    tests::node_test::test_node_is_named_but_aliased_as_anonymous
    tests::parser_test::test_parsing_after_editing_tree_that_depends_on_column_values
```

The biggest changes required were due to `StreamingIterator` having no `collect` method.
We could implement it with an extended trait but it would require to introduce `Clone` to `QueryMatch` which we might not want.  
For now, I'd advise against it and simply use the slightly more verbose syntax to work with the `StreamingIterator`.

pr0me:
fixed wasm tests and merged recent commits from master:

> `cargo test`
> `cargo test --features=wasm`
> `script/test`
> `script/test-wasm`

all finish successfully locally.

pr0me:
bump

merged most recent commits and successfully ran tests locally again

pr0me:
failing check is a lint in a file my commits did not touch.

how is this handled? should I just resolve this on my branch? how are other PRs getting merged with this being present? maybe I'm missing something, sorry

pr0me:
merged with upstream and fixed linting errors.
successfully reran all tests locally + `cargo clippy --workspace --all-targets -- -D warnings`

@amaanq if you have any opinions, suggestions or requests on this PR I'd appreciate it :)
I just think this is a pretty serious issue that was prevalent for way too long and would love to see it fixed.

pr0me:
bump 
@ObserverOfTime maybe you can help.
any feedback on this, anything you need from me?
would love to see this merged soon.

thanks

amaanq:
Yeah I know I hadn't said anything in a while, but I refrained from touching this because it wasn't a trivial merge - there was a lot of stuff missing, e.g. `StreamingIterator` wasn't even implemented on `QueryCaptures`, and implementing it there brought in a whole slew of issues elsewhere in the highlight crate and others (we **really** need to decouple the highlight and tags crates).

pr0me:
agreed, not trivial.
this was my first PR to this project so I expected that there were code sites needing modifications that I would've had missed, that's why I asked for feedback, would've been happy to help :)

anyways, thanks for your work on this and the merge!

sminez:
I've just been trying to work with this for implementing parsing and highlighting in a text editor I'm writing and it caught me out that I need to add another dependency in order to be able to make use of the return type from `QueryCursor::matches`. Is it possible to re-export the `StreamingIterator` trait from this crate at all? Or at the very least, document that you need to add the dependency for the return type of `matches` to be at all useful.

woodruffw:
Yeah, documenting this somewhere would be really helpful -- I ran into the same issue as @sminez, where a routine bump of tree-sitter now requires a third-party trait that isn't re-exported.

maxbrunsfeld:
We should re-export it I guess. Or make the methods also available as inherent methods, in addition to the trait impl.

Issue #3479, opened 2024-07-21
`test --update` doesn't work properly with test attributes


withamoliver:
### Problem

Multiple issues with test attributes when using `tree-sitter test --update`:

1. If you use `:skip` then the test is always deleted from the test file.
2. If you use `:error` then the test is always deleted from the test file.
3. If the `:platform` doesn't match then the test is deleted from the test file.
4. Test attributes are always deleted from the test file.

There are multiple paths through `run_tests` which return without pushing to the `corrected_entries`, e.g. for `:skip`:
https://github.com/tree-sitter/tree-sitter/blob/800f2c41d0e35e4383172d7a67a16f3933b86039/cli/src/test.rs#L343-L350

The test attributes look to be getting deleted because they are never included in the `corrected_entries` which gets written back to the test file.

### Steps to reproduce

`:skip` bug (similar for `:platform(x)` with a platform `x` which doesn't match the current platform):
1. Add `:skip` to a test.
2. Run `tree-sitter test -u`.
3. The test is deleted from the test file.

`:error` bug:
1. Add `:error` to a test.
2. Run `tree-sitter test -u`.
3. The test is deleted from the test file.

Test attributes deleted bug:
1. Add `:platform(x)` to a test where `x` matches the current platform.
2. Run `tree-sitter test -u`.
3. The `:platform(x)` attribute is deleted from the test file.

### Expected behavior

For `tree-sitter test --update` to be useful with test attributes, it should behave similarly to if a test attribute wasn't specified.

When a test is skipped due to `:skip` or a mismatched `:platform` then the test should be written back to the file with the same contents as before.

For `:error` it might make sense to write back the same content with an omitted parse tree after the `---` even if the original file contained one, as per the documentation. 

Test attributes shouldn't be deleted when writing back to the file.

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.22.6 (800f2c41d0e35e4383172d7a67a16f3933b86039)

### Operating system/version

macOS 14.4.1

withamoliver:
The same thing happens when a test is skipped due to not matching a `--filter <FILTER>`, `--include <INCLUDE>` or when matching an `--exclude <EXCLUDE>`.

cathaysia:
As a supplement. test -u will also cause ||| to be lost. that is :

````
=================|||
test name
=================|||
```
will become:

```
==================
test name
==================
```

Issue #3238, opened 2024-04-01
tree-sitter-cli@0.22.2 breaks tree-sitter-css tests


jeandudey:
### Problem

The tests in tree-sitter-css fails with tree-sitter-cli 0.22.2. Also reported on: https://github.com/tree-sitter/tree-sitter-css/issues/47.

After playing a bit with the test file I've found that the problem might not be the grammar but tree-sitter-cli itself as when removing the colons from the test name it works as intended.

For example, changing from this:

https://github.com/tree-sitter/tree-sitter-css/blob/02b4ee757654b7d54fe35352fd8e53a8a4385d42/test/corpus/selectors.txt#L104-L106

To this:

```
=========================
slotted pseudo element
=========================
```

Works without problems.

### Steps to reproduce

```
git clone https://github.com/tree-sitter/tree-sitter-css --depth 1
cd tree-sitter-css; tree-sitter generate; tree-sitter test
```

### Expected behavior

Tests should run without problem.

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.22.2

### Operating system/version

GNU Guix 4d79a9cd6b5f0d8c5afbab0c6b70ae42740d5470

sogaiu:
May be [this](https://github.com/tree-sitter/tree-sitter/pull/3060/commits/51431c53c57e4ae3789a8a071df2c44c8895c3c1) is related?

ObserverOfTime:
FWWI the [test](https://github.com/tree-sitter-grammars/tree-sitter-test) parser works (albeit more strict with suffixes). Maybe it's time to use it internally. ;)

ObserverOfTime:
There won't be any cyclic dependencies when #3069 is merged since parsers will only depend on `tree-sitter-language`. `tree-sitter-cli` will depend on the (not yet published) `tree-sitter-test` crate which will include the generated files, even when they're removed from version control in the future. If Nix/Guix has its own special way of building `tree-sitter-cli` then it may also need its own special solution.

jeandudey:
It'd still be a cyclic dependency for Guix since we remove generated files from grammar repositories, i.e.: generated bindings, `grammar.json`, `parser.c`, etc. So to build `tree-sitter-test` one needs an existing `tree-sitter-cli` to generate those files to then build the newer `tree-sitter-cli`.

ObserverOfTime:
[...]

PR #3513, opened 2024-08-04
refactor(cli): using tree-sitter-test parse copus test

cathaysia:
This is a pre PR. Due to the large size of the changes, I am not sure if this PR will be accepted.
The purpose of this PR is to fix: #3479 

If this type of PR is acceptable, I will continue to work.


ObserverOfTime:
See https://github.com/tree-sitter/tree-sitter/issues/3238#issuecomment-2029932388

cathaysia:
I see Guix mentioned a lot there. I don't quite understand why Guix is ​​relevant to this issue. Theoretically Guix should just need to be built with cargo. Why is there a circular dependency?

cathaysia:
It would be nice if tree-sitter-test could also be merged into this repository, so there would be no need to maintain two sets of code.

Last week I built a tool similar to tree-sitter test at my work using tree-sitter-test. The difference from the current test implementation logic is that the current test implementation seems to rewrite the entire corpus test file. With tree-sitter-test, we can just replace the output node, and the implementation is quite simple, with less than fifty lines of code. Maybe the we can also do this in this PR.

clason:
> It would be nice if tree-sitter-test could also be merged into this repository, so there would be no need to maintain two sets of code.

Quite the opposite; there's already too many loosely related pieces of code in this repository. IMO the right approach would be to separate them out further and set up (and test!) explicit integrations properly.

piero1100:
> This is a pre PR. Due to the large size of the changes, I am not sure if this PR will be accepted. The purpose of this PR is to fix #3479
> 
> If this type of PR is acceptable, I will continue to work.
> 
> I copied the entire tree-sitter-test over here because I ran into a tree-sitter dependency issue. I don't know how to fix it yet.
> 
> block by:
> 
> * [bug: cannot parse `Code with dashes` tree-sitter-grammars/tree-sitter-test#3](https://github.com/tree-sitter-grammars/tree-sitter-test/issues/3)
> * [bug: cannot parse title with newline tree-sitter-grammars/tree-sitter-test#4](https://github.com/tree-sitter-grammars/tree-sitter-test/issues/4)
> * [bug: cannot parse title with suffix tree-sitter-grammars/tree-sitter-test#5](https://github.com/tree-sitter-grammars/tree-sitter-test/issues/5)

cathaysia:
close as no plan

PR #3541, opened 2024-08-21
build(lib): build using cmake

ObserverOfTime:
One more attempt, now that wasmtime requires cmake.

More things to consider if this is acceptable:

- [ ] Add a few more lines (and an extra file?) for [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) to work
- [ ] Move the dev targets to cmake and remove the makefile
- [ ] Use cmake in `build.rs` (with or without the [crate](https://crates.io/crates/cmake))
- [ ] Do the same for the parser bindings


clason:
I don't think that's a net positive. Note that the issue is not the transitive reliance on the `cmake` crate but the need to maintain the cmake code -- especially fending off constant PRs trying to add support for niche cases or making use of supposedly more convenient recently introduced features.

(Also, it's not a great sign that each of the four -- so far -- PRs adding CMake support has essentially been completely different... And for reference, the [CMakeLists.txt](https://github.com/neovim/neovim/blob/master/cmake.deps/cmake/TreesitterCMakeLists.txt)  that Neovim uses to build the lib is less than a dozen lines.)

ObserverOfTime:
> And for reference, the [CMakeLists.txt](https://github.com/neovim/neovim/blob/master/cmake.deps/cmake/TreesitterCMakeLists.txt) that Neovim uses to build the lib is less than a dozen loc. (That's without wasmtime, granted, but that only adds a couple loc on top.))

Neovim also does not (need to) set the metadata, soname, compile flags, nor does it create the pkgconfig file.

clason:
Well, yes; that's kinda my point: I'm missing a clear problem statement and a delineated scope for this.

ObserverOfTime:
It's necessary for Windows users and still easier to maintain than the current Makefile.

dundargoc:
> (Also, it's not a great sign that each of the four -- so far -- PRs adding CMake support has essentially been completely different... And for reference, the [CMakeLists.txt](https://github.com/neovim/neovim/blob/master/cmake.deps/cmake/TreesitterCMakeLists.txt?rgh-link-date=2024-08-21T09%3A18%3A53Z) that Neovim uses to build the lib is less than a dozen loc. (That's without wasmtime, granted, but that only adds a couple loc on top.))

To be fair the other cmake PRs were kinda wack and added way to much nonsense. I'd argue this implementation is pretty bare bones. I think we can shave off a few lines here and there but I don't see any glaring/inherent problems with the implementation on a first glance.

I read the comments but I feel like the reasoning for this PR is somewhat conflicting and was wondering if someone could clarify. The PR description mentions wasmtime requires cmake but the comments are instead discussing the benefits of make vs cmake and I'm trouble connecting the dots between these two discussions. If wasmtime *requires* downstreams consumers to use cmake, and treesitter wants to use wasmtime, then surely treesitter *has* to use cmake?

clason:
> If wasmtime requires downstreams consumers to use cmake, and treesitter wants to use wasmtime, then surely treesitter has to use cmake?

Not quite; there's a difference between _depending_ on cmake (transitively, via wasmtime-c-api) and _using_ cmake for your own build. In brief, the situation is

* `wasmtime-c-api` uses `cmake` for their own configuration; that means any Rust project including the  `wasmtime-c-api` crate also has to pull in the `cmake` crate as a dependency; other than that there is _zero_ effect on tree-sitter, in particular the build scripts (as evidenced that bumping wasmtime to 24.0.0 needed no build changes).
* The PR here proposes replacing the `Makefile` used for building the `tree-sitter` lib with a CMake script (or just adding the latter) -- this is completely orthogonal to whether wasmtime-c-api is configured via CMake or not, so has to stand on its own merits (which I personally don't see, but obviously I'm missing something).

dundargoc:
IMO the cmake file looks good and it's a positive for downstream consumers who also use cmake. Don't see a problem with it.

dundargoc (approved):

amaanq:
I'm okay with adding it (@maxbrunsfeld?)

amaanq (approved):

lewis6991 (approved):

lib/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(tree-sitter
+        VERSION "0.22.6"
```
clason, line 4:
```suggestion
        VERSION "0.23.0"
```
(unless you have pre-release versioning -- which you should consider)

lib/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+  endif(NOT DEFINED CACHE{WASMTIME_LIBRARY})
+
+  target_compile_definitions(tree-sitter PUBLIC TREE_SITTER_FEATURE_WASM)
+  target_include_directories(tree-sitter SYSTEM PRIVATE "${WASMTIME_INCLUDE_DIR}")
+  target_link_libraries(tree-sitter PRIVATE "${WASMTIME_LIBRARY}")
+endif(TREE_SITTER_FEATURE_WASM)
+
+set_target_properties(tree-sitter
+                      PROPERTIES
+                      C_STANDARD 11
```
clason, line 46: For the record, this is only necessary if you enable wasmtime (would affect Neovim if we dropped our own file @dundargoc)

dundargoc: I believe that is correct (IIRC not using c11 for treesitter throws warnings which we silence while not using c11 for wasmtime throws errors). I personally think setting c11 to be minimum version is fine tbh, but we could conditionally change C_STANDARD if necessary.

Issue #3454, opened 2024-07-07
Unable to share TSWasmEngine between multiple TSWasmStore


CGamesPlay:
### Problem

While investigating adding wasm support to `py-tree-sitter`, I noticed that it isn't possible to reuse TSWasmStore objects. The goal is to create a WASM store, load a bunch of languages into it, and then parse a bunch of files in those languages.

- **Attempt 1**: Create a store, load all languages into it, create a parser for each language, share the store between all parsers.
  **Problem 1**: The parser assumes ownership over the passed store and frees it in [ts_parser_set_wasm_store](https://github.com/tree-sitter/tree-sitter/blob/3095fbe07bbcc605910b7ad8bbd6313e0a07e678/lib/src/parser.c#L2146) (called from `ts_parser_delete`).
- **Attempt 2**: Create a store, load all languages into it, create a parser for each language, using a new store for each parser (inspired by [this comment](https://github.com/tree-sitter/tree-sitter/blob/3095fbe07bbcc605910b7ad8bbd6313e0a07e678/lib/include/tree_sitter/api.h#L1198)).
  **Problem 2**: The store assumes ownership over its engine and frees it in [ts_wasm_store_delete](https://github.com/tree-sitter/tree-sitter/blob/3095fbe07bbcc605910b7ad8bbd6313e0a07e678/lib/src/wasm_store.c#L921).
- **Attempt 3**: Same as before, but using a separate engine for each parser.
  **Problem 3:** It is [not possible](https://github.com/bytecodealliance/wasmtime/blob/a068bfe06435a048502bf4afc426451dab5294af/crates/wasmtime/src/runtime/instance.rs#L255) to use a type from another engine in a store .
- **Attempt 4**: For each parser, create a separate engine, dedicated store, load all relevant languages into the store.
  **Problem 4**: Non-ergonomic and inefficient.

Now, I don't know the point of sharing a store between languages. I assume it's because embedded languages need to be in the same store, but I'm just speculating. I can certainly see value in have a different store for each parser, to control memory usage. Based on this, I'm going to suggest that the ideal architecture allows parsers, languages, and stores to share the underlying resources with reference counting (this is already how things are implemented [internally in wasmtime](https://github.com/bytecodealliance/wasmtime/blob/a068bfe06435a048502bf4afc426451dab5294af/crates/c-api/include/doc-wasm.h#L20)).

1. In wasmtime, create a C API to expose cloning engine references.
[...]

PR #3542, opened 2024-08-22
clone wasm store engine

CGamesPlay:
This resolves https://github.com/tree-sitter/tree-sitter/issues/3454.

This brings the usage of wasmtime::Engine in line with how wasmtime intends it to be used. All wasmtime functions that receive an Engine always receive an `&Engine`, never an owned `Engine`.  They are always responsible for cloning the reference if they need it.

This brings the usage of wasmtime::Engine in line with how TSParser treats TSLanguages: when setting a language to the parser, the parser is responsible for cloning the reference to the TSLanguage. It is counterintuitive for TSParser to have different behavior when receiving wasmtime_engine_t.

C API users also expect this behavior, see "Memory Management" [here](https://docs.wasmtime.dev/c-api/wasm_8h.html). Talking about the C API: without this change, failing to clone the `wasmtime_engine_t` (which, again, is never something API users need to do in wasmtime) and then reusing the engine in multiple TSLanguages results in a use after free. With this change, failing to call `wasm_engine_delete` on your owned Engine results in a memory leak. Memory leaks are safer than use-after-free.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -1198,7 +1198,7 @@ impl Loader {
     }
 
     #[cfg(feature = "wasm")]
-    pub fn use_wasm(&mut self, engine: tree_sitter::wasmtime::Engine) {
+    pub fn use_wasm(&mut self, engine: &tree_sitter::wasmtime::Engine) {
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
Thank you!

PR #3544, opened 2024-08-24
fix(bindings): update go bindings

amaanq:
Now with the advent of https://github.com/tree-sitter/go-tree-sitter Go users will have a much easier time using tree-sitter :slightly_smiling_face:


ObserverOfTime:
You forgot to update the bindings list. :)

PR #3548, opened 2024-08-24
fix!: revert interning of a sequence or choice of a single rule

amaanq:
This reverts #2577, as mentioned in #3456, it's too intrusive, albeit it corrects "UB" when it comes to querying these nodes.


clason:
For the record, the change makes sense for that reason; just the way it was introduced was unfortunate (not marked as breaking; misleading "warning"). A fresh PR (after the release) would not have the same issues.

PR #3559, opened 2024-08-29
feat: add an API to time out query executions

### LINKED ISSUE #1957: Support timeout for query.
Some large files may be parsed fine but it will takes too long to query. It may be hard to optimize the query logic to accelerate the query? But a timeout option can allow us to simply ignore those strange files.

The problematic file can be find here https://github.com/google/google-ctf/blob/master/third_party/edk2/MdeModulePkg/Library/BrotliCustomDecompressLib/common/dictionary.c

amaanq:
Closes #1957

## Problem

Currently, if a predicate is unlikely to match on the Rust side, a sizable query against a very large file can take forever, and ends up hanging, even if many matches were returned from the C side before. This is because if the result doesn't satisfy the predicates [here](https://github.com/tree-sitter/tree-sitter/blob/92330caa080d4cd75c618c485bdb1cf3a2fb4976/lib/binding_rust/lib.rs#L2661), the loop continues without ever having a chance to exit until a match is returned *with* the predicate evaluating to true.

## Solution

This PR adds a pair of API functions, `ts_query_cursor_set_timeout_micros` and `ts_query_cursor_timeout_micros`, to limit how long query execution is allowed to take, thereby negating the chance of a hang to occur. The implementation is very similar to that of the parser timeout API, in that we set the `end_clock` of a `QueryCursor` to the time of `clock_now()` + `self->timeout_duration` in `ts_query_cursor_exec`. Then, in `ts_query_cursor__advance`, an additional condition is added to determine whether or not to return, that being if the `end_clock` isn't 0 and the time of `clock_now()` is greater than it. We do not always perform this check as it can be expensive in a hot code path, so instead we do this every 100 iterations, like in the parser version.

The test in the Rust side demonstrates the timeout working by using a JS source file of a simple function definition repeated 1000x, and we are querying for this function definition. Ordinarily, we'd expect 1000 matches to return. However, if we set a timeout of 1000us, or 1ms, we will get less than 1000 matches, as the timeout will be reached before execution is finished. Afterwards, we set the timeout micros to 0 and expect to get all 1000 matches, as passing in zero does not halt execution after a certain amount of elapsed time.


clason:
The `_micros` is probably unnecessary unless you plan on adding a `_nanos` version later ;)

amaanq:
> The `_micros` is probably unnecessary unless you plan on adding a `_nanos` version later ;)

It's for parity with the parser functions' names :sweat_smile:

sogaiu:
Symmetry!

maxbrunsfeld (approved):
Great work. Looks really nice.

PR #3566, opened 2024-08-30
fix(lib): peek at the next sibling when iterating to find the child that contains a given descendant

### LINKED ISSUE #1378: Empty node .parent points to immediately preceding non-empty node
As far as I can tell, the logic of .parent doesn't work for empty nodes because of the way it determines whether a node contains another node.  From looking at the [code](https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/node.c#L478) it seems to do it by checking whether an iterator that has just yielded a potential parent, is now at or beyond the end byte of the node we want to get the parent of.

The issue is that with empty nodes, this check will be true of many nodes going down the tree.

For example, with the following HTML:

```html
<script></script>
```

We have a `script_element` with three children: a `start_tag`, an empty `raw_text`, and an `end_tag`.

When searching for the parent of the `raw_text`, we start at the root node and keep descending into its children while the current child contains the passed node.  This check returns true for the `script_element`, but also the `start_tag` and the `>` -- their end bytes are all at or beyond the end byte of the empty `raw_text`.  This is intuitively true as it makes sense to say that any given node "contains" a zero-length string directly after itself.

My current solution is to skip empty nodes while traversing the tree, but since this incurs a bit of a performance hit I thought it might be nice to have some way of avoiding the issue within the tree-sitter code.  I don't know how feasible/possible it would be to fix the issue without actually maintaining pointers internally, but I think there are a couple of other solutions that could work:

- Have tree-sitter `throw` when trying to get the parent of an empty node, as an unsupported feature, then:
- Add `nextNonEmptySibling`, `firstNonEmptyChild`, `lastNonEmptyChild`, `nonEmptyChildren` pointers
- Add an option for `nextSibling` etc. to automatically skip empty nodes
- Add an option to completely omit empty nodes from the generated tree

What do you think?  Is there any reason not to just go for the last option of omitting empty nodes altogether, and moving the burden of checking for node existence onto user code?

### LINKED ISSUE #1647: ts_node_parent() returns incorrect parent when zero-width child has a previous sibling
In this case, we're parsing this incomplete Python code `if :` with this sexp:
```
(module (if_statement condition: (MISSING identifier) consequence: (block)))
```

Here's the test script:

```
from tree_sitter import Language, Parser

PY_LANGUAGE = Language('build/python.so', 'python')

parser = Parser()
parser.set_language(PY_LANGUAGE)

tree = parser.parse(b'if :')

def walk(node):
    yield node
    for child in node.children:
        yield from walk(child)

for node in walk(tree.root_node.children[0]):
    print('node', node)
    parent = node.parent
    if parent:
        print('parent', parent)
    print()
```

Here's the relevant output. `Node id=0x600002cdc660 kind=identifier` is the expression, which is zero-width because the code is incomplete. It's a child of the `if_statement`, but it reports its parent as the `"if"` token due to this bug.

```
node <Node id=0x600002cdc7e0 kind=if_statement, start_point=(0, 0), end_point=(0, 4)>
parent <Node id=0x6000008dc140 kind=module, start_point=(0, 0), end_point=(0, 4)>

node <Node id=0x6000023dc070 kind="if", start_point=(0, 0), end_point=(0, 2)>
parent <Node id=0x600002cdc7e0 kind=if_statement, start_point=(0, 0), end_point=(0, 4)>

node <Node id=0x600002cdc660 kind=identifier, start_point=(0, 2), end_point=(0, 2)>
parent <Node id=0x6000023dc070 kind="if", start_point=(0, 0), end_point=(0, 2)>
```

I believe the issue is due to this line in `ts_node_parent()`: https://github.com/tree-sitter/tree-sitter/blob/bd06b1a8b37e66a875002bede0dcc9419d8de307/lib/src/node.c#L478

This line recurses into the first node where the end byte overlaps the current node's end byte. But that will only hold _sometimes_ for a zero width node - the zero width node could be an end node (at any depth) after recursing, or could be a sibling of the node we're recursing into, or could be a child (at any depth) of a following sibling.

(I assume in theory a zero-width node could also have both siblings and children that are zero-width nodes, complicating this)

### LINKED ISSUE #1872: `parent` of 0width node is `previousSibling`
running `node.parent` on a 0width node returns the previous sibling instead of the parent

0width node: `node.text == ''`
node is named

```js
alias(
    token.immediate(
        repeat(
            choice(
                /\\./,
                /[^\\"\r\n]+/,
            ),
        ),
    ),
    $.value
),
```

### LINKED ISSUE #3271: `ts_node_parent` can still return wrong parent for a zero-width node
### Problem

Related to #3185.

Sorry if I misunderstood something in how tree-sitter works, but it looks like `ts_node_parent()` sometimes returns the previous sibling (or its descendant) of a zero-width node instead of its parent. This can happen when the previous sibling has children, since currently `ts_node_parent()` descends into the first node that has child nodes and can contain the given node:

https://github.com/tree-sitter/tree-sitter/blob/4d8b031bcc218977187575e22181106d48f82b86/lib/src/node.c#L523

I made a modified version of python grammar that could produce such a scenario (python.patch). It allows function definitions without ":" at the end, so that `parameters` node could be a sibling of a zero-width `body`.

### Steps to reproduce

<details>

<summary>test.patch</summary>

```diff
diff --git a/cli/src/tests/node_test.rs b/cli/src/tests/node_test.rs
index 193b4562..d4c987db 100644
--- a/cli/src/tests/node_test.rs
+++ b/cli/src/tests/node_test.rs
@@ -250,7 +250,7 @@ fn test_node_parent_of_child_by_field_name() {
 
 #[test]
 fn test_parent_of_zero_width_node() {
-    let code = "def dupa(foo):";
+    let code = "def dupa(foo)";
 
     let mut parser = Parser::new();
     parser.set_language(&get_language("python")).unwrap();
@@ -258,12 +258,13 @@ fn test_parent_of_zero_width_node() {
     let tree = parser.parse(code, None).unwrap();
     let root = tree.root_node();
     let function_definition = root.child(0).unwrap();
-    let block = function_definition.child(4).unwrap();
+    let block = function_definition.child(3).unwrap();
     let block_parent = block.parent().unwrap();
 
     assert_eq!(block.to_string(), "(block)");
-    assert_eq!(block_parent.kind(), "function_definition");
-    assert_eq!(block_parent.to_string(), "(function_definition name: (identifier) parameters: (parameters (identifier)) body: (block))");
+    assert_eq!(block_parent.kind(), "parameters");
+    assert_eq!(block_parent.to_string(), "(parameters (identifier))");
+    assert_eq!(function_definition.to_string(), "(function_definition name: (identifier) parameters: (parameters (identifier)) body: (block))");
 }
 
 #[test]
```

</details>

<details>

<summary>python.patch</summary>

```diff
diff --git a/src/grammar.json b/src/grammar.json
index 56fdf8e..dac95ec 100644
--- a/src/grammar.json
+++ b/src/grammar.json
@@ -1647,18 +1647,22 @@
                     "type": "SYMBOL",
                     "name": "type"
                   }
+                },
+                {
+                    "type": "STRING",
+                    "value": ":"
                 }
               ]
             },
+            {
+                "type": "STRING",
+                "value": ":"
+            },
             {
               "type": "BLANK"
             }
           ]
         },
-        {
-          "type": "STRING",
-          "value": ":"
-        },
         {
           "type": "FIELD",
           "name": "body",
```

</details>

Put both files into the home directory and run this:

```console
git clone https://github.com/tree-sitter/tree-sitter
cd tree-sitter
git reset --hard cbcb51b857
git apply ~/test.patch
script/fetch-fixtures
git -C test/fixtures/grammars/python/ reset --hard a22761025cdac
git -C test/fixtures/grammars/python/ apply ~/python.patch
script/generate-fixtures
script/test
```

All tests pass.

### Expected behavior

(modified) `test_parent_of_zero_width_node()` should fail since `parameters` node is not `body`'s immediate parent.

tree-sitter 0.22.2 (cbcb51b8575ca58f8bfde4f5e78cd08d562faa81)

### Operating system/version

Ubuntu 22.04.3 LTS (WSL 2)

amaanq:
Closes #1378
Closes #1647
Closes #1872
Closes #3271

## Problem

Currently, the internal node child iterator which is used to find the parent or find the child containing a descendant is buggy when it comes to siblings that are zero-width tokens (ZWT) that are *adjacent* to the prior sibling. This has come up numerous times across many issues, and is frankly not an easy problem to solve. The issue is that the loop inside `ts_node_child_containing_descendant` does not advance to any more siblings if the current child node is one that touches the end boundary of the target subnode. This is problematic because if the ZWT sibling that is adjacent to this prior sibling *is* in fact the subnode, or *does* in fact contain the subnode, the returned data is incorrect. Therefore, there are many cases to consider when it comes to these ZWTs.

A) The case where we have a ZWT following a node of which said node's children contains the target subnode. Let's say our target node is A.2 in the figure below (2nd child of node A):
```
      A          B
(0    :    10)(10:10)
......................
 A.1     A.2   <------- Target
(0:3)   (3:10)
```

B) The case where the ZWT *is* the target subnode, so in the above diagram, that'd be `B`

C) The case where the ZWT is comprised of children that are all ZWTs, where one of these children is in fact the target subnode. I'm not sure if a grammar exists that has this construct.

## Solution

A function that peeks at the next sibling to see if it's an adjacent ZWT *and* is relevant has been added after the first if condition of the inner do-while loop of `ts_node_child_containing_descendant`. We fast-forward to process as *many* adjacent ZWTs as we can, checking to see if either a recursive call of `ts_node_child_containing_descendant` gives us a non-null node (case C) OR if the node id of the next ZWT(s) is **equal** to the target subnode's id (case B). The reason I looped in case A is because my initial implementation actually *failed* on that case (it was more like an iterator that only checked the siblings without descending into the prior sibling's subtrees).

This implementation is now a DFS (calling `ts_node_child_containing_descendant` inside this loop) that is performed `n` times where `n` is the number of ZWTs adjacent to the sibling + the current node. If no match was found or the node wasn't exactly the target subnode then we restore `self` with the `old` node that was the current node *prior* to entering this loop. This fixes case *A*. If we were to take Python for example and use the test case [here](https://github.com/tree-sitter/tree-sitter/issues/3271#issuecomment-2079772923), the parent of Python's `)` node in a `parameters` node would be `function_definition` and not `parameters` like we'd expect. Though, this test is a bit unfeasible to use here because it requires a patch to the Python grammar. So, I opted for the example in #1378, which is now fixed with this PR.

I will admit that this loop + the function `ts_node_child_iterator_next_sibling_is_empty_adjacent` is not the prettiest solution (and probably not the most possible performant one either), but I could not think of another way to solve all the cases I outlined above that was more performant & clearer, but I'm open to feedback :)

### DIFF
--- a/cli/src/tests/node_test.rs
+++ b/cli/src/tests/node_test.rs
@@ -290,6 +290,16 @@ fn test_parent_of_zero_width_node() {
         function_definition
     );
     assert_eq!(function_definition.child_containing_descendant(block), None);
+
+    let code = "<script></script>";
+    parser.set_language(&get_language("html")).unwrap();
+
+    let tree = parser.parse(code, None).unwrap();
+    let root = tree.root_node();
+    let script_element = root.child(0).unwrap();
+    let raw_text = script_element.child(1).unwrap();
+    let parent = raw_text.parent().unwrap();
+    assert_eq!(parent, script_element);
 }
 
 #[test]
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -12,6 +12,8 @@ typedef struct {
   const TSSymbol *alias_sequence;
 } NodeChildIterator;
 
+static inline bool ts_node__is_relevant(TSNode self, bool include_anonymous);
+
 // TSNode - constructors
 
 TSNode ts_node_new(
@@ -99,6 +101,21 @@ static inline bool ts_node_child_iterator_next(
   self->position = length_add(self->position, ts_subtree_size(*child));
   self->child_index++;
   return true;
+}
+
+// This will return true if the next sibling is a zero-width token that is adjacent to the current node and is relevant
+static inline bool ts_node_child_iterator_next_sibling_is_empty_adjacent(NodeChildIterator *self, TSNode previous) {
+  if (!self->parent.ptr || ts_node_child_iterator_done(self)) return false;
+  if (self->child_index == 0) return false;
+  const Subtree *child = &ts_subtree_children(self->parent)[self->child_index];
+  TSSymbol alias = 0;
+  if (!ts_subtree_extra(*child)) {
+    if (self->alias_sequence) {
+      alias = self->alias_sequence[self->structural_child_index];
+    }
+  }
+  TSNode next = ts_node_new(self->tree, child, self->position, alias);
+  return ts_node_end_byte(previous) == ts_node_end_byte(next) && ts_node__is_relevant(next, true);
 }
 
 // TSNode - private
@@ -530,6 +547,24 @@ TSNode ts_node_child_containing_descendant(TSNode self, TSNode subnode) {
       ) {
         return ts_node__null();
       }
+
+      // Here we check the current self node and *all* of its zero-width token siblings that follow.
+      // If any of these nodes contain the target subnode, we return that node. Otherwise, we restore the node we started at
+      // for the loop condition, and that will continue with the next *non-zero-width* sibling.
+      TSNode old = self;
+      // While the next sibling is a zero-width token
+      while (ts_node_child_iterator_next_sibling_is_empty_adjacent(&iter, self)) {
+        TSNode current_node = ts_node_child_containing_descendant(self, subnode);
+        // If the target child is in self, return it
+        if (!ts_node_is_null(current_node)) {
+          return current_node;
+        }
+        ts_node_child_iterator_next(&iter, &self);
+        if (self.id == subnode.id) {
+          return ts_node__null();
+        }
+      }
+      self = old;
     } while (iter.position.bytes < end_byte || ts_node_child_count(self) == 0);
   } while (!ts_node__is_relevant(self, true));
 


ObserverOfTime:
Backport or can this break something?

PR #3570, opened 2024-09-01
fix(lib): backtrack to the last relevant iterator if no child was found

### LINKED ISSUE #2012: Incorrect result of ts_node_first_child_for_byte
There are 2 reports of `ts_node_first_child_for_byte` returning what appears to be an incorrect result:

* https://github.com/tree-sitter/tree-sitter-bash/issues/139
* https://github.com/sogaiu/tree-sitter-clojure/issues/32

I investigated both issues and posted [a writeup](https://github.com/tree-sitter/tree-sitter-bash/issues/139#issuecomment-1364675904) to the fomer issue explaining my findings.  ATM, I think `ts_node_first_child_for_byte` doesn't handle certain specific cases.

I went searching for use of `ts_node_first_child_for_byte`, but apart from Emacs 29+, I did not find any use.  Does anyone know of other uses of this function?

cc @casouri @dannyfreeman

amaanq:
Closes https://github.com/tree-sitter/tree-sitter-bash/issues/139
Closes #2012

## Problem

The current implementation of `ts_node__first_child_for_byte` iterates over a node's children, descending down until it finds a node that is *relevant* and its end byte is past the goal byte. The problem is, if a node that is *not* relevant (e.g. an auxiliary repeat node, or a hidden node) spans a range that is before and after the goal byte and has child nodes, the iterator descends down to this node and checks each child. However, it might be the case that the relevant child(ren) are all *before* the goal byte, with the only one being *after* the goal byte being a hidden node. And so, the function returns a null node because it thinks  no such node exists past this byte, but what it fails to do is "backtrack" back to the highest (shallowest?) iterator *that has more siblings to check*. If we were at the end of the line with our non-relevant node, sure, returning null makes sense then. But in the issue linked above, that is not the case, and there *is* a relevant child beyond the auxiliary repeat symbols that is after the goal byte.

## Solution

Before descending down into a non-relevant node's children, we save the current iterator into a `last_iterator` variable. If it turns out we have a `last_iterator` that has more siblings *to be checked*, we restore this iterator and continue the loop. So, in the case linked above with the Bash parser, once we descend down the auxiliary symbols and past the relevant node that is before the goal byte (to the non-relevant node that is after the goal byte), we restore the iterator back to the last iterator with more siblings to check, and continue iterating through those siblings. This fixes the incorrect behavior outlined in the linked issue, but the Bash grammar has changed significantly since then, so it's no longer reproducible, hence why I didn't add a test.

cc @sogaiu I think you mentioned the Clojure parser suffers from the same issue. Would you mind checking this branch out and seeing if the issue is fixed with this commit? thanks :)

### DIFF
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -304,21 +304,35 @@ static inline TSNode ts_node__first_child_for_byte(
   TSNode node = self;
   bool did_descend = true;
 
+  NodeChildIterator last_iterator;
+  bool has_last_iterator = false;
+
   while (did_descend) {
     did_descend = false;
 
     TSNode child;
     NodeChildIterator iterator = ts_node_iterate_children(&node);
+  loop:
     while (ts_node_child_iterator_next(&iterator, &child)) {
       if (ts_node_end_byte(child) > goal) {
         if (ts_node__is_relevant(child, include_anonymous)) {
           return child;
         } else if (ts_node_child_count(child) > 0) {
+          if (iterator.child_index < ts_subtree_child_count(ts_node__subtree(child))) {
+            last_iterator = iterator;
+            has_last_iterator = true;
+          }
           did_descend = true;
           node = child;
           break;
         }
       }
+    }
+
+    if (!did_descend && has_last_iterator) {
+      iterator = last_iterator;
+      has_last_iterator = false;
+      goto loop;
     }
   }
 


sogaiu:
Thanks for the heads up.

I gave the code in https://github.com/tree-sitter/tree-sitter/pull/3570/commits/e1592fb1d794bc8db9388fc2da566bdfd88f5b6b a try.

Unfortunately, I don't see a difference for the test code we had.  It's possible I messed up somewhere ofc (^^;

---

I put together a repository [here](https://github.com/sogaiu/tree-sitter-issue-2012) that shows what I tried.

The repository contains `main.c` -- code that @dannyfreeman came up with (with slight tweaks) to demonstrate the issue.  The code and output are described [here](https://github.com/sogaiu/tree-sitter-clojure/issues/32).

The repository has a couple of build scripts to build the code in two ways:

1. Using tree-sitter before pr-3570 (4387e44b)
2. Using tree-sitter with pr-3570 (e1592fb1)

Diffing the output of both doesn't show any differences here.

amaanq:
Yeah, I made a mistake in my initial implementation - should be fixed now @sogaiu, I tested the Clojure issue myself and it seems fixed :)

sogaiu:
@amaanq It seems to be working now.

Thanks a lot!

maxbrunsfeld (approved):

PR #3590, opened 2024-09-06
fix(cli): keep skipped tests unchanged in the test/corpus

jetjinser:
Currently, when using `--update` and filter (`--include`, `--exclude`, ..) together, the `tree-sitter test` would only write test updates that were not skipped to the test file, and *lose* the skipped ones.
This is wrong, the skipped test should be retained unchanged.

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, HashSet},
+    collections::BTreeMap,
     ffi::OsStr,
     fs,
     io::{self, Write},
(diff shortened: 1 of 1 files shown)


amaanq:
Hey @jetjinser thanks for the PR. I updated it because I wanted to avoid the output like I described [here](https://github.com/tree-sitter/tree-sitter/pull/2902#issuecomment-1931350277), but then the issue arises where updating deletes skipped tests. So I refactored your PR a bit to account for that :slightly_smiling_face: thanks for the catch!

PR #3605, opened 2024-09-08
feat!: remove the `build-wasm` subcommand

amaanq:
`build --wasm` should be used instead

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -37,7 +37,6 @@ enum Commands {
     InitConfig(InitConfig),
     Generate(Generate),
     Build(Build),
-    BuildWasm(BuildWasm),
     Parse(Parse),
     Test(Test),
     Fuzz(Fuzz),
@@ -122,19 +121,6 @@ struct Build {
     pub reuse_allocator: bool,
     #[arg(long, short = '0', help = "Compile a parser in debug mode")]
     pub debug: bool,
-}
-
-#[derive(Args)]
-#[command(about = "Compile a parser to WASM", alias = "bw")]
-struct BuildWasm {
-    #[arg(
-        short,
-        long,
-        help = "Run emscripten via docker even if it is installed locally"
-    )]
-    pub docker: bool,
-    #[arg(index = 1, num_args = 1, help = "The path to output the wasm file")]
-    pub path: Option<String>,
 }
 
 #[derive(Args)]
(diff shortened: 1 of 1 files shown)


sogaiu:
I know this has been warned about for a bit, but is it a significant cost to keep `build-wasm` working (with the deprecation / removal warnings intact) and may be undocument it (e.g. don't have it show up via `tree-sitter --help`) instead?

As I understand it, this is my interpretation of what happened to the `web-ui` subcommand -- it's now called `playground`, but it's still possible to use it:

```
$ tree-sitter web-ui --help
Start local playground for a parser in the browser

Usage: tree-sitter playground [OPTIONS]

Options:
  -q, --quiet                        Don't open in default browser
      --grammar-path <GRAMMAR_PATH>  Path to the directory containing the grammar and wasm files
  -h, --help                         Print help
```

I think [the current warning has been in place for about 5 months](https://github.com/tree-sitter/tree-sitter/blame/278526ef75dec32e97cfbf6f18f4d299dd2c279e/cli/src/main.rs#L548)?  Does that seem right?

The period of time folks might have had a chance to see the warning seems rather on the short side and I don't think there is a good way to estimate how many who rely on `build-wasm` already are aware that this change has been coming.

Perhaps for most cases, once you know all you would need to do is to change `build-wasm` to `build --wasm`?

I have code that lets me work with multiple versions of the `tree-sitter` cli (currently back to about v0.20.9), and that means if such code is going to continue to support some older versions, it must account for the change in API.  This is doable, but it would be nicer to not to have to do this, or at least make a change later.

clason:
You will have to remove obsolete/deprecated features _some time_, ideally with a (written down) fixed strategy (warn in version X, remove in version X+1). If you _have_ such a strategy, you should _not_ deviate from it because that only creates confusion and reduces the value of the strategy.

Also, as you write, calling code can just check for version and use the appropriate API; it's just a single flag.

amaanq:
Yeah, like Clason said, I don't think the exact time a warning has been out has a bearing on when a deprecated command is removed, but rather a major version bump after the deprecation notice has been out, and 5 months + a full major version (0.23) is enough time for active users of tree-sitter to be aware of this (plus, 0.24 isn't coming out tomorrow with this removal).

As for `web-ui`, that's an alias, not a different command - they are the exact same command whereas `build` and `build-wasm` are different commands.

PR #3608, opened 2024-09-08
build(bindings): add CMakeLists.txt file

ObserverOfTime:
Follow-up to #3541.

The ABI version for the soname must be defined as a variable since the `parser.c` file may be missing during the configuration stage (the Makefile actually has the same issue), and it can't be read or set during the build stage.

### DIFF
--- a/cli/src/templates/cmakelists.txt
+++ b/cli/src/templates/cmakelists.txt
@@ -1,0 +1,54 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(tree-sitter-PARSER_NAME
+        VERSION "0.0.1"
+        DESCRIPTION "CAMEL_PARSER_NAME grammar for tree-sitter"
+        HOMEPAGE_URL "https://github.com/tree-sitter/tree-sitter-PARSER_NAME"
+        LANGUAGES C)
+
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+
+set(TREE_SITTER_ABI_VERSION ABI_VERSION_MAX CACHE STRING "Tree-sitter ABI version")
+if(NOT ${TREE_SITTER_ABI_VERSION} MATCHES "^[0-9]+$")
+    unset(TREE_SITTER_ABI_VERSION CACHE)
+    message(FATAL_ERROR "TREE_SITTER_ABI_VERSION must be an integer")
+endif()
+
+find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
+
+add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/parser.c"
+                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
+                   COMMAND "${TREE_SITTER_CLI}" generate src/grammar.json
+                            --abi=${TREE_SITTER_ABI_VERSION}
+                   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                   COMMENT "Generating parser.c")
+
+add_library(tree-sitter-PARSER_NAME src/parser.c)
+if(EXISTS src/scanner.c)
+  target_sources(tree-sitter-PARSER_NAME PRIVATE src/scanner.c)
+endif()
+target_include_directories(tree-sitter-PARSER_NAME PRIVATE src)
+
+set_target_properties(tree-sitter-PARSER_NAME
+                      PROPERTIES
+                      C_STANDARD 11
+                      POSITION_INDEPENDENT_CODE ON
+                      SOVERSION "${TREE_SITTER_ABI_VERSION}.${PROJECT_VERSION_MAJOR}")
+
+configure_file(bindings/c/tree-sitter-PARSER_NAME.pc.in
+               "${CMAKE_CURRENT_BINARY_DIR}/tree-sitter-PARSER_NAME.pc" @ONLY)
+
+include(GNUInstallDirs)
+
+install(FILES bindings/c/tree-sitter-PARSER_NAME.h
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/tree_sitter")
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/tree-sitter-PARSER_NAME.pc"
+        DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/pkgconfig")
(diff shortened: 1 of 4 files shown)


cli/src/generate/templates/cmakelists.txt
```diff
@@ -0,0 +1,53 @@
...
+install(FILES bindings/c/tree-sitter-PARSER_NAME.h
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/tree_sitter")
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/tree-sitter-PARSER_NAME.pc"
+        DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/pkgconfig")
+install(TARGETS tree-sitter-PARSER_NAME
+        LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+
+add_custom_target(test "${TREE_SITTER_CLI}" test
+                  WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                  COMMENT "tree-sitter test")
```
dundargoc, line 53:
```suggestion
                  COMMENT "tree-sitter test")
                  
                  # vim: set ft=cmake:
```

cli/src/generate/templates/cmakelists.txt
```diff
@@ -0,0 +1,53 @@
...
+                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
+                   COMMAND "${TREE_SITTER_CLI}" generate --no-bindings
+                           --abi=${TREE_SITTER_ABI_VERSION} src/grammar.json
+                   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                   COMMENT "Generating parser.c")
+
+add_library(tree-sitter-PARSER_NAME src/parser.c)
+if(EXISTS src/scanner.c)
+  target_sources(tree-sitter-PARSER_NAME PRIVATE src/scanner.c)
+endif(EXISTS src/scanner.c)
```
dundargoc, line 30: I always found adding the text inside the `endif()` parenthesis goofy. Do you find it more readable than without?

cli/src/generate/templates/cmakelists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+set(TREE_SITTER_ABI_VERSION ABI_VERSION_MAX CACHE STRING "Tree-sitter ABI version")
+if(NOT ${TREE_SITTER_ABI_VERSION} MATCHES "^[0-9]+$")
+    unset(TREE_SITTER_ABI_VERSION CACHE)
+    message(FATAL_ERROR "TREE_SITTER_ABI_VERSION must be an integer")
+endif(NOT ${TREE_SITTER_ABI_VERSION} MATCHES "^[0-9]+$")
+
+find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
+
+add_custom_target(generate DEPENDS src/parser.c)
```
dundargoc, line 19: I don't fully understand this. I can't see where this target is used. Is this meant to be used manually by a user?

dundargoc (approved):

cli/src/generate/templates/cmakelists.txt
ObserverOfTime: The main target depends on `src/parser.c` and will call this target if it is missing.

cli/src/generate/templates/cmakelists.txt
```diff
@@ -0,0 +1,53 @@
...
+if(EXISTS src/scanner.c)
+  target_sources(tree-sitter-PARSER_NAME PRIVATE src/scanner.c)
+endif(EXISTS src/scanner.c)
+target_include_directories(tree-sitter-PARSER_NAME PRIVATE src)
+
+set_target_properties(tree-sitter-PARSER_NAME
+                      PROPERTIES
+                      C_STANDARD 11
+                      POSITION_INDEPENDENT_CODE ON
+                      SOVERSION "${PROJECT_VERSION_MAJOR}.${TREE_SITTER_ABI_VERSION}")
```
ObserverOfTime, line 37: Shall we bikeshed on whether the TS ABI version should be the major or the minor?

cli/src/generate/templates/cmakelists.txt
ObserverOfTime: Definitely not in this case.

cli/src/generate/templates/cmakelists.txt
dundargoc: I still suspect the target `generate` isn't needed and that only `add_custom_command` is required but I don't know how to test this.

ObserverOfTime: Maybe, but this allows users to run it manually.

maxbrunsfeld:
I don't think CMake should be needed for compiling grammars. Can you explain the advantage of using it vs just plain Make?

ObserverOfTime:
It works on Windows with half the lines of code.

cli/src/templates/cmakelists.txt
```diff
@@ -0,0 +1,54 @@
...
+endif()
+
+find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
+
+add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/parser.c"
+                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
+                   COMMAND "${TREE_SITTER_CLI}" generate src/grammar.json
+                            --abi=${TREE_SITTER_ABI_VERSION}
+                   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                   COMMENT "Generating parser.c")
```
ObserverOfTime, line 24: With a `--no-parser` flag for `generate` we could add another intermediate target for `src/grammar.json`.

maxbrunsfeld:
Ok, that's very valid.

The big downside to me is that, while Make is installed by default on any posix system, CMake isn't, and different systems have different *versions* of CMake. For instance, in order to build `wasmtime-c-api`, I found that I couldn't even use the version of CMake that was available by default in the Debian registry of some systems I was working on.

Do you have sense of what CMake feature we can limit ourselves to, to make sure that things build on any version of CMake in common use?

ObserverOfTime:
We don't need anything special so it should work with any available CMake version.
Also, I restored the Makefile.

PR #3610, opened 2024-09-08
fix: properly handle utf8 code points for highlight and tag assertions

### LINKED ISSUE #3464: Parser returns wrong end character because of bigger length of utf8 Cyrillic letters
### Problem

Good day everyone

Here [my rule](https://github.com/redexp/tree-sitter-familymarkup/blob/589b6e54e13f27286ca20b203fe7e8eff230f1bd/grammar.js#L83) of my parser. With Latin alphabet everything works great, but if I try to change to some Cyrillic letters then parser returns wrong end char for `name` rule. Every Cyrillic letter it will return as 2 letters - so for example word `Name` will be `(name [0, 0] - [0, 4])` but `Паша` - `(name [0, 0] - [0, 8])` when should be same as `Name (name [0, 0] - [0, 4])`

### Steps to reproduce

```
git clone git@github.com:redexp/tree-sitter-familymarkup.git
cd tree-sitter-familymarkup
npm i
npm test
```
will be successful test

But if change first `Name` to `Паша` in `test/highlight/main.fm` then test will fail

### Expected behavior

any letter char has same position as latin chars

tree-sitter 0.22.6 (b40f342067a89cd6331bf4c27407588320f3c263)

### Operating system/version

Ubuntu 22.04

amaanq:
Closes #3464

## Problem

Tree-sitter's `Point`s return byte offsets for the given source code, which is ordinarily fine, except for when users want to perform highlight or test assertions by commenting *below* the given source code and using carats to signify the offset of where a given capture should match. This is okay when our source code is ASCII, but is problematic if it's UTF8 text. Consider this example:

```js
var y̆y̆y̆y̆ = function() {}
//  ^ function 
//         ^ keyword
```

The offset of our function assertion is at byte offset 4, this lines up with the name `y̆y̆y̆y̆` which is okay. But consider the key word assertion which sits at byte offset 11. This fails because on the line above it, the byte offset 11 actually corresponds to the 3rd y̆ in the `y̆y̆y̆y̆` name, because each `y̆` is *three* bytes long. This makes writing tests for text in non-utf8 characters *very* annoying and difficult, since we'd have to shift the carats over *12* characters instead of 4 to account for the length of `y̆y̆y̆y̆`. So for the prior example to work, it'd look something like this:

```js
var y̆y̆y̆y̆ = function() {}
//  ^ function 
//                  ^ keyword
```

Which, from a user's perspective, doesn't look right at all.

## Solution

When it comes to highlight and tag assertions, we should translate each TS point to a UTF8 point under the hood. We only care about translating the *column* since rows should not be affected (newlines are still newlines). We use `bstr` to create a wrapper around the source and iterate over the graphemes, and return how many graphemes were seen for the `Utf8Point`'s column once we hit the end byte offset. This function has `O(m) + O(n)` time complexity, where m is the number of lines and n is the column length, so this is not a huge performance penalty to correct the user experience.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,7 +99,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "itertools",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -126,6 +126,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40723b8fb387abc38f4f4a37c09073622e41dd12327033091ef8950659e6dc0c"
+dependencies = [
+ "memchr",
+ "regex-automata",
+ "serde",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


amaanq:
> This is a breaking change for the rust lib.

how? only the cli crate is touched

PR #3615, opened 2024-09-09
feat: add `root` field in node-types.json

### LINKED ISSUE #3574: Add root node information to `node-types.json`
### Problem

If I understand it correctly, the root node is the first `rules` entry in the `grammar.js`. The problem is that entries in `node-types.json` are sorted (by name?) and it is therefore not possible to find out which node type the root is.

The information about the root node can be helpful to generate typed bindings which directly allow to go from 'tree' → 'typed tree' / 'root node'.
Currently the user either manually has to specify the root node when generating typed bindings, or there is a non-type-safe gap in the bindings between 'tree' → 'typed node'.

### Expected behavior

`node-types.json` should contain information about the root node.

Implementing this by preserving the node types order (and then assuming the first node type is the root) is probably not a good solution because:
- Consumers of `node-types.json` cannot know then if they are processing an older version where the order was not preserved, and the first node is _not_ the root node
- It would not work well if multiple root nodes are supported in the future (https://github.com/tree-sitter/tree-sitter/issues/870)

Maybe a solution could be to add `"root": true` to the `node-types.json` entries?
Not sure how well this would work with dynamic root nodes (https://github.com/tree-sitter/tree-sitter/issues/711), but in case there are still 'default' root nodes and you can just override them at runtime, the default root node information would probably still be useful.

amaanq:
Closes #3574

## Problem

The nodes in `node-types.json` are not ordered like those in `grammar.json`, so it's difficult to figure out which node is the root node.

## Solution

A root field is added only to the node that is a root node, that is, the first node in the variables.

### DIFF
--- a/cli/src/generate/node_types.rs
+++ b/cli/src/generate/node_types.rs
@@ -36,6 +36,8 @@ pub struct NodeInfoJSON {
     #[serde(rename = "type")]
     kind: String,
     named: bool,
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
+    root: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
     fields: Option<BTreeMap<String, FieldInfoJSON>>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -475,6 +477,7 @@ pub fn generate_node_types_json(
                     .or_insert_with(|| NodeInfoJSON {
                         kind: variable.name.clone(),
                         named: true,
+                        root: false,
                         fields: None,
                         children: None,
                         subtypes: None,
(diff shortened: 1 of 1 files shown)


cli/src/generate/node_types.rs
```diff
@@ -36,6 +36,8 @@ pub struct NodeInfoJSON {
     #[serde(rename = "type")]
     kind: String,
     named: bool,
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
```
ObserverOfTime, line 39:
```suggestion
    #[serde(skip_serializing_if = "core::ops::Not::not")]
```

ObserverOfTime (approved):

binhle113:
[](url)

PR #3616, opened 2024-09-09
feat(api): expose function to retrieve symbol metadata

ribru17:
Even though we can already get `ts_language_symbol_type`, sometimes it would be nice to get the actual metadata for a symbol, e.g. to check if it is a supertype (which is counted as an auxiliary node type by `ts_language_symbol_type`). See https://github.com/neovim/neovim/pull/30224 for an example of some reasoning behind wanting this change.

---

Another different way of tracking supertypes nicely would maybe be to count them as visible nodes (since they can also be queried just like normal named nodes). So maybe it would entail a much smaller change to https://github.com/tree-sitter/tree-sitter/blob/8667e3ea0c10a7d6803240c2c611cbfaa4ca8605/cli/src/generate/render.rs#L485? 

The con of this PR is an API change, and the con of the other solution is that perhaps we still want supertypes to be considered auxiliary nodes. But I think the latter con is much smaller. Now that I think about it, it sounds like that might be a better way to go about things... but let me know what you all think.


amaanq:
Why not expose a `ts_node_is_supertype` function instead? That way we're not exposing too much of the internal types instead, and it's very similar to other existing node functions

ribru17:
Nice, that makes sense. I changed it to check symbols instead of nodes because it seemed to make more sense given other functions, but let me know if this was a misstep

PR #3641, opened 2024-09-16
Support darwin cross-compile

stackmystack:
This is (very long) follow up on https://github.com/tree-sitter/tree-sitter/pull/1824 .

It seems to me that `-dumpmachine` is supported by both `GCC` and `clang`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ SONAME_MAJOR := $(word 1,$(subst ., ,$(VERSION)))
 SONAME_MINOR := $(word 2,$(subst ., ,$(VERSION)))
 
 # OS-specific bits
-ifeq ($(shell uname),Darwin)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ SONAME_MAJOR := $(word 1,$(subst ., ,$(VERSION)))
 SONAME_MINOR := $(word 2,$(subst ., ,$(VERSION)))
 
 # OS-specific bits
-ifeq ($(shell uname),Darwin)
+ifeq ($(findstring darwin, $(shell $(CC) -dumpmachine)), darwin)
```
ObserverOfTime, line 37:
```suggestion
ifneq ($(findstring darwin,$(shell $(CC) -dumpmachine)),)
```

stackmystack:
Thanks, done.

Issue #3643, opened 2024-09-16
0.22.6 breaks test file suffix feature


ahelwer:
### Problem

For the [tree-sitter-tlaplus](https://github.com/tlaplus-community/tree-sitter-tlaplus) grammar, upgrading from tree-sitter 0.22.5 to 0.22.6 causes 61 corpus tests to fail (out of around 250).

### Steps to reproduce

```
git clone https://github.com/tlaplus-community/tree-sitter-tlaplus
cd tree-sitter-tlaplus
npm install
./node_modules/.bin/tree-sitter --version
./node_modules/.bin/tree-sitter test
git fetch origin
git checkout ts_0_22_6_upgrade
git diff main
npm install
./node_modules/.bin/tree-sitter --version
./node_modules/.bin/tree-sitter test
```

### Expected behavior

I expected tests that passed in one patch release to pass in the next patch release I suppose (the same failure occurs with 0.23.0). This might have to do with the external scanner, were there any breaking changes that happened around that?

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.22.6

### Operating system/version

linux

ahelwer:
Further experimentation shows this is almost certainly due to the test file suffix feature; quoting https://tree-sitter.github.io/tree-sitter/creating-parsers#command-test

>If your language’s syntax conflicts with the === and --- test separators, you can optionally add an arbitrary identical suffix (in the below example, |||) to disambiguate them

As far as I know the TLA+ parser is the only one using this feature. The test file parser seemingly fails to find the terminating `----|||` test input marker.

ahelwer:
Bisecting, I found the commit introducing the bug is 33045ee89803448700f7590cf427a1c3a3802d14 by @amaanq.

maxbrunsfeld:
Thanks for the report and for investigating @ahelwer

PR #3645, opened 2024-09-17
Added failing unit test for corpus test file parser

ahelwer:
See #3643

This test will pass if 33045ee89803448700f7590cf427a1c3a3802d14 is reverted, but that presumably breaks other functionality.

You can run the test with `cargo test test_parse_tlaplus_corpus_test_suffixes -- --include-ignored`

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -1291,6 +1291,77 @@ NOT A TEST HEADER
     }
 
     #[test]
+    #[ignore = "Failing; see https://github.com/tree-sitter/tree-sitter/issues/3643"]
+    fn test_parse_tlaplus_corpus_test_suffixes() {
+        let first_input = r"
+
+---- MODULE Test ----
+op ==
+    /\ 1
+    /\ 2
+====
+
+        ";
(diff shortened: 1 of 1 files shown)


ahelwer:
Thanks for the fix! Do you think this test would be valuable, though?

amaanq:
> Thanks for the fix! Do you think this test would be valuable, though @amaanq? Since the TLA+ grammar is the main consumer of this feature and another test for this feature failed to catch the breakage.

I added a minimal reproducible example to the tests :) the issue was 3 or more equals signs in the input followed by another test

ahelwer:
Sure! I'm mostly worried about catching the next breakage.

amaanq:
:thinking: what aspect of your test would you like covered that I hadn't covered? if there's a specific structure of the test you'd like tested, then sure we could add it if it actually tests something meaningful & not already covered

ahelwer:
Mostly I'm just interested in catching things that we can't think of ahead of time.

ahelwer:
Basically the way I see it:
1. We had a test that seemed comprehensive for the suffix feature
2. A change was made to the regex which broke the feature despite the test passing; this was not noticed for a few months
3. The issue was fixed and a test was added for that specific issue

Regexes are hard and it is easy to imagine there could be some modification that breaks this feature while both existing tests pass. These tests take less than a millisecond to run so I don't see why we wouldn't want to just put real-world input as a test. Otherwise I foresee this same scenario playing out again within a couple years.

PR #3664, opened 2024-09-22
refactor(tests): do not use `.as_bytes().len()` on strings

samueltardieu:
`.len()` on a string already returns the number of bytes.


amaanq:
thanks

PR #3665, opened 2024-09-22
build: simplify some cmake logic

dundargoc:
It's not needed to add logic around the find modules as they generally
behave like you want them, such as skipping a search if the cache
variable is already defined.

Also add only `src/lib.c` to the library since it by design includes the
necessary files to build the library. We could remove `src/lib.c` and
just add the necessary files as it is usually done, but if we already
have such a file it makes sense to use it to ensure it works as
expected.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -6,41 +6,29 @@ project(tree-sitter
         HOMEPAGE_URL "https://tree-sitter.github.io/tree-sitter/"
         LANGUAGES C)
 
-if(NOT MSVC)
-  set(CMAKE_C_FLAGS "-O3 -Wall -Wextra -Wshadow -Wno-unused-parameter -pedantic")
-endif(NOT MSVC)
-
 option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
 option(TREE_SITTER_FEATURE_WASM "Enable the Wasm feature" OFF)
 
-file(GLOB TS_SOURCE_FILES src/*.c)
-list(REMOVE_ITEM TS_SOURCE_FILES "${PROJECT_SOURCE_DIR}/src/lib.c")
-
-add_library(tree-sitter ${TS_SOURCE_FILES})
+add_library(tree-sitter src/lib.c)
 
 target_include_directories(tree-sitter PRIVATE src src/wasm include)
+if(NOT MSVC)
+  target_compile_options(tree-sitter PRIVATE -O3 -Wall -Wextra -Wshadow -Wno-unused-parameter -pedantic)
+endif()
 
 if(TREE_SITTER_FEATURE_WASM)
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
> It's not needed to add logic around the find modules as they generally
> behave like you want them, such as skipping a search if the cache
> variable is already defined.

But the messages are printed.

lib/CMakeLists.txt
```diff
@@ -6,41 +6,29 @@ project(tree-sitter
...
-    find_library(WASMTIME_LIBRARY wasmtime
-                 REQUIRED)
-    message(CHECK_PASS "found")
-  endif(NOT DEFINED CACHE{WASMTIME_LIBRARY})
+  find_path(WASMTIME_INCLUDE_DIR wasmtime.h
+            PATHS ENV DEP_WASMTIME_C_API_INCLUDE
+            REQUIRED)
+
+  find_library(WASMTIME_LIBRARY wasmtime
+               REQUIRED)
```
dundargoc, line 25: We could use https://github.com/neovim/neovim/blob/master/cmake/FindWasmtime.cmake if we want to have a minimum/exact wasmtime version check. If I'm not mistaked then neovim and tree-sitter should require the same wasmtime versions, but pinging @clason for confirmation.

dundargoc:
> But the messages are printed.

Hmm, you're right. I just realized `find_path` and `find_library` doesn't print out this information like I assumed, this information is only printed out if used by `find_package` and `FindWasmtime.cmake` like neovim does. If so, the following will be printed out:

```
-- Found Wasmtime: /opt/homebrew/include (found version "25.0.0")
```

I will add this instead if you think this is a good enough replacement, otherwise let's just restore it like it was. I'll let you make the call which you prefer.

dundargoc:
This isn't that important tbh, and I don't want to take away from more important PRs. Closing.

amaanq:
I don't think it's a bad thing though to simplify logic, I'm just not a cmake expert

lib/CMakeLists.txt
clason: Well, obviously the intersection of valid ranges cannot be empty... But it is a very good idea to ensure that a compatible version is found since distros will 1. require using system wasmtime and 2. yolo upgrade every chance they get (which is once a month on the 20th).

From what it looks like, breaking changes are rare even in major version updates (unless we trigger them...) but you'll never know. If Neovim goes full wasm, then _not_ being able to rely on builds having a known-good version would be a dealbreaker.

dundargoc: Makes sense. Cmake requires using `find_package`/`FindWasmtime.cmake` but that makes the code somewhat more complex as seen here. If it's not strictly needed now but later then we can revisit this later on when we need this.

PR #3667, opened 2024-09-23
chore: remove `compile_flags.txt`

amaanq:
This is no longer needed for LSPs to pick up the build configuration given that we have a Makefile & CMakeList.txt

### DIFF
--- a/lib/compile_flags.txt
+++ b/lib/compile_flags.txt
@@ -1,4 +1,0 @@
--std=c11
--Isrc/wasm
--Iinclude
--D TREE_SITTER_FEATURE_WASM


maxbrunsfeld:
What language server do you use? For me, clangd does not seem to automatically read a Makefile or CMakeList.txt.

amaanq:
:sweat_smile: sorry, I forgot I generated a `compile_commands.json` locally and relied on that instead. The problem I faced was that `compile_flags.txt` seemed to take precedence over `compile_commands.json`, and I found that annoying because running `goto definition` on a function that is not already in an open buffer goes to the *declaration* of the function, and not the *definition*, e.g. if I run goto def on `ts_language_symbol_metadata` in `node.c` without having `language.c` open, it goes to the decl in `language.h` when I want it to go to the def in `language.c`. What I now realize is that my `compile_commands.json` was one level above `compile_flags.txt` (in the root dir of the repo), but placing it in the same dir as `compile_flags.txt` makes clangd *prefer* it, which is exactly what I want.

I can add it back, but I think generating `compile_commands.json` is much more useful for local development than `compile_flags.txt` with clangd. What do you think?

maxbrunsfeld:
Yeah, I'm fine using cmake to generate the compile_commands.json.

PR #3679, opened 2024-09-24
fix: disallow empty string literals in rules

### LINKED ISSUE #700: Empty string as input to choice causes `parse` to hang
Example:

```js
boolean_literal: $ => choice('true', 'false', '')
```

Would be nice if this can be rejected on `generate` instead of  hanging on `parse`.

amaanq:
Closes #700

Empty strings (`''`) should not be allowed unless it's used in the first rule, since they cause infinite loops otherwise.


clason:
This broke https://github.com/alemuller/tree-sitter-re2c @alemuller

PR #3689, opened 2024-09-27
feat: move generate logic to its own crate

### LINKED ISSUE #3357: Split tree_sitter_cli::generate off into its own crate

amaanq:
Closes #3357


ObserverOfTime:
Wouldn't it be better to do this _after_ moving the bindings to another command?

amaanq:
> Wouldn't it be better to do this _after_ moving the bindings to another command?

Hm, sure, I'll do that first

PR #3694, opened 2024-09-27
feat!: move generation of grammar files to an `init` command

### LINKED ISSUE #3321: generate should not create .editorconfig (et al.), that should be part of an init command
### Problem

.editorconfig is a configuration file to help developers in creating a TSG. `generate` is to generate a TSG once it's already been coded by a developer.

Currently, the CLI checks to see if the file exists, and if it does not, it generates one from the template. It does this with other files, too. ("The first time you run tree-sitter generate, it will also generate a few other files for bindings")

If the developer does not *want* an .editorconfig file (or one of the others), there doesn't seem to be a way to prevent `generate` from adding one.

The TSG manual online indicates `generate` is the command both for *initializing/scaffolding* a project and for *generating* code for parsing every time you make changes to the grammar. ("After making changes to your grammar, just run tree-sitter generate again.")

However, some of what `generate` does is more properly organized as init code. Copying template .editorconfig, .gitignore, .gitattributes, `parser.c`, the Rust bindings, etc. files are all properly initialization code, not code for generating/building.

I'd be happy to file a PR for this, and I think it makes logical sense. I can't recall an application I've used where scaffolding and building were the same command.

### Steps to reproduce

Look in `[cli/src/generate/grammar_files.rs](https://github.com/tree-sitter/tree-sitter/blob/32cfceec6200ea5e4295e5182cb6d69045397333/cli/src/generate/grammar_files.rs#L29)`

Observe the code

```rust
// Write .editorconfig file
    missing_path(repo_path.join(".editorconfig"), |path| {
        generate_file(path, EDITORCONFIG_TEMPLATE, language_name)
    })?;
```

### Expected behavior

There should be a new command, `tree-sitter init` that does the scaffolding. `generate` should not do anything that isn't necessary for *generating* the grammar.json, etc. much like how one would expect `tree-sitter build` not to create a non-transient `.editorconfig`.

tree-sitter 0.20.8

### Operating system/version

macOS 12.0.1

amaanq:
Closes #3321

### Problem

Oftentimes, when users run generate, they get a whole slew of files created or updated, which is annoying when the author likely doesn't want that to happen. As more files get added or changed, this becomes really chaotic and annoying.

### Solution

The generate subcommand should stick to solely generating a parser and headers, and we should instead rely on an `init` subcommand that is responsible for generating grammar files. This will be extended upon in a future PR when a dedicated json config file is introduced, where we can then prompt the user for specific inputs for stuff that they want to have in their grammar repo.


maxbrunsfeld (approved):

clason:
How do you update the bindings (e.g., with a new tree-sitter version) now? Call `init` again?

clason:
And this is again a needlessly breaking change -- couldn't you keep `--no-bindings` around as a no-op?

Scorg:
Why the heck do you want every grammar to duplicate build system and binding files with all potential bugs and defects? As have been pointed above, what if an update happens to these files? Would you go around all the grammar repos and update them? Would you force third parties monitor this repo for changes (the likely scenario)?

This decision is really not though through. I would imagine that the common use case is to take the grammar file, generate the parser with bindings for the target language, and ship it with the tool. Or even statically link it.

PR #3696, opened 2024-09-28
build: bump deps

amaanq:


clason:
No wasmtime bump? (Current is 25.0.1, which Neovim already tracks.)

Cargo.toml
```diff
@@ -34,15 +34,20 @@ codegen-units = 1    # Maximum size reduction optimizations.
...
-inherits = "optimize"
-strip = false
+[profile.release-dev]
+inherits = "release"
+lto = false
+debug = true
+debug-assertions = true
+overflow-checks = true
+incremental = true
+codegen-units = 256
```
clason, line 44: Sneaky manual change!

amaanq: (it's very useful for debugging and profiling, more so than the old one)

clason: Sure, no objection to the change itself; I'm just pointing out that it's hidden in an unrelated PR (which is bad for discoverability and bisecting).

PR #3698, opened 2024-09-28
feat(init): add an update flag

ObserverOfTime:
This makes binding updates explicit.

### DIFF
--- a/cli/src/init.rs
+++ b/cli/src/init.rs
@@ -100,7 +100,11 @@ fn insert_after(
     entries.into_iter().collect()
 }
 
-pub fn generate_grammar_files(repo_path: &Path, language_name: &str) -> Result<()> {
+pub fn generate_grammar_files(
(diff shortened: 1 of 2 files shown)


cli/src/init.rs
```diff
@@ -100,14 +100,19 @@ fn insert_after(
...
 
-pub fn generate_grammar_files(repo_path: &Path, language_name: &str) -> Result<()> {
+pub fn generate_grammar_files(
+    repo_path: &Path,
+    language_name: &str,
+    allow_update: bool,
+) -> Result<()> {
     let dashed_language_name = language_name.to_kebab_case();
 
     // TODO: remove legacy code updates in v0.24.0
```
ObserverOfTime, line 110: It's time.

clason:
Can you re-add the `--no-bindings` flag (as a no-op, with a deprecation warning) while you're at it?

PR #3699, opened 2024-09-28
fix(generate): add a no-op `--no-bindings` flag

ObserverOfTime:
This prevents downstream scripts from breaking immediately upon update.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -66,6 +66,8 @@ struct Generate {
     pub grammar_path: Option<String>,
     #[arg(long, short, help = "Show debug log during generation")]
     pub log: bool,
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -448,6 +450,9 @@ impl Init {
 
 impl Generate {
     fn run(self, mut loader: loader::Loader, current_dir: &Path) -> Result<()> {
+        if self.no_bindings {
+            eprint!("The --no-bindings flag is now a no-op and will be removed in v0.25.0");
```
clason, line 454:
```suggestion
            eprint!("The --no-bindings flag is no longer used and will be removed in v0.25.0");
```

clason:
I would consider this a `fix` ("restore flag for bwc") ;)

PR #3700, opened 2024-09-29
feat: move tree-sitter configuration to dedicated file

### LINKED ISSUE #3637: Move tree-sitter configuration to dedicated file
### Problem

The tree-sitter CLI is currently configured through `package.json`, which increases its coupling with Node.js.

### Expected behavior

The configuration should be moved to a dedicated  `tree-sitter.json` file.
This file will contain all the basic metadata necessary to create/update a parser project.

<details>
<summary>Proposed schema</summary>

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "grammars": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string",
            "description": "The name of the grammar.",
            "pattern": "^[a-z0-9_]+$"
          },
          "camelcase": {
            "type": "string",
            "description": "The name converted to CamelCase.",
            "pattern": "^\\w+$",
            "examples": [
              "Rust",
              "HTML"
            ],
            "$comment": "This is used in the description and the class names."
          },
          "scope": {
            "type": "string",
            "description": "The TextMate scope that represents this language.",
            "pattern": "^(source|text)(\\.\\w+)+$",
            "examples": [
              "source.rust",
              "text.html"
            ]
          },
          "path": {
            "type": "string",
            "default": ".",
            "description": "The relative path to the directory containing the grammar."
          },
          "external-files": {
            "type": "array",
            "description": "The relative paths to files that should be checked for modifications during recompilation.",
            "items": {
              "type": "string"
            },
            "minItems": 1
          },
          "file-types": {
            "type": "array",
            "description": "An array of filename suffix strings.",
            "items": {
              "type": "string"
            },
            "minItems": 1
          },
          "highlights": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "array",
                "items": {
                  "type": "string"
                },
                "minItems": 2
              }
            ],
            "default": "queries/highlights.scm",
            "description": "The path(s) to the grammar's highlight queries."
          },
          "injections": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "array",
                "items": {
                  "type": "string"
                },
                "minItems": 2
              }
            ],
            "default": "queries/injections.scm",
            "description": "The path(s) to the grammar's injection queries."
          },
          "locals": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "array",
                "items": {
                  "type": "string"
                },
                "minItems": 2
              }
            ],
            "default": "queries/locals.scm",
            "description": "The path(s) to the grammar's local variable queries."
          },
          "tags": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "array",
                "items": {
                  "type": "string"
                },
                "minItems": 2
              }
            ],
            "default": "queries/tags.scm",
            "description": "The path(s) to the grammar's code navigation queries."
          },
          "injection-regex": {
            "type": "string",
            "format": "regex",
            "description": "A regex pattern that will be tested against a language name in order to determine whether this language should be used for a potential language injection site."
          },
          "first-line-regex": {
            "type": "string",
            "format": "regex",
            "description": "A regex pattern that will be tested against the first line of a file in order to determine whether this language applies to the file."
          },
          "content-regex": {
            "type": "string",
            "format": "regex",
            "description": "A regex pattern that will be tested against the contents of the file in order to break ties in cases where multiple grammars matched the file."
          }
        },
        "required": [
          "name",
          "scope"
        ]
      },
      "minItems": 1
    },
    "metadata": {
      "type": "object",
      "properties": {
        "version": {
          "type": "string",
          "description": "The current version of the project.",
          "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
          "$comment": "The CLI will use this version to update package.json, Cargo.toml, pyproject.toml, Makefile."
        },
        "license": {
          "type": "string",
          "default": "MIT",
          "description": "The project's license."
        },
        "description": {
          "type": "string",
          "description": "The project's description.",
          "examples": [
            "Rust grammar for tree-sitter"
          ]
        },
        "links": {
          "type": "object",
          "properties": {
            "repository": {
              "type": "string",
              "format": "uri",
              "description": "The project's repository."
            },
            "homepage": {
              "type": "string",
              "format": "uri",
              "description": "The project's homepage."
            }
          },
          "required": [
            "repository"
          ]
        },
        "authors": {
          "type": "array",
          "items": {
            "type": "object",
            "description": "The project's author(s).",
            "properties": {
              "name": {
                "type": "string"
              },
              "email": {
                "type": "string",
                "format": "email"
              },
              "url": {
                "type": "string",
                "format": "uri"
              }
            },
            "required": [
              "name"
            ]
          },
          "minItems": 1
        },
        "namespace": {
          "type": "string",
          "description": "The namespace for the Java & Kotlin packages.",
          "default": "io.github.tree-sitter",
          "$comment": "Used as is in the Maven/Gradle group name and transformed accordingly for the package names and directories (e.g. io.github.treesitter.jtreesitter.html - src/main/java/io/github/treesitter/jtreesitter/html)."
        }
      },
      "required": [
        "version",
        "links"
      ]
    },
    "bindings": {
      "type": "object",
      "description": "The language bindings that will be generated.",
      "properties": {
        "c": {
          "type": "boolean",
          "default": true,
          "const": true,
          "$comment": "Always generated"
        },
        "go": {
          "type": "boolean",
          "default": true
        },
        "java": {
          "type": "boolean",
          "default": false
        },
        "kotlin": {
          "type": "boolean",
          "default": false
        },
        "node": {
          "type": "boolean",
          "default": true,
          "const": true,
          "$comment": "Always generated (for now)"
        },
        "python": {
          "type": "boolean",
          "default": true
        },
        "rust": {
          "type": "boolean",
          "default": true,
          "const": true,
          "$comment": "Always generated"
        },
        "swift": {
          "type": "boolean",
          "default": true
        }
      }
    }
  },
  "required": [
    "grammars",
    "metadata"
  ]
}
```

</details>

amaanq:
Closes #3637

### Problem

The current way to configure a tree-sitter grammar repo is by adding a `tree-sitter` section to the `package.json`. This is unideal because it increases coupling to node and confuses newcomers as to why changing stuff in a `package.json` file influences CLI behavior. Adding more fields to the `package.json` also increases complexity of a file that, ideally, should stick to node-related configuration data only. Moving to a dedicated file will make it clear that all tree-sitter related configuration is held in this file, and it also makes it easier for us to automatically derive more data to populate in other bindings files (license info, author, etc).

### Solution

We will now migrate to a dedicated `tree-sitter.json` configuration file. This file is created and populated when a user runs `tree-sitter init`, of which now `init` prompts the user for input much like `npm init` does. Additionally, if `init` is ran and a `package.json` file *is* detected, the user is prompted for whether or not they'd like the CLI to automatically migrate their config to the new `tree-sitter.json` file. The package.json is parsed and the key fields for migration are extracted, including information such as the author, version, description, name, and repository url.

The loader now holds all the relevant data structures that represent the old package.json tree-sitter fields and new tree-sitter.json fields. Compatibility is retained for the old `tree-sitter` field in a `package.json` file until 0.25, at which point that will be removed. 

I'm putting this PR up now and not once I've finished the docs so that people can take a look and review it.

### TODO

- [x] Doc updates

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -188,7 +188,8 @@ pub struct Metadata {
     pub license: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
-    pub authors: Vec<Author>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub authors: Option<Vec<Author>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub links: Option<Links>,
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -25,6 +25,7 @@ use libloading::{Library, Symbol};
 use once_cell::unsync::OnceCell;
 use path_slash::PathBufExt as _;
 use regex::{Regex, RegexBuilder};
+use semver::Version;
 use serde::{Deserialize, Deserializer, Serialize};
 use tree_sitter::Language;
 #[cfg(any(feature = "tree-sitter-highlight", feature = "tree-sitter-tags"))]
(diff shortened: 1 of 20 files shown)


cli/loader/src/lib.rs
```diff
@@ -48,6 +49,330 @@ pub struct Config {
...
+pub struct Author {
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub email: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub url: Option<String>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Links {
```
WillLillis, line 199: Would it make sense to use something like [http::Uri](https://docs.rs/http/latest/http/uri/struct.Uri.html#) for these fields (and `Author.url` above) so the tool can ensure urls are valid at creation time? It looks like there's already some similar checks to make sure the `Author.email` field is ok.

amaanq: Yeah good point, but I'd rather not bring in another dependency *just* for validation here

WillLillis: Fair enough!

amaanq: using `url::Url` works though :grin:

cli/src/main.rs
```diff
@@ -435,17 +441,216 @@ impl InitConfig {
...
-            .map(|x| x.to_string_lossy().to_ascii_lowercase())
-        {
-            if let Some(language_name) = dir_name
-                .strip_prefix("tree-sitter-")
-                .or_else(|| Some(dir_name.as_ref()))
-            {
-                generate_grammar_files(current_dir, language_name, self.update)?;
+        let configure_json = if current_dir.join("tree-sitter.json").exists() {
+            Confirm::with_theme(&ColorfulTheme::default())
+                .with_prompt("It looks like you already have a `tree-sitter.json` file. Do you want to re-configure it?")
```
ObserverOfTime, line 446: Not sure if this is necessary.

cli/src/main.rs
```diff
@@ -435,17 +441,216 @@ impl InitConfig {
...
+                .interact()?
+        } else if current_dir.join("package.json").exists() {
+            if Confirm::with_theme(&ColorfulTheme::default())
+                .with_prompt("It looks like you have a `package.json` already. If you have a `tree-sitter` field in it, do you want to auto-migrate to the new `tree-sitter.json` config file?")
```
ObserverOfTime, line 450: Maybe we should migrate unprompted with a warning?

cli/src/main.rs
```diff
@@ -435,17 +441,216 @@ impl InitConfig {
...
+                    .with_prompt("What file types are associated with your language? (space or comma separated, e.g. `.py, .pyw`)")
+                    .default(format!(".{name}"))
+                    .validate_with(|input: &String| {
+                        if input.split(',')
+                            .flat_map(|part| part.split(' '))
+                            .all(|ext| ext.starts_with('.'))
+                        {
+                            Ok(())
+                        } else {
+                            Err("Each file type must start with a period")
```
ObserverOfTime, line 525: Not necessarily. Filenames can be filetypes too.

cli/src/main.rs
```diff
@@ -435,17 +441,216 @@ impl InitConfig {
...
+                        }
+                    })
+                    .interact_text()
+                    .map(|ft| {
+                        let mut set = HashSet::new();
+                        for ext in ft.split(',').flat_map(|part| part.split(' ')) {
```
ObserverOfTime, line 531: We should pick one of the two.

cli/src/templates/_cargo.toml
```diff
@@ -1,8 +1,9 @@
 [package]
 name = "tree-sitter-PARSER_NAME"
-description = "CAMEL_PARSER_NAME grammar for tree-sitter"
+description = "PARSER_DESCRIPTION"
 version = "0.0.1"
-license = "MIT"
+authors = ["PARSER_AUTHOR_NAME PARSER_AUTHOR_EMAIL_ANGLED"]
```
ObserverOfTime, line 5: What if the author is not provided? (also applies to the other files)

cli/src/templates/package.json
```diff
@@ -1,9 +1,13 @@
...
   "name": "tree-sitter-PARSER_NAME",
   "version": "0.0.1",
-  "description": "CAMEL_PARSER_NAME grammar for tree-sitter",
+  "description": "PARSER_DESCRIPTION",
   "repository": "github:tree-sitter/tree-sitter-PARSER_NAME",
-  "license": "MIT",
+  "license": "PARSER_LICENSE",
+  "author": {
+    "name": "PARSER_AUTHOR_NAME",
+    "email": "PARSER_AUTHOR_EMAIL"
```
ObserverOfTime, line 9: This should also include the author URL, unless we get rid of it.

ObserverOfTime (commented):
This is what I had in mind:

- If `package.json` exists, the configuration should be moved to `tree-sitter.json` (with a warning)
- If `tree-sitter.json` (or another file specified via `-c` / `--config`) exists, it should read it
- If neither file exists, it should prompt the user for the basic info (with defaults):
  ```
  Parser name: foo (from cwd)
  CamelCase parser name: Foo
  Description: Foo grammar for tree-sitter
  Repository URL: https://github.com/tree-sitter/tree-sitter-foo
  TextMate Scope: source.foo
  File types (space-separated): .foo
  Version: 0.1.0
  License: MIT
  Author name:
  Author email:
  Author URL:
  ```

cli/loader/src/lib.rs
```diff
@@ -48,6 +50,194 @@ pub struct Config {
...
+    String(String),
+    Object { url: String },
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct PackageJSON {
+    pub name: String,
+    pub version: Version,
+    pub description: Option<String>,
+    pub author: Option<PackageJSONAuthor>,
```
WillLillis, line 99: Do we want an optional `maintainers` field here? When I was testing with tree-sitter-c, I noticed that your info wasn't migrated over.

docs/section-3-creating-parsers.md
```diff
@@ -103,6 +90,80 @@ You now have a working parser.
...
+##### The `metadata` field
+
+This field contains information that tree-sitter will use to populate relevant bindings' files, especially their versions. A future
+`bump-version` and `publish` subcommand will leverage this version information as well. Typically, this will all be set up when you
+run `tree-sitter init`, but you are welcome to update it as you see fit.
+
+* `version` (required) - The current version of your grammar, which should follow [semver](https://semver.org)
+* `license` - The license of your grammar, which should be a valid [SPDX license](https://spdx.org/licenses)
+* `description` - The brief description of your grammar
+* `authors` (required) - An array of objects that contain a `name` field, and optionally an `email` and `url` field. Each field is a string
```
WillLillis, line 150: Since the

cli/loader/src/lib.rs
amaanq: yeah for the migration we could add that

cli/src/templates/grammar.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * @file CAMEL_PARSER_NAME grammar for tree-sitter
```
ObserverOfTime, line 2:
```suggestion
 * @file PARSER_DESCRIPTION
```

cli/src/templates/pyproject.toml
```diff
@@ -4,18 +4,19 @@ build-backend = "setuptools.build_meta"
...
-  "Typing :: Typed"
+  "Typing :: Typed",
 ]
+authors = [{ name = "PARSER_AUTHOR_NAME", email = "PARSER_AUTHOR_EMAIL" }]
 requires-python = ">=3.9"
-license.text = "MIT"
+license.text = "PARSER_LICENSE"
 readme = "README.md"
 
 [project.urls]
```
ObserverOfTime, line 22: The URL needs to be updated.

ObserverOfTime (commented):
These template files also need to be updated:
- `__init__.py` (description)
- `binding_test.go` (stripped URL)
- `go.mod` (stripped URL)

cli/src/init.rs
```diff
@@ -100,20 +120,305 @@ fn insert_after(
...
+                        }
+                    }
+                })
+                .collect(),
+            links: Some(Links {
+                repository: old_config
+                    .repository
+                    .map(|r| match r {
+                        PackageJSONRepository::String(s) => {
+                            if let Some(stripped) = s.strip_prefix("github:") {
```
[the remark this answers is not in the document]
ObserverOfTime: `user/repo` is also treated as a GitHub link.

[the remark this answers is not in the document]
amaanq: right, good point

cli/src/init.rs
```diff
@@ -100,20 +123,320 @@ fn insert_after(
...
+                                let repo_name = stripped.trim();
+                                Url::parse(&format!("https://github.com/{repo_name}"))
+                            } else if let Some(stripped) = s.strip_prefix("gitlab:") {
+                                let repo_name = stripped.trim();
+                                Url::parse(&format!("https://gitlab.com/{repo_name}"))
+                            } else if Regex::new(r"\w+/\w+").unwrap().is_match(&s) {
+                                Url::parse(&format!("https://github.com/{s}"))
```
ObserverOfTime, line 324:
```suggestion
                                Url::parse(&format!("https://github.com/{stripped}"))
                            } else if Regex::new(r"^[\w.-]+/[\w.-]+$").unwrap().is_match(&s) {
                                Url::parse(&format!("https://github.com/{s}"))
                            } else if let Some(stripped) = s.strip_prefix("gitlab:") {
                                Url::parse(&format!("https://gitlab.com/{stripped}"))
```

ObserverOfTime (approved):

ObserverOfTime:
We forgot to add a `PARSER_VERSION` placeholder.

amaanq:
done

bm-w:
This PR broke my usage of `tree-sitter-loader`. The migration strategy only works when using the Tree-sitter CLI, and then only when the parser repo is the current directory. In my case I’m using `tree-sitter-loader` by itself, in which case there is no “migration”, but just the pre-existing fallback that parses a very reduced `LanguageConfiguration` from `src/grammar.json`.

And it turns out that even if I can manually select such a reduced `LanguageConfiguration` based on the name, there is no way to load a `Language` from it (`Loader::language_by_id` is private, and all the public methods don’t work because the reduced `LanguageConfiguration` lacks the required data).

PR #3717, opened 2024-10-02
build(cmake): link wasmtime dependencies

ObserverOfTime:
Only for static builds. See https://docs.wasmtime.dev/c-api/.
Though it doesn't seem like cmake actually links them? :thinking:

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -6,10 +6,6 @@ project(tree-sitter
         HOMEPAGE_URL "https://tree-sitter.github.io/tree-sitter/"
         LANGUAGES C)
 
-if(NOT MSVC)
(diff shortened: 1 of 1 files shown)


lib/CMakeLists.txt
```diff
@@ -1,15 +1,11 @@
 cmake_minimum_required(VERSION 3.13)
 
 project(tree-sitter
-        VERSION "0.23.0"
+        VERSION "0.23.2"
```
ObserverOfTime, line 4: Don't forget to bump this too. ;)

clason: Don't forget that the version _on this branch_ is (or should be) 0.24.0 -- point releases only come from the release-0.23 branch.

soz-ai (changes requested):
A

PR #3718, opened 2024-10-02
fix(init): do not migrate `package.json` on error

amaanq:


clason:
This breaks haskell_persistent on `tree-sitter generate`:
```
Failed to find `tree-sitter` section in package.json, unable to migrate
Failed to migrate package.json
thread 'main' panicked at cli/generate/src/render.rs:259:63:
no entry found for key
```

clason:
Also on simple `tree-sitter build`:
```
Failed to migrate package.json

Caused by:
    missing field `name` at line 3 column 1
```

I _really_ think this migration should be opt-in, at least for a version. It's unacceptable that a **completely unrelated** command fails now.

PR #3733, opened 2024-10-04
build(cmake): correct library scopes

ObserverOfTime:
For consumers of static tree-sitter + wasmtime.

---

PRIVATE: needed by the library but not its consumers
PUBLIC: needed by the library and its consumers
INTERFACE: needed only by the library's consumers

https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-command-scope

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.13)
 
 project(tree-sitter
         VERSION "0.24.1"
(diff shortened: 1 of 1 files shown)


lib/CMakeLists.txt
```diff
@@ -44,22 +36,27 @@ if(TREE_SITTER_FEATURE_WASM)
...
     message(CHECK_START "Looking for wasmtime library")
-    find_library(WASMTIME_LIBRARY wasmtime
-                 REQUIRED)
+    if(BUILD_SHARED_LIBS)
+      find_library(WASMTIME_LIBRARY wasmtime REQUIRED)
+    elseif(MSVC)
+      find_library(WASMTIME_LIBRARY wasmtime.lib REQUIRED)
+    else()
+      find_library(WASMTIME_LIBRARY libwasmtime.a REQUIRED)
+    endif()
```
dundargoc, line 45: This is a bit sus, since `find_library(WASMTIME_LIBRARY wasmtime REQUIRED)` will find both shared and static libraries if we don't add some more code. Also `REQUIRED` was only added in version 3.18, so we shouldn't use that.

ObserverOfTime: But it will look for the shared library first.

ObserverOfTime: We can bump the minimum version to 3.18. It's the [oldest available version](https://pkgs.org/search/?q=cmake).

dundargoc: Didn't max say he uses potato computers that run ancient debians or something? I don't personally mind either way though.

clason: Buster just went EOL this year, so that may have been before that.

amaanq: I don't mind raising the minimum requirement to 3.18

dundargoc: > But it will look for the shared library first.

I don't think this is defined actually, it differs from platform to platform IIRC. But I'll take your word for it if you've tested this out.

dundargoc (approved):

lib/CMakeLists.txt
dundargoc: Wait, I think neovim might have to bump mimimum to 3.18 if treesitter bumps to 3.18 if I'm not mistaken. @clason I'm thinking correctly here, right?

clason: I think so, yes, unless we want to go back to our own CMake files.

dundargoc: That feels like a step in the wrong direction. Could we not just check if library was not found and then throw error in this PR? It's should be one line of code.

clason: (I think raising to 3.18 is not out of the question, for the record.)

dundargoc: Hmm, I see. In that case let's go ahead with this then.

clason: But if we don't have to, it's less discussion, so even better ;)

ObserverOfTime: > I don't think this is defined actually, it differs from platform to platform IIRC.

From https://cmake.org/cmake/help/latest/command/find_library.html

> Each library name given to the NAMES option is first considered as a library file name and then considered with platform-specific prefixes (e.g. lib) and suffixes (e.g. .so). Therefore one may specify library file names such as libfoo.a directly. This can be used to locate static libraries on UNIX-like systems.

dundargoc: I don't think that text says shared libraries will always be searched first, but it was a nitpick in the first place anyway. Let's go ahead with the PR.

PR #3738, opened 2024-10-04
build(zig): add optional wasmtime dependency

ObserverOfTime:
And support compiling a shared library

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,20 +1,116 @@
 const std = @import("std");
 
-pub fn build(b: *std.Build) void {
-    var lib = b.addStaticLibrary(.{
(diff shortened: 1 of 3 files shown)


build.zig
```diff
@@ -1,18 +1,113 @@
...
+    .target = target,
+    .optimize = optimize,
+    .link_libc = true,
+  }) else b.addSharedLibrary(.{
+    .name = "tree-sitter",
+    .pic = true,
+    .target = target,
+    .optimize = optimize,
+    .link_libc = true,
+  });
```
ObserverOfTime, line 22: It's also possible to specify the version, but it requires the patch, which we don't want.

PR #3740, opened 2024-10-05
feat!: properly handle UTF-16 endianness encoding

amaanq:
### Problem

Tree-sitter always treats UTF16 code as little endian. This is problematic since UTF16 text can be big endian encoded, which cannot be configured in the lib or CLI.

### Solution

A new enum member for UTF16-BE (`TSInputEncodingUTF16BE`) is added in `TSInputEncoding`, and the old `TSInputEncodingUTF16` member was renamed to `TSInputEncodingUTF16LE` in the lib. In the unicode header, we add an endian header that is portable (taken from https://gist.github.com/panzi/6856583) and separate decoding of UTF16-LE and UTF16-BE into two separate functions.

In the Rust side, we deprecate the `Parser::parse_utf16` and `Parser::parse_utf16_with` functions in favor of the `_le` and `_be` counterparts, to be removed in 0.26.

> [!NOTE]  
> On some platforms, the macro `_DEFAULT_SOURCE` needs to be defined. But this gets convoluted since depending on the order of header inclusions and macro definitions, `_POSIX_C_SOURCE` could be overwritten by a newer value, or the lib fails to build. As such, it's better to define these macros in the build process, and not in the actual source or header files themselves. This way we avoid having to define `_POSIX_C_SOURCE` and `_DEFAULT_SOURCE` in several places, and avoid having the former be overridden by the stdlib.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -100,24 +100,42 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
     let time = Instant::now();
 
     #[inline(always)]
-    fn is_utf16_bom(bom_bytes: &[u8]) -> bool {
-        bom_bytes == [0xFF, 0xFE] || bom_bytes == [0xFE, 0xFF]
+    fn is_utf16_le_bom(bom_bytes: &[u8]) -> bool {
+        bom_bytes == [0xFF, 0xFE]
     }
 
-    let tree = match opts.encoding {
-        Some(encoding) if encoding == ffi::TSInputEncodingUTF16 => {
-            let source_code_utf16 = source_code
-                .chunks_exact(2)
-                .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
-                .collect::<Vec<_>>();
-            parser.parse_utf16(&source_code_utf16, None)
+    #[inline(always)]
+    fn is_utf16_be_bom(bom_bytes: &[u8]) -> bool {
+        bom_bytes == [0xFE, 0xFF]
+    }
+
+    let encoding = match opts.encoding {
+        None if source_code.len() >= 2 => {
+            if is_utf16_le_bom(&source_code[0..2]) {
+                Some(ffi::TSInputEncodingUTF16LE)
+            } else if is_utf16_be_bom(&source_code[0..2]) {
+                Some(ffi::TSInputEncodingUTF16BE)
+            } else {
+                None
+            }
         }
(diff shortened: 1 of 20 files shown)


lib/CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ set_target_properties(tree-sitter
                       POSITION_INDEPENDENT_CODE ON
                       SOVERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}")
 
+add_definitions(-D_POSIX_C_SOURCE=200112L -D_DEFAULT_SOURCE)
```
ObserverOfTime, line 74:
```suggestion
target_compile_definitions(tree-sitter PRIVATE _POSIX_C_SOURCE=200112L _DEFAULT_SOURCE)
```

ObserverOfTime (changes requested):

cli/src/parse.rs
```diff
@@ -100,24 +100,43 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+        bom_bytes == [0xFF, 0xFE]
     }
 
+    #[inline(always)]
+    fn is_utf16_be_bom(bom_bytes: &[u8]) -> bool {
+        bom_bytes == [0xFE, 0xFF]
+    }
+
+    // if encoding is none and it's got an LE or BE bom, assign to LE or BE
+    opts.encoding = match opts.encoding {
```
WillLillis, line 113: Nitpick, but would it be better to assign the chosen encoding to a local here and pass `opts` as an immutable reference? Doesn't seem to make a difference in performance (tried with some large and small files) but imo does help with readability (since that's the only part of the struct being modified).


<details>
<summary>Bench</summary>

image

</details>

clason:
Seems this broke builds on Windows? Can't test unfortunately. Is this not covered by CI?

clason:
Also, the endian headers might be outdated, see https://github.com/mikepb/endian.h/issues/4

amaanq:
> Seems this broke builds on Windows? Can't test unfortunately. Is this not covered by CI?

Windows x86 and x86_64 are tested in CI 🤔 that's a useful issue though!

clason:
> Windows x86 and x86_64 are tested in CI

Weeell... They are, but they skip building the lib, which is exactly what broke here.

(Now that there's a cmake file, there's no reason to do that any longer -- frankly, I would have expected that PR to enable them, as Windows support was the primary reason given for adding it.)

clason:
Also broken is BSD...

PR #3742, opened 2024-10-05
Misc template improvements

ObserverOfTime:


cli/src/templates/gitignore
```diff
@@ -1,18 +1,22 @@
...
+Cargo.lock
 
 # Node artifacts
 build/
 prebuilds/
 node_modules/
-*.tgz
+package-lock.json
+pnpm-lock.yaml
+yarn.lock
```
amaanq, line 11: These shouldn't be ignored imo, at least not Cargo.olck. Not sure what the convention is in JS land but it's really useful to have a lockfile that ensures reproducible builds based on exact version numbers

ObserverOfTime: Yes, but the repository is already full of generated files. I'd rather not add a bunch of lockfiles on top.

clason: I'm sorry, but that is the silliest reason I have ever heard.

PR #3758, opened 2024-10-06
feat: move scripts to xtasks

amaanq:
Supersedes #3136

~~Still not done, but a lot more fleshed out than the last PR. The main benefit here is that this will work on any platform, without requiring a cmd script for windows.~~


xtask/src/main.rs
```diff
@@ -1,35 +1,232 @@
...
+    "docker.io/emscripten/emsdk:",
+    include_str!("../../cli/loader/emscripten-version")
+);
+
+fn main() {
+    let result = run();
+    if let Err(err) = &result {
+        // Ignore BrokenPipe errors
+        if let Some(error) = err.downcast_ref::<std::io::Error>() {
+            if error.kind() == std::io::ErrorKind::BrokenPipe {
```
ObserverOfTime, line 140: Why?

amaanq: copied from [cli/src/main.rs](https://github.com/tree-sitter/tree-sitter/blob/5c6445edea1efaa2cac45b2fd16c42441896a11a/cli/src/main.rs#L1272), because of https://github.com/tree-sitter/tree-sitter/pull/1135. Though I think xtask commands are less likely to be piped, I don't see why not to have this

PR #3762, opened 2024-10-08
feat: Add CST pretty-printer for parser output

WillLillis:
This PR adds pretty-printing functionality for the `parse` command that can be accessed via `--cst` or `-c`.  Here's a quick example using a bare-bones C file:

```C
int main(int argc, char *argv[])
{
    return EXIT_SUCCESS;
}
```

Normal output:

image

Pretty-printed:

image

`ERROR` and `MISSING` nodes are highlighted in red (as well as invalid UTF8). For example, removing the semicolon from the above C code gives the following:

image

(EDIT) ~~TODO: Configurable color scheme~~

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -22,7 +22,7 @@ use tree_sitter_cli::{
     highlight,
     init::{generate_grammar_files, get_root_path, migrate_package_json, JsonConfigOpts},
     logger,
-    parse::{self, ParseFileOptions, ParseOutput},
+    parse::{self, ParseFileOptions, ParseOutput, ParseTheme},
     playground, query, tags,
     test::{self, TestOptions},
     test_highlight, test_tags, util, wasm,
@@ -801,6 +801,16 @@ impl Parse {
             ParseOutput::Normal
         };
 
+        let parse_theme = if color {
+            config
+                .get::<tree_sitter_cli::parse::Config>()
+                .with_context(|| "Failed to parse CST theme")?
+                .parse_theme
+                .into()
+        } else {
+            ParseTheme::blank()
+        };
+
         let encoding = self.encoding.map(|e| match e {
             Encoding::Utf8 => ffi::TSInputEncodingUTF8,
             Encoding::Utf16LE => ffi::TSInputEncodingUTF16LE,
@@ -876,6 +886,7 @@ impl Parse {
                 encoding,
                 open_log: self.open_log,
                 no_ranges: self.no_ranges,
+                parse_theme: &parse_theme,
             };
 
             let parse_result = parse::parse_file_at_path(&mut parser, &opts)?;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -183,6 +183,12 @@ struct Parse {
         help = "Output the parse data in XML format"
     )]
     pub output_xml: bool,
+    #[arg(
+        long = "cst",
+        short = 'c',
+        help = "Output the parse data in a pretty-printed CST format"
+    )]
+    pub output_cst: bool,
     #[arg(long, short, help = "Show parsing statistic")]
     pub stat: bool,
     #[arg(long, help = "Interrupt the parsing process by timeout (µs)")]
@@ -787,6 +793,8 @@ impl Parse {
             ParseOutput::Dot
         } else if self.output_xml {
             ParseOutput::Xml
+        } else if self.output_cst {
+            ParseOutput::Cst
         } else if self.quiet {
             ParseOutput::Quiet
         } else {
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -803,12 +803,13 @@ impl Parse {
 
         let parse_theme = if color {
             config
-                .get::<tree_sitter_cli::parse::Config>()
+                .get::<parse::Config>()
                 .with_context(|| "Failed to parse CST theme")?
                 .parse_theme
+                .unwrap_or_default()
                 .into()
         } else {
-            ParseTheme::blank()
+            ParseTheme::empty()
         };
 
         let encoding = self.encoding.map(|e| match e {
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -655,6 +655,7 @@ fn cst_render_node(
             )?;
         }
         let kind_color = if node.is_error() {
+            write!(stdout, "{}", paint(opts.parse_theme.error, "•"))?;
             opts.parse_theme.error
         } else if node.is_extra() || node.parent().is_some_and(|p| p.is_extra()) {
             opts.parse_theme.extra
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -567,9 +567,7 @@ fn write_node_text(
     text_info: (char, usize, usize),
 ) -> Result<()> {
     let (quote, row_width, indent_level) = text_info;
-    if (!source.is_empty() && source.find('\n') == Some(source.len() - 1))
-        || source.find('\n').is_none()
-    {
+    if source.ends_with('\n') || source.find('\n').is_none() {
         write!(
             stdout,
             "{quote}{}{quote}",
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -6,8 +6,9 @@ use std::{
     time::{Duration, Instant},
 };
 
-use anstyle::AnsiColor;
+use anstyle::{AnsiColor, Color, RgbColor};
 use anyhow::{anyhow, Context, Result};
+use bstr::ByteSlice;
 use tree_sitter::{ffi, InputEdit, Language, LogType, Parser, Point, Range, Tree, TreeCursor};
 
 use super::util;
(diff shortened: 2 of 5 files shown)


cli/src/parse.rs
```diff
@@ -393,6 +431,99 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+                node.kind()
+            )
+        )?;
+    } else if let Ok(text) = node.utf8_text(source_code) {
+        writeln!(
+            stdout,
+            "\"{}\"",
+            paint(
+                Some(AnsiColor::Green),
+                &text.replace('\n', "\\n").to_string()
```
amaanq, line 510: handle `\r`, `\t`, `\0`, `\\`, `\v`, and `\f` too

amaanq, line 510: If the text spans multiple lines, the renderer looks pretty ugly with it intruding on row info. I think each line should be processed separately.

docs/section-4-syntax-highlighting.md
```diff
@@ -58,6 +58,25 @@ The Tree-sitter highlighting system works by annotating ranges of source code wi
 
 In your config file, the `"theme"` value is an object whose keys are dot-separated highlight names like `function.builtin` or `keyword`, and whose values are JSON expressions that represent text styling parameters.
 
+### Parse Theme
+
+The Tree-sitter `parse` command will output a pretty-printed CST when the `--cst` option is used. You can control which colors are used for various parts of the tree in your configuration file:
+
+```json
+{
+  "parse-theme": {
```
WillLillis, line 67: Would it be a good idea to set one of these to `null` to indicate that assigning each field is optional?

cli/src/parse.rs
```diff
@@ -39,11 +42,104 @@ impl fmt::Display for Stats {
...
+        Self {
+            node_kind: resolve_val(value.node_kind, default.node_kind),
+            node_text: resolve_val(value.node_text, default.node_text),
+            field: resolve_val(value.field, default.field),
+            token: resolve_val(value.token, default.token),
+            row_color: resolve_val(value.row_color, default.row_color),
+            row_color_named: resolve_val(value.row_color_named, default.row_color_named),
+            extra: resolve_val(value.extra, default.extra),
+            error: resolve_val(value.error, default.error),
+        }
```
amaanq, line 133: two nits - `val_or_default` is a better name imo, and the `Into` chain is a bit crazy

```suggestion
        let val_or_default = |val: Option<Rgb>, default: Option<Color>| -> Option<Color> {
            val.map_or(default, |v| Some(Color::Rgb(v.into())))
        };
        let default = Self::default();

        Self {
            node_kind: val_or_default(value.node_kind, default.node_kind),
            node_text: val_or_default(value.node_text, default.node_text),
            field: val_or_default(value.field, default.field),
            token: val_or_default(value.token, default.token),
            row_color: val_or_default(value.row_color, default.row_color),
            row_color_named: val_or_default(value.row_color_named, default.row_color_named),
            extra: val_or_default(value.extra, default.extra),
            error: val_or_default(value.error, default.error),
        }
```

cli/src/parse.rs
```diff
@@ -39,11 +42,104 @@ impl fmt::Display for Stats {
...
+    pub token: Option<Color>,
+    pub row_color: Option<Color>,
+    pub row_color_named: Option<Color>,
+    pub extra: Option<Color>,
+    pub error: Option<Color>,
+}
+
+impl ParseTheme {
+    #[must_use]
+    pub const fn blank() -> Self {
```
amaanq, line 59: I think `empty` or `none` fit better, up to you

```suggestion
    pub const fn empty() -> Self {
```

cli/src/main.rs
```diff
@@ -787,12 +793,24 @@ impl Parse {
...
+            ParseOutput::Cst
         } else if self.quiet {
             ParseOutput::Quiet
         } else {
             ParseOutput::Normal
         };
 
+        let parse_theme = if color {
+            config
+                .get::<tree_sitter_cli::parse::Config>()
```
amaanq, line 806:
```suggestion
                .get::<parse::Config>()
```

docs/section-4-syntax-highlighting.md
amaanq: add a note instead to clearly convey that - `Note: setting the value to null will render the text without any color` or something along those lines

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+
+    for c in source.chars() {
+        if let Some(esc) = escape_invisible(c) {
+            escaped.push_str(esc);
+        } else {
+            escaped.push(c);
+        }
+    }
+
+    escaped
```
amaanq, line 554: A couple of suggestions in one, you don't need to call reserve after creating a new `String`, `with_capacity` is a shorthand version of that, and we can also just leverage `fold` since we're accumulating into the String and nothing else.

```suggestion
    source
        .chars()
        .fold(String::with_capacity(source.len()), |mut acc, c| {
            if let Some(esc) = escape_invisible(c) {
                acc.push_str(esc);
            } else {
                acc.push(c);
            }
            acc
        })
```

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+            }
+        }
+    }
+
+    Ok(())
+}
+
+fn render_node_range(opts: &ParseFileOptions, cursor: &TreeCursor, row_width: usize) -> String {
+    let node = cursor.node();
+    let is_named = node.is_named();
```
amaanq, line 606: you called `node.is_named()` in the caller of this function - just pass the value of the first call into here instead.

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+        } else {
+            opts.parse_theme.node_kind
+        };
+        write!(stdout, "{} ", paint(kind_color, node.kind()))?;
+        if node.child_count() == 0 {
+            let color = if source_code.is_utf8() {
+                opts.parse_theme.node_text
+            } else {
+                opts.parse_theme.error
+            };
```
amaanq, line 658: Non-utf8 doesn't mean we should paint it as an `error` imo, other encodings exist.

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+            opts.parse_theme.error
+        } else {
+            opts.parse_theme.node_text
+        };
+        if node.is_missing() {
+            write!(
+                stdout,
+                "{}",
+                &format!("{}: \"{}\"", paint(color, "MISSING"), node.kind())
+            )?;
```
amaanq, line 681: split this into two writes - you paint the MISSING as red correctly, but you forgot to paint the node kind. also, thinking about this, using a "warning" color for missing might be better to distinguish it from error nodes, like a dark orange. another idea might be to add a red `•` before all nodes where `has_error` is true

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+    opts: &ParseFileOptions,
+    stdout: &mut StdoutLock<'static>,
+    cursor: &TreeCursor,
+    source: &str,
+    color: Option<impl Into<Color> + Copy>,
+    quote: char,
+    row_width: usize,
+    indent_level: usize,
+) -> Result<()> {
+    if source.find('\n') == Some(source.len() - 1)
```
amaanq, line 568: this subtraction will overflow if the source is empty

cli/src/parse.rs
```diff
@@ -393,6 +526,177 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
...
+                cursor,
+                &String::from_utf8_lossy(&source_code[node.start_byte()..node.end_byte()]),
+                color,
+                '`',
+                row_width,
+                indent_level,
+            )?;
+        }
+    } else {
+        let color = if node.is_missing() || !source_code.is_utf8() {
```
amaanq, line 671: same as earlier, non-utf8 text isnt inherently erroneous, and missing might be better modeled as a "warning"

cli/src/parse.rs
```diff
@@ -39,11 +42,104 @@ impl fmt::Display for Stats {
...
+    fn from(val: Rgb) -> Self {
+        Self(val.0, val.1, val.2)
+    }
+}
+
+#[derive(Debug, Copy, Clone, Default, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct Config {
+    pub parse_theme: ParseThemeRaw,
+}
```
amaanq, line 102: make this an `Option`, we do not break user configs unless necessary

cli/src/parse.rs
WillLillis: Ah, that's much nicer

cli/src/parse.rs
WillLillis: Yeah I definitely felt a little weird writing that many `into`s, much happier with this version

docs/section-4-syntax-highlighting.md
WillLillis: Settled on using the default color for entries that are omitted or set to `null`.

amaanq (approved):

binhle113 (approved):

PR #3774, opened 2024-10-11
ci: major overhaul

ObserverOfTime:
- Simplify some workflow steps and auxiliary scripts
- Build library using cmake when not cross-compiling
- Try to fetch fixtures from cache first
- Use `actions-rust-lang/setup-rust-toolchain`

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -1,24 +1,23 @@
-name: 'Cache'
-description: "This action caches fixtures"
+name: Cache
+
+description: This action caches fixtures
+
 outputs:
   cache-hit:
-    description: 'Cache hit'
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ OBJ := $(SRC:.c=.o)
 
 # define default flags, and override to append mandatory flags
 ARFLAGS := rcs
-CFLAGS ?= -O3 -Wall -Wextra -Wshadow -pedantic
+CFLAGS ?= -O3 -Wall -Wextra -Wshadow -Wno-unused-parameter -pedantic
```
amaanq, line 28: You can just remove the parameter from `ts_parser__call_keyword_lex_fn`, seems like it's no longer used.

Scutua (approved):

Scutua (approved):

PR #3776, opened 2024-10-12
perf(loader): improve language lookup speed

amaanq:
### Problem

While running a `release-dev` build to test changes at a faster rate, I noticed that the `parse` subcommand felt like it was hanging for a little bit. Turns out, it was around ~100ms after checking with `hyperfine`:
image

 I decided to poke around and see what was causing the slowdown with a flamegraph, and got something that looked like this:
 
image

It becomes pretty evident that the majority of the time spent here is in serde's `deserialize`, which is called in `find_language_configurations_at_path` in the loader crate. If we take a look, we see there are three spots we use serde's deserialize: [here](https://github.com/tree-sitter/tree-sitter/blob/fe92e978f922b7c8caf8d43604d39f419f963749/cli/loader/src/lib.rs#L1136), [here](https://github.com/tree-sitter/tree-sitter/blob/fe92e978f922b7c8caf8d43604d39f419f963749/cli/loader/src/lib.rs#L1148), and [here](https://github.com/tree-sitter/tree-sitter/blob/fe92e978f922b7c8caf8d43604d39f419f963749/cli/loader/src/lib.rs#L1243).

The first case is not the problem, since we are loading a relatively small file, but the other two are loading a grammar.json file, and in fact, it's the same file being deserialized *twice*. It becomes pretty clear that the first one isn't needed, we deserialize `grammar.json` just to read the name field when we've already successfully parser a config file - we should use the name in the config file instead.

However, this does not solve the ~100ms avg parse subcommand time, because, as one would expect, release-like profiles likely (don't quote me) optimize the two deserialize calls into one as an optimization, since it's the same path we're reading from. So, the real problem is that we're reading potentially thousands of lines long grammar files *just* to extract the field name, and nothing else.

### Solution

Instead of reading & deserializing thousands of lines into memory to extract one field, we can rely on the invariant that if a grammar.json was generated with tree-sitter cli (why wouldn't it be? ;) ), the name field is *always* within the first three lines. We can just read three lines and extract the name field with a simple regex. The results are great: with `release-dev`, the `parse` subcommand goes down from 100ms to 2.3ms, and is just under 2ms with the `release` profile on a Ryzen 9 9950x:
image
image

Another small finding that was observed when playing around with serde's deserialization is that `serde_json::from_reader` with a `BufReader` is a bit slower than `serde_json::from_str` with `fs::read_to_string`. In my testing, for `release-dev` it was ~3.5ms, compared to 2.3ms without it, so every usage of `from_reader` + `BufReader` was replaced with `from_str` + `fs::read_to_string`.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -21,6 +21,7 @@ use anyhow::Error;
 use anyhow::{anyhow, Context, Result};
 use fs4::fs_std::FileExt;
 use indoc::indoc;
+use lazy_static::lazy_static;
 use libloading::{Library, Symbol};
 use once_cell::unsync::OnceCell;
 use path_slash::PathBufExt as _;
@@ -37,6 +38,10 @@ use tree_sitter_highlight::HighlightConfiguration;
 #[cfg(feature = "tree-sitter-tags")]
 use tree_sitter_tags::{Error as TagsError, TagsConfiguration};
 use url::Url;
+
+lazy_static! {
+    static ref GRAMMAR_NAME_REGEX: Regex = Regex::new(r#""name":\s*"(.*?)""#).unwrap();
+}
 
 pub const EMSCRIPTEN_TAG: &str = concat!("docker.io/emscripten/emsdk:", env!("EMSCRIPTEN_VERSION"));
 
@@ -143,11 +148,7 @@ pub struct TreeSitterJSON {
 impl TreeSitterJSON {
     #[must_use]
     pub fn from_file(path: &Path) -> Option<Self> {
-        if let Ok(file) = fs::File::open(path.join("tree-sitter.json")) {
-            Some(serde_json::from_reader(file).ok()?)
-        } else {
-            None
-        }
+        serde_json::from_str(&fs::read_to_string(path.join("tree-sitter.json")).ok()?).ok()
     }
 
     #[must_use]
@@ -194,7 +195,6 @@ pub struct Metadata {
     pub authors: Option<Vec<Author>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub links: Option<Links>,
-    // #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(skip)]
     pub namespace: Option<String>,
 }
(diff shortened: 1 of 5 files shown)


clason:
This broke `tree-sitter build` for a bunch of grammars:
```
[nvim-treesitter/install/glimmer_typescript] error: Error during "tree-sitter build": thread 'main' panicked at cli/src/main.rs:776:18:
called `Result::unwrap()` on an `Err` value: Failed to parse the language name from grammar.json at /Users/clason/.cache/nvim/tree-sitter-glimmer_typescript/src/grammar.json

[nvim-treesitter/install/glimmer_javascript] error: Error during "tree-sitter build": thread 'main' panicked at cli/src/main.rs:776:18:
called `Result::unwrap()` on an `Err` value: Failed to parse the language name from grammar.json at /Users/clason/.cache/nvim/tree-sitter-glimmer_javascript/src/grammar.json
...
[nvim-treesitter/install/slang] error: Error during "tree-sitter build": thread 'main' panicked at cli/src/main.rs:776:18:
called `Result::unwrap()` on an `Err` value: Failed to parse the language name from grammar.json at /Users/clason/.cache/nvim/tree-sitter-slang/src/grammar.json
...
[nvim-treesitter/install/hlsl] error: Error during "tree-sitter build": thread 'main' panicked at cli/src/main.rs:776:18:
called `Result::unwrap()` on an `Err` value: Failed to parse the language name from grammar.json at /Users/clason/.cache/nvim/tree-sitter-hlsl/src/grammar.json
...
[nvim-treesitter/install/qmljs] error: Error during "tree-sitter build": thread 'main' panicked at cli/src/main.rs:776:18:
called `Result::unwrap()` on an `Err` value: Failed to parse the language name from grammar.json at /Users/clason/.cache/nvim/tree-sitter-qmljs/src/grammar.json
```
(Due to missing "inherits" field, from a quick sample.)

cli/loader/src/lib.rs
```diff
@@ -1280,6 +1247,36 @@ impl Loader {
...
+            .with_context(|| {
+                format!(
+                    "Failed to read the first three lines of grammar.json at {}",
+                    grammar_path.display()
+                )
+            })?
+            .join("\n");
+
+        let name = GRAMMAR_NAME_REGEX
+            .captures(&first_three_lines)
```
clason, line 1268: I don't think that optimization is safe; the first three lines are not guaranteed to contain that name, e.g.,
https://github.com/tree-sitter-grammars/tree-sitter-slang/blob/dd991eb3b6957a33d9044e0f5914588f7f449a78/src/grammar.json#L2-L6

(Current CLI versions no longer generate such a -- bugged -- `grammar.json`, but these still exist in the wild as my crater test showed.)

PR #3778, opened 2024-10-12
refactor: clean up remaining scripts

ObserverOfTime:
- [x] Move cliff.toml to .github folder
- [x] Migrate generate-wasm-exports-list to Makefile
- [ ] Migrate generate-unicode-categories-json to xtask
  - [ ] Produces the same `unicode-categories.json`
  - [ ] Produces the same `unicode-properties.json`
  - [ ] Produces the same `unicode-category-aliases.json`
  - [x] Produces the same `unicode-property-aliases.json`


amaanq:
#3404 should get rid of the need for the unicode files

amaanq:
Superseded by #3831

PR #3780, opened 2024-10-13
feat: remove the need of an external JS runtime for processing grammars

### LINKED ISSUE #465: Remove the need of NodeJS for generating parsers
There are no things that JavaScript can do that can't be done with Rust. It makes possible the next logical step - get rid of NodeJS for generating parsers from the description as well.

There is a number of JavaScript parsers in Rust:
- https://github.com/ratel-rust/ratel-core
- https://github.com/dherman/esprit
- https://github.com/jasonwilliams/boa
- probably more

amaanq:
Closes #465

### Problem

Currently, hard-requiring an external JS runtime for processing grammars can be problematic for projects that want to build grammars from source. It'd be nice to have an opt-in feature in the CLI that embeds a small JS runtime, pretty much just enough for us to evaluate a grammar. For that, [QuickJS](https://bellard.org/quickjs) is a great choice.

### Solution

This PR adds [`rquickjs`](https://github.com/DelSkayn/rquickjs) to support natively executing a grammar's JavaScript code. It is gated behind the `qjs-rt` feature (QuickJS runtime), and is invoked only when an argument of `native` is passed to the `--js-runtime` argument of the `generate` subcommand. One feature that becomes very easy to implement with this runtime is the ability to add tree-sitter's parser directories to the runtime's file resolver. What this allows downstream tools to do is not even have to run `npm install` and rely on `node_modules` for resolving imports of grammars that inherit other grammars.

The only downside to this approach is that ESM is a requirement for the QJS runtime to work. This requirement is not a strct breaking change as grammars that use ESM do not break the normal `tree-sitter generate` command - that still functions as is with `node`, `deno`, and `bun`, however, any grammars that inherit a grammar that migrates to ESM will also have to migrate if they are looking to update the base grammar.

I'm putting up this PR as a draft PR now so that interested users can get a heads up, try it out themselves, and offer feedback if they'd like.

TODO:

- [ ] Uploaded release artifacts should include this by default, I'd hope every platform can build this just okay
- [ ] Add more basic stdlib support? Currently, console.log works and outputs whatever the user logs for debugging
- [ ] Add warning in CLI for grammars that are still using CommonJS to migrate to ESM, though this will never become an error
- [ ] Document this feature
- [ ] Document advisory to use ESM instead of CommonJS (should be drop-in, unless inheriting from another grammar)
- [ ] Test on grammars in the wild (requires manual changes to verify it works)


ObserverOfTime:
We should wait for DelSkayn/rquickjs#369 for better Windows support.

> Add warning in CLI for grammars that are still using CommonJS to migrate to ESM

Let's not do that for an optional feature.

cli/generate/src/dsl.js
```diff
@@ -466,6 +466,7 @@ function checkPrecedence(value) {
 }
 
 function getEnv(name) {
+  if (globalThis.native) return __ts_grammar_path;
```
ObserverOfTime, line 469: We should add an env module and use it here (or just use quickjs's `std.getenv` if rquickjs allows that).

cli/generate/src/lib.rs
```diff
@@ -245,6 +271,113 @@ fn load_js_grammar_file(grammar_path: &Path, js_runtime: Option<&str>) -> Result
...
+
+            globals.set(
+                "__print",
+                Function::new(ctx.clone(), print)?.with_name("__print")?,
+            )?;
+
+            ctx.eval::<(), _>(
+                r"
+                globalThis.console = {
+                    log(...v) {
```
ObserverOfTime, line 338: We will also need `error` (and maybe `warn`) at the very least.

sogaiu:
I tried https://github.com/tree-sitter/tree-sitter/commit/5607f8459fdc566522fed45fd20a9935c3c7c42b with tree-sitter-c and tree-sitter-janet-simple successfully.  Output produced via `native` appeared to be the same as that via `node`.

Appreciate the ability to do this via env vars, thanks!

```
$ TREE_SITTER_JS_RUNTIME=native ../tree-sitter/target/debug/tree-sitter generate
```

---

Thanks to @WillLillis for the help with accessing the gated feature by using `--features=qjs-rt` when building with cargo.

maxbrunsfeld:
This is great. Very happy about migrating to ESM.

crates/cli/Cargo.toml
```diff
@@ -30,7 +30,9 @@ name = "benchmark"
 harness = false
 
 [features]
+default = ["qjs-rt"]
```
ObserverOfTime, line 33: Should we enable it by default?

crates/generate/Cargo.toml
```diff
@@ -20,17 +20,23 @@ path = "src/generate.rs"
...
 indexmap.workspace = true
 indoc.workspace = true
 log.workspace = true
+pathdiff = { version = "0.2.3", optional = true }
 regex.workspace = true
 regex-syntax.workspace = true
+rquickjs = { version = "0.9.0", optional = true, features = [
+  "bindgen",
+  "full-async",
+] }
```
ObserverOfTime: 1. Should we enable `loader` to move module loading from JS to Rust?
2. Can we enable `bindgen` only in [unsupported platforms](https://github.com/DelSkayn/rquickjs#supported-platforms)?
3. Do we actually need `full-async`?

crates/cli/Cargo.toml
amaanq: yeah, it only adds 1mb to the binary

ObserverOfTime: 1MB is actually a lot but that's not my point.

Considering this is still experimental in a way, we shouldn't enable the feature by default (except in our release binaries).
After polishing it throughout 0.26, we can enable the feature in 0.27 and maybe make it the default runtime in 0.28 (or 1.0?).

crates/generate/Cargo.toml
amaanq: Yeah, I enabled the loader and removed `full-async`. We have 4 unsupported platforms in ci, I'd rather spend a few extra seconds for a release build to generate the bindings rather than write:

```
[target.'cfg(any(target_arch = "powerpc", target_arch = "arm", all(target_os = "windows", target_arch = "x86"), target_arch = "aarch64"))'.dependencies]
rquickjs = { version = "0.9.0", optional = true, features = ["bindgen", "loader", "phf"] }
```

ObserverOfTime: With the loader feature we should be able to handle the loading in Rust like in the [example].

[example]: https://github.com/DelSkayn/rquickjs/blob/master/examples/module-loader/src/main.rs

crates/cli/Cargo.toml
amaanq: It's not experimental, and it's not the default runtime, users *have* to opt in with `--js-runtime=native`. There isn't much to polish.

clason: Just for the sake of clarity: setting this flag has no effect on the `ts g src/grammar.json` codepath, right?

amaanq: Nope not at all, you *have* to explicitly generate from the js and pass in `--js-runtime=native`, so the only time the code changes in here are in effect is if you run `ts g --js-runtime=native`.

clason:
For reference: This will not help with grammars that require `npm` to pull in dependencies, right (meaning, they'll still need `npm install` before you can generate)? Have you tested `--js-runtime=native` with one of those?

amaanq:
Yeah, it wouldn't. I thought about adding a shim that checks the package json for the required `tree-sitter-xxx` dependencies, fetches them from npm, and then loads the sub-grammar's js file (this would be done recursively if the sub-grammar has a dependency too), but I think that's doing too much.

crates/cli/Cargo.toml
clason: Sorry, I meant to ask specifically "what happens if you do `ts g --js-runtime=native src/grammar.json`?"

clason:
But  `ts g --js-runtime=native` still works for those parsers _after_ you've done `npm install`, right?

amaanq:
> But `ts g --js-runtime=native` still works for those parsers _after_ you've done `npm install`, right?

Yes!

crates/cli/Cargo.toml
amaanq: Nothing different, it has no effect if the grammar path ends in `json`

ObserverOfTime:
We can completely decouple generation from Node:

1. Encourage users to make use of git submodules for dependencies.
2. Add support for importing from https URLs (maybe sandboxed somehow?).
3. Resolve dependency paths/URLs from `tree-sitter.json`.

(Only the first one is necessary.)

crates/generate/src/quickjs.rs
```diff
@@ -0,0 +1,480 @@
...
+    result
+}
+
+fn load_module_from_content<'a>(
+    ctx: &Ctx<'a>,
+    path: &Path,
+    contents: &str,
+) -> rquickjs::Result<Value<'a>> {
+    if path.extension().is_some_and(|ext| ext == "json") {
+        return ctx.eval::<Value, _>(format!("JSON.parse({contents:?})"));
```
ObserverOfTime, line 224: This is vulnerable to injection.

crates/generate/src/quickjs.rs
```diff
@@ -0,0 +1,480 @@
...
+    module_obj.set("exports", exports.clone())?;
+
+    let require_func = ctx.globals().get::<_, Function>("require")?;
+    let dirname = path
+        .parent()
+        .map_or_else(|| ".".to_string(), |p| p.to_string_lossy().to_string());
+    let filename = path.to_string_lossy().to_string();
+
+    let wrapper =
+        format!("(function(exports, require, module, __filename, __dirname) {{ {contents} }})");
```
ObserverOfTime, line 246: This is also vulnerable to injection.

amaanq:
> We can completely decouple generation from Node:
> 
> 1. Encourage users to make use of git submodules for dependencies.
> 2. Add support for importing from https URLs (maybe sandboxed somehow?).
> 3. Resolve dependency paths/URLs from `tree-sitter.json`.
> 
> (Only the first one is necessary.)

The only acceptable one is git submodules. I think it'd be fine to have grammars use that instead of npm.

crates/generate/src/quickjs.rs
amaanq: It's no more vulnerable than shelling out to node and executing arbitrary code.

clason:
"I'll solve this using submodules" -- "Now you have two problems".

(The _real_ problem is depending on another grammar, which introduces a coupling that downstream users can't manage.)

amaanq:
> "I'll solve this using submodules" -- "Now you have two problems".
> 
> (The _real_ problem is depending on another grammar, which introduces a coupling that downstream users can't manage.)

We could advise vendoring the base grammar.js itself, though that doesn't really help with the coupling issue

clason:
Personally (and for what little that is worth), I believe extending grammars is a mistake. Just copy and adapt, same with queries. The house of cards that is the C family firmly convinced me of that.

clason:
(None of this is commentary on this PR, which is a clear positive, even if it doesn't make any difference to nvim-treesitter and friends.)

amaanq:
> Personally (and for what little that is worth), I believe extending grammars is a mistake. Just copy and adapt, same with queries. The house of cards that is the C family firmly convinced me of that.

Yeah I agree. I think copying over the grammar.js and "importing" that locally would be better (instead of merging the two in one file, so porting changes can be done more easily).

amaanq:
Grammars that don't generate (beyond those requiring npm):

- vim (requires `fs`)
- unison (implicit global)
- slim (implicit global)
- proto (implicit global)
- haskell_persistent (implicit global, I have a year-old PR fixing this that's not merged)
- fennel (requires `lodash`, lol...)

I can PR fixes to the 3 using implicit globals, but the others require non-trivial fixes.


Anyway, this PR is really ready now.

WillLillis (approved):

ObserverOfTime:
I can handle the Vim grammar.

clason:
The vim grammar needs _a lot_ more love (including to the scanner), so please no drive-by "fixups". And all this is irrelevant to the PR here, which is targeting grammar authors, not Neovim or nvim-treesitter.

clason:
And _please_ wait for 0.26 to be released and tested in the wild before you go around converting grammars.

On that note, I must say that I find the warning on current parsers somewhat irritating: "Better compatibility" with _what_? Certainly not tree-sitter, for which these were written? The way it is, grammar authors get bugged constantly to make a change without any clear benefit. We should try to _reduce_ friction for our users, not add to it...

(Again, this is not meant to throw shade on the new feature here, which is great!)

amaanq:
> And _please_ wait for 0.26 to be released and tested in the wild before you go around converting grammars.
> 
> On that note, I must say that I find the warning on current parsers somewhat irritating: "Better compatibility" with _what_? Certainly not tree-sitter, for which these were written? The way it is, grammar authors get bugged constantly to make a change without any clear benefit. We should try to _reduce_ friction for our users, not add to it...
> 
> (Again, this is not meant to throw shade on the new feature here, which is great!)

Some grammars already use ESM in the wild, e.g. xml - it's not that this explicitly enabled ESM, it's just that until recently, nodejs (the default js runtime) did not support ESM, so we were kinda stuck with CommonJS. Since node v20 is the LTS version now (has been since April I believe), we're now able to use ESM. However, I'm not really in a rush to convert any grammars :^).

I agree the message can be improved, sorry about that. It provides better compatibility with JS runtimes, tooling & typescript (though you shouldn't write a grammar in typescript without providing a grammar.js file...), the syntax is better, and it's "official". For example, QuickJS, the runtime I added in this PR, doesn't even support CommonJS (I had to add in a shim for `require`...).

clason:
No worries, honestly trying to be constructive here.

Personally, I would suggest to
1. state clearly in the documentation that ESM is the official standard for grammars (though CommonJS is supported for backwards compatibility);
2. remove the warning, at least for 0.26;
3. _if_ and when you consider dropping support for CommonJS, _then_ start a deprecation process that begins with printing a warning (and making Node v20 a hard requirement for external runtimes).

I think it would make sense to draw up _and publicize_ a roadmap for transitioning to native runtime (such that external runtimes are the exception for niche grammars that require more stdlib or whatever); the warning should be part of this gradual process.

amaanq:
Yeah that makes sense, mentioning it in the docs would be far less intrusive.

PR #3781, opened 2024-10-13
build: treat incompatible pointer warning as error

ObserverOfTime:
See #3779

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
           - { platform: windows-x64       , target: x86_64-pc-windows-msvc      , os: windows-latest , features: wasm  }
           - { platform: windows-x86       , target: i686-pc-windows-msvc        , os: windows-latest                   }
           - { platform: macos-arm64       , target: aarch64-apple-darwin        , os: macos-14       , features: wasm  }
-          - { platform: macos-x64         , target: x86_64-apple-darwin         , os: macos-12       , features: wasm  }
+          - { platform: macos-x64         , target: x86_64-apple-darwin         , os: macos-13       , features: wasm  }
 
           # Cross compilers for C library
           - { platform: linux-arm64       , cc: aarch64-linux-gnu-gcc           , ar: aarch64-linux-gnu-ar   }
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -135,26 +135,41 @@ jobs:
...
 
     - name: Build C library (CMake)
       if: ${{ !matrix.use-cross }}
-      run: |-
+      run: |
         cmake -S lib -B build \
           -DBUILD_SHARED_LIBS=OFF \
           -DCMAKE_BUILD_TYPE=Debug \
-          -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
+          -DCMAKE_C_COMPILER=clang \
```
ObserverOfTime, line 159: Using clang to avoid an [incorrect GCC diagnostic](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83584).

ObserverOfTime:
Is this supposed to be used anywhere?
https://github.com/tree-sitter/tree-sitter/blob/178c5d93f4ca8c962dc977365148148edbce1f24/lib/src/wasm_store.c#L162

ObserverOfTime:
I don't understand what's wrong with the linker…

amaanq:
> Is this supposed to be used anywhere?
> 
> https://github.com/tree-sitter/tree-sitter/blob/178c5d93f4ca8c962dc977365148148edbce1f24/lib/src/wasm_store.c#L162

:thinking: don't think so, and looks like it was added in 2022, so my guess is it's a remnant of the code from back then and can be removed.

lib/src/tree_cursor.c
```diff
@@ -253,7 +256,11 @@ bool ts_tree_cursor_goto_last_child(TSTreeCursor *self) {
         return false;
     }
   }
-  return false;
+#ifdef _MSC_VER
+  __assume(0);
+#else
+  __builtin_unreachable();
+#endif
```
amaanq, line 263: Can we revert this and ignore the warning MSVC unnecessarily complains about?

ObserverOfTime: I mean this is technically an optimisation.

amaanq: a compiler from this millennia would know that this is unreachable code regardless

BobbieCarroll:
Redownload for a interrupt for the first time you have the right access to
my account and the app says it's not working with me and I'm just not
interested in the fact of other stuff.

PR #3783, opened 2024-10-13
revert: "Memory errors in wasm_store"

ObserverOfTime:
This reverts commit 178c5d93f4ca8c962dc977365148148edbce1f24.


clason:
They _will_ be in v26, though, which is released next week, so keep that in mind.

ObserverOfTime:
It will be unreverted with the update. ;)

PR #3791, opened 2024-10-15
fix(cli): check more lines in grammar.json loader

clason:
# Problem

The assumption that `grammar.json` contains the language name in the first three lines is wrong, since a sizable number of parsers store an inherited language name character by character _before_ the language name. This breaks building these parsers with `tree-sitter build`.

# Solution

Raise the number of lines to 20, which is still O(1) with respect to grammar size. This can be dropped again once all(!) grammars are regenerated so the inherited language is stored _after_ the language name.

Fixup for #3776

@amaanq

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -1252,9 +1252,10 @@ impl Loader {
             format!("Failed to open grammar.json at {}", grammar_path.display())
         })?;
 
-        let first_three_lines = BufReader::new(file)
+        let first_lines = BufReader::new(file)
             .lines()
-            .take(3)
+            // todo(clason): reduce to 3 after all parsers are fixed
+            .take(20)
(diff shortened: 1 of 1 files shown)


amaanq:
The only offender is `glimmer_javascript` and `glimmer_typescript`, which are by the same author & which both have PRs up to fix them, so I don't think this is worth it.

amaanq:
*If* they're not merged by the time a release is due, we can get this in, no point atm since only master is affected.

clason:
You are a bit too cavalier about breaking other people's projects that provide value to tree-sitter, you know.

amaanq:
2/346 breaking is not a strong argument...

clason:
It is if it's your project (and don't forget that the number was much higher -- and a bunch of grammars are _still_ broken **including with the last release**).

PR #3797, opened 2024-10-17
Fix refcounting in `Parser::language()` getter

Mrmaxmeier:
`ts_parser_language` doesn't do any refcounting, so we can't return the resulting pointer as an owned `Language` object (or would have to clone the LanguageRef's target to keep API compatibility).


amaanq:
Sorry, just saw this. Do you mind writing up a minimal reproducer?

PR #3798, opened 2024-10-17
Remove unnecessary referencing/dereferencing from build_parse_table

414owen:
These were probably optimized away, and in any case are only run once, per CLI run, but may as well remove them.


amaanq:
thanks!

PR #3799, opened 2024-10-17
feat: update python bindings

ObserverOfTime:
- [x] Drop Python 3.9 support
- [x] Support [free-threaded](https://docs.python.org/3.13/howto/free-threading-extensions.html) build
- [x] Migrate existing bindings
- [x] Wait for py-tree-sitter 0.24

### DIFF
--- a/cli/src/templates/py-binding.c
+++ b/cli/src/templates/py-binding.c
@@ -8,6 +8,13 @@ static PyObject* _binding_language(PyObject *Py_UNUSED(self), PyObject *Py_UNUSE
     return PyCapsule_New(tree_sitter_LOWER_PARSER_NAME(), "tree_sitter.Language", NULL);
 }
 
+static struct PyModuleDef_Slot slots[] = {
+#ifdef Py_GIL_DISABLED
+    {Py_mod_gil, Py_MOD_GIL_NOT_USED},
+#endif
+    {0, NULL}
+};
+
 static PyMethodDef methods[] = {
     {"language", _binding_language, METH_NOARGS,
      "Get the tree-sitter language for this grammar."},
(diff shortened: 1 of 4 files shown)


cli/src/templates/py-binding.c
```diff
@@ -8,6 +8,13 @@ static PyObject* _binding_language(PyObject *Py_UNUSED(self), PyObject *Py_UNUSE
     return PyCapsule_New(tree_sitter_LOWER_PARSER_NAME(), "tree_sitter.Language", NULL);
 }
 
+static struct PyModuleDef_Slot slots[] = {
+#if PY_MINOR_VERSION >= 13
+    {Py_mod_gil, Py_MOD_GIL_NOT_USED},
```
amaanq, line 13: Did you try this out without `Py_GIL_DISABLED` defined? It doesn't fail to compile if the system Python version is 3.13, because of this part in setup.py:

```py
if limited_api := not get_config_var("Py_GIL_DISABLED"):
    macros.append(("Py_LIMITED_API", "0x030A0000"))
```

The error I get is:

```
      bindings/python/tree_sitter_javascript/binding.c:13:6: error: ‘Py_mod_gil’ undeclared here (not in a function)
         13 |     {Py_mod_gil, Py_MOD_GIL_NOT_USED},
            |      ^~~~~~~~~~
      bindings/python/tree_sitter_javascript/binding.c:13:18: error: ‘Py_MOD_GIL_NOT_USED’ undeclared here (not in a function)
         13 |     {Py_mod_gil, Py_MOD_GIL_NOT_USED},
            |                  ^~~~~~~~~~~~~~~~~~~
```

If it's set to `0x030D0000` then it works, but Python bindings will *not* build with this change, so this should be updated.

amaanq (changes requested):

amaanq (approved):

PR #3803, opened 2024-10-19
feat!: bump internal abi to 15

amaanq:


lib/include/tree_sitter/api.h
```diff
@@ -26,13 +26,13 @@ extern "C" {
...
  */
-#define TREE_SITTER_LANGUAGE_VERSION 14
+#define TREE_SITTER_LANGUAGE_VERSION 15
 
 /**
  * The earliest ABI version that is supported by the current version of the
  * library.
  */
-#define TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION 13
+#define TREE_SITTER_MIN_COMPATIBLE_LANGUAGE_VERSION 14
```
clason, line 35: Is this necessary? A number of parsers in the wild are still on ABI 13.

amaanq: ABI 14 is more than two years old at this point

clason: So? What harm does this do?

amaanq: Ok, we can keep the runtime support for 13, but I'll remove allowing the CLI to generate ABI 13. If the compatibility code between 14 and 15 becomes too unwieldy with 13 in the mix later on, I'll revisit removing runtime support then.

clason: Yes, that is a good compromise.

PR #3812, opened 2024-10-20
fix(dsl): fix types for RuleBuilder

### LINKED ISSUE #3811: RuleBuilder in dsl.d.ts seems incorrect
### Problem

According to https://github.com/tree-sitter/tree-sitter/blob/60c5057617c0b2ca3804a362594c484728e3e5bf/cli/npm/dsl.d.ts#L43-L46, `RuleBuilder` has a required second parameter, however at https://github.com/tree-sitter/tree-sitter-c/blob/79284a14a0d7e4495b095b38acdd8a27acbe870e/grammar.js#L1404, `content` is a `RuleBuilder<string>` and is used with only one parameter. This result in an error when checking the file with TypeScript.

Should the type be updated to
```ts
type RuleBuilder<RuleName extends string> = (
  $: GrammarSymbols<RuleName>,
  previous?: Rule,
) => RuleOrLiteral;
```
or is this parameter actually required, and this is a bug in tree-sitter-c?

### Steps to reproduce

With VS Code installed:
```
git clone https://github.com/tree-sitter/tree-sitter-c.git
code tree-sitter-c
```
Then open grammar.js and look for errors.

### Expected behavior

TypeScript finds no errors when checking the file.

tree-sitter 0.24.3

### Operating system/version

Windows 11 build 22631

joelspadin:
The second parameter to RuleBuilder should be optional. Fixes #3811.

### DIFF
--- a/cli/npm/dsl.d.ts
+++ b/cli/npm/dsl.d.ts
@@ -42,7 +42,7 @@ type GrammarSymbols<RuleName extends string> = {
 
 type RuleBuilder<RuleName extends string> = (
   $: GrammarSymbols<RuleName>,
-  previous: Rule,
+  previous?: Rule,
 ) => RuleOrLiteral;
 
 type RuleBuilders<


amaanq:
nice, thank you!

PR #3816, opened 2024-10-22
feat(loader): implement `language_for_configuration`

bm-w:
While diagnosing a breaking change introduced by v0.24 (specifically #3700; I left a [comment](/tree-sitter/tree-sitter/pull/3700#issuecomment-2427922990) there), I found that there is actually no way to directly load a `Language` from a given `LanguageConfiguration` (such as one returned from `Loader::get_all_language_configurations`). This PR trivially remedies that.

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -602,6 +602,13 @@ impl Loader {
         }
     }
 
+    pub fn language_for_configuration(
+        &self,
+        configuration: &LanguageConfiguration,
+    ) -> Result<Language> {
+        self.language_for_id(configuration.language_id)
+    }
+
     fn language_for_id(&self, id: usize) -> Result<Language> {
         let (path, language, externals) = &self.languages_by_id[id];
         language


cli/loader/src/lib.rs
```diff
@@ -600,6 +600,13 @@ impl Loader {
         }
     }
 
+    pub fn language_for_configuration(
+        &self,
+        configuraiton: &LanguageConfiguration,
+    ) -> Result<Language> {
+        self.language_for_id(configuraiton.language_id)
```
ObserverOfTime, line 607:
```suggestion
        configuration: &LanguageConfiguration,
    ) -> Result<Language> {
        self.language_for_id(configuration.language_id)
```

amaanq:
What's the use case for this? It sounds like whatever grammar(s) you're using should be migrating to the new config

bm-w:
Yes, they should, and maybe they could, and I wish that they would, but if wishes were horses then beggars would ride, and all that. In the meantime something as trivial as this would help cross a few gaps. In my particular case I’m stuck on some older grammars that I don’t control, and while forking is an option it’d be nice if there was a single simple solution such as the one I’m proposing.

PR #3818, opened 2024-10-22
test: run cmake tests with ctest

ObserverOfTime:
This does add a lot more lines to `CMakeLists.txt`. Maybe we should only do this for bindings?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,8 +99,8 @@ test:
 	cargo xtask generate-fixtures
 	cargo xtask test
 
-test_wasm:
-	cargo xtask generate-fixtures-wasm
+test-wasm:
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
I'll just put it out there that I'd really like to minimize the amount of CMake code in this repo, for both bindings and the core. Do we need to be able to run these tests and commands through CMake?

dundargoc:
I don't quite understand the reasoning for this if I'm being honest. Cmake makes sense for the library so consumers can easily use it (because if TS doesn't use cmake then they will necessarily need to roll out their own version if they want to support windows). But from a quick glance this seems more for TS:s own testing which doesn't require cmake?

Issue #281, opened 2019-02-17
Document external scanners


fcurts:
Please document the API contract of external scanners. They seem to be required for parsing most general-purpose languages. Even after looking at other scanners, I haven't been able to figure out how exactly they work (i.e., I can't get my scanner to work correctly).

fcurts:
My biggest question is what `mark_end` is for. Initially I thought that `mark_end` followed by `advance` could be used for multi-character lookahead, but this doesn't seem to be the case.
So far I haven't found a way to write a grammar for my language that produces the desired parse tree without looking ahead multiple characters in the scanner (which doesn't seem to be supported).

Aerijo:
You can lookahead multiple characters by calling `advance`, like in [existing external scanners](https://github.com/tree-sitter/tree-sitter-html/blob/22e568ea752624791d795746c4c027969a65f182/src/scanner.cc#L119).

fcurts:
@Aerijo By "looking ahead multiple characters" I mean that when I return `true`, none of these characters have been consumed. Is this possible, and if so, how?

Aerijo:
If you 
- advance several times, 
- mark the end, 
- give it a node type, and
- return `true` from the scanning function

it should add the node to the tree and stay where it advanced to. If you return false, it will go back to where it started because the scan is considered a failure.

Do you have a link to your grammar I can look at to see what's happening?

fcurts:
What you describe is not what I'm looking for (and doesn't seem to require calling `mark_end`).

I'd like to consume some characters, then look ahead multiple characters without consuming them, then return `true`, and it should proceed from the position after the last consumed character.

What I'm trying to parse is Swift 5 string literals with custom string delimiters (https://github.com/apple/swift-evolution/blob/master/proposals/0200-raw-string-escaping.md#customized-escape-delimiters). Unfortunately, I can't share my grammar.

maxbrunsfeld:
Yeah this is exactly what `mark_end` does, and we use it for this purpose in several existing scanners.

The default behavior of an external scanner is that, when you return `true`, it will create a token whose type is given by `lexer->result_symbol`, and consume all of the characters for which you called `advance`. Sometimes you may want to look ahead at some characters without necessarily consuming characters. You can do this by calling `lexer->mark_end`.

If you call `martk_end` and then later call `lexer->advance` one or more times, the resulting token will *not* consume those characters, unless you call `mark_end` again later.

maxbrunsfeld:
[...]

PR #3830, opened 2024-10-26
Add overview to external scanner documentation

dbaynard:
The documentation in 190b219be5368a04200668ea469bc4fc5bf98357 is helpful yet would benefit from a high level overview, as provided in #281. That overview is presented here.

I'm sure there are further improvements that could be made. If you are minded to merge, I'd recommend a technical reviewer check the language is accurate, yet further improvements not block this PR.

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -833,6 +833,8 @@ Many languages have some tokens whose structure is impossible or inconvenient to
 
 Tree-sitter allows you to handle these kinds of tokens using *external scanners*. An external scanner is a set of C functions that you, the grammar author, can write by hand in order to add custom logic for recognizing certain tokens.
 
+#### `src/scanner.c` API
+
 To use an external scanner, there are a few steps. First, add an `externals` section to your grammar. This section should list the names of all of your external tokens. These names can then be used elsewhere in your grammar.
 
 ```js
@@ -866,8 +868,9 @@ enum TokenType {
 ```
 
 Finally, you must define five functions with specific names, based on your language's name and five actions: *create*, *destroy*, *serialize*, *deserialize*, and *scan*.
+The first four concern how tree-sitter operates the scanner, the last how the scanner tokenizes the input stream.
 
-#### Create
+##### Create
 
 ```c
 void *tree_sitter_my_language_external_scanner_create(void) {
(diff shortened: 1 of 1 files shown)


docs/section-3-creating-parsers.md
```diff
@@ -919,6 +922,23 @@ It is good practice to explicitly erase your scanner state variables at the star
 
 #### Scan
 
+This function is responsible for recognizing external tokens. It should return `true` if a token was recognized, and `false` otherwise. It is called with a "lexer" struct.
```
WillLillis, line 925: Question for the maintainers:

Should "lexer" be inside quotes as it currently stands, or inside back ticks?

dbaynard: Thanks for the quick response.

I kept it in quotes here as it seems like a categorization rather than a name. Later I use backticks to make clear it matches the `lexer` argument to `tree_sitter_my_language_external_scanner_scan`.

Alexizx00789999 (approved):

amaanq:
I added the relevant changes and added you as a co-author in #4035 @dbaynard, the docs had been rewritten quite a bit so I found that easier to do, thank you for the PR though!

dbaynard:
I'm glad you've merged changes, though I feel you've lost something in doing so. The point was that an overview was missing — in extending the overview from four short lines to four chunky paragraphs (using words like "optionally") I fear it defeats the purpose of this PR.

Thankfully, I've had to put off my tree sitter scanner writing, and I've completely forgotten how to do it. When I return to it, if I think the issue remains, I'll open a new PR.

amaanq:
Well, yes, you don't have to call `mark_end` at all - it's optional, but gives you some benefits. I'm not sure what's so clunky about that.

dbaynard:
You haven't asked a question but I'll answer the one that is implied: in what ways do I think the merged text is worse than the version here?

1.  The use of "typically" gives the common case. It's important to keep this simple, in a summary, by which I mean do not introduce modalities. The merged version adds too much detail, thereby complicating understanding. 
2.  The merged text uses scare quotes, around "peek ahead," and this grammar use usually indicates a better choice of words is possible. By contrast, the changes in this PR introduce a short standalone paragraph explaining the logic, using clear language. (It also has more detail in this non-summary part, which would benefit those new to writing scanners, i.e. the audience for the documentation.)
3.  The merged version includes typos and inconsistent punctuation. It would have benefitted from a review.

Documentation is best written by someone who understands the project and who understands what it's like _not_ to understand the project. I found this aspect unnecessarily tough to understand, _because the documentation lacked the latter insight_. Hence the PR.

That aside, I'm happy to hear there have been larger documentation changes. Good! I look forward to reading them next time I'm working on tree-sitter projects.

PR #3831, opened 2024-10-26
feat(generate)!: use `regex_syntax::Hir` for expanding regexes

amaanq:
Supersedes #3404
Supersedes #3778

### Problem

Currently, the way tree-sitter processes regexes is by hard-coding the unicode category and property tables. This is problematic for several reasons, since *we* have to constantly update these tables every time a new unicode version is released, and they are *massive* files. We are also relying on `regex_syntax::Ast`, which means we process a lot of the regex components ourselves.

### Solution

Instead of using `regex_syntax::Ast`, we can use `regex_syntax::Hir` and let the crate parse our regex into a more higher level IR. This abstracts a lot of the manual processing we were doing before, and gives us the character ranges we need for the most part. Another benefit is that this removes the need for hard-coding the unicode tables in the generate crate, which reduces the binary size by **7 megabytes**! Also, we no longer need to keep a script for generating the unicode categories, so we have *no more* bash scripts; bye bye `script` folder.

That's not to say the translation was a smooth process, there were a couple of edge cases worth pointing out. For one, `regex_syntax::Hir` uses `unicode` by default, which I ran into issues with when it came to simple character class escapes (`\w`, `\s`, and `\d`). The problem is that it gives us the *full* unicode range for these class escapes, when oftentimes that is not the intention by the grammar author. So, we could disable unicode, but the problem we run into then is if any unicode classes are used, the parser fails, and the max range of byte classes is `0xFF`, which breaks the assumption that it's always `0x10FFFF` (aka `char::MAX`). So instead, we perform a simple replacement before processing regexes that replaces these 3 **simple** character classes with an alternation. So, `\s` becomes `[\t\r ]`, `\d` becomes `[0-9]`, etc and we keep unicode on. This leads to practically no changes across the majority of parsers, with some having *very* minor changes likely due to the unicode range(s) changing from v15 to v16.

Secondly, and weirdly, the letter `s` alone generates three characters in case-insensitive mode: `s`, `S`, and `ſ` (long S). I found this to be strange, and practically almost never intended, so this long S is filtered out if detected.


cli/generate/src/prepare_grammar/expand_tokens.rs
```diff
@@ -114,8 +95,25 @@ impl NfaBuilder {
...
-                let ast = parse::Parser::new().parse(s)?;
-                self.expand_regex(&ast, next_state_id, f.contains('i'))
+                // With unicode enabled, `\w`, `\s` and `\d` expand to character sets that are much
+                // larger than intended, so we replace them with the actual
+                // character sets they should represent. If the full unicode range
+                // of `\w`, `\s` or `\d` are needed then `\p{L}`, `\p{Z}` and `\p{N}` should be
+                // used.
+                let s = s
+                    .replace(r"\w", r"[0-9A-Za-z_]")
+                    .replace(r"\s", r"[\t-\r ]")
```
ObserverOfTime, line 105: This does not match [JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes#types). It should be:
`[\f\n\r\t\v\u0020\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`

amaanq: It never matched JS in the first place.

PR #3833, opened 2024-10-27
feat: add the ability to specify a custom decode function

### LINKED ISSUE #717: Support custom text decoding function

amaanq:
Supersedes #2400
Closes #717

### DIFF
--- a/lib/binding_rust/bindings.rs
+++ b/lib/binding_rust/bindings.rs
@@ -35,9 +35,13 @@ pub struct TSQueryCursor {
 pub struct TSLookaheadIterator {
     _unused: [u8; 0],
 }
+pub type DecodeFunction = ::core::option::Option<
+    unsafe extern "C" fn(string: *const u8, length: u32, code_point: *mut i32) -> u32,
+>;
 pub const TSInputEncodingUTF8: TSInputEncoding = 0;
 pub const TSInputEncodingUTF16LE: TSInputEncoding = 1;
 pub const TSInputEncodingUTF16BE: TSInputEncoding = 2;
+pub const TSInputEncodingCustom: TSInputEncoding = 3;
 pub type TSInputEncoding = ::core::ffi::c_uint;
 pub const TSSymbolTypeRegular: TSSymbolType = 0;
 pub const TSSymbolTypeAnonymous: TSSymbolType = 1;
@@ -71,6 +75,7 @@ pub struct TSInput {
         ) -> *const ::core::ffi::c_char,
     >,
     pub encoding: TSInputEncoding,
+    pub decode: DecodeFunction,
 }
 #[repr(C)]
 #[derive(Debug, Copy, Clone)]
@@ -212,7 +217,7 @@ extern "C" {
     ) -> *mut TSTree;
 }
 extern "C" {
-    #[doc = " Use the parser to parse some source code and create a syntax tree, with some options.\n\n See [`ts_parser_parse`] for more details."]
+    #[doc = " Use the parser to parse some source code and create a syntax tree, with some options.\n\n See [`ts_parser_parse`] for more details.\n\n See [`TSParseOptions`] for more details on the options."]
     pub fn ts_parser_parse_with_options(
         self_: *mut TSParser,
         old_tree: *const TSTree,
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -210,6 +210,12 @@ type ParseProgressCallback<'a> = &'a mut dyn FnMut(&ParseState) -> bool;
 /// A callback that receives the query state during query execution.
 type QueryProgressCallback<'a> = &'a mut dyn FnMut(&QueryCursorState) -> bool;
 
+pub trait Decode {
+    /// A callback that decodes the next code point from the input slice. It should return the code
+    /// point, and how many bytes were decoded.
+    fn decode(bytes: &[u8]) -> (i32, u32);
+}
+
 /// A stateful object for walking a syntax [`Tree`] efficiently.
 #[doc(alias = "TSTreeCursor")]
 pub struct TreeCursor<'cursor>(ffi::TSTreeCursor, PhantomData<&'cursor ()>);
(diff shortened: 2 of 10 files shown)


lib/binding_rust/bindings.rs
```diff
@@ -185,6 +189,14 @@ extern "C" {
         input: TSInput,
     ) -> *mut TSTree;
 }
+extern "C" {
+    pub fn ts_parser_parse_custom(
```
WillLillis, line 193: Missing docs here?

lib/include/tree_sitter/api.h
```diff
@@ -271,6 +280,13 @@ TSTree *ts_parser_parse(
   TSInput input
 );
 
+TSTree *ts_parser_parse_custom(
```
WillLillis, line 283: If docs are added to the location of my above comment, they should probably be added here too.

lib/binding_rust/bindings.rs
amaanq: outdated w changes, will address adding docs after 3843

lib/src/lexer.c
```diff
@@ -87,7 +89,11 @@ static void ts_lexer__get_lookahead(Lexer *self) {
     self->input.encoding == TSInputEncodingUTF8 ? ts_decode_utf8 :
     self->input.encoding == TSInputEncodingUTF16LE ? ts_decode_utf16_le : ts_decode_utf16_be;
 
-  self->lookahead_size = decode(chunk, size, &self->data.lookahead);
+  if (self->input.encoding == TSInputEncodingCustom) {
+    self->lookahead_size = self->input.decode(self->input.payload, chunk, size, &self->data.lookahead);
+  } else {
+    self->lookahead_size = decode(chunk, size, &self->data.lookahead);
+  }
```
maxbrunsfeld, line 96: Can we unify this conditional with the one above, that initially defines `decode`?

amaanq: I don't think we can, since in the user-specified `decode` function we accept a payload argument that allows us to store arbitrary data (one use case being FFI), but the other decode functions don't have this argument

maxbrunsfeld: Oh, interesting, I didn't notice that. What do you think about *not* having that payload argument? It looks like it isn't used in any of your example use cases in `parser_test.rs`, and it kinda makes sense that unicode decoding should probably be a pure function of the bytes.

In the context of an FFI binding, I don't think it would ever be a good idea to have a payload. We definitely don't want to be doing FFI calls for every single character.

lib/include/tree_sitter/api.h
```diff
@@ -86,6 +97,7 @@ typedef struct TSParseState {
 typedef struct TSParseOptions {
   void *payload;
   bool (*interrupt_callback)(TSParseState *state);
+  DecodeFunction decode_callback;
```
maxbrunsfeld, line 100: Why is the `DecodeFunction` included in both `TSInput` and `TSParseOptions`? Could it just be a property of the input?

lib/src/lexer.c
maxbrunsfeld: I see your point about not wanting to expose `extern "C"` functions to the Rust API. But since you shouldn't really need a payload at the Rust layer, could you avoid the need for that indirection by requiring that the `Encoding` be specified *statically* via a trait?

Another benefit to this is that I think it's pretty confusing for `decode` and `read` to take the same payload. This way, this custom `decode` doesn't have anything to do with any concept of a payload.

```rust
trait Encoding {
    fn decode(bytes: &[u8]) -> (i32, u32);
}

impl Parser {
    fn parse_custom_encoding<T, F, E>(
        &mut self,
        callback: &mut F,
        old_tree: Option<&Tree>
    )
    where
        E: Encoding,         // <- new type parameter for the custom encoding
        T: AsRef<[u8]>,
        F: FnMut(usine, Point) -> T,
    {
        // At compile time, create a C-compatible callback that calls the custom `decode` method.
        extern "C" fn decode_fn<E: Encoding>(data: *const u8, len: usize, char_len: *mut u32) -> i32 {
            let (c, len) = T::decode(std::slice::from_raw_parts(data, len));
            if let Some(char_len) = char_len.as_mut() {
                *char_len = len;
            }
            c
        }

        // ...

        let c_input = ffi::TSInput {
            payload: core::ptr::addr_of_mut!(payload).cast::<c_void>(),
            read: Some(read::<T, F>),
            encoding: ffi::TSInputEncodingCustom,

            // Use this custom decode callback
            decode: decode_fn::<T>(),
        };

        // ...
    }
}

```

lib/binding_rust/lib.rs
```diff
@@ -752,7 +767,7 @@ impl Parser {
         old_tree: Option<&Tree>,
         options: Option<ParseOptions>,
     ) -> Option<Tree> {
-        type Payload<'a, F, T> = (&'a mut F, Option<T>);
+        type Payload<'a, F, T> = (&'a mut F, Option<DecodeCallback<'a>>, Option<T>);
```
maxbrunsfeld, line 770: Overall, I don't get why the decode callback would be part of the Options. It doesn't really make sense to say `parse_utf16` with a custom payload. Same thing for the other `parse` APIs, which are otherwise documented as taking UTF8 input.

maxbrunsfeld (approved):

clason:
So that function is now mandatory? (This is a breaking change.)

What should I do if I _don't_ want to specify a custom decode function?

PR #3835, opened 2024-10-27
Replace nerd font symbols with Unicode symbols, and count skipped tests correctly

keidax:
I noticed the test attributes introduced in https://github.com/tree-sitter/tree-sitter/pull/3060 use Unicode PUA code points. These require Nerd Fonts to render properly. I propose replacing these code points with more compatible Unicode symbols. Here's how this looks in a couple different terminal emulators without Nerd Fonts installed:
| Before PR | After PR |
| ------------- | ------------- |
| image| image  |
| image | image |

I also found the 🖵️ symbol a bit unclear when indicating the `platform` attribute, so I replaced that with a simple `P`.

### DIFF
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -358,7 +358,7 @@ fn run_tests(
 
             if attributes.skip {
                 println!(
-                    "{:>3}.  {}",
+                    "{:>3}. ⌀ {}",
                     opts.test_num,
                     paint(opts.color.then_some(AnsiColor::Yellow), &name),
                 );
@@ -367,7 +367,7 @@ fn run_tests(
 
             if !attributes.platform {
                 println!(
-                    "{:>3}.  {}",
+                    "{:>3}. ⌀ {}",
                     opts.test_num,
                     paint(opts.color.then_some(AnsiColor::Magenta), &name),
                 );
@@ -387,7 +387,7 @@ fn run_tests(
                 if attributes.error {
                     if tree.root_node().has_error() {
                         println!(
-                            "{:>3}.  {}",
+                            "{:>3}. ✓ {}",
                             opts.test_num,
                             paint(opts.color.then_some(AnsiColor::Green), &name)
                         );
@@ -418,7 +418,7 @@ fn run_tests(
                             ));
                         }
                         println!(
-                            "{:>3}.  {}",
+                            "{:>3}. ✗ {}",
                             opts.test_num,
                             paint(opts.color.then_some(AnsiColor::Red), &name)
                         );
--- a/cli/src/test.rs
+++ b/cli/src/test.rs
@@ -362,6 +362,7 @@ fn run_tests(
                     opts.test_num,
                     paint(opts.color.then_some(AnsiColor::Yellow), &name),
                 );
+                opts.test_num += 1;
                 return Ok(true);
             }
 
@@ -371,6 +372,7 @@ fn run_tests(
                     opts.test_num,
                     paint(opts.color.then_some(AnsiColor::Magenta), &name),
                 );
+                opts.test_num += 1;
                 return Ok(true);
             }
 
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
`P` is not clear either.

keidax:
Yeah, I think `🞜` and `⭙` are not particularly intuitive. I considered `🖵` for `platform` but it usually renders as 2 cells wide, while the rest are a single cell. Another choice is using `⌀` for a test skipped by either `:skip` or `:platform`, since they are still distinguishable by color.

If emoji would be acceptable, I might suggest something like this:
image
But that's a pretty big visual change, and the emoji don't have consistent widths on different terminal emulators.

ObserverOfTime:
Emojis are not acceptable.

kylegoetz:
This makes a lot of sense now! Working on a separate PR for the test code, I legitimately thought the icons that were other than standard pass/fail were meant to be nondescript boxes!

kylegoetz:
Does it make sense that you need a different icon for "skipped because marked 'skip'" versus "skipped because wrong platform"? Why not just use the same icon for both but color differently or something (I'm not even convinced someone running tests needs to know why a test was skipped; just that it was skipped)?

I'm not sure there is a visually useful icon that indicates "skipped because it's not the right platform" so why try to find one? No matter what, the user is going to have to go to the TS manual/website and look for a writeup about what this color or icon means. Is there even any writeup that explains it?

keidax:
@kylegoetz Thank you for explaining your thoughts on this -- that helped clarify my approach.

I made some new changes and rebased, so that:

#### The same skip icon is used for any skipped test.
I agree that trying to encode the skip reason in an icon is not useful. I'm leaning towards keeping the distinct colors for now, so you can tell at a glance if e.g. a large swath of tests are skipped because you're using Windows.

#### There's no special icon for "passed with the :error attribute". 
Anyone encountering the previous icon for the first time would likely need to look into the specific test or documentation to understand the icon's purpose.

I considered printing ✗ for a passing test with `:error`, but decided against it. The test overview doesn't need to be concerned with what conditions caused a test to pass, only if it was a pass or fail. Plus, ✗ on pass is going to be confusing for anyone with `NO_COLOR=1`.

#### Skipped tests don't have a test number.
Probably just an oversight before

How this looks now:
image

keidax:
@ObserverOfTime @amaanq Please let me know if there are any further changes needed to move this forward.

cli/src/test.rs
```diff
@@ -358,17 +358,15 @@ fn run_tests(
 
             if attributes.skip {
                 println!(
-                    "{:>3}.  {}",
-                    opts.test_num,
+                    "     ⌀ {}",
```
amaanq, line 361: Why remove the format specifier?

keidax: `test_num` isn't incremented for skipped tests, so currently if several tests are skipped in a row, it looks like
```
      1.  skipped test 1
      1.  skipped test 2
      1.  skipped test 3
      1.  skipped test 4
      1.  skipped test 5
      1.  skipped test 6
      1. ✓ test that runs 1
      2. ✓ test that runs 2
```

I think it looks better to hide the test number in this case.

If you'd prefer to have the test number incremented for all tests, including skipped ones, I can make that change.

amaanq: ah that's a good catch - let's have the test number incremented always, because we can parse tests with `tree-sitter parse -n <test_number>`, so knowing the number always is helpful

kylegoetz: Commenting to follow because the changes to test reports have made my json report PR no longer viable and I'll have to update its behavior once this is merged

cli/src/test.rs
```diff
@@ -600,8 +600,6 @@ fn run_tests(
                 }
             }
 
-            opts.test_num += skipped_tests;
```
keidax, line 603: I removed this because it was double-counting skipped tests

keidax (commented):
Tests numbers should now be consistent, whether skipped or not.

I manually confirmed these commands behave as expected:
`tree-sitter parse -n <test number>`
`tree-sitter test --include <pattern> --update`
`tree-sitter test --exclude <pattern> --update`

I would like to add unit tests for this behavior, but that's probably beyond my skill with Rust atm

amaanq (approved):

amaanq:
thanks @keidax

PR #3843, opened 2024-10-29
feat: add options to parsing and query execution

### LINKED ISSUE #2541: Add `ts_parser_parse_interruptible` C API call
# Motivation

Currently Tree-sitter is able to interrupt its parsing process in several ways:
* By a cancellation flag.
* By a timeout.

But there may be other cases:
* Someone wants to interrupt early on a first error: #2289 
* Interrupt due to a decoding error in an input text _(that's a one of reasons for `TSParserInput` struct mentioned bellow)_.

All of such cases make sense but adding new and new API calls increases Tree-sitter's API surface and adds additional logic that triggers every time on an interruption requirement check point _(for now every 100 parser steps)_.

The `ts_parser_parse_interruptible` C API call would receive an additional callback parameter that would check any interruption condition implemented on _a caller side_.

If to add the `ts_parser_parse_interruptible` C API call it would be possible:
* To interrupt parsing on any condition that would be managed on the caller side.
* To deprecate the next C API calls because all of their functionality would be implementable externally in a check callback:
  * `ts_parser_cancellation_flag`
  * `ts_parser_set_cancellation_flag`
  * `ts_parser_timeout_micros`
  * `ts_parser_set_timeout_micros`
* To minimize a condition check logic inside of Tree-sitter's C core.
* To make Tree-sitter core [_optionally_ dependable](https://github.com/tree-sitter/tree-sitter/pull/2540#issue-1858123862) from:
  * The time `<time.h>` header.
  * The `clock_gettime` libc call [on Linux](https://github.com/tree-sitter/tree-sitter/pull/2515#issue-1854946650) and similar on other platforms.

To cover the #2289 case Tree-sitter would start to manage a struct `TSParserState` with some flags and pass it as a parameter to the callback received by the `ts_parser_parse_interruptible` API call. Such struct would be easily extendable in the future by adding more flags if it would be needed. Currently there would be a one field called an `error_count`.

```c
struct TSParserState {
  uint32_t error_count,
}
```
---
## Alternative

I'd even suggest to go further and implement a one universal `ts_parser_parse_ex` API call that would look like:
```c
TSTree *ts_parser_parse_ex(
  TSParser *self,
  const TSTree *old_tree,
  TSParserInput *input,
  TSParserConfig *config
);
```
The `config` parameter may be `NULL` what would lead to a default parser behavior.

The `TSParserInput` would be a new struct that supersedes `TSInput` struct but would improve a design of how input decoding works. _(Will be described separately)._

The `TSParserConfig` struct would look like:
```c
struct TSParserConfig {
  bool (*interrupt_check)(TSParserState *parser_state);
}
```

This would allow to have a one universal parsing call with an access to the parsing logic and all other simpler calls like `ts_parser_parse`, `ts_parser_parse_string` and `ts_parser_parse_string_encoding` would be just thin wrappers on top of it.

_If to go in this way then adding the  `ts_parser_parse_interruptible` API call wouldn't be needed because its functionality would be a part of `TSParserConfig` struct._

amaanq:
Closes #2541
Would like #3833 to extend upon this

### Problem

Currently, users have a hard time having manual control over when parsing or query execution is canceled. The API does offer a few ways to control this, typically by setting a timeout in microseconds or by setting a cancellation flag. However, this approach is very rigid, doesn't give the users flexibility to easily change this on the fly at runtime, and ties us to OS-specific time headers.

### Solution

In the core library, two new functions, `ts_parser_parse_with_options` & `ts_query_cursor_exec_with_options`, were added. These functions take in the same parameters as their -`_with_options` variants, plus an additional `parse_options` parameter of type `TSParseOptions` for the parse function, and a `query_options` parameter of type `TSQueryCursorOptions *`.

Currently, these options consist of a `payload` field, and an `interrupt_callback` field. If the interrupt callback is *not* null, it is invoked to check whether or not to halt parsing or query execution. The callback is to be set by the user, and takes in one parameter, `state`, of type `TSParseState`/`TSQueryCursorState`, which currently only has a `payload` field (which makes working with FFI code easier). This type can easily be extended upon, in case we want to add additional data later on.

The plan is to deprecate the `timeout_micros` and `cancellation_flag` functions for the next version, and then eventually get rid of these functions + the platform-specific time code (`clock.h`) in the following version.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -116,6 +116,82 @@ pub struct Parser(NonNull<ffi::TSParser>);
 pub struct LookaheadIterator(NonNull<ffi::TSLookaheadIterator>);
 struct LookaheadNamesIterator<'a>(&'a mut LookaheadIterator);
 
+/// A stateful object that is passed into a [`ParseProgressCallback`]
+/// to pass in the current state of the parser.
+pub struct ParseState(NonNull<ffi::TSParseState>);
+
+impl ParseState {
+    #[must_use]
+    pub fn current_byte_offset(&self) -> usize {
+        unsafe { self.0.as_ref() }.current_byte_offset as usize
+    }
+}
+
+/// A stateful object that is passed into a [`QueryProgressCallback`]
+/// to pass in the current state of the query execution.
+pub struct QueryCursorState(NonNull<ffi::TSQueryCursorState>);
+
+impl QueryCursorState {
+    #[must_use]
+    pub fn current_byte_offset(&self) -> usize {
+        unsafe { self.0.as_ref() }.current_byte_offset as usize
+    }
+}
+
+#[derive(Default)]
+pub struct ParseOptions<'a> {
+    pub progress_callback: Option<ParseProgressCallback<'a>>,
+}
+
+impl<'a> ParseOptions<'a> {
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[must_use]
+    pub fn progress_callback<F: FnMut(&ParseState) -> bool>(mut self, callback: &'a mut F) -> Self {
+        self.progress_callback = Some(callback);
+        self
+    }
+}
+
+#[derive(Default)]
+pub struct QueryCursorOptions<'a> {
+    pub progress_callback: Option<QueryProgressCallback<'a>>,
+}
+
+impl<'a> QueryCursorOptions<'a> {
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[must_use]
+    pub fn progress_callback<F: FnMut(&QueryCursorState) -> bool>(
+        mut self,
+        callback: &'a mut F,
+    ) -> Self {
+        self.progress_callback = Some(callback);
+        self
+    }
+}
+
+struct QueryCursorOptionsDrop(*mut ffi::TSQueryCursorOptions);
+
+impl Drop for QueryCursorOptionsDrop {
+    fn drop(&mut self) {
+        unsafe {
+            if !(*self.0).payload.is_null() {
+                drop(Box::from_raw(
+                    (*self.0).payload.cast::<QueryProgressCallback>(),
+                ));
+            }
+            drop(Box::from_raw(self.0));
+        }
+    }
+}
+
 /// A type of log message.
 #[derive(Debug, PartialEq, Eq)]
 pub enum LogType {
@@ -125,8 +201,14 @@ pub enum LogType {
 
 type FieldId = NonZeroU16;
 
-/// A callback that receives log messages during parser.
+/// A callback that receives log messages during parsing.
 type Logger<'a> = Box<dyn FnMut(LogType, &str) + 'a>;
+
+/// A callback that receives the parse state during parsing.
+type ParseProgressCallback<'a> = &'a mut dyn FnMut(&ParseState) -> bool;
+
+/// A callback that receives the query state during query execution.
+type QueryProgressCallback<'a> = &'a mut dyn FnMut(&QueryCursorState) -> bool;
 
 /// A stateful object for walking a syntax [`Tree`] efficiently.
 #[doc(alias = "TSTreeCursor")]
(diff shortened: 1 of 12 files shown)


amaanq:
Ok this is now ready. I'm really happy with how the ergonomics for the Rust bindings turned out, and other bindings will have similar mechanisms involved for passing in the interrupt callback. This will also easily extend to passing in the decode callback too in #3833

lib/binding_rust/lib.rs
```diff
@@ -116,6 +116,68 @@ pub struct Parser(NonNull<ffi::TSParser>);
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[must_use]
+    pub fn interrupt_callback<F: FnMut(&QueryCursorState) -> bool + 'a>(
+        mut self,
+        callback: F,
+    ) -> Self {
+        self.interrupt_callback = Some(Box::new(callback));
```
maxbrunsfeld, line 161: Do we need to box here, or can we just store a `&'a mut dyn FnMut() -> bool`

lib/binding_rust/lib.rs
```diff
@@ -116,6 +116,68 @@ pub struct Parser(NonNull<ffi::TSParser>);
...
+
+impl<'a> ParseOptions<'a> {
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[must_use]
+    pub fn interrupt_callback<F: FnMut(&ParseState) -> bool + 'a>(mut self, callback: F) -> Self {
+        self.interrupt_callback = Some(Box::new(callback));
```
maxbrunsfeld, line 140: Do we need to box here, or can we just store a `&'a mut dyn FnMut() -> bool`?

maxbrunsfeld (commented):
This is definitely a good change. I left a few questions.

lib/include/tree_sitter/api.h
```diff
@@ -79,6 +79,15 @@ typedef struct TSInput {
   TSInputEncoding encoding;
 } TSInput;
 
+typedef struct TSParseState {
+  void *payload;
+} TSParseState;
```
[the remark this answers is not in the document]
amaanq: One *suggestion* in the linked issue was knowing the current status of errors detected, but I feel that that's a bit hard to nail down properly since the parser *could* in theory recover with an error-free tree in the event an error was detected. This was more to set us up with a struct that can be amended later on, so that the signature of these callbacks does not change, and it retains ABI compatibility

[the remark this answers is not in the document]
maxbrunsfeld: Ok, that use case does make sense. But what would get passed back in `ParseState`? I would think that you'd want to control that in the caller: like `ParseOptions::new().error_callback(my_callback)`.

lib/include/tree_sitter/api.h
```diff
@@ -79,6 +79,15 @@ typedef struct TSInput {
...
+
+typedef struct TSParseOptions {
+  void *payload;
+  bool (*interrupt_callback)(TSParseState *state);
```
maxbrunsfeld, line 88: What do you think of calling this something like `progress_callback`, since it's called multiple times during normal parsing?

And actually, we could pass your current byte offset to let you track the progress. That's a good use case for `TSParserState` struct!

lib/include/tree_sitter/api.h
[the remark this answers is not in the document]
amaanq: Yeah, that makes a lot more sense, I can't really think of any stateful information that should be passed into the interrupt callback then, so I'll remove these wrapper structs

lib/include/tree_sitter/api.h
amaanq: That's really useful! Can we apply this to the `TSQueryCursorState` too with the current node's start byte?

maxbrunsfeld: Yeah, that's a good idea. Might have to grab the tree-cursor's node or something.

amaanq: Updated :grin:

kennet31526 (approved):

PR #3847, opened 2024-10-29
ci: add bindgen workflow

amaanq:
### Problem

Currently, we could update our C code without updating the corresponding Rust bindgen header located at `lib/binding_rust/bindings.rs`. We should have a CI check in place that ensures these two files are not out of sync.

### Solution

This PR adds a CI job that verifies there are no changes after running `cargo xtask generate-bindings` when `api.h` is changed.

[Successful run](https://github.com/tree-sitter/tree-sitter/actions/runs/11582325859/job/32245115197?pr=3847) (there *was* an error) because the bindings weren't updated in #3184

### DIFF
--- a/.github/workflows/bindgen.yml
+++ b/.github/workflows/bindgen.yml
@@ -1,0 +1,24 @@
+name: Check Bindgen Output
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  check-bindgen:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up stable Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Generate bindings
+        run: cargo xtask generate-bindings
+
+      - name: Check if the bindgen output changed
+        run: git diff --exit-code lib/binding_rust/bindings.rs
--- a/lib/binding_rust/bindings.rs
+++ b/lib/binding_rust/bindings.rs
@@ -761,6 +761,10 @@ extern "C" {
     ) -> TSStateId;
 }
 extern "C" {
+    #[doc = " Get the name of this language. This returns `NULL` in older parsers."]
+    pub fn ts_language_name(self_: *const TSLanguage) -> *const ::core::ffi::c_char;
+}
+extern "C" {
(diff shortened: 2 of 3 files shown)


.github/workflows/bindgen.yml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_call:
+
+jobs:
+  check-bindgen:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
amaanq: yeah good point it's not, I initially had it compare against the base ref which was unnecessary, thanks

.github/workflows/bindgen.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  check-bindgen:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Fetch the target branch
+        run: |
+          git fetch origin ${{ github.base_ref }}
```
dundargoc, line 15: `base_ref` is `master`, right? You could just add 

```
with:
  ref: ${github.base_ref}
```

to the `actions/checkout` action

amaanq: Yeah I realized I don't actually need this 😅

dundargoc: Or, if the branch is supposed to be the PR branch then it should be removed completely ofc.

.github/workflows/bindgen.yml
```diff
@@ -0,0 +1,25 @@
...
+          toolchain: stable
+
+      - name: Generate bindings
+        run: cargo xtask generate-bindings
+
+      - name: Check if the bindgen output changed
+        run: |
+          if git status --porcelain | grep -q "lib/binding_rust/bindings.rs"; then
+            git diff --exit-code
+          fi
```
ObserverOfTime, line 25:
```suggestion
        run: git diff --exit-code lib/binding_rust/bindings.rs
```

dundargoc (approved):

.github/workflows/ci.yml
```diff
@@ -33,5 +33,8 @@ jobs:
   sanitize:
     uses: ./.github/workflows/sanitize.yml
 
+  bindgen:
+    uses: ./.github/workflows/bindgen.yml
+
```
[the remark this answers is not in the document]
amaanq: Yeah we don't really need to do it like this

PR #3852, opened 2024-10-30
Code-format repro commands in bug report template

samestep:
Currently the issue template is a bit confusing since its placeholder text for "Steps to reproduce" is a sequence of commands, but does not have any triple backticks. This PR adds triple backticks to that placeholder to reduce confusion.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -13,7 +13,7 @@ body:
     attributes:
       label: "Steps to reproduce"
       placeholder: |
-        ```
+        ```sh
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -13,9 +13,11 @@ body:
     attributes:
       label: "Steps to reproduce"
       placeholder: |
+        ```
```
amaanq, line 16: would adding sh add some basic shell syntax highlighting? might be good to add
```suggestion
        ```sh
```

amaanq:
appreciate the PRs @samestep :slightly_smiling_face:

PR #3856, opened 2024-10-30
fix: pass all fields to `tree-sitter.json`

samestep:
Currently `tree-sitter init` completely ignores the provided `file_types` and `url`. This PR fixes that.


amaanq:
:facepalm: thank you!

PR #3855, opened 2024-10-30
docs: recommend `npm install` for IntelliSense

samestep:
Fixes #3854.

PR #3857, opened 2024-10-30
docs: fix missing links

samestep:
Followup on #3855; sorry for the silly mistake!


amaanq:
My bad on letting that slip :sweat_smile:

PR #3864, opened 2024-11-02
fix: disallow tokens that match the empty string

amaanq:
Supersedes #2446

This keeps the behavior the same as non-token rules, and prevents an infinite loop bug

Closes https://github.com/tree-sitter/tree-sitter/issues/98
Closes https://github.com/tree-sitter/tree-sitter/issues/1111
Closes https://github.com/tree-sitter/tree-sitter/issues/1271
Fixes the memory explosion cases in https://github.com/tree-sitter/tree-sitter/issues/1156


maxbrunsfeld:
It might be better to do this after converting rules to Nfa, so that you could also check if a regex matches the empty string. You could check if an accepting state is reachable from the starting state.

Issue #3412, opened 2024-06-11
`ts_node_descendant_for_point_range()` and `ts_node_descendant_for_byte_range()` with misspecified ranges


DavisVaughan:
### Problem

Consider these (currently failing) tests added to your suite of tests
https://github.com/DavisVaughan/tree-sitter/commit/b5d6223289481d2f350073351d4ead5c8f77071b

```rust
    // Range is misspecified, return self - byte query
    let self_node = array_node
        .named_descendant_for_byte_range(5, 3)
        .unwrap();
    assert_eq!(self_node, array_node);

    // Range is misspecified, return self - point query
    let self_node = array_node
        .named_descendant_for_point_range(Point::new(6, 6), Point::new(6, 4))
        .unwrap();
    assert_eq!(self_node, array_node);
```

Note how `start > end` in both cases. I think I would expect that if `start > end`, then there should be an early exit in these functions that just returns `self`.

Currently I think they are returning the smallest node that contains `start`? For example, the point range test errors with:

```
  left: {Node string (6, 4) - (6, 7)}
 right: {Node document (2, 0) - (9, 0)}
```

implying that `self_node` is `string (6, 4) - (6, 7)`, which seems a bit misleading

### Steps to reproduce

See above

### Expected behavior

Probably return `self` in these cases?

### Tree-sitter version (tree-sitter --version)

tree-sitter main

### Operating system/version

macOS 13.6.5

PR #3865, opened 2024-11-02
fix(lib): Add range checks for some functions

WillLillis:
It's possible to pass in an invalid range to `ts_node_descendant_for_point_range()` and `ts_node_descendant_for_byte_range()`, leading to non-intuitive/ incorrect results. This PR adds early checks for such incorrect ranges and returns a null node in such cases. 

The same potential issue is also present in `ts_query_cursor_set_byte_range()` and `ts_query_cursor_set_point_range()`. These functions now return a `bool` to indicate whether the range was set.

Closes https://github.com/tree-sitter/tree-sitter/issues/3412

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -1996,6 +1996,7 @@ fn test_query_matches_within_byte_range() {
             ]
         );
 
+        // An end byte of zero indicates there is no end
         let matches =
             cursor
                 .set_byte_range(12..0)
--- a/lib/binding_web/binding.js
+++ b/lib/binding_web/binding.js
@@ -1280,6 +1280,13 @@ class Query {
     if (typeof matchLimit !== 'number') {
       throw new Error('Arguments must be numbers');
     }
+    if (endIndex != 0 && startIndex > endIndex) {
+      throw new Error('`startIndex` cannot be greater than `endIndex`');
+    }
+    if (endPosition != ZERO_POINT && (startPosition.row > endPosition.row ||
+      (startPosition.row == endPosition.row && startPosition.column > endPosition.row))) {
+      throw new Error('`startPosition` cannot be greater than `endPosition`');
+    }
 
     marshalNode(node);
 
@@ -1344,6 +1351,13 @@ class Query {
   ) {
     if (typeof matchLimit !== 'number') {
       throw new Error('Arguments must be numbers');
+    }
+    if (endIndex != 0 && startIndex > endIndex) {
+      throw new Error('`startIndex` cannot be greater than `endIndex`');
+    }
+    if (endPosition != ZERO_POINT && (startPosition.row > endPosition.row ||
+      (startPosition.row == endPosition.row && startPosition.column > endPosition.row))) {
+      throw new Error('`startPosition` cannot be greater than `endPosition`');
     }
 
     marshalNode(node);
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -1089,11 +1089,20 @@ void ts_query_cursor_set_timeout_micros(TSQueryCursor *self, uint64_t timeout_mi
 uint64_t ts_query_cursor_timeout_micros(const TSQueryCursor *self);
 
 /**
- * Set the range of bytes or (row, column) positions in which the query
- * will be executed.
+ * Set the range of bytes in which the query will be executed.
+ *
+ * This will return `false` if the start byte is greater than the end byte, otherwise
+ * it will return `true`.
  */
-void ts_query_cursor_set_byte_range(TSQueryCursor *self, uint32_t start_byte, uint32_t end_byte);
-void ts_query_cursor_set_point_range(TSQueryCursor *self, TSPoint start_point, TSPoint end_point);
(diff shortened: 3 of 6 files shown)


lib/src/query.c
```diff
@@ -3103,28 +3103,34 @@ void ts_query_cursor_exec_with_options(
...
+bool ts_query_cursor_set_point_range(
   TSQueryCursor *self,
   TSPoint start_point,
   TSPoint end_point
 ) {
   if (end_point.row == 0 && end_point.column == 0) {
     end_point = POINT_MAX;
+  } else if (point_gt(start_point, end_point)) {
+    return false;
   }
```
amaanq, line 3130: remove the `else`, since at this point we've set the end point to `POINT_MAX`, so the `gt` comparison will evaluate to false
```suggestion
  if (point_gt(start_point, end_point)) {
    return false;
  }
```

lib/src/query.c
```diff
@@ -3103,28 +3103,34 @@ void ts_query_cursor_exec_with_options(
...
+bool ts_query_cursor_set_byte_range(
   TSQueryCursor *self,
   uint32_t start_byte,
   uint32_t end_byte
 ) {
   if (end_byte == 0) {
     end_byte = UINT32_MAX;
+  } else if (start_byte > end_byte) {
+    return false;
   }
```
amaanq, line 3115: same thing

```suggestion
  if (start_byte > end_byte) {
    return false;
  }
```

amaanq (commented):
also, update the web bindings

lib/binding_web/binding.js
```diff
@@ -1280,6 +1280,13 @@ class Query {
     if (typeof matchLimit !== 'number') {
       throw new Error('Arguments must be numbers');
     }
+    if (endIndex != 0 && startIndex > endIndex) {
+      throw new Error('startIndex cannot be greater than endIndex');
+    }
+    if (endPosition != ZERO_POINT && (startPosition.row > endPosition.row ||
```
WillLillis, line 1286: Logic here is duplicated below. Wasn't sure if it should be pulled into a helper/ where that would live in the file.

cli/src/tests/query_test.rs
```diff
@@ -1996,6 +1996,7 @@ fn test_query_matches_within_byte_range() {
             ]
         );
 
+        // Zero end byte is treated like no end byte.
```
amaanq, line 1999:
```suggestion
        // An end byte of zero is indicates there is no end
```

lib/binding_web/binding.js
```diff
@@ -1280,6 +1280,13 @@ class Query {
     if (typeof matchLimit !== 'number') {
       throw new Error('Arguments must be numbers');
     }
+    if (endIndex != 0 && startIndex > endIndex) {
+      throw new Error('startIndex cannot be greater than endIndex');
```
amaanq, line 1284:
```suggestion
      throw new Error('`startIndex` cannot be greater than `endIndex`');
```

lib/binding_web/binding.js
```diff
@@ -1280,6 +1280,13 @@ class Query {
...
+    }
+    if (endPosition != ZERO_POINT && (startPosition.row > endPosition.row ||
+      (startPosition.row == endPosition.row && startPosition.column > endPosition.row))) {
+      throw new Error('startPosition cannot be greater than endPosition');
```
amaanq, line 1288:
```suggestion
      throw new Error('`startPosition` cannot be greater than `endPosition`');
```

lib/binding_web/binding.js
amaanq: I think it's fine

lib/binding_web/binding.js
```diff
@@ -1345,6 +1352,13 @@ class Query {
...
+    }
```
amaanq, line 1361: same comments as above

lib/include/tree_sitter/api.h
```diff
@@ -1092,8 +1092,8 @@ uint64_t ts_query_cursor_timeout_micros(const TSQueryCursor *self);
  * Set the range of bytes or (row, column) positions in which the query
  * will be executed.
  */
-void ts_query_cursor_set_byte_range(TSQueryCursor *self, uint32_t start_byte, uint32_t end_byte);
```
amaanq, line 1095: Comment what the `bool` return value is

lib/include/tree_sitter/api.h
```diff
@@ -1091,9 +1091,11 @@ uint64_t ts_query_cursor_timeout_micros(const TSQueryCursor *self);
 /**
  * Set the range of bytes or (row, column) positions in which the query
  * will be executed.
+ *
+ * If the provided range was set, return `true`. Otherwise, return `false`.
```
amaanq, line 1095: Clearer explanation

"This will return false if the start byte/point is greater than the end byte/point, otherwise it will return true.`

Also, do a favor and have the doc comments be associated to each function individually, (adjust byte/point semantics as needed). This will help w/ the rust doc comments in `bindings.rs`

amaanq (approved):
Good work @WillLillis!

PR #3868, opened 2024-11-02
fix(generate): do not set the unit reduction symbol if it's in the extras array

### LINKED ISSUE #990: Parser generation can run into an infinite loop
I have created a grammar for which `tree-sitter generate` seems to run indefinitely. This appears to be related to the `_comment` rule, for which inclusion is allowed anywhere in the tree. I have removed all other rules and the grammar is as follows:
```js
module.exports = grammar({
  name: 'bug-reproduction',

  rules: {
    source_file: $ => repeat(choice($.foo, $._comment)),

    _comment: $ => choice(
      /\/\/.*\n/m,
      $._blockComment,
    ),
    _blockComment: $ => seq(
      "/*",
      $._blockCommentContents,
      repeat(seq(
        optional("/"),
        $._blockComment,
        $._blockCommentContents,
      )),
      optional("*"),
      "*/",
    ),
    _blockCommentContents: $ => /([^\/\*]|\/[^\*]|\*[^\/]|\\\/|\*\\\/)+/,

    foo: $ => 'foo',
  },
  extras: $ => [$._comment],
});
```

I have also trimmed down the comment rule in a way that preserves this issue:
```js
module.exports = grammar({
  name: 'bug-reproduction',

  rules: {
    source_file: $ => repeat(choice($.foo, $._comment)),

    _comment: $ => choice(
      /\/\/.*\n/m,
    ),

    foo: $ => 'foo',
  },
  extras: $ => [$._comment],
});
```
The top-level declaration includes `_comment` as a choice as a work-around to issue #768. At this point removing the `choice` in the `_comment` rule fixes the problem.

I have found the following fixes to this problem so that the grammar including both types of comments generates as expected:
* Renaming the rule from `_comment` to `comment`. Unfortunately this also implies the inclusion of the comment nodes into the tree and I would rather not handle them in every possible location.
* Renaming the rule to `comment` and introducing a new rule `_commend: $ => seq($.comment, '')`. This seems to imply the problem is somehow related to the use of choice. I have read in other issues that empty tokens can cause problems with error recovery, so this doesn't seem like a true fix either. Defining the rule as `seq($.comment)` does not fix the issue.
* I can factor out the leading slash from line and block comments and rewrite the grammar as follows:
```js
module.exports = grammar({
  name: 'bug-reproduction',

  rules: {
    source_file: $ => repeat(choice($.foo, $._comment)),

    _comment: $ => seq("/", choice(
      /\/.*\n/m,
      $._blockCommentWithoutSlash,
    )),
    _blockComment: $ => seq('/', $._blockCommentWithoutSlash),
    _blockCommentWithoutSlash: $ => seq(
      "/*",
      $._blockCommentContents,
      repeat(seq(
        optional("/"),
        $._blockComment,
        $._blockCommentContents,
      )),
      optional("*"),
      "*/",
    ),
    _blockCommentContents: $ => /([^\/\*]|\/[^\*]|\*[^\/]|\\\/|\*\\\/)+/,

    foo: $ => 'foo',
  },
  extras: $ => [$._comment],
});
```
~~While this works as a temporary work-around, I would hardly consider it a fix.~~ I believe this isn't a complete fix since this would now allow inserting whitespace between the first slash and (second slash OR `*`).

With some debugging I have managed to track the indefinitely running step to be `remove_unit_reductions`, specifically [this loop](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/build_tables/minimize_parse_table.rs#L102-L115). Looking at this code I can see how it could get stuck in an infinite loop, but I'm not really sure why the name would influence this behavior.

### LINKED ISSUE #1062: Generate hangs when extra rule is aliased
```js
module.exports = grammar({
    name: 'test',

    extras: $ => [ $._whitespace, $._split ],

    rules: {

        source_file: $ => repeat(
            seq($.list,';')
        ),

        list: $ => seq(
            $.identifier,
            repeat(seq(
                $._whitespace,
                $.identifier
            ))
        ),

        identifier: $ => /[a-zA-Z_]\w*/,

        // hangs:
        _split: $ => alias(token(seq('\\', '\n')), '\\'),
        // works:
        //_split: $ => token(seq('\\', '\n')),

        _whitespace: $ => /[\t ]/,

    }
});
```

```
$ tree-sitter --version
tree-sitter 0.19.4 (76862e281b3aa5d93b471bfdf903527fec4f9b29)
$ tree-sitter generate 
```
tree-sitter generate don't finish

### LINKED ISSUE #1165: Hang with hidden rule in extras
```js
module.exports = grammar({
  name: 'bugcase',

  extras: $ => [
    $._extra,
  ],

  rules: {
    source_file: $ => repeat('hello'),

    _extra: $ => $.space,

    space: $ => ' ',
  }
});
```

Renaming `_extra` to `extra` does not cause a hang.

Full test case: https://gist.github.com/CyberShadow/34b3f9f0ec3c20d68f3cd30fd7a38f01 (`npm install && node_modules/.bin/tree-sitter generate`).

#1062 looks related.

### LINKED ISSUE #1371: Hidden rule with alias causes an endless loop on generation if put in `externals`
So when working on [tree-sitter-viml](https://github.com/vigoux/tree-sitter-viml) and especially [this PR](https://github.com/vigoux/tree-sitter-viml/pull/12) I came across the bug mentioned in the title.

After some work on minimizing the example, here is the minimal `grammar.js` that I could make:
```js
module.exports = grammar({
  name: 'foo',

  extras: ($) => [$._weird_alias],

  rules: {
    script: ($) => repeat1('foobar'),
    _weird_alias: ($) => alias('bar', '\\'),
  },
});
```

# Reproducting steps

1. Put this file in your `grammar.js`
2. `tree-sitter generate`
3. Observe the hang

amaanq:
- Closes #990
- Closes #1062
- Closes #1165
- Closes #1371

### Problem

When a hidden rule is added to the extras array, parser generation hangs, as observed in the linked issues.

### Solution

The reason it hangs is because the unit reduction function tries to update a state that contains a non terminal extra, and when doing so, the closure passed in [here](https://github.com/tree-sitter/tree-sitter/blob/310a9f0704aeb8d9b1e32ff2bf9b6bd03c8032eb/cli/generate/src/build_tables/minimize_parse_table.rs#L113) receives a `GotoAction::ShiftExtra`, which should not even be processed in here. So, we just skip any symbols that are in the grammar's `extra_symbols` array here.

### DIFF
--- a/cli/generate/src/build_tables/minimize_parse_table.rs
+++ b/cli/generate/src/build_tables/minimize_parse_table.rs
@@ -70,18 +70,17 @@ impl Minimizer<'_> {
                             production_id: 0,
                             symbol,
                             ..
-                        } => {
-                            if !self.simple_aliases.contains_key(symbol)
-                                && !self.syntax_grammar.supertype_symbols.contains(symbol)
-                                && !aliased_symbols.contains(symbol)
-                                && self.syntax_grammar.variables[symbol.index].kind
-                                    != VariableType::Named
-                                && (unit_reduction_symbol.is_none()
-                                    || unit_reduction_symbol == Some(symbol))
-                            {
-                                unit_reduction_symbol = Some(symbol);
-                                continue;
-                            }
+                        } if !self.simple_aliases.contains_key(symbol)
+                            && !self.syntax_grammar.supertype_symbols.contains(symbol)
+                            && !self.syntax_grammar.extra_symbols.contains(symbol)
+                            && !aliased_symbols.contains(symbol)
+                            && self.syntax_grammar.variables[symbol.index].kind
+                                != VariableType::Named
+                            && (unit_reduction_symbol.is_none()
+                                || unit_reduction_symbol == Some(symbol)) =>
+                        {
+                            unit_reduction_symbol = Some(symbol);
+                            continue;
                         }
                         _ => {}
                     }


clason:
Sooo.... did we add these tests for nothing?

ObserverOfTime:
Actually it was #3864 that broke a parser.
https://github.com/tree-sitter/tree-sitter/actions/runs/11640072144/job/32418580730#step:10:496

clason:
That broke generation (and was fixed in tree-sitter-ruby); the windows build failure is more worrying.

amaanq:
The windows build failure is spurious, and unrelated to this PR.

clason:
So rerun the tests to make sure. If you merge PRs with failing tests, you might as well not have them.

amaanq:
Well, for one, the windows CI fails pretty regularly and inconsistently..

clason:
...in tree-sitter build. might want to look into that.

maxbrunsfeld:
Nice job @amaanq

PR #3876, opened 2024-11-03
fix(docs): add Elixir language binding

mwhitworth:
Although these are at an experimental stage, most of the interface is covered


ObserverOfTime:
Please build upon the C API rather than the Rust bindings.

PR #3879, opened 2024-11-03
feat: allow css classes in HTML highlighter output, rather than inline styles

### LINKED ISSUE #553: Add flag to output css classes for CLI tree-sitter highlight
Hello,

tree-sitter seems to be a great project that can also be used to overcome the limitations of regex based highlighting.

I'm trying to use it as a (drop-in) replacement for e.g. [Prism](https://github.com/PrismJS/prism/) or [Pygments](https://pygments.org/)/[Chroma](https://github.com/alecthomas/chroma).

__It would be great to have a flag to output CSS classes instead of inline styles.__  
I've looked at the source code but my Rust knowledge is not sufficient to fully implement this – but I think it should not be too hard.

Inline style
```html
<span style='color: #5f00d7;'>function</span>
```

With CSS classes
```html
<span class='keyword'>function</span>
```

Ideally, there could be additional CLI flags for `tree-sitter highlight` to 
* output only a HTML snippet or a standalone HTML page (see https://github.com/tree-sitter/tree-sitter/pull/551) and to 
* have a table layout or alternatively a `<pre><code><spans…>` layout w/o line-numbers (see the other issue I opened: https://github.com/tree-sitter/tree-sitter/issues/554)

What do you think about adding a flag to output CSS classes instead of inline styles to generate output similar to the one above?

lygaret:
fixes #553

when the `--data-highlight` flag is passed along with the `--html` flag, the highlighter will inject `data-tree-sitter-highlight` attributes into the `span` tags, rather than pulling inline styles from the theme.

this makes it significantly easier to use from downstream tools, because we can apply css to the output directly, rather than having the user provide TS with a theme.

only matches which have theme entries are considered for attributes.

Example Output _with_ flag
```html
<span data-tree-sitter-highlight="punctuation.bracket">(</span>
<span data-tree-sitter-highlight="function.builtin">sequence-map</span> 
 path-&gt;note _
<span data-tree-sitter-highlight="punctuation.bracket">)</span>
<span data-tree-sitter-highlight="punctuation.bracket">)</span>
<span data-tree-sitter-highlight="punctuation.bracket">)</span>
```

Original Output _without_ flag
```html
<span style='color: #4e4e4e'>(</span>
<span style='font-weight: bold;color: #005fd7'>sequence-map</span>
 path-&gt;note _
<span style='color: #4e4e4e'>)</span>
<span style='color: #4e4e4e'>)</span>
<span style='color: #4e4e4e'>)</span>
```

### DIFF
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -407,11 +407,10 @@ pub fn html(
                 theme.styles[highlight.0]
                     .css
                     .as_ref()
-                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes())
+                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes()),
             );
             output.extend("'".as_bytes())
-        }
-        else {
+        } else {
             output.extend(b"class='");
             let mut parts = theme.highlight_names[highlight.0].split('.').peekable();
             while let Some(part) = parts.next() {
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -409,9 +409,15 @@ pub fn html(
             );
         }
         else {
-            output.extend(b"data-tree-sitter-highlight=\"");
-            output.extend(theme.highlight_names[highlight.0].as_bytes());
-            output.extend(b"\"");
+            output.extend(b"class='");
+            let mut parts = theme.highlight_names[highlight.0].split('.').peekable();
+            while let Some(part) = parts.next() {
+                output.extend(part.as_bytes());
+                if parts.peek().is_some() {
+                    output.extend(b" ");
+                }
+            }
+            output.extend(b"'");
         }
     })?;
 
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -378,14 +378,18 @@ pub fn ansi(
     Ok(())
 }
 
+pub struct HtmlOptions {
+    pub inline_styles: bool,
+    pub quiet: bool,
+    pub print_time: bool,
+}
+
 pub fn html(
     loader: &Loader,
     theme: &Theme,
     source: &[u8],
     config: &HighlightConfiguration,
-    quiet: bool,
-    inline_styles: bool,
-    print_time: bool,
+    opts: &HtmlOptions,
     cancellation_flag: Option<&AtomicUsize>,
 ) -> Result<()> {
     use std::io::Write;
@@ -401,15 +405,15 @@ pub fn html(
 
     let mut renderer = HtmlRenderer::new();
     renderer.render(events, source, &move |highlight, output| {
-        if inline_styles {
+        if opts.inline_styles {
             output.extend(b"style='");
             output.extend(
                 theme.styles[highlight.0]
                     .css
                     .as_ref()
-                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes())
+                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes()),
             );
-            output.extend(b"'")
+            output.extend(b"'");
         } else {
             output.extend(b"class='");
             let mut parts = theme.highlight_names[highlight.0].split('.').peekable();
@@ -423,7 +427,7 @@ pub fn html(
         }
     })?;
 
-    if !quiet {
+    if !opts.quiet {
         writeln!(&mut stdout, "<table>")?;
         for (i, line) in renderer.lines().enumerate() {
             writeln!(
@@ -436,7 +440,7 @@ pub fn html(
         writeln!(&mut stdout, "</table>")?;
     }
 
-    if print_time {
+    if opts.print_time {
         eprintln!("Time: {}ms", time.elapsed().as_millis());
     }
 
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -383,6 +383,7 @@ pub fn html(
     source: &[u8],
     config: &HighlightConfiguration,
     quiet: bool,
+    inline_styles: bool,
     print_time: bool,
     cancellation_flag: Option<&AtomicUsize>,
 ) -> Result<()> {
@@ -398,11 +399,20 @@ pub fn html(
     })?;
 
     let mut renderer = HtmlRenderer::new();
-    renderer.render(events, source, &move |highlight| {
-        theme.styles[highlight.0]
-            .css
-            .as_ref()
-            .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes())
+    renderer.render(events, source, &move |highlight, output| {
+        if inline_styles {
+            output.extend(
+                theme.styles[highlight.0]
+                    .css
+                    .as_ref()
+                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes())
+            );
+        }
+        else {
+            output.extend(b"data-tree-sitter-highlight=\"");
+            output.extend(theme.highlight_names[highlight.0].as_bytes());
+            output.extend(b"\"");
+        }
     })?;
 
     if !quiet {
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -270,7 +270,7 @@ fn hex_string_to_rgb(s: &str) -> Option<(u8, u8, u8)> {
 }
 
 fn style_to_css(style: anstyle::Style) -> String {
-    let mut result = "".to_string();
+    let mut result = String::new();
     let effects = style.get_effects();
     if effects.contains(Effects::UNDERLINE) {
         write!(&mut result, "text-decoration: underline;").unwrap();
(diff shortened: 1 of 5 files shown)


lygaret:
Changed the attribute highlighter callback to pass the buffer directly, since I couldn't figure out how to return the fully formatted data attribute up; I'm not a rust programmer really.

If there are changes, or a different approach, please let me know, but I'd very much like to be able to use this tool in scripts without having to translate a bunch of themes downstream.

Thanks!

ObserverOfTime:
Can you do this with classes instead?

lygaret:
> Can you do this with classes instead?

Not without doing a translation pass on the highlighted key names, since the periods in the type names make them invalid as CSS clas names (eg. `class="punctuation.bracket"` isn't valid afaiu)

lygaret:
> Not without doingwithout doing a translation pass on the highlighted key names, since the periods in the type names make them invalid as CSS clas names (eg. `class="punctuation.bracket"` isn't valid afaiu).

Thinking this through a bit more, I could see doing a translation by parts: where`variable.parameter` becomes two classnames: `variable variable-parameter` in order to get the same ability to target the top level.

So,

```html
<span class="variable variable-parameter">
vs.
<span data-tree-sitter-highlight="variable.parameter">
```

If that's preferable, I can work on that, but I think it'd be a bigger change in order to build up the class strings in the theme, the same way it currently builds up style strings.

ObserverOfTime:
Just split them. `@foo.bar` is supposed to inherit from `@foo` anyway.

lygaret:
> Just split them and let both highlights be applied

Just making sure I understand, the preferred output is: `<span class="variable parameter">`?

ObserverOfTime:
Yes.

lygaret:
Got sidetracked with US election and work related activities, coming back to this in the next few days.

lygaret:
Also, I don't see a `scripts` directory, and hence can't install fixtures, so `cargo test` fails for me locally. I'm not sure how to move forward on that one.

Thank you!

---

[Docs specify](https://tree-sitter.github.io/tree-sitter/contributing):

> Before you can run the tests, you need to clone some grammars that are used for testing:
>
>     script/fetch-fixtures
>
> To test any changes you’ve made to the CLI, you can regenerate these parsers using your current CLI code:
>
>     script/generate-fixtures
>

lygaret:
@ObserverOfTime hey, just checking in on this; does the split-up on `.` work for you?

ObserverOfTime (commented):
The styles are not applied though.

lygaret:
> The styles are not applied though.

Inline styles are the default give `--html`; setting `--css-classes` will output class names _instead_. 

I don't think it should do both styles and class names: the inline styles will take precedence, so there wouldn't be any point to outputting classes. 

I guess I'm confused as to what else you're looking for.

lygaret:
> > The styles are not applied though.
> 
> Inline styles are the default give `--html`; setting `--css-classes` will output class names _instead_. 
> 
> I don't think it should do both styles and class names: the inline styles will take precedence, so there wouldn't be any point to outputting classes. 
> 
> I guess I'm confused as to what else you're looking for.

Importantly, my use case would be supplying styles that target the classes separately, not including them directly or in CSS output in the generated HTML. 

IMO, if I needed explicit styling in the output, the inline styles would have been fine.

ObserverOfTime:
The styles are defined in the user's `config.json` and should be applied even when outputting classes.

```html
<style>
.variable { color: blue; }
.variable.builtin { color: green; }
</style>
```

lygaret:
> The styles are defined in the user's `config.json` and should be applied even when outputting classes.
> 
> ```html
> <style>
> .variable { color: blue; }
> .variable.builtin { color: green; }
> </style>
> ```

I'm sorry, I very much disagree; at least I don't need them, and that will blow up the scope of this change a ton. 

Is there a way we can add the classes with the assumption that if you're going to display the HTML directly, you shouldn't change anything and continue using inline styles?

lygaret:
> and that will blow up the scope of this change a ton.

I guess I should see how big a change it would be before making that statement; I'll dig in.

That said, if we're going to include a style block based on the user's config, should I remove the flag, and just make this the default output of the tool? Having inline-vs-classes as an option doesn't make a ton of sense to me, unless classes is more for getting parsing info.

That's why I started with a `data-` attribute actually, because I don't care at all about the styling coming out of the tool, and can drop the output into another HTML file where it will pick up styles.

cli/src/main.rs
```diff
@@ -1204,6 +1210,19 @@ impl Highlight {
                     }
                 }
 
+                if html_mode && !quiet {
+                    println!("  <style>");
+                    let names  = theme_config.theme.highlight_names.iter();
+                    let styles = theme_config.theme.styles.iter();
+                    for (name, style) in names.zip(styles) {
+                        if let Some(css) = &style.css {
+                            println!("    .{} {{ {} }}", name, css);
```
lygaret, line 1219: this is working, because the `.` of the highlight name separator works as the child class name.

This is incidental though, and will only work if the format of highlight names doesn't change, and doesn't include illegal characters for CSS class names; all of the grammars I've found work, but this feels ugly and lucky to me.

ie, not sure how better to sanitize highlight names into css classes, just letting it through for now

lygaret:
current output:

```sh
$ cargo run highlight --html --css-classes ./build.zig | head -50 | pbcopy -
```

```html
<!doctype HTML>
<head>
  <title>Tree-sitter Highlighting</title>
  <style>
    body {
      font-family: monospace
    }
    .line-number {
      user-select: none;
      text-align: right;
      color: rgba(27,31,35,.3);
      padding: 0 10px;
    }
    .line {
      white-space: pre;
    }
  </style>
  <style>
    .property { color: #af0000 }
    .attribute { font-style: italic;color: #af0000 }
    .type.builtin { font-weight: bold;color: #005f5f }
    .constant.builtin { font-weight: bold;color: #875f00 }
    .module { color: #af8700 }
    .variable.builtin { font-weight: bold; }
    .type { color: #005f5f }
    .comment { font-style: italic;color: #875f00 }
    .punctuation.bracket { color: #4e4e4e }
    .number { font-weight: bold;color: #875f00 }
    .string { color: #008700 }
    .tag { color: #000087 }
    .punctuation.delimiter { color: #4e4e4e }
    .variable.parameter { text-decoration: underline; }
    .function.builtin { font-weight: bold;color: #005fd7 }
    .string.special { color: #008787 }
    .constructor { color: #af8700 }
    .keyword { color: #5f00d7 }
    .function { color: #005fd7 }
    .constant { color: #875f00 }
    .operator { font-weight: bold;color: #4e4e4e }
  </style>

</head>
<body>

<table>
<tr><td class=line-number>1</td><td class=line><span class='keyword'>const</span> <span class='module'>std</span> <span class='operator'>=</span> <span class='keyword'>@import</span><span class='punctuation bracket'>(</span><span class='string'>&quot;std&quot;</span><span class='punctuation bracket'>)</span><span class='punctuation delimiter'>;</span>
</td></tr>
<tr><td class=line-number>2</td><td class=line>
</td></tr>
```

ObserverOfTime:
> That said, if we're going to include a style block based on the user's config, should I remove the flag, and just make this the default output of the tool? Having inline-vs-classes as an option doesn't make a ton of sense to me, unless classes is more for getting parsing info.

Maybe.

cli/src/highlight.rs
```diff
@@ -268,7 +270,7 @@ fn hex_string_to_rgb(s: &str) -> Option<(u8, u8, u8)> {
 }
 
 fn style_to_css(style: anstyle::Style) -> String {
-    let mut result = "style='".to_string();
+    let mut result = "".to_string();
```
WillLillis, line 273: Probably just want to use `String::new()` here

cli/src/highlight.rs
```diff
@@ -399,11 +401,27 @@ pub fn html(
...
+    renderer.render(events, source, &move |highlight, output| {
+        if inline_styles {
+            output.extend("style='".as_bytes());
+            output.extend(
+                theme.styles[highlight.0]
+                    .css
+                    .as_ref()
+                    .map_or_else(|| "".as_bytes(), |css_style| css_style.as_bytes()),
+            );
+            output.extend("'".as_bytes())
```
WillLillis, line 413: Consistency needed where applicable with the b"" syntax used in the else block below

cli/src/main.rs
```diff
@@ -361,6 +361,11 @@ struct Query {
 struct Highlight {
     #[arg(long, short = 'H', help = "Generate highlighting as an HTML document")]
     pub html: bool,
+    #[arg(
+        long,
+        help = "When generating HTML, use css classes, rather than inline styles"
```
WillLillis, line 366: Remove the second comma

lygaret:
Hey @ObserverOfTime, @WillLillis, is there anything else y'all are looking for? Don't want this to get abandoned if I can help it.

WillLillis:
Changes LGTM stylistically, but I'll defer to Observer for the remainder of the review.

WillLillis (approved):

ObserverOfTime (approved):

lygaret:
with this approved, is there anything left for me to do? at least get the workflows running so I can fix anything that comes out of CI?

Issue #2354, opened 2023-06-28
allowing double barrelled extensions to be declared in `file-types` , in `package.json`


EmranMR:
Hello 👋

I had a feature request and was hoping that someone could help me out. 
Currently we can only declare **one** `file-types` or array of (multiple) `file-types` in `package.json` for the `tree-sitter` property. This is all good for most languages, frameworks and template engines but, there are few frameworks out there that use double barrelled extensions such as `source.blade.php` see EmranMR/tree-sitter-blade#6.

So in the case of `blade.php`, I could only use 'php' as the `file-types` however that  would badly conflict with the `tree-sitter-php` that I am trying to inject inside `blade.php` when using commands like `tree-sitter highlight`

if I do:

```json
"file-types": ["blade.php"]
```

and then do `tree-sitter dump-languages` I get:
```
file_types: [".blade.php"]
content_regex: None
injection_regex: Some(blade)
```

It seems that `tree-sitter` has picked up but apparently that is not the case as "`tree-sitter` uses simple language by a file-type look up, using the first match for parsing. As well as not allowing multiple suffixes."

Thanks
Emran 😊

PR #3880, opened 2024-11-03
feat(loader): support multi-barreled file extensions

WillLillis:
The problem is that `PathBuf::extension()` only returns the rightmost extension from a provided path. That is, if we have a filename such as `foo.tar.gz.`, calling `.extension()` will only yield `gz`. The solution to this is just to repeatedly call off extensions from a given filename until there aren't any left. This fix will cover any number of extensions to a filename.

Closes https://github.com/tree-sitter/tree-sitter/issues/2354

### DIFF
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -521,11 +521,15 @@ impl Loader {
             .and_then(|n| n.to_str())
             .and_then(|file_name| self.language_configuration_ids_by_file_type.get(file_name))
             .or_else(|| {
-                path.extension()
-                    .and_then(|extension| extension.to_str())
-                    .and_then(|extension| {
-                        self.language_configuration_ids_by_file_type.get(extension)
-                    })
+                let mut path = path.to_owned();
+                let mut extensions = Vec::with_capacity(2);
+                while let Some(extension) = path.extension() {
+                    extensions.push(extension.to_str()?.to_string());
+                    path = PathBuf::from(path.file_stem()?.to_os_string());
+                }
+                extensions.reverse();
+                self.language_configuration_ids_by_file_type
+                    .get(&extensions.join("."))
             });
 
         if let Some(configuration_ids) = configuration_ids {
--- a/cli/src/tests/detect_language.rs
+++ b/cli/src/tests/detect_language.rs
@@ -89,6 +89,120 @@ fn detect_language_by_first_line_regex() {
     );
 }
 
+#[test]
+fn detect_langauge_by_double_barrel_file_extension() {
+    let blade_dir = tree_sitter_dir(
+        r#"{
+  "grammars": [
+    {
+      "name": "blade",
+      "path": ".",
+      "scope": "source.blade",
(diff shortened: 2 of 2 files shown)


ObserverOfTime (commented):
Make sure this doesn't break files without an extension (e.g. `ssh_config`) or without a name (e.g. `.gitignore`). _Note that this appears to be broken in 0.24.3 but not in master._

WillLillis:
> Make sure this doesn't break files without an extension (e.g. `ssh_config`) or without a name (e.g. `.gitignore`). _Note that this appears to be broken in 0.24.3 but not in master._

Good point. Both of these cases seem to be working with this PR, but I added some tests to help prevent future regressions. Both of these seem to work without reaching these changes because `language_configuration_for_filename` checks for matches from the `file-types` field with the file's full name before trying to do anything with extensions.

cli/loader/src/lib.rs
```diff
@@ -521,11 +521,18 @@ impl Loader {
...
+                while let Some(extension) = path.extension() {
+                    full_extension = format!(
+                        "{}{}{full_extension}",
+                        extension.to_str()?,
+                        if full_extension.is_empty() { "" } else { "." }
+                    );
+                    path = PathBuf::from(path.file_stem()?.to_os_string());
+                }
+                self.language_configuration_ids_by_file_type
+                    .get(&full_extension)
```
amaanq, line 535: for clarity and avoiding allocations on every call to format, store a `Vec` of extensions, and push to it on each iteration while there is an extension, then just join them with a `.` at the end

cli/loader/src/lib.rs
```diff
@@ -521,11 +521,20 @@ impl Loader {
...
-                    .and_then(|extension| {
-                        self.language_configuration_ids_by_file_type.get(extension)
-                    })
+                let mut path = path.to_owned();
+                let mut extensions = Vec::new();
+                while let Some(extension) = path.extension() {
+                    extensions.push(extension.to_str()?.to_string());
+                    path = PathBuf::from(path.file_stem()?.to_os_string());
+                }
+                let full_extension = extensions
```
WillLillis: This felt a bit awkward, but using `VecDeque` with `push_front` doesn't work since `.join` isn't implemented on that container. I tried two approaches but left them in separate commits.

cli/loader/src/lib.rs
```diff
@@ -521,11 +521,15 @@ impl Loader {
             .and_then(|n| n.to_str())
             .and_then(|file_name| self.language_configuration_ids_by_file_type.get(file_name))
             .or_else(|| {
-                path.extension()
-                    .and_then(|extension| extension.to_str())
-                    .and_then(|extension| {
-                        self.language_configuration_ids_by_file_type.get(extension)
-                    })
+                let mut path = path.to_owned();
+                let mut extensions = std::collections::VecDeque::new();
```
WillLillis, line 525: Wasn't sure if this was worth a `use` at the top of the file since it's only being used here.

PR #3881, opened 2024-11-04
Link to Emacs’ Tree-sitter documentation

mavit:
Since Emacs 29, Tree-sitter is a built-in feature not requiring a third-party Emacs module.


ObserverOfTime:
Maybe we should add a new section for editors that use tree-sitter.

PR #3887, opened 2024-11-06
feat: query missing nodes with (MISSING)

### LINKED ISSUE #606: Allow MISSING in tree queries
When using tree queries, types added by tree-sitter itself, such as `ERROR` are fine in a query like `(ERROR) @error`.

Using the type `MISSING` as in `(MISSING) @missing`, however, results in an error (at least on the Python and Web binding, but it should likely be the same for the core, since they're just bindings).

The error I get is:

```
...
File "../venv/lib/python3.8/site-packages/tree_sitter/__init__.py", line 92, in query
    return _language_query(self.language_id, source)
NameError: Invalid node type MISSING
```

There doesn't seem to be any direct error thrown by the web binding, but it isn't working as with other node types.

### LINKED ISSUE #650: Document how MISSING nodes work
If I parse a file that contains a couple of errors using `tree-sitter parse corpus/medium.rpc` I get this:

```
...
    (structField [25, 2] - [25, 32]
      structFieldName: (fieldName [25, 2] - [25, 25])
      structFieldType: (type [25, 27] - [25, 31]))
    (ERROR [26, 2] - [26, 33]))
  (enum [29, 0] - [32, 1]
    enumName: (userTypeDeclaration [29, 5] - [29, 8])
    (enumVariant [30, 2] - [30, 4]
      enumVariantName: (variantName [30, 2] - [30, 3]))
    (enumVariant [31, 2] - [31, 7]
      enumVariantName: (variantName [31, 2] - [31, 3])
      enumVariantType: (userTypeUse [31, 4] - [31, 5]))))
corpus/medium.rpc       0 ms    (MISSING "," [24, 39] - [24, 39])
```

The first `ERROR` makes sense to me (although I have no idea how to get an actual error message from it). But I can't figure out the `MISSING` node. It doesn't seem to be part of the tree itself, and I can't find it using a query like `(MISSING) @missing`. Is there any way to access it so I can show a VSCode diagnostic?

ribru17:
Closes #606

Also closes #650 I believe? The first task in [this comment](https://github.com/tree-sitter/tree-sitter/issues/650#issuecomment-647688573) has not been addressed, but I believe this is a separate issue

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -806,7 +806,7 @@ fn test_query_matches_capturing_missing_nodes() {
             r#"
             (MISSING
               ; Comments should be valid
-            ) @missing-node
+            ) @missing
             (MISSING
               ; Comments should be valid
               ";"
@@ -825,7 +825,7 @@ fn test_query_matches_capturing_missing_nodes() {
             //                    ^ MISSING semicolon here
             ",
             &[
-                (0, vec![("missing-node", "")]),
+                (0, vec![("missing", "")]),
                 (1, vec![("missing-semicolon", "")]),
             ],
         );
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -804,6 +804,7 @@ fn test_query_matches_capturing_missing_nodes() {
                 c();
 
                 if (*) d();
+                //   ^ MISSING identifier here
               }
             }
             ",
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -193,6 +193,36 @@ fn test_query_errors_on_invalid_syntax() {
             ]
             .join("\n")
         );
+
+        // MISSING keyword with full pattern
+        assert_eq!(
+            Query::new(
+                &get_language("c"),
+                r"(MISSING (function_declarator (identifier))) "
+            )
+            .unwrap_err()
+            .message,
+            [
+                r"(MISSING (function_declarator (identifier))) ",
+                r"         ^",
+            ]
+            .join("\n")
+        );
+
+        // MISSING keyword with multiple identifiers
+        assert_eq!(
+            Query::new(
+                &get_language("c"),
+                r"(MISSING function_declarator function_declarator) "
+            )
+            .unwrap_err()
+            .message,
+            [
+                r"(MISSING function_declarator function_declarator) ",
+                r"                             ^",
+            ]
+            .join("\n")
+        );
     });
 }
 
@@ -771,7 +801,20 @@ fn test_query_matches_capturing_error_nodes() {
 fn test_query_matches_capturing_missing_nodes() {
     allocations::record(|| {
         let language = get_language("javascript");
-        let query = Query::new(&language, "(MISSING) @missing-node").unwrap();
+        let query = Query::new(
+            &language,
+            r#"
+            (MISSING
+              ; Comments should be valid
+            ) @missing-node
+            (MISSING
+              ; Comments should be valid
+              ";"
+              ; Comments should be valid
+              ) @missing-semicolon
+            "#,
+        )
+        .unwrap();
 
         // Missing anonymous nodes
         assert_query_matches(
@@ -781,14 +824,18 @@ fn test_query_matches_capturing_missing_nodes() {
             x = function(a) { b; } function(c) { d; }
             //                    ^ MISSING semicolon here
             ",
-            &[(0, vec![("missing-node", "")])],
+            &[
+                (0, vec![("missing-node", "")]),
+                (1, vec![("missing-semicolon", "")]),
+            ],
         );
 
         let language = get_language("c");
         let query = Query::new(
             &language,
-            "(field_expression (identifier) (MISSING) @field-ident)
-            (MISSING) @broad-missing",
+            "(MISSING field_identifier) @missing-field-ident
+            (MISSING identifier) @missing-ident
+            (MISSING) @missing-anything",
         )
         .unwrap();
 
@@ -809,9 +856,10 @@ fn test_query_matches_capturing_missing_nodes() {
             }
             ",
             &[
-                (0, vec![("field-ident", "")]),
-                (1, vec![("broad-missing", "")]),
-                (1, vec![("broad-missing", "")]),
+                (0, vec![("missing-field-ident", "")]),
+                (2, vec![("missing-anything", "")]),
+                (1, vec![("missing-ident", "")]),
+                (2, vec![("missing-anything", "")]),
             ],
         );
     });
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -768,6 +768,55 @@ fn test_query_matches_capturing_error_nodes() {
 }
 
 #[test]
+fn test_query_matches_capturing_missing_nodes() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+        let query = Query::new(&language, "(MISSING) @missing-node").unwrap();
+
+        // Missing anonymous nodes
+        assert_query_matches(
+            &language,
+            &query,
+            "
+            x = function(a) { b; } function(c) { d; }
+            //                    ^ MISSING semicolon here
+            ",
+            &[(0, vec![("missing-node", "")])],
+        );
+
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            "(field_expression (identifier) (MISSING) @field-ident)
+            (MISSING) @broad-missing",
+        )
+        .unwrap();
+
+        // Missing named nodes
+        assert_query_matches(
+            &language,
+            &query,
+            "
+            int main() {
+              if (a.) {
+              //    ^ MISSING field_identifier here
+                b();
+                c();
+
+                if (*) d();
+              }
+            }
+            ",
+            &[
+                (0, vec![("field-ident", "")]),
+                (1, vec![("broad-missing", "")]),
+                (1, vec![("broad-missing", "")]),
+            ],
+        );
+    });
+}
+
+#[test]
 fn test_query_matches_with_extra_children() {
     allocations::record(|| {
         let language = get_language("ruby");
--- a/docs/section-2-using-parsers.md
+++ b/docs/section-2-using-parsers.md
@@ -617,6 +617,24 @@ For example, this pattern would match any node inside a call:
 (call (_) @call.inner)
 ```
 
+#### Special Nodes
+
+When the parser encounters text it does not recognize, it represents this node
+as `(ERROR)` in the syntax tree. These error nodes can be queried just like
+normal nodes:
+
+```scheme
+(ERROR) @error-node
+```
+
+Similarly, if a parser is able to recover from erroneous text by inserting a missing token and then reducing, it will insert that missing node in the final tree so long as that tree has the lowest error cost. These missing nodes appear as seemingly normal nodes in the tree, but they are zero tokens wide, and are a property of the actual terminal node that was inserted, instead of being its own kind of node. These special missing nodes can be queried using `(MISSING)`:
+
+```scheme
+(MISSING) @missing-node
+```
+
+This is useful when attempting to detect all syntax errors in a given parse tree, since these missing node are not captured by `(ERROR)` queries.
+
 #### Anchors
 
 The anchor operator, `.`, is used to constrain the ways in which child patterns are matched. It has different behaviors depending on where it's placed inside a query.
--- a/docs/section-2-using-parsers.md
+++ b/docs/section-2-using-parsers.md
@@ -633,7 +633,12 @@ Similarly, if a parser is able to recover from erroneous text by inserting a mis
 (MISSING) @missing-node
 ```
 
-This is useful when attempting to detect all syntax errors in a given parse tree, since these missing node are not captured by `(ERROR)` queries.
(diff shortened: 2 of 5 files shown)


ribru17:
Wasm tests fail with the following:

```

 Exception during run: Error: Cannot find module '/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/tree-sitter.js'. Please verify that the package.json has a valid "main" entry
    at tryPackage (node:internal/modules/cjs/loader:494:19)
    at Function._findPath (node:internal/modules/cjs/loader:795:18)
    at Function._resolveFilename (node:internal/modules/cjs/loader:1235:27)
    at Function._load (node:internal/modules/cjs/loader:1075:27)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)
    at Module.require (node:internal/modules/cjs/loader:1340:12)
    at require (node:internal/modules/helpers:141:16)
    at Object.<anonymous> (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/test/helper.js:1:16)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1689:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)
    at cjsLoader (node:internal/modules/esm/translators:263:5)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:196:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:268:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:543:26)
    at async formattedImport (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/node_modules/mocha/lib/nodejs/esm-utils.js:9:14)
    at async exports.requireOrImport (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/node_modules/mocha/lib/nodejs/esm-utils.js:42:28)
    at async exports.loadFilesAsync (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/node_modules/mocha/lib/nodejs/esm-utils.js:100:20)
    at async singleRun (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/node_modules/mocha/lib/cli/run-helpers.js:162:3)
    at async exports.handler (/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/node_modules/mocha/lib/cli/run.js:375:5) {
  code: 'MODULE_NOT_FOUND',
  path: '/home/rileyb/Documents/CodeProjects/tree-sitter/lib/binding_web/package.json',
  requestPath: '..'
```

I feel like this is unrelated?

ribru17:
Ok `make test` passes, but `make test_wasm` does not because the `generate-fixtures-wasm` target does not exist. I believe this is why the above wasm output happened. I don't know how to proceed from here, or how other PRs were passing these tests to be honest

ObserverOfTime:
```diff
--- a/Makefile
+++ b/Makefile
@@ -99,8 +99,8 @@ test:
 	cargo xtask generate-fixtures
 	cargo xtask test
 
-test_wasm:
-	cargo xtask generate-fixtures-wasm
+test-wasm:
+	cargo xtask generate-fixtures --wasm
 	cargo xtask test-wasm
```

ribru17:
This does not fix the error of `tree-sitter.js` not existing. Was this deleted recently?

818Nawaf (approved):

amaanq:
Sorry for taking a while to get to this. I don't think we should have a new builtin symbol for missing nodes, since they are *not* represented by their own node type, but rather by a field in each subtree. I think this needs to be reworked, in the query parser when you detect that there's no symbol name, meaning just a generic `(MISSING)` node, that can be represented by the wildcard symbol which must be a missing node. "Missing" nodes are not a separate construct on their own, but rather are a property of a given symbol.

ribru17:
That makes sense! I have amended the PR to reflect your feedback.

Initially I added the builtin missing symbol as I thought it would simplify the code, but it looks like not having it even gives less LOC :)

cli/src/tests/query_test.rs
```diff
@@ -737,6 +737,57 @@ fn test_query_matches_capturing_error_nodes() {
...
+fn test_query_matches_capturing_missing_nodes() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+        let query = Query::new(
+            &language,
+            "
+            (MISSING) @missing-node
+            ",
+        )
+        .unwrap();
```
amaanq, line 750: single line
```suggestion
        let query = Query::new(
            &language,
            "(MISSING) @missing-node",
        )
        .unwrap();
```

rustfmt might make the whole call on one line then

cli/src/tests/query_test.rs
```diff
@@ -737,6 +737,57 @@ fn test_query_matches_capturing_error_nodes() {
...
+
+        // Missing anonymous nodes
+        assert_query_matches(
+            &language,
+            &query,
+            "if ({a: 'b'} {c: 'd'}) {
+  x = function(a) { b; } function(c) { d; }
+}",
```
amaanq, line 758: Can we comment what the missing node is and where with a carat? It's not immediately obvious to anyone else *what* exactly is missing in this test. Also, try to minimize the test case, there's too much going on that is not needed to test that querying a missing node works.

cli/src/tests/query_test.rs
```diff
@@ -737,6 +737,57 @@ fn test_query_matches_capturing_error_nodes() {
...
+            &query,
+            "int main() {
+  if (a.) {
+    b();
+    c();
+
+    if (*) d();
+  }
+}",
+            &[
```
amaanq, line 782: Unlike the prior comment, this test case is a lot more minimal, let's just comment where the missing nodes are and what is missing, since these missing nodes are ZWTs and we have no clue what's missing.

docs/section-2-using-parsers.md
```diff
@@ -615,6 +615,28 @@ For example, this pattern would match any node inside a call:
...
+normal nodes:
+
+```scheme
+(ERROR) @the-error
+```
+
+Similarly, if a parser is almost able to parse a document but it detects a
+missing node, it will insert that missing node into the tree to preserve the
+rest of the document. These missing nodes appear as regular nodes in the tree
+(but with zero width), and can be queried using the `(MISSING)` alias:
```
amaanq, line 631: Adjust the copy

```suggestion
Similarly, if a parser is able to recover from erroneous text by inserting a missing token and then reducing, it will insert that missing node in the final tree so long as that tree has the lowest error cost. These missing nodes appear as seemingly normal nodes in the tree, but they are zero tokens wide, and are a property of the actual terminal node that was inserted, instead of being its own kind of node. These special missing nodes can be queried using `(MISSING)`:
```

docs/section-2-using-parsers.md
```diff
@@ -615,6 +615,28 @@ For example, this pattern would match any node inside a call:
...
 ```
 
+#### Special Nodes
+
+When the parser encounters text it does not recognize, it represents this node
+as `(ERROR)` in the syntax tree. These error nodes can be queried just like
+normal nodes:
+
+```scheme
+(ERROR) @the-error
```
amaanq, line 625: Consistency
```suggestion
(ERROR) @error-node
```

docs/section-2-using-parsers.md
```diff
@@ -615,6 +615,28 @@ For example, this pattern would match any node inside a call:
...
+missing node, it will insert that missing node into the tree to preserve the
+rest of the document. These missing nodes appear as regular nodes in the tree
+(but with zero width), and can be queried using the `(MISSING)` alias:
+
+```scheme
+(MISSING) @missing-node
+```
+
+This is useful when trying to detect all syntax errors, since missing nodes will
+not be covered by `(ERROR)` queries.
```
amaanq, line 638:
```suggestion
This is useful when attempting to detect all syntax errors in a given parse tree, since these missing node are not captured by `(ERROR)` queries.
```

lib/binding_web/test/language-test.js
```diff
@@ -36,7 +36,7 @@ describe('Language', () => {
     });
 
     it('handles invalid inputs', () => {
-      assert.equal(null, JavaScript.nodeTypeForId(-1));
+      assert.equal(null, JavaScript.nodeTypeForId(-12));
```
amaanq, line 39: shouldn't need this

lib/src/query.c
```diff
@@ -328,6 +329,7 @@ static const TSQueryError PARENT_DONE = -1;
 static const uint16_t PATTERN_DONE_MARKER = UINT16_MAX;
 static const uint16_t NONE = UINT16_MAX;
 static const TSSymbol WILDCARD_SYMBOL = 0;
+static const TSSymbol MISSING_SYMBOL = ts_builtin_sym_error - 1;
```
amaanq, line 332: remove?

lib/src/query.c
```diff
@@ -2323,6 +2325,8 @@ static TSQueryError ts_query__parse_pattern(
         // Parse the wildcard symbol
         if (length == 1 && node_name[0] == '_') {
           symbol = WILDCARD_SYMBOL;
+        } else if (!strncmp(node_name, "MISSING", length)) {
+          symbol = MISSING_SYMBOL;
```
amaanq, line 2329: instead of "MISSING_SYMBOL" (this is semantically *incorrect*), just add a bool "is_missing" with symbol.

xtask/src/test.rs
```diff
@@ -122,8 +123,32 @@ pub fn run_wasm() -> Result<()> {
...
     }
 
-    let output = Command::new(npm).arg("test").output()?;
-    bail_on_err(&output, &format!("Failed to run {npm} test"))?;
+    println!("Running {npm} test...");
+    let mut child = Command::new(npm)
+        .arg("test")
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()?;
```
amaanq, line 131: Spawn inherits std{in,out,err} from the parent, so piping them and then later printing them shouldn't be necessary

xtask/src/test.rs
```diff
@@ -122,8 +123,32 @@ pub fn run_wasm() -> Result<()> {
...
+        let reader = io::BufReader::new(stderr);
+        for line in reader.lines() {
+            eprintln!("{}", line?);
+        }
+    }
+
+    let status = child.wait()?;
+    if !status.success() {
+        anyhow::bail!("{npm} test failed");
+    }
```
amaanq, line 151: We have `bail_on_err`, so actually you could call `.output` to hide stdout/err by default, only print err and exit if there's an error, and then print stdout if it's fine. Look at `run_wasm_exports`

lib/src/query.c
```diff
@@ -3758,6 +3767,9 @@ static inline bool ts_query_cursor__advance(
             }
             if (!has_supertype) node_does_match = false;
           }
+          if (step->is_missing) {
+            node_does_match = is_missing;
+          }
```
amaanq, line 3772: I would think this should be higher up, because this would overwrite this being false in case there's later siblings and this is supposed to be the last child. I would put it inside the if/else check of the wildcard symbol, basically adding the missing part onto the conditional in each branch.

amaanq (changes requested):

lib/binding_web/test/language-test.js
ribru17: Currently, this test gives

```
AssertionError: expected null to equal 'ERROR'
```

this is because the `ts_builtin_sym_error` is equal to `-1`. Perhaps a separate PR should assert the node type id is nonnegative in the web bindings (all bindings?)

amaanq: Weird, I can't reproduce this at all, with your branch or master...

cli/src/tests/query_test.rs
```diff
@@ -737,6 +737,55 @@ fn test_query_matches_capturing_error_nodes() {
...
+                c();
+
+                if (*) d();
+              }
+            }
+            ",
+            &[
+                (0, vec![("field-ident", "")]),
+                (1, vec![("broad-missing", "")]),
+                (1, vec![("broad-missing", "")]),
```
amaanq, line 783: comment the other 2, just to be clear

lib/binding_web/test/language-test.js
ribru17: Weird... I have been having some issues with my local test running so it is probably a quirk on my end. I have removed the change

lib/src/query.c
```diff
@@ -3737,7 +3746,11 @@ static inline bool ts_query_cursor__advance(
...
           if (step->symbol == WILDCARD_SYMBOL) {
-            node_does_match = !node_is_error && (is_named || !step->is_named);
+            if (step->is_missing) {
+              node_does_match = is_missing;
+            } else {
+              node_does_match = !node_is_error && (is_named || !step->is_named);
+            }
           } else {
             node_does_match = symbol == step->symbol;
           }
```
amaanq, line 3756: One last question, didn't we want to support specific missing nodes too? like `(MISSING identifier)`?

ribru17: Of course :) Yeah that would be really helpful... would you prefer I work on it on this PR or try in a second one?

ribru17: Ok I have just gone ahead and implemented this (in a third commit, for easier change viewing)

amaanq (approved):

PR #3924, opened 2024-11-10
docs: remove CHANGELOG.md

dundargoc:
It doesn't really serve useful purpose as this information should
already available in the release notes. Furthermore, it is misleading
since it will always be out of date unless it is updated on every PR.


clason:
That would be a valid argument *if there were release notes*... 😞

PR #3926, opened 2024-11-10
doc(xtask): update doc to use xtask, not scripts

### LINKED ISSUE #3895: contributions guide in docs mentions scripts that don't exist
### Problem

Currently, the [contributing section in the documentation](https://tree-sitter.github.io/tree-sitter/contributing#testing) mentions building the WASM module and fetching fixtures for tests through scripts in the `scripts` directory, which doesn't exist.

> Before you can run the tests, you need to clone some grammars that are used for testing:
>
>     script/fetch-fixtures

After digging around, I found that it has been simplified to simply `make test` and `make test_wasm`

### Steps to reproduce

* clone repo
* read docs
* try to run `scripts/generate_fixtures`
* be confused

### Expected behavior

The contribution page should mention the Makefile

tree-sitter 0.25.0 (2b877fe0b4a67fd1a55e92dee2a45efd66118abd)

### Operating system/version

Darwin mac.lan 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:00:32 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6030 arm64

crvdgc:
Follow-up of dbe8bbf. Also removed `-l` flag since it's not used anymore.

Closes #3895.

### DIFF
--- a/cli/src/tests/corpus_test.rs
+++ b/cli/src/tests/corpus_test.rs
@@ -115,6 +115,12 @@ pub fn test_language_corpus(
     skipped: Option<&[&str]>,
     language_dir: Option<&str>,
 ) {
+    if let Some(filter) = LANGUAGE_FILTER.as_ref() {
+        if language_name != filter {
+            return;
+        }
+    }
+
     let language_dir = language_dir.unwrap_or_default();
 
     let grammars_dir = fixtures_dir().join("grammars");
@@ -341,7 +347,7 @@ fn test_feature_corpus_files() {
         let language_name = language_name.to_str().unwrap();
 
         if let Some(filter) = LANGUAGE_FILTER.as_ref() {
-            if language_name != filter.as_str() {
+            if language_name != filter {
                 continue;
             }
         }
--- a/docs/section-6-contributing.md
+++ b/docs/section-6-contributing.md
@@ -32,7 +32,7 @@ cd tree-sitter
 Optionally, build the WASM library. If you skip this step, then the `tree-sitter playground` command will require an internet connection. If you have emscripten installed, this will use your `emcc` compiler. Otherwise, it will use Docker or Podman:
 
 ```sh
-./script/build-wasm
+cargo xtask build-wasm
 ```
 
 Build the Rust libraries and the CLI:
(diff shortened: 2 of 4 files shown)


docs/section-6-contributing.md
```diff
@@ -48,48 +48,42 @@ This will create the `tree-sitter` CLI executable in the `target/release` folder
...
-script/test
+cargo xtask test
 ```
 
 Similarly, to test the WASM binding, you need to compile these parsers to WASM:
 
 ```sh
-script/generate-fixtures-wasm
-script/test-wasm
+cargo xtask generate-fixtures-wasm
```
WillLillis, line 69: This isn't a command, I think you meant `cargo xtask generate-fixtures --wasm`?

crvdgc: Fixed.

docs/section-6-contributing.md
```diff
@@ -48,48 +48,42 @@ This will create the `tree-sitter` CLI executable in the `target/release` folder
...
+cargo xtask test test_does_something
 ```
 
 You can run the tests under the debugger (either `lldb` or `gdb`) using the `-g` flag:
 
 ```sh
-script/test test_does_something -g
-```
-
-Part of the Tree-sitter test suite involves parsing the _corpus_ tests for several different languages and performing randomized edits to each example in the corpus. If you just want to run the tests for a particular _language_, you can pass the `-l` flag. And if you want to run a particular _example_ from the corpus, you can pass the `-e` flag:
```
amaanq, line 89: Hm the `-e` flag still exists, so we could retain this info minus the sentence mentioning the `-l` flag.

crvdgc: Hi, I've played with these two options a bit and discovered that maybe both `-l` and `-e` should remain.

For `-l`, the language filter was not in effect for `cli::corpus_test::test_language_corpus`. So I added an early return if the filter is set and the language is not the filtered language. The `xtask test` would set the `TREE_SITTER_LANGUAGE` environment variable to enable language filtering.

For `-e`, the correct environment variable should `TREE_SITTER_EXAMPLE_INCLUDE` rather than `TREE_SITTER_EXAMPLE`.

I've fixed both problems and tested that the environment variables are in effect for the corpus tests.

cli/src/tests/corpus_test.rs
```diff
@@ -115,6 +115,12 @@ pub fn test_language_corpus(
     skipped: Option<&[&str]>,
     language_dir: Option<&str>,
 ) {
+    if let Some(filter) = LANGUAGE_FILTER.as_ref() {
+        if language_name != filter.as_str() {
```
amaanq, line 119: Don't think you *need* the `as_str` call

```suggestion
        if language_name != filter {
```

docs/section-6-contributing.md
amaanq: Yeah that's great. `-e` was already there but `-l` was removed before the migration to `xtasks`. Thank you for the fixes!

cli/src/tests/corpus_test.rs
crvdgc: Thanks for the comment. Also fixed the part I was copying from.

amaanq:
thank you @crvdgc!

PR #3938, opened 2024-11-12
feat: get supertype info

### LINKED ISSUE #2081: Supertypes in the Rust API?
Hey!

With the same context as https://github.com/tree-sitter/tree-sitter/pull/2080; I'm trying to match on all `_expression` nodes to replace them with a placeholder.

However, I can't find a way to access a node's supertypes from the [`Node`](https://docs.rs/tree-sitter/0.20.9/tree_sitter/struct.Node.html).

Am I missing something?

ribru17:
Introduces a new function that takes in a supertype symbol and returns all associated subtypes. Can be used by query.c to give better errors for invalid subtypes, as well as downstream applications like the query LSP to give better diagnostics.

Fixes #2081

### DIFF
--- a/cli/generate/src/render.rs
+++ b/cli/generate/src/render.rs
@@ -1,6 +1,6 @@
 use std::{
     cmp,
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     fmt::Write,
     mem::swap,
 };
@@ -9,6 +9,7 @@ use super::{
     build_tables::Tables,
     grammars::{ExternalToken, LexicalGrammar, SyntaxGrammar, VariableType},
     nfa::CharacterSet,
+    node_types::ChildType,
     rules::{Alias, AliasMap, Symbol, SymbolType, TokenSet},
     tables::{
         AdvanceAction, FieldLocation, GotoAction, LexState, LexTable, ParseAction, ParseTable,
@@ -80,6 +81,8 @@ struct Generator {
     reserved_word_sets: Vec<TokenSet>,
     reserved_word_set_ids_by_parse_state: Vec<usize>,
     field_names: Vec<String>,
+    supertype_symbol_map: BTreeMap<Symbol, Vec<ChildType>>,
+    supertype_map: BTreeMap<String, Vec<ChildType>>,
 
     #[allow(unused)]
     abi_version: usize,
@@ -114,6 +117,10 @@ impl Generator {
 
         self.add_non_terminal_alias_map();
         self.add_primary_state_id_list();
+
+        if self.abi_version >= ABI_VERSION_WITH_RESERVED_WORDS && !self.supertype_map.is_empty() {
+            self.add_supertype_map();
+        }
 
         let buffer_offset_before_lex_functions = self.buffer.len();
 
@@ -224,33 +231,24 @@ impl Generator {
             for alias in &production_info.alias_sequence {
                 // Generate a mapping from aliases to C identifiers.
                 if let Some(alias) = &alias {
-                    let existing_symbol = self.parse_table.symbols.iter().copied().find(|symbol| {
-                        self.default_aliases.get(symbol).map_or_else(
-                            || {
-                                let (name, kind) = self.metadata_for_symbol(*symbol);
-                                name == alias.value && kind == alias.kind()
-                            },
-                            |default_alias| default_alias == alias,
-                        )
-                    });
-
                     // Some aliases match an existing symbol in the grammar.
-                    let alias_id = if let Some(existing_symbol) = existing_symbol {
-                        self.symbol_ids[&self.symbol_map[&existing_symbol]].clone()
-                    }
-                    // Other aliases don't match any existing symbol, and need their own
-                    // identifiers.
-                    else {
-                        if let Err(i) = self.unique_aliases.binary_search(alias) {
-                            self.unique_aliases.insert(i, alias.clone());
+                    let alias_id =
+                        if let Some(existing_symbol) = self.symbols_for_alias(alias).first() {
+                            self.symbol_ids[&self.symbol_map[existing_symbol]].clone()
                         }
+                        // Other aliases don't match any existing symbol, and need their own
+                        // identifiers.
+                        else {
+                            if let Err(i) = self.unique_aliases.binary_search(alias) {
+                                self.unique_aliases.insert(i, alias.clone());
+                            }
 
-                        if alias.is_named {
-                            format!("alias_sym_{}", self.sanitize_identifier(&alias.value))
-                        } else {
-                            format!("anon_alias_sym_{}", self.sanitize_identifier(&alias.value))
-                        }
-                    };
+                            if alias.is_named {
+                                format!("alias_sym_{}", self.sanitize_identifier(&alias.value))
+                            } else {
+                                format!("anon_alias_sym_{}", self.sanitize_identifier(&alias.value))
+                            }
+                        };
 
                     self.alias_ids.entry(alias.clone()).or_insert(alias_id);
                 }
@@ -288,6 +286,18 @@ impl Generator {
                 self.reserved_word_sets.len() - 1
             };
             self.reserved_word_set_ids_by_parse_state.push(id);
+        }
+
+        if self.abi_version >= ABI_VERSION_WITH_RESERVED_WORDS {
+            for (supertype, subtypes) in &self.supertype_symbol_map {
+                if let Some(supertype) = self.symbol_ids.get(supertype) {
+                    self.supertype_map
+                        .entry(supertype.clone())
+                        .or_insert_with(|| subtypes.clone());
+                }
+            }
+
+            self.supertype_symbol_map.clear();
         }
 
         // Determine which states should use the "small state" representation, and which should
@@ -404,6 +414,7 @@ impl Generator {
             "#define PRODUCTION_ID_COUNT {}",
             self.parse_table.production_infos.len()
         );
+        add_line!(self, "#define SUPERTYPE_COUNT {}", self.supertype_map.len());
         add_line!(self, "");
     }
 
@@ -689,7 +700,7 @@ impl Generator {
 
         add_line!(
             self,
-            "static const TSFieldMapSlice ts_field_map_slices[PRODUCTION_ID_COUNT] = {{",
+            "static const TSMapSlice ts_field_map_slices[PRODUCTION_ID_COUNT] = {{",
         );
         indent!(self);
         for (production_id, (row_id, length)) in field_map_ids.into_iter().enumerate() {
@@ -719,6 +730,83 @@ impl Generator {
                     add!(self, ", .inherited = true");
                 }
                 add!(self, "}},\n");
+            }
+            dedent!(self);
+        }
+
+        dedent!(self);
+        add_line!(self, "}};");
+        add_line!(self, "");
+    }
+
+    fn add_supertype_map(&mut self) {
+        add_line!(
+            self,
+            "static const TSSymbol ts_supertype_symbols[SUPERTYPE_COUNT] = {{"
+        );
+        indent!(self);
+        for supertype in self.supertype_map.keys() {
+            add_line!(self, "{supertype},");
+        }
+        dedent!(self);
+        add_line!(self, "}};\n");
+
+        add_line!(
+            self,
+            "static const TSMapSlice ts_supertype_map_slices[] = {{",
+        );
+        indent!(self);
+        let mut row_id = 0;
+        let mut supertype_ids = vec![0];
+        let mut supertype_string_map = BTreeMap::new();
+        for (supertype, subtypes) in &self.supertype_map {
+            supertype_string_map.insert(
+                supertype,
+                subtypes
+                    .iter()
+                    .flat_map(|s| match s {
+                        ChildType::Normal(symbol) => vec![self.symbol_ids.get(symbol).cloned()],
+                        ChildType::Aliased(alias) => {
+                            self.alias_ids.get(alias).cloned().map_or_else(
+                                || {
+                                    self.symbols_for_alias(alias)
+                                        .into_iter()
+                                        .map(|s| self.symbol_ids.get(&s).cloned())
+                                        .collect()
+                                },
+                                |a| vec![Some(a)],
+                            )
+                        }
+                    })
+                    .flatten()
+                    .collect::<BTreeSet<String>>(),
+            );
+        }
+        for (supertype, subtypes) in &supertype_string_map {
+            let length = subtypes.len();
+            add_line!(
+                self,
+                "[{supertype}] = {{.index = {row_id}, .length = {length}}},",
+            );
+            row_id += length;
+            supertype_ids.push(row_id);
+        }
+        dedent!(self);
+        add_line!(self, "}};");
+        add_line!(self, "");
+
+        add_line!(
+            self,
+            "static const TSSymbol ts_supertype_map_entries[] = {{",
+        );
+        indent!(self);
+        for (i, (_, subtypes)) in supertype_string_map.iter().enumerate() {
+            let row_index = supertype_ids[i];
+            add_line!(self, "[{row_index}] =");
+            indent!(self);
+            for subtype in subtypes {
+                add_whitespace!(self);
+                add!(self, "{subtype},\n");
             }
             dedent!(self);
         }
@@ -1462,6 +1550,9 @@ impl Generator {
         add_line!(self, ".state_count = STATE_COUNT,");
         add_line!(self, ".large_state_count = LARGE_STATE_COUNT,");
         add_line!(self, ".production_id_count = PRODUCTION_ID_COUNT,");
+        if self.abi_version >= ABI_VERSION_WITH_RESERVED_WORDS {
+            add_line!(self, ".supertype_count = SUPERTYPE_COUNT,");
+        }
         add_line!(self, ".field_count = FIELD_COUNT,");
         add_line!(
             self,
@@ -1482,6 +1573,11 @@ impl Generator {
             add_line!(self, ".field_names = ts_field_names,");
             add_line!(self, ".field_map_slices = ts_field_map_slices,");
             add_line!(self, ".field_map_entries = ts_field_map_entries,");
+        }
+        if !self.supertype_map.is_empty() && self.abi_version >= ABI_VERSION_WITH_RESERVED_WORDS {
+            add_line!(self, ".supertype_map_slices = ts_supertype_map_slices,");
+            add_line!(self, ".supertype_map_entries = ts_supertype_map_entries,");
+            add_line!(self, ".supertype_symbols = ts_supertype_symbols,");
         }
         add_line!(self, ".symbol_metadata = ts_symbol_metadata,");
         add_line!(self, ".public_symbol_map = ts_symbol_map,");
@@ -1633,6 +1729,23 @@ impl Generator {
                 (&token.name, token.kind)
             }
         }
+    }
+
+    fn symbols_for_alias(&self, alias: &Alias) -> Vec<Symbol> {
+        self.parse_table
+            .symbols
+            .iter()
+            .copied()
+            .filter(move |symbol| {
+                self.default_aliases.get(symbol).map_or_else(
+                    || {
+                        let (name, kind) = self.metadata_for_symbol(*symbol);
+                        name == alias.value && kind == alias.kind()
+                    },
+                    |default_alias| default_alias == alias,
+                )
+            })
+            .collect()
     }
 
     fn sanitize_identifier(&self, name: &str) -> String {
@@ -1802,6 +1915,7 @@ pub fn render_c_code(
     lexical_grammar: LexicalGrammar,
     default_aliases: AliasMap,
     abi_version: usize,
+    supertype_symbol_map: BTreeMap<Symbol, Vec<ChildType>>,
 ) -> String {
     assert!(
         (ABI_VERSION_MIN..=ABI_VERSION_MAX).contains(&abi_version),
@@ -1819,6 +1933,7 @@ pub fn render_c_code(
         lexical_grammar,
         default_aliases,
         abi_version,
+        supertype_symbol_map,
         ..Default::default()
     }
     .generate()
--- a/cli/src/tests/language_test.rs
+++ b/cli/src/tests/language_test.rs
@@ -95,3 +95,100 @@ fn test_symbol_metadata_checks() {
         }
     }
 }
+
+#[test]
+fn test_supertypes() {
+    let language = get_language("rust");
+    let supertypes = language.supertypes();
+
+    assert_eq!(supertypes.len(), 5);
+    assert_eq!(
+        supertypes
+            .iter()
+            .filter_map(|&s| language.node_kind_for_id(s))
+            .map(|s| s.to_string())
+            .collect::<Vec<String>>(),
+        vec![
+            "_expression",
+            "_literal",
+            "_literal_pattern",
+            "_pattern",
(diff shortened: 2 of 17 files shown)


818Nawaf (approved):

amaanq:
@ribru17 I'll close this out until it's ready, no offense to you but the constant notification everytime you push is a _little_ annoying 😅

clason:
You can unsubscribe, you know. Pushing to a closed PR breaks GitHub.

clason:
(Marking as ready was me fat fingering on mobile.)

amaanq:
> You can unsubscribe, you know. Pushing to a closed PR breaks GitHub. 

I would like to be subscribed, *when* it's ready.

ribru17:
Yeah I think it is a really annoying GitHub issue, I'll re-mark as draft and stop pushes haha. (Though I think this is almost ready.) Sorry!

clason:
Yes, it's a GH issue. But pushing to a closed PR prevents it from being opened again.

ribru17:
Alright from my perspective it looks like this is pretty much done, but I am having some perplexing issues that look like some data offset weirdness (i.e. `field_count` returns `max_alias_sequence_length`, `supertype_count` returns `field_count`, etc.). May need some help with that because I can't tell where it's coming from, but in my limited understanding I feel like 90% of the heavy lifting is done

ribru17:
Note that the nvim-treesitter CI check here will inherently fail because this changes the `TSLanguage` struct. In the future it would be nice to run this check building Neovim with the local version of TS rather than just unpacking the Neovim tarball, but this sounds like a job for the mighty @dundargoc :) Running Neovim with the updated local version of TS on my branch shows things working still.

clason:
> Note that the nvim-treesitter CI check here will inherently fail because this changes the TSLanguage struct. 

Which means that nvim-treesitter will fail in the wild, too, so the check is valuable. (Note that the `generate` workflow explicitly generates parsers with ABI 14 to match the tree-sitter lib Neovim is built with.)

lib/include/tree_sitter/api.h
```diff
@@ -1166,6 +1166,26 @@ uint32_t ts_language_symbol_count(const TSLanguage *self);
...
+const TSSymbol *ts_language_supertypes(const TSLanguage *self);
+
+/**
+ * Get a list of all subtype symbol names for a given supertype symbol.
+*/
+void ts_language_supertype_map(
+  const TSLanguage *self,
+  TSSymbol supertype,
+  const char *const **start,
+  const char *const **end
```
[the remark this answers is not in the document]
ribru17: Good point. I just copied `ts_language_field_map` without thinking but that is a better implementation

ribru17:
@ObserverOfTime do you have any idea why the rust bindings have some (I am guessing) data offset issues mentioned in this https://github.com/tree-sitter/tree-sitter/pull/3938#issuecomment-2477906461? I ported these changes to Neovim locally and everything worked when just called from C directly

ObserverOfTime:
Nope.

ribru17:
Nvm I just forgot to regenerate fixtures

cli/generate/src/render.rs
```diff
@@ -694,6 +703,73 @@ impl Generator {
...
     }
 
+    fn add_supertype_map(&mut self) {
+        add_line!(
+            self,
+            "static const TSSymbol ts_supertypes[SUPERTYPE_COUNT] = {{"
+        );
+        indent!(self);
+        for supertype in self.supertype_map.keys() {
+            add_line!(self, "{},", supertype);
```
WillLillis, line 713: Really pikcy nit, but I think `supertype` can be moved inside the braces to match the style in the in the rest of this function.

ribru17: Thanks!

ribru17:
@amaanq this is ready now, sorry for all the pushes :)

lib/include/tree_sitter/api.h
```diff
@@ -1166,6 +1166,25 @@ uint32_t ts_language_symbol_count(const TSLanguage *self);
...
+
+/**
+ * Get a list of all supertype symbols for the language.
+*/
+const TSSymbol *ts_language_supertypes(const TSLanguage *self);
+
+/**
+ * Get a list of all subtype symbol names for a given supertype symbol.
+*/
+uint16_t ts_language_supertype_map(
```
ribru17, line 1182: Note: should it be documented that this function returns the amount of subtype symbols? It would be helpful but the downside is that it wouldn't match the rust binding docstring (which obviously has no need to return the length, it just returns a vector)

clason:
This needs to be gated behind an ABI 15 check; tree-sitter CLI is supposed to be able to generate ABI 14 parsers on demand (which Neovim -- like any downstream in existence -- demands).

ribru17:
No, bwc doesn't matter for this project 

Serious reply: what would this look like exactly?

clason:
> No, bwc doesn't matter for this project

Yes, I am starting to get that impression. If this is you saying "fuck downstream, fuck you, and fuck the horse you rode in on", message received. I might as well archive nvim-treesitter then.

clason:
(I am very serious, by the way.)

ribru17:
> I might as well archive nvim-treesitter then.

No more work! :partying_face: 

> what would this look like exactly?

Just to be explicit since tone doesn't translate over text, I meant this out of genuine curiosity/seeking advice

ribru17:
> (I am very serious, by the way.)

You're bluffing! Archiving nvim-treesitter is madness. At least without considering forking tree-sitter first (neo-tree-sitter? problematic for when tree-sitter needs `neo` bindings in 2035)

amaanq:
Well this _should_ be gated behind ABI 15 tbh, similar to how the primary states are gated behind ABI 14

Yuho1985:
**translation**

ribru17:
Of course, sorry didn't mean to send the impression that I was dismissing that statement. Just unsure on the literal logic of how to do that :sweat_smile: but I will look into how it was done with primary states, thanks

ribru17:
Building neovim with local TS changes show no segfaults with lalrpop, not sure why it's failing here. (Also the changes are gated past that check's ABI anyway?) In any case the checks are now passing for all other parsers, with ABI gate in place.

lib/include/tree_sitter/api.h
ObserverOfTime: Just edit the Rust docstring then?

cli/generate/src/render.rs
```diff
@@ -115,7 +116,7 @@ impl Generator {
         self.add_non_terminal_alias_map();
         self.add_primary_state_id_list();
 
-        if !self.supertype_map.is_empty() {
+        if !self.supertype_map.is_empty() && self.abi_version >= ABI_VERSION_WITH_SUPERTYPE_INFO {
```
ObserverOfTime, line 119: Can the entire `supertype_map` initialisation be gated behind the ABI version check?
And shouldn't the C API also have a similar check?

clason:
>  In any case the checks are now passing for all other parsers, with ABI gate in place.

No, that's not true. First, the generate step fails immediately, so there's definitely something wrong. (And not sure why `lalrpop` is the first -- which does not mean only one! -- to fail; it uses a custom scanner, but I'm not seeing any other relevant differences.)

More importantly, I just tested `tree-sitter generate --abi 14` on your branch and the generated files are full of the new supertype contents (even though the `parser.c` claims -- wrongly -- `ABI 14`).

ribru17:
Thanks for the review, the help is appreciated, truly. nvim-ts tests are passing (except for one due to curl timeout) and core tests are all over the place, but failures look like out of date fixtures (I have not confirmed this!)

cli/generate/src/build_tables/build_parse_table.rs
```diff
@@ -899,12 +907,13 @@ impl<'a> ParseTableBuilder<'a> {
...
 
-        if let Some(index) = self
-            .parse_table
-            .production_infos
-            .iter()
-            .position(|seq| *seq == production_info)
-        {
+        if let Some(index) = self.parse_table.production_infos.iter().position(|seq| {
+            // Only check for alias sequence and field map equality. Otherwise the presence of the
+            // supertype map my increment the ProductionInfoId by one which throws off the indices
```
WillLillis, line 912: my -> may here?

ribru17: :facepalm: thx!

ribru17:
@amaanq This is ready now from what I can see. Previous failures were due to incorrect wasm store address locations, now that has been fixed and all failing tests are using the cached fixtures, while all tests which regenerated fixtures are passing

ribru17:
Windows ci failure os related to this issue I believe: https://github.com/graalvm/setup-graalvm/issues/66

BobbieCarroll:
Set up new password

lib/include/tree_sitter/api.h
```diff
@@ -1166,6 +1166,26 @@ uint32_t ts_language_symbol_count(const TSLanguage *self);
...
 
+/**
+ * Get the number of supertype symbols in the language.
+*/
+uint32_t ts_language_supertype_count(const TSLanguage *self);
+
+/**
+ * Get a list of all supertype symbols for the language.
+*/
+const TSSymbol *ts_language_supertypes(const TSLanguage *self);
```
amaanq, line 1177: Instead of exposing a function for just getting the # of supertypes (which I question the usefulness of), why not instead have the caller pass in a pointer to a `uint32_t`, called `length`, which is written to in the function body. This will let the user know how many supertypes there are, see `ts_tree_included_ranges` for an example of what I mean. The main reason being this will set the precedence for us to add `_count` functions for other node types, which we don't want to add.

lib/include/tree_sitter/api.h
```diff
@@ -1166,6 +1166,26 @@ uint32_t ts_language_symbol_count(const TSLanguage *self);
...
+
+/**
+ * Get a list of all subtype symbol names for a given supertype symbol. Returns
+ * the size of said list.
+*/
+uint16_t ts_language_supertype_map(
+  const TSLanguage *self,
+  TSSymbol supertype,
+  const char *const **start
```
amaanq, line 1186: I don't think we should be writing C-style strings, as this function is giving us the subtype symbol names, of which we can't really do much with. It would make a lot more sense imo, to store the supertype info in the parser as a map of symbol ids to an array of symbol ids, which is a lot better for the parser binary size and memory usage (esp in wasm). With the symbol ids, they are a lot more useful to us and we can always fetch the symbol names with this array of ids.

lib/src/language.c
```diff
@@ -24,6 +24,29 @@ uint32_t ts_language_state_count(const TSLanguage *self) {
   return self->state_count;
 }
 
+uint32_t ts_language_supertype_count(const TSLanguage *self) {
+  return self->version >= LANGUAGE_VERSION_WITH_SUPERTYPE_INFO ? self->supertype_count : 0;
```
amaanq, line 28: Gate this with `LANGUAGE_VERSION_WITH_RESERVED_WORDS` instead, and remove the other define

lib/src/language.c
```diff
@@ -24,6 +24,29 @@ uint32_t ts_language_state_count(const TSLanguage *self) {
...
+) {
+  if (self->version < LANGUAGE_VERSION_WITH_SUPERTYPE_INFO || !ts_language_symbol_metadata(self, supertype).supertype) {
+    *start = NULL;
+    return 0;
+  }
+
+  TSSupertypeMapSlice slice = self->supertype_map_slices[supertype];
+  *start = &self->supertype_map_entries[slice.index];
+  return slice.length;
+}
```
amaanq, line 48: Like I mentioned earlier, let's return an array of the symbol ids instead.

lib/src/parser.h
```diff
@@ -31,6 +31,11 @@ typedef struct {
   uint16_t length;
 } TSFieldMapSlice;
 
+typedef struct {
+  uint16_t index;
+  uint16_t length;
+} TSSupertypeMapSlice;
```
amaanq, line 37: Instead of duplicating a type definition, rename `TSFieldMapSlice` to `TSMapSlice`, and leave a comment explaining that it's used for looking up the index and length of either the field map slices or supertype map slices.

cli/generate/src/render.rs
```diff
@@ -1483,6 +1565,11 @@ impl Generator {
             add_line!(self, ".field_map_slices = ts_field_map_slices,");
             add_line!(self, ".field_map_entries = ts_field_map_entries,");
         }
+        if !self.supertype_map.is_empty() && self.abi_version >= ABI_VERSION_WITH_SUPERTYPE_INFO {
```
amaanq, line 1568: Use `ABI_VERSION_WITH_RESERVED_WORDS`

cli/generate/src/render.rs
```diff
@@ -1483,6 +1565,11 @@ impl Generator {
...
+            add_line!(self, ".supertype_map_slices = ts_supertype_map_slices,");
+            add_line!(self, ".supertype_map_entries = ts_supertype_map_entries,");
+            add_line!(self, ".supertypes = ts_supertypes,");
```
amaanq, line 1571: I think calling this field `supertype_symbols` would be clearer, and then the corresponding const array `ts_supertype_symbols`

cli/src/tests/language_test.rs
```diff
@@ -95,3 +95,67 @@ fn test_symbol_metadata_checks() {
         }
     }
 }
+
+#[test]
+fn test_supertype_data() {
```
amaanq, line 100: reminder (mostly for me): when you've updated the definition of subtypes_for_supertype to return the array of symbol ids, this test should be adjusted to map the symbol ids to their names and then compare that list

lib/src/language.h
```diff
@@ -12,6 +12,7 @@ extern "C" {
 
 #define LANGUAGE_VERSION_WITH_RESERVED_WORDS 15
 #define LANGUAGE_VERSION_WITH_PRIMARY_STATES 14
+#define LANGUAGE_VERSION_WITH_SUPERTYPE_INFO 15
```
amaanq, line 15: remove and use `LANGUAGE_VERSION_WITH_PRIMARY_STATES` instead

cli/generate/src/render.rs
```diff
@@ -728,6 +740,73 @@ impl Generator {
         add_line!(self, "");
     }
 
+    fn add_supertype_map(&mut self) {
+        add_line!(
+            self,
+            "static const TSSymbol ts_supertypes[SUPERTYPE_COUNT] = {{"
```
amaanq, line 746: `ts_supertype_symbols` like i mentioned earlier

cli/generate/src/build_tables/build_parse_table.rs
```diff
@@ -55,6 +55,7 @@ struct ParseTableBuilder<'a> {
     syntax_grammar: &'a SyntaxGrammar,
     lexical_grammar: &'a LexicalGrammar,
     variable_info: &'a [VariableInfo],
+    subtype_map: &'a Vec<(Symbol, Vec<ChildType>)>,
```
amaanq, line 58: After thinking about this, why do we map the symbol to `ChildType`? Like I mentioned earlier, conceptually, it'd make more sense to map the supertype symbol ids to the subtype symbol ids, thus meaning we don't actually need the symbol names. That indirection should happen outside of this relationship, via `ts_language_symbol_name`, so in here, I would think we would store a mapping of the symbol to an array of symbol ids that can be a potential subtype to the given supertype. With that in mind, please refactor this, it should clean up this section of the code quite a bit :) as I was really confused by the change in signature of `generate_node_types_json` to return the `ChildType` info, but really, we don't need that data.

cli/generate/src/render.rs
```diff
@@ -728,6 +740,73 @@ impl Generator {
...
+        let mut supertype_string_map = BTreeMap::new();
+        for (supertype, subtypes) in &self.supertype_map {
+            supertype_string_map.insert(
+                supertype,
+                subtypes
+                    .iter()
+                    .map(child_type_to_string)
+                    .collect::<BTreeSet<String>>(),
+            );
+        }
```
amaanq, line 775: Continuing on with what I said about mapping symbol ids instead, here we wouldn't need the child type info to get the symbol string name, avoiding the need for this mapping of symbol or alias to the name.

cli/generate/src/render.rs
```diff
@@ -1635,6 +1722,16 @@ impl Generator {
...
 
+    fn string_for_symbol(&self, symbol: &Symbol) -> String {
+        self.sanitize_string(
+            self.default_aliases
+                .get(symbol)
+                .map_or(self.metadata_for_symbol(*symbol).0, |alias| {
+                    alias.value.as_str()
+                }),
+        )
+    }
```
amaanq, line 1733: Shouldn't be needed with what I've echoed earlier about using symbol ids

amaanq (changes requested):
Sorry for taking so long to get to this. I'm glad you've taken this up, it'd be really nice to have supertype and subtype info programmatically in the bindings, but the only major architectural change imo is that this map should be a mapping of supertype symbol ids to subtype symbol ids, not supertype symbol ids to subtype symbol names

lib/src/parser.h
ribru17: Will the renaming of the `TSFieldMapSlice` be breaking when generating for a previous ABI?

cli/generate/src/render.rs
ribru17: I had aimed for a supertype id -> subtype ids relationship initially and met some challenges. I might need your help with it, at the time it seemed even impossible. There are some cases (when aliases are involved) that we have subtypes aliased to an otherwise non-existant symbol name. Consider this rule in `tree-sitter-rust`:

```js
    _type: $ => choice(
      $.abstract_type,
      $.reference_type,
      $.metavariable,
      $.pointer_type,
      $.generic_type,
      $.scoped_type_identifier,
      $.tuple_type,
      $.unit_type,
      $.array_type,
      $.function_type,
      $._type_identifier,
      $.macro_invocation,
      $.never_type,
      $.dynamic_type,
      $.bounded_type,
      $.removed_trait_bound,
      alias(choice(...primitiveTypes), $.primitive_type),
    ),
```

here `_type` is a supertype. `primitiveTypes` is just a list of literals. `$.primitive_type` is not a real symbol, it only exists for this alias. In this case, from the parser.c side, there is no symbol ID for `primitive_type`. Thus what do we do here in this mapping? Surely we still want to record it somewhere since we can query `(primitive_type)`

lib/src/parser.h
amaanq: Nah it shouldn't, because the layout didn't change.

If you're compiling a parser generated with the type renamed in parser.c, and you're using an old header, obviously that'd break, but that *shouldn't* happen, because everytime you regenerate, everything in `src` is "regenerated". So, they should never be "out of sync", and you can use a parser compiled with the old name and with the new name in the same application, because the layout is the exact same.

cli/generate/src/render.rs
amaanq: Ah, that's a good point...aliasing tokens is a little different. I think what we'd want here is to do one of two things:

- use the "base" token that the other tokens link back to, because they're all "hidden", in this case it'd be `anon_sym_u8` as it's just the first one
- add them all

I'm inclined to say we should add them all, as conceptually, we have n different tokens that are all aliased back to primitive type, and these tokens all have their symbol ids available for inspection

ribru17: Gotcha, that makes sense. I'm working on a solution that adds them all right now and it isn't as complicated as I thought initially

lib/src/parser.h
ribru17: Ok cool thanks, that makes sense

ribru17:
Thanks for the review, I've found a solution that maps all node ids now. The only thing I had to keep was the `ChildType` type for the subtype map, since we need the occasional `Alias` information in addition to the `Symbol`s. But I refactored that logic out of the main `node_types` generator so hopefully the code is clearer now.

ribru17:
Test failures due to out of date fixtures I believe. All failing tests use cached fixtures and all passing tests regenerated.

Side question, is there a way to manually tell CI to regenerate fixtures or ignore cache for a workflow run?

WillLillis:
> Test failures due to out of date fixtures I believe. All failing tests use cached fixtures and all passing tests regenerated.

I think you're right, it looks the cache hit on your last [force push](https://github.com/tree-sitter/tree-sitter/compare/ed45118493d3e069559d4a717ab4f3d896e53d37..98fb139c2fb07db916ea74eedc3c0465c6a5688a) because it only modified `lib/src/tree_cursor.h`.

The fixtures cache is indexed by a hash of several files, but doesn't include that one:

https://github.com/tree-sitter/tree-sitter/blob/490f79bca2bdea3334a92e0c5af0b41462473ee3/.github/actions/cache/action.yml#L19-L23

ribru17:
Ahh, thank you that makes sense. Yeah I had modified that as a mistake during some debugging... :sweat_smile: Seems like `lib/src/parser.h` should be included in that list

amaanq:
@ribru17 I cleaned up some of the code, though some notable changes are:

- I completely removed the usage of production infos to store supertype info, because conceptually, a field is associated with the children a given parent at runtime, but we don't store this relationship globally because fields are not directly linked to one set of symbols, whereas supertypes are for all supertypes (they must have one visible child only), and we're doing a lot of redundant cloning to store the exact same "global" set of supertypes for each production info, when we don't need to

- I changed the signature of `ts_language_supertype_map` to not take a double pointer, since the reason the internal field map version does is because it's used more like a range rather than a list, and also because it's a littler nicer for downstream consumers

If this looks good to you we can get this in :)

ribru17:
Looks good to me, thanks for the help!

ribru17:
One last thing to note now that we have moved the function to return symbol IDs: say we have a supertype:

```js
supertype_sym: $ => choice (/* ...stuff */ alias($.symbol, $.unique_alias)),
```

right now the map will return the symbol ID for both $.symbol and the $.unique_alias, even though we cannot query `supertype_sym/symbol`, only `supertype_sym/unique_alias`. In this case would it be possible for e.g. `query.c` to differentiate visibility in this way? Before with returning strings we only had `"unique_alias"` here

amaanq:
Yeah true, why not just keep the alias symbol? I think we'd just remove the extra logic for fetching the underlying symbol of an alias in `get_supertype_symbol_map`

ribru17:
If we do that, for the rust rule `alias(choice(...primitiveTypes), $.primitive_type),` we no longer get any symbols that correspond to the `primitive_type` alias, sadly. After inspecting `parser.c`, maybe it would be enough to keep the current implementation and read the names from `ts_symbol_names`? (And also maybe expose that data in the rust bindings)?

amaanq:
Ah right, aliasing a terminal is a little different than a non terminal (there's a slight optimization made here), so we can remove the fetching logic when getting the supertype map, and then, if there's no aliased symbol that exists in `alias_ids` (which is for non-terminal aliases), we'll just do a lookup of the parse table's symbols and fetch all the underlying terminal symbols that map to this alias.

amaanq (approved):

ribru17:
Everything is working perfectly in my tests as well, thx again for the help :smile:

amaanq:
cool, this is great to have - thanks @ribru17!

maxbrunsfeld:
@ribru17 @amaanq - Just a heads up that backwards-compatibility wasn't handled correctly in this PR, in the case of WASM. If Tree-sitter 0.25.1 tries to load a WASM-compiled language that was generated by an earlier version of Tree-sitter, a crash will occur when executing [this line](https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/wasm_store.c#L1335).

I'm going to push up a fix shortly.

ribru17:
Ack, sorry about that. Thanks for pushing a fix!

PR #3939, opened 2024-11-13
feat(cli): show parse times in testing output

### LINKED ISSUE #3929: Corpus test: time elapsed
### Problem

I'm not sure if it would be beneficial for corpus tests to emit how much time it took to run the test.

I've had a test that took a long time, and it turned out to be an undetected parsing issue that resulted in a slower parse.

When looking at inspiration for a good-looking design for an HTML report for corpus tests, I was reminded that in most testing suites, time to execute is a datum that is collected.

I wanted to start some discussion about it.

### Expected behavior

Capture and include elapsed time per test in test output.

WillLillis:
This adds timing information in two forms for the cli's `test` subcommand. By default, the CLI will now print out some timing stats at  the end of a run. For example, running `tree-sitter test` on `tree-sitter-c` now looks like:

```sh
ambiguities:
      1. ✓ pointer declarations vs expressions
      2. ✓ casts vs multiplications
      3. ✓ function-like type macros vs function calls
     ...
     83. ✓ Attributes
  types:
     84. ✓ Primitive types
     85. ✓ Type modifiers

Total parses: 85; successful parses: 85; failed parses: 0; success percentage: 100.00%; average speed: 2058 bytes/ms

syntax highlighting:
    ✓ keywords.c (3 assertions)
    ✓ names.c (20 assertions)

```

~~In addition, running `tree-sitter test --show-times` will display the parsing time for each individual test case.~~ The use case outlined in #3929 describes finding tests with slow parse rates. Simply printing the parse times isn't useful in this regard (tests vary a lot from run to run even on the same hardware). Providing the parse rate is more useful, but still leaves the user with a lot of guess work and manual checking. A solution I arrived at is to check for statistically significant slow parse rates. The process is basically as follows:

1. When `tree-sitter test` is run, parse all relevant test cases and record their adjusted (logarithmic) parse rates.
- Log parse rates are recorded in order to mitigate against the effect of outliers, which leads to a large standard deviation, making our statistical tests useless. This is just a heuristic.
2. Calculate the mean and standard deviation of all parse log parse rates.
3. Run all the tests normally. When each test is run, if its adjusted parsing rate is more than 3 standard deviations below the mean parsing rate (the "Empirical Rule"), a warning is printed alongside the test name.

By default the `--stat outliers-and-total` option is passed, meaning that running `tree-sitter test` will display a warning if any tests have a slow parse rate. If `--stat all` is passed, the parse rates for every test will be displayed as well as any outliers. Finally, passing `--stat total-only` will only display the total parsing statistics at the end.

This could look like the following for tree-sitter-c, which has one test that occasionally triggers this check:

```sh
❯ ts_dev test
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
  ambiguities:
      1. ✓ pointer declarations vs expressions
      2. ✓ casts vs multiplications
      3. ✓ function-like type macros vs function calls
     ...
     57. ✓ Top Level Empty Expression Statement -- Warning: Slow parse rate (333.333bytes/ms)
     ...
     83. ✓ Attributes
  types:
     84. ✓ Primitive types
     85. ✓ Type modifiers

Total parses: 85; successful parses: 85; failed parses: 0; success percentage: 100.00%; average speed: 2143 bytes/ms

syntax highlighting:
    ✓ keywords.c (3 assertions)
    ✓ names.c (20 assertions)
```

And here's a screenshot of the `--stat all` flag being passed for good measure:

image

Closes #3929

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -24,7 +24,7 @@ use tree_sitter_cli::{
     logger,
     parse::{self, ParseFileOptions, ParseOutput, ParseTheme},
     playground, query, tags,
-    test::{self, parse_tests, TestOptions},
+    test::{self, parse_tests, TestOptions, TestStats},
     test_highlight, test_tags, util, version, wasm,
 };
 use tree_sitter_config::Config;
@@ -274,8 +274,8 @@ struct Test {
     pub config_path: Option<PathBuf>,
     #[arg(long, help = "Force showing fields in test diffs")]
     pub show_fields: bool,
-    #[arg(long, help = "Check tests for slow parse rates")]
-    pub check_rates: bool,
+    #[arg(long, help = "Show parsing statistics")]
+    pub stat: Option<TestStats>,
     #[arg(short, long, help = "Force rebuild the parser")]
     pub rebuild: bool,
     #[arg(long, help = "Show only the pass-fail overview tree")]
@@ -938,6 +938,7 @@ impl Test {
     fn run(self, mut loader: loader::Loader, current_dir: &Path) -> Result<()> {
         let config = Config::load(self.config_path)?;
         let color = env::var("NO_COLOR").map_or(true, |v| v != "1");
+        let stat = self.stat.unwrap_or_default();
 
         loader.debug_build(self.debug_build);
         loader.force_rebuild(self.rebuild);
@@ -967,7 +968,9 @@ impl Test {
         let test_corpus_dir = test_dir.join("corpus");
         if test_corpus_dir.is_dir() {
             let languages = languages.iter().map(|(l, n)| (n.as_str(), l)).collect();
-            let outlier_stats = if self.check_rates {
+            let outlier_stats = if stat == TestStats::TotalOnly {
+                None
+            } else {
                 let mut rates = Vec::new();
                 test::get_test_parsing_rate(
                     &mut parser,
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -274,6 +274,8 @@ struct Test {
     pub config_path: Option<PathBuf>,
     #[arg(long, help = "Force showing fields in test diffs")]
     pub show_fields: bool,
+    #[arg(long, help = "Check tests for slow parse rates")]
+    pub check_rates: bool,
```
amaanq, line 278: same as parse for consistency, `stat`

cli/src/test.rs
```diff
@@ -622,6 +653,59 @@ fn count_subtests(test_entry: &TestEntry) -> usize {
...
+    Ok(())
+}
+
+// Parse time is interpreted in μs before converting to ms to avoid truncation issues
+// Parse rates often have several outliers, leading to a large standard deviation. Taking
+// the log of these rates serves to "flatten" out the distribution, yielding a more
+// usable standard deviation for finding statistically significant slow parse rates
+// NOTE: This is just a heuristic
+#[must_use]
+pub fn adjusted_parse_rate(tree: &tree_sitter::Tree, parse_time: std::time::Duration) -> f64 {
```
amaanq, line 705: import these, `Tree`, and `Duration` are clear enough to not necessitate the full qualifier here imo. But this is really cool!

cli/src/main.rs
```diff
@@ -955,10 +957,36 @@ impl Test {
...
+                    parse_tests(&test_corpus_dir)?,
+                    &mut rates,
+                    &languages,
+                )?;
+                if rates.is_empty() {
+                    None
+                } else {
+                    let avg: f64 = rates.iter().sum::<f64>() / rates.len() as f64;
+                    let std_dev = {
+                        let variance: f64 = rates.iter().map(|t| (t - avg).powi(2)).sum::<f64>()
```
amaanq, line 979: probably don't need to declare the type (`: f64`)? could be wrong

amaanq (commented):
looks cool, but maybe we can restructure this to have multiple "options" for showing stats, like I mentioned offline.

WillLillis:
It looks like there's already a fix for the WASM CI failures in https://github.com/tree-sitter/tree-sitter/pull/3887/commits/5c90d58b94f1ef3da74cf4e42de6a00c67280dcb. I can wait until that's merged in and rebase off of master, or just copy it in if that would be preferred.

WillLillis:
(In case anyone is curious)

Performance comparison against master for the original approach in this PR:

image

After [ee28322](https://github.com/tree-sitter/tree-sitter/pull/3939/commits/ee28322a784d892bb285e017d886074e41c8aa23) (only runs the test corpus once):

image

Issue #3891, opened 2024-11-07
Generate pretty test report


kylegoetz:
### Problem

It's possible a user will have a workflow like:

1. write grammar code
2. tree-sitter test
3. scan through test failures for easy wins, fix those, then start looking for harder failures, and finally look for failures where there are actual ERROR messages mid-parse tree

This can be challenging with the way `tree-sitter test` currently just dumps failures to stdout.

### Expected behavior

It would be nice to have a command or flag like `tree-sitter test --report` that instead builds an HTML report of the tests and pipes that to a browser with something like `bcat` (or dumps to file that can be opened, though this is more friction)

I've begun writing an NPM package that could be added to a tree-sitter package as a devDep that I found extremely useful for ironing out bugs with my `tree-sitter-unison` package. It's invoked with something like `tree-sitter test 2>&1 | tree-sitter-test-report` where the report package parses the combination of stdout and stderr to generate a table of results.

(Attached is a screenshot of an ugly first draft I used the past few days that was very helpful to me
<img width="1208" alt="Screenshot 2024-11-07 at 10 21 49 AM" src="https://github.com/user-attachments/assets/48431959-d595-4f7c-b147-63c8090b2bbf">
.)

However, this is limited to what `test` sends to stdout and stderr. It'd be cool if you could expand a row, pass *or* failure, to see the underlying test including expected result.

I'm able to implement this for failures in my JS package, but since successes don't get sent to stdout, and the `test` results don't include *filenames* that contain the tests, I can't get the expected output from file so cleanly in an external package.

Hence why I think it'd be nice to have in tree-sitter itself as part of the Rust code. I think it'd be a real boon to people developing grammars to have this for a comprehensive, easy-to-read overview of parsing tests.

### Plan

I'm thinking this would be the ideal, but am hoping for feedback:

- [x] `tree-sitter test --report` runs tests and dumps JSON data to stdout
- [x] this data contains the filename for a test suite + children, which are (presumably) a list of tests with a status (`pass`, `fail`, `skipped`, `no_platform`(??)), and if there's a failure, the diff of expected and actual
- [x] corpus tests
- [ ] highlighting tests
- [ ] any other test types

kylegoetz:
being able to expand a row to see the actual test (or mouseover or something), and FAILs would show a left-right diff

ObserverOfTime:
A JSON report would be more useful. You can then convert that to whatever you want.

kylegoetz:
[...]

PR #3941, opened 2024-11-13
JSON test report

kylegoetz:
Addresses #3891

More can be done. But I wanted to start the PR and get feedback.

With this PR, if one executes `git test --report` it will dump a JSON of test results to stdout instead of the default stdout formatting.

It only does corpus tests right now. The output is akin to this format:

```typescript
type TestResults = {
  "name": "corpus",
  "file_path": null,
  "children": (Group|Example)[]
}

type Group = {
  "name": string,
  "file_path": string,
  "children": (Group|Example)[]
}

type Example = {
  "name": string,
  "results": Result[],
}

type Result = {
  "name": string,
  "status": "fail",
  "actual_output" string,
  "diff": DiffLine[]
} | {
  "name": string,
  "status": "pass",
  "actual": string,
} | {
  "name": string,
  "status": "skipped"
}

type DiffLine = {
  tag: 'delete' | 'insert' | 'equal',
  value: string,
}
```

One can generate HTML from this that includes the ability to click on a test name and get directed to the file (in browser) with the name of the test centered and highlighted (but this code is currently not included in the PR; it's a Node project I haven't packed for NPM yet. It is a single HTML file that could be templated and included in TS as something like `tree-sitter test --html-report` that would take the JSON and dump it into the template and then log the template's HTML to screen.

A few screenshots, one of the JSON output, and one of a simple HTML report that could be built-in if desired:

Screenshot 2024-11-13 at 2 44 39 PM
Screenshot 2024-11-13 at 2 44 47 PM
Screenshot 2024-11-13 at 2 46 13 PM

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -278,6 +278,8 @@ struct Test {
     pub rebuild: bool,
     #[arg(long, help = "Show only the pass-fail overview tree")]
     pub overview_only: bool,
+    #[arg(long, help = "Generate a JSON report of the test results")]
+    pub report: bool,
 }
 
 #[derive(Args)]
@@ -972,6 +974,7 @@ impl Test {
                 test_num: 1,
                 show_fields: self.show_fields,
                 overview_only: self.overview_only,
+                generate_report: self.report,
             };
 
             test::run_tests_at_path(&mut parser, &mut opts)?;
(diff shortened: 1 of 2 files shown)


kylegoetz:
I'd wondered about how TS iterates through languages and runs the tests for each language. When I test here, the languages in the test opts has no name, and is the language I'm running the test from. But since TS iterates over the languages, that is why the JSON for Example.results is an array rather than an object.

There's also some `no_platform` code, and I wasn't sure what that is.

kylegoetz:
The code could be simplified if the "default" way of logging tests to stdout were converted from all those println calls to an accumulator that dumps the entire thing to stdout at the end (which is what the json does). Then the two code branches (json and the "default") could be merged into a single one.

But I was loathe to make that change myself.

amaanq:
> The code could be simplified if the "default" way of logging tests to stdout were converted from all those println calls to an accumulator that dumps the entire thing to stdout at the end (which is what the json does). Then the two code branches (json and the "default") could be merged into a single one.
>
>But I was loathe to make that change myself.

If you're willing to put in the effort, that'd be great and totally accepted, but in its current state, the delta from where this PR is at and where it'd ideally be is quite large - here are my thoughts

1. we should allow people to just output results to a JSON file, for parsing and testing - this flag would be something like `json-report`, and then it can be combined with a flag like `open-report` that is similar to `open-log` for `parse`, in which this UI to view your results shows up (something similar to clang's `scan-build` would be cool).

2. the json data **should** be typed - just spamming `json!` everywhere is very error prone

kylegoetz:
I've got a new code branch I'm working on this. Right now passing `--report` will log a JSON object containing corpus, highlighting, and tag test results, and without it, the behavior is unchanged from current `tree-sitter test` as far as I can tell.

I wrote a lot of unit tests to nail down the behavior.

To do is create a flag to export the JSON optionally to a file and have the flag that will open it in a viewer. I prefer having the ability to dump to stdout instead of to a file because some users (this includes me) will want to pipe the JSON to a viewer instead of writing to a file and opening it with the viewer, and then needing to remember to delete that file later.

https://github.com/kylegoetz/tree-sitter/tree/feat/json

PR #3945, opened 2024-11-15
fix: smarter escape detection for invalid anonymous nodes

ribru17:
Right now, the current quotation escape checker fails in the case that there is an anonymous node that is just an escaped backslash (it thinks the backslash escapes the quote, when really it is just an escaped backslash itself. See the added test case for an example of this).

This commit ensures the node identification logic keeps track of the number of backslashes seen so it can accurately determine if the quotation is escaped or not.

See https://github.com/neovim/neovim/pull/31127/files

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -2353,27 +2353,24 @@ impl Query {
                 ffi::TSQueryErrorNodeType | ffi::TSQueryErrorField | ffi::TSQueryErrorCapture => {
                     let suffix = source.split_at(offset).1;
                     let in_quotes = source.as_bytes()[offset - 1] == b'"';
-                    let mut end_offset = suffix.len();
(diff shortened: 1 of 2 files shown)


lib/binding_rust/lib.rs
```diff
@@ -2353,27 +2353,26 @@ impl Query {
...
-                            }
-                            c if !in_quotes
-                                && (c.is_whitespace() || c == '(' || c == ')' || c == ':') =>
-                            {
-                                Some(i)
+                    let end_offset = suffix
+                        .find({
+                            let mut backslashes = 0;
+                            move |c| {
+                                if in_quotes {
```
amaanq, line 2360: you can move the assignment of backslashes outside of the closure, and the closure doesn't have to explicitly `move` anything, otherwise this looks great

amaanq:
nice, thank you

PR #3947, opened 2024-11-16
feat: generate schema in tree-sitter.json

### LINKED ISSUE #3903: Add a `$schema` property to generated `tree-sitter.json` files
### Problem

When the `tree-sitter.json` file was proposed (in #3637) the comment has a (collapsed) [JSON Schema](https://json-schema.org/) which contains the expected schema of this file.

### Expected behavior

It would be useful for the pre-generated `tree-sitter.json` generated by `tree-sitter init` to include a [`$schema` property so that editors can automatically show completion and documentation when editing the file](https://code.visualstudio.com/Docs/languages/json#_json-schemas-and-settings).

Thanks!

ribru17:
Fixes #3903


cli/loader/src/lib.rs
```diff
@@ -140,6 +140,8 @@ pub struct LanguageConfigurationJSON {
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct TreeSitterJSON {
+    #[serde(rename = "$schema")]
+    pub schema: String,
```
amaanq, line 144: Option so it doesn't break backwards compatibility, and we can omit from tests

ribru17: Ah thx good call with option. Also when making this a static str I get this error I am unsure how to resolve:

```
1. temporary value dropped while borrowed
   creates a temporary value which is freed while still in use [E0716]
2. argument requires that borrow lasts for `'static` [E0716]
```

In here:

```rust
    pub fn from_file(path: &Path) -> Result<Self> {
        Ok(serde_json::from_str(&fs::read_to_string(
            path.join("tree-sitter.json"),
        )?)?)
    }
```

amaanq: Ah yeah, String is fine then

amaanq (approved):

PR #3952, opened 2024-11-18
fix(bindings): include headers & queries in python sdist

ObserverOfTime:
Figured out a way to fix tree-sitter/tree-sitter-python#285 without an extra file.

### DIFF
--- a/cli/src/init.rs
+++ b/cli/src/init.rs
@@ -6,6 +6,7 @@ use std::{
 
 use anyhow::{anyhow, Context, Result};
 use heck::{ToKebabCase, ToShoutySnakeCase, ToSnakeCase, ToUpperCamelCase};
+use indoc::indoc;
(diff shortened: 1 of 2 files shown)


cli/src/init.rs
```diff
@@ -624,9 +635,46 @@ pub fn generate_grammar_files(
...
+                                        self.filelist.recursive_include("queries", "*.scm")
+                                        self.filelist.include("src/tree_sitter/*.h")
+
+
+                                setup(
+                                "#},
+                            )
+                            .replace(
+                                "\"bdist_wheel\": BdistWheel\n",
+                                // XXX: indoc drops the indent for some reason
```
amaanq, line 669: Yeah, because indoc trims the first line's leading space and uses that as a guide for indenting, so the second line isn't indented as the first line wasn't. The comment isn't needed

PR #3963, opened 2024-11-25
Feat/json

kylegoetz:
Addresses #3891 

introduces two flags to `tree-sitter test`:

`--report <REPORT>` which allows a user to create a JSON report stored in a given file. If the user types `:stdout:` as the filename, it will dump the JSON to stdout, which enables a user to pipe the output to the receiving application of their choice (like a test report generator that converts JSON to an HTML report of their choice)

`--open-report`, which opens the JSON report in a browser. Some browsers have fancy JSON display abilities that can be useful here. Though it would be nice to have an idea of a JSON program that could be configured to open instead of it going to the browser. Could be in a future PR.

The JSON report includes results for corpus, highlighting, and tag tests.

This is an improvement over a previous (closed by me) PR that had suggestions from @amaanq. We now have the ability to open the output in an application (although right now it opens in browser)

The JSON is also typed now, which is safer than spamming the serve `json!` macro everywhere.

I could see there being an --html-report flag, too, that is essentially a pre-ordained HTML template with the JSON results dumped into it, and then auto-open that in browser, but I think it'd be better as a future feature so this PR doesn't become even larger. As it is, there were a lot of necessary changes in order to create JSON *or* "standard" test results.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -497,6 +497,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "downcast"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1435fa1053d8b2fbbe9be7e97eca7f33d37b28409959813daefc1446a14247f1"
+
+[[package]]
 name = "either"
 version = "1.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


kylegoetz:
If this diff is too big, would it be palatable to have a PR converting the `println!` all over the place in the tests into structured data that accumulates everything and then prints at the end?

And then introduce flags for JSON stuff that are just another serialization of this same structured data?

Just in case there's paranoia about so many changes...

WillLillis:
Hey! Thanks for taking the time to put this together, the idea to add a report option to the `test` sub command is really good! Given how large this change set is however, I think we’ll need to spend some more time discussing avenues for implementation in the issue. This will be a helpful data point, but there are enough open questions and issues (e.g. functions commented out without explanation, error handling in some places, and the addition of a mocking dependency) that I don’t think we can move forward with this PR currently stands.

kylegoetz:
I appreciate being encouraged to write this and only after spending hours on it, being told never mind.

PR #3987, opened 2024-12-05
feat: add error information in the progress callback

criminosis:
Hi Tree-Sitter team!

Taking another stab at this since [my last PR](https://github.com/tree-sitter/tree-sitter/pull/2289#event-12066393031), leveraging the upcoming callback paradigm introduced by https://github.com/tree-sitter/tree-sitter/pull/3843

In https://github.com/tree-sitter/tree-sitter/issues/2541 my notion of exiting early didn't seem to make the cut when the issue was closed. In the PR this [thread](https://github.com/tree-sitter/tree-sitter/pull/3843#discussion_r1823579755) between @amaanq & @maxbrunsfeld briefly talked about it. I got the impression from having a brief [conversation](https://github.com/tree-sitter/tree-sitter/issues/2541#issuecomment-2502557752) with @amaanq that my prior attempt was errant from trying to infer an error too early.

This PR instead is attempting to report to the callback that an error node will be added to the tree.

To be clear that's the scenario I'm interested in. Because if there's a portion of the tree that is just an error node I'm no longer interested in the tree at all and would prefer to cancel processing rather than waste time executing further.

I also noted [this site](https://github.com/tree-sitter/tree-sitter/blob/69d977d73648010d7060001fa518f3198a41a7e5/lib/src/parser.c#L1493) as being an alternative, but wasn't sure if that'd be preferable to what I ended up pushing up.

As always happy to make changes

### DIFF
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -1029,6 +1029,35 @@ fn test_parsing_with_timeout_during_balancing() {
     });
 }
 
+#[test]
+fn test_parsing_with_timeout_when_error_detected() {
+    let mut parser = Parser::new();
+    parser.set_language(&get_language("json")).unwrap();
+
+    // Parse an infinitely-long array, but insert an error after 1000 characters.
+    let mut offset = 0;
+    let erroneous_code = "!,";
+    let tree = parser.parse_with_options(
+        &mut |i, _| match i {
+            0 => "[",
+            1..=1000 => "0,",
+            _ => erroneous_code,
+        },
+        None,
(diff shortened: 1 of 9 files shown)


criminosis:
Formatting should be fixed now 👍

Alexizx00789999 (approved):

amaanq:
Sorry for taking a while to get to this, the implementation was "not right" to say the least, but this is totally a use case we *should* support. I've augmented this by renaming the field to `has_error`, and changing the logic to only mark this as true if the error is *definite*, meaning **all** stack versions have a non-zero error cost.

amaanq (approved):

Issue #1654, opened 2022-02-16
Aliasing to a Grammar Symbol without a Production Produces an AST without a Corresponding Node in node-types.json


berchn:
I've run into this issue recently in tree-sitter-python in regards to the following commit:
https://github.com/tree-sitter/tree-sitter-python/commit/3324e923485177d72c1a87d315f3b31bf2294585

Notice that `alias($.expression, $.as_pattern_target)` and `$.as_pattern_target` does not have a production. In node-types.json, `as_pattern` specifies `as_pattern_target` as an accepted type for its `alias` field, but `as_pattern_target` does not exist as a type in node-types.json.

Is this behavior intended?

berchn:
Furthermore, this is also attempting to alias a grammar symbol which is specified as a supertype. Does it make sense to allow this?

Marcono1234:
Possibly related to #455

But still the case for the latest Tree-sitter version:
- https://github.com/tree-sitter/tree-sitter-python/blob/v0.23.0/grammar.js#L711
- https://github.com/tree-sitter/tree-sitter-python/blob/v0.23.0/src/node-types.json#L396

Just based on the `node-types.json` it is impossible to tell what `as_pattern_target` refers to.

clason:
Please test the latest nightly; there were some significant changes to how supertypes are entered into `node_types.json`.

Marcono1234:
Thanks! I have tried v0.27.0 `tree-sitter generate` on tree-sitter/tree-sitter-python, and it included `"type": "as_pattern_target"` now as desired.

PR #3988, opened 2024-12-06
fix(cli): add error message for supertypes used as alias

wangweixuan:
Currently, aliasing a non-terminal rule to a supertype rule causes the cli to panic at the generation of `node-types.json`.

Reproducible example:
```js
module.exports = grammar({
  name: "foo",

  supertypes: ($) => [$._x],

  rules: {
    source_file: ($) => alias($.y, $._x),
    _x: ($) => "x",               // _x is a supertype
    y: ($) => field("f", "y"),    // y is a non-terminal
  },
});
```
Panic at: https://github.com/tree-sitter/tree-sitter/blob/69d977d73648010d7060001fa518f3198a41a7e5/cli/generate/src/node_types.rs#L530

This PR adds an error message during grammar preparation to disallow supertypes used as alias.

Notes:
- Aliasing a terminal rule to a supertype rule is currently accepted, but the produced `node-types.json` is incorrect (the supertype does not mention the terminal symbol as a subtype). This PR makes that an error.
- Aliasing a supertype rule to another supertype is currently accepted, but the produced `node-types.json` is incorrect (the supertypes are not joined). This PR makes that an error too.

Possibly related: https://github.com/tree-sitter/tree-sitter/issues/1654#issuecomment-1041909372

### DIFF
--- a/cli/generate/src/prepare_grammar/intern_symbols.rs
+++ b/cli/generate/src/prepare_grammar/intern_symbols.rs
@@ -114,10 +114,20 @@ impl Interner<'_> {
                 Ok(Rule::Seq(result))
             }
             Rule::Repeat(content) => Ok(Rule::Repeat(Box::new(self.intern_rule(content, name)?))),
-            Rule::Metadata { rule, params } => Ok(Rule::Metadata {
-                rule: Box::new(self.intern_rule(rule, name)?),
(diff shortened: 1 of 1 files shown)


amaanq:
I would think this should be allowed, but the aliased rule would be visible, unlike a supertype...

PR #3989, opened 2024-12-06
chore(binding_rust): mark `set_language` as `must_use`

ugur-a:
its `Result` should ideally be checked by the caller

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -579,6 +579,7 @@ impl Parser {
     /// [`Language::version`] and compare it to this library's
     /// [`LANGUAGE_VERSION`] and [`MIN_COMPATIBLE_LANGUAGE_VERSION`] constants.
     #[doc(alias = "ts_parser_set_language")]
+    #[must_use]
     pub fn set_language(&mut self, language: &Language) -> Result<(), LanguageError> {
         let version = language.version();
         if (MIN_COMPATIBLE_LANGUAGE_VERSION..=LANGUAGE_VERSION).contains(&version) {


ugur-a:
One call just call the function and ignore its output (especially since, in this case, the `Ok` variant is just `()`). By adding `must_use` you make it necessary* for the caller to `unwrap` the `Result`.

\* if they want to avoid the corresponding lint firing, which they probably do

WillLillis:
Marking `set_language` with `#[must_use]` just adds a redundant warning for the reason Max gave.

<Details>
<Summary>Minimal repro</Summary>

```rust
use tree_sitter::Parser;

fn main() {
    parser.set_language(&tree_sitter_c::LANGUAGE.into()).unwrap();
}
```

</Details>

Before:

```sh
warning: unused `Result` that must be used
 --> src/main.rs:5:5
  |
5 |     parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this `Result` may be an `Err` variant, which should be handled
  = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
  |
5 |     let _ = parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     +++++++

warning: `must_use_test` (bin "must_use_test") generated 1 warning
```


After:

```sh
warning: unused `Result` that must be used
 --> src/main.rs:5:5
  |
5 |     parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this `Result` may be an `Err` variant, which should be handled
  = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
  |
5 |     let _ = parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     +++++++

warning: unused return value of `Parser::set_language` that must be used
 --> src/main.rs:5:5
  |
5 |     parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
help: use `let _ = ...` to ignore the resulting value
  |
5 |     let _ = parser.set_language(&tree_sitter_c::LANGUAGE.into());
  |     +++++++

warning: `must_use_test` (bin "must_use_test") generated 2 warnings
```

maxbrunsfeld:
Thanks for the concrete test @WillLillis .

The must_use would be redundant here.

PR #3992, opened 2024-12-09
3991/word token

### LINKED ISSUE #3991: Error message regarding "word token" is a bit confusing, could be improved
### Problem

The following error message caused confusion on Discord today:

`Non-terminal symbol 'identifier' cannot be used as the word token`

Two users with Tree Sitter experience interpreted this as being about the use of `token` (i.e., `the word "token"`) as opposed to the use of the property `word` (i.e., `the "word" token`)

I suggest (drawing on my previous experience as an editor) that this be clarified by adding quotation marks around "word" to make it clear the issue is with a token having value `word` as opposed to a word having value `token`.

### Steps to reproduce

Unnecessary, as it's a rule of English punctuation.

### Expected behavior

I expect "word" to be in quotation marks.

0.24.3

### Operating system/version

macOS 15

kylegoetz:
I added quotation marks around `word` in every occurrence of `"word token"` that was about the token called `word`, including in comments. This should avoid confusion for people creating grammars who run into this bug and are using `token(...)` in their code.

See discussion [here](https://discord.com/channels/1063097320771698699/1063097321648312354/1315715723502555206) for an example

**Edit** Oh this would close #3991

### DIFF
--- a/cli/generate/src/build_tables/mod.rs
+++ b/cli/generate/src/build_tables/mod.rs
@@ -202,10 +202,10 @@ fn populate_used_symbols(
     parse_table.symbols.push(Symbol::end());
     for (i, value) in terminal_usages.into_iter().enumerate() {
         if value {
-            // Assign the grammar's word token a low numerical index. This ensures that
+            // Assign the grammar's "word" token a low numerical index. This ensures that
             // it can be stored in a subtree with no heap allocations, even for grammars with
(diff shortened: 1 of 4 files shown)


amaanq:
Not really a fan of this, I don't think this really clarifies anything since it's pretty well-known the word token is it's own distinct concept, and the error in the given issue is misleading, since the problem really stemmed from using the word token as the start rule. (#4004 would be a more appropriate solution)

PR #3999, opened 2024-12-10
enhancement: do not validate e-mail address during grammar init

### LINKED ISSUE #3994: tree-sitter init email parsing behavior should change
### Problem

The current email parsing just warns the user if the email address entered during initialization is invalid. It doesn't explain why.

Furthermore, the current email validator requires two things: at least one `@` and a `.`. That's the whole validation scheme.

This technically produces false negatives. The RFC specification for email addresses requires the domain part to be a dot-separated list of DNS labels **or an IPv4 or IPv6 address** surrounded by brackets.

As such, a domain name like `jsmith@[IPv6:2001:db8::1]`, which contains no dots, is a valid email address that would be rejected by `tree-sitter init`.

### Steps to reproduce

`tree-sitter init`

for email address, input jsmith@[IPv6:2001:db8::1] (which is a valid email address)

### Expected behavior

Personally, because email validation is very complex, I would recommend not attempting any attempt to validate email addresses: if the validator isn't correct, to what end does one exist in the codebase?

But if we were to keep email validation, then I recommend getting rid of the dot requirement (because it's incorrect validation), keeping the @ requirement, and then updating the error message to mention that an `@` is required (so the user knows exactly what they did wrong).

Technically the `@` isn't even required, but without one, the address is assumed to be local, which makes no sense for any publicly-available package (although internal grammars for companies conceivably could be local-only).

For reference, here are some developers who have written about why one should not bother trying to validate email addresses (unless you're writing an email server):

* https://blog.sinjakli.co.uk/2011/02/13/email-address-validation-please-stop/
* https://davidcel.is/articles/stop-validating-email-addresses-with-regex/ (recommending only look for the `@`)
* https://news.ycombinator.com/item?id=11414749 (HackerNews discussion about why one shouldn't bother parsing email addresses with links to a few blogs written by participants in the discussion)

0.24.3

### Operating system/version

macOS 15.0.1

kylegoetz:
fixes and will close #3994

Removed email validation and added explicit type parameter (`: String`) a few lines down because linter recommended it.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -586,11 +586,11 @@ impl Init {
             };
 
             let email = || {
-                Input::with_theme(&ColorfulTheme::default())
+                Input::<String>::with_theme(&ColorfulTheme::default())
                     .with_prompt("Author email")
                     .allow_empty(true)
                     .interact_text()
-                    .map(|e: String| (!e.trim().is_empty()).then_some(e))
+                    .map(|e| (!e.trim().is_empty()).then_some(e))
             };
 
             let url = || {
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -588,20 +588,9 @@ impl Init {
             let email = || {
                 Input::with_theme(&ColorfulTheme::default())
                     .with_prompt("Author email")
-                    .validate_with({
-                        move |input: &String| -> Result<(), &str> {
-                            if (input.contains('@') && input.contains('.'))
-                                || input.trim().is_empty()
-                            {
-                                Ok(())
-                            } else {
-                                Err("This is not a valid email address")
-                            }
-                        }
-                    })
                     .allow_empty(true)
                     .interact_text()
-                    .map(|e| (!e.trim().is_empty()).then_some(e))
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -588,20 +588,9 @@ impl Init {
             let email = || {
                 Input::with_theme(&ColorfulTheme::default())
```
amaanq, line 589:
```suggestion
                Input::<String>::with_theme(&ColorfulTheme::default())
```

cli/src/main.rs
```diff
@@ -588,20 +588,9 @@ impl Init {
...
-                                Ok(())
-                            } else {
-                                Err("This is not a valid email address")
-                            }
-                        }
-                    })
                     .allow_empty(true)
                     .interact_text()
-                    .map(|e| (!e.trim().is_empty()).then_some(e))
+                    .map(|e: String| (!e.trim().is_empty()).then_some(e))
```
amaanq, line 593:
```suggestion
                    .map(|e| (!e.trim().is_empty()).then_some(e))
```

PR #4004, opened 2024-12-13
docs: clarify how the entrypoint for parsing is determined

### LINKED ISSUE #4003: Docs: update to clarify which rule is treated as the entry for a source file
### Problem

Docs don't explain this, which creates confusion. See [here](https://discord.com/channels/1063097320771698699/1063097321648312354/1316888006367383684) for discussion pursuant to a mistaken understanding about this.

### Expected behavior

Docs should clarify which rule is treated as the entry for a source file. As of now, my understanding is that it's the first rule, but it's unclear to me how "first" is defined. JavaScript has a counterintuitive definition of "first property" for JS objects that changed in 2015 from "there is no guaranteed ordering of properties in an object" to something different (ordering is integer keys sorted in increasing order, then string keys sorted by insertion order, then symbol keys sorted by insertion order.

kylegoetz:
closes #4003 by explaining which property in `rules` is the entry point for parsing.

### DIFF
--- a/docs/section-3-creating-parsers.md
+++ b/docs/section-3-creating-parsers.md
@@ -557,6 +557,8 @@ Although languages have very different constructs, their constructs can often be
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/section-3-creating-parsers.md
```diff
@@ -557,6 +557,8 @@ Although languages have very different constructs, their constructs can often be
 }
 ```
 
+One important thing to note up front is that the entry point for grammar parsing is defined as the first property in the `rules` object. The formal specification for what constitutes the "first property" is found in the ECMAScript 2015 specification. However, since grammars typically only contain string-based property names, it is safe to assume that the top-most key-value pair in `rules` will be treated as the entry point. In the examples on this page, this rule is called `source_file`, but it can be anything. [Other](https://github.com/tree-sitter/tree-sitter-haskell/blob/2c2fcb22e3b939ed59ae8d6f10500c3213c68670/grammar.js) [grammars](https://github.com/kylegoetz/tree-sitter-unison/blob/3c97db76d3cdbd002dfba493620c2d5df2fd6fa9/grammar.js) have named this property after the programming language whose grammar is described by the file.
```
ObserverOfTime, line 560: Too verbose.
```suggestion
One important thing to note up front is that the entry point for grammar parsing is defined as the first property in the `rules` object. In the examples on this page, this rule is called `source_file`, but it can be anything.
```

amaanq:
Superseded by #4028

PR #4005, opened 2024-12-13
fix: mark command aliases as visible

ribru17:
This will document them in `--help` and generate shell completions for them in `tree-sitter complete`.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -61,14 +61,18 @@ enum Commands {
 struct InitConfig;
 
 #[derive(Args)]
-#[command(about = "Initialize a grammar repository", alias = "i")]
+#[command(about = "Initialize a grammar repository", visible_alias = "i")]
 struct Init {
     #[arg(long, short, help = "Update outdated files")]
     pub update: bool,
 }
 
 #[derive(Args)]
-#[command(about = "Generate a parser", alias = "gen", alias = "g")]
+#[command(
+    about = "Generate a parser",
+    visible_alias = "gen",
+    visible_alias = "g"
+)]
 struct Generate {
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
If completions for e.g. `tree-sitter g --<TAB>` work without making the aliases visible,
then I'd rather not fill the command suggestions with aliases.

ribru17:
Not sure if this is possible right now, https://github.com/clap-rs/clap/issues/4265 has some discussion on this

amaanq:
I agree that the aliases in the help menu are pretty noisy with this change (output of `tree-sitter --help`), and this currently doesn't affect completions.

ribru17:
I think this is just a bug with the nushell completions; it seems like there is no diff with that completion file between the two changes but bash and zsh does have an updated completion script

Personally, I like the `--help` changes, I don't think it adds too much clutter. I think it is nice to know about command shorthands but maybe it is slightly annoying having them after `tree-sitter <TAB>`

amaanq:
Yeah good point, the alias issue does seem to be nushell specific. Though I would say a better spot to put alias information (along with a more detailed explanation/command overview) would be the docs (which really need a rewrite :tm:).

amaanq:
Well, the docs have been rewritten to include a cli reference page (which includes aliases for commands), I'm now *really* not a fan of exposing the aliases in the CLI (especially in completions)...

ribru17:
Sure that makes sense. Maybe this should be closed in favor of a PR to clap that provides alias completions *only after they have been inputted*? E.g., don't complete `tree-sitter <g>`, but upon typing `tree-sitter g <Tab>` there will be completions?

PR #4010, opened 2024-12-16
fix(cli): handle nested tags test files

### LINKED ISSUE #3955: nested tag test panic strangeness
### Problem

nesting folders in a test/tags causes a panic that language not found for a path

### Steps to reproduce

1. clone kylegoetz/tree-sitter-unison
2. create a test/tags/foo
3. run tree-sitter test
4. observe panic
5. rename *directory* `foo` to `foo.u`
6. run tree-sitter test
7. observe no panic

So something about the tag tests expect paths to have names that end in the file extension of an available language

### Expected behavior

I expect paths to not have to have a mock file extension matching an available grammar language

0.24.3

### Operating system/version

macos 15

amaanq:
- Closes #3955

### DIFF
--- a/cli/src/test_tags.rs
+++ b/cli/src/test_tags.rs
@@ -49,47 +49,84 @@ pub fn test_tags(
     directory: &Path,
     use_color: bool,
 ) -> Result<()> {
-    let mut failed = false;
     println!("tags:");
+    test_tags_indented(loader, loader_config, tags_context, directory, use_color, 2)
+}
+
+pub fn test_tags_indented(
+    loader: &Loader,
+    loader_config: &Config,
+    tags_context: &mut TagsContext,
+    directory: &Path,
+    use_color: bool,
+    indent_level: usize,
+) -> Result<()> {
+    let mut failed = false;
+
     for tag_test_file in fs::read_dir(directory)? {
         let tag_test_file = tag_test_file?;
         let test_file_path = tag_test_file.path();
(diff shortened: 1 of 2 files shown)


cli/src/test_tags.rs
```diff
@@ -49,47 +49,84 @@ pub fn test_tags(
...
-                        test_file_name.to_string_lossy().as_ref()
-                    ),
-                );
+        print!(
+            "{indent:indent_level$}",
+            indent = "",
+            indent_level = indent_level * 2
+        );
+        if test_file_path.is_dir() && test_file_path.read_dir()?.next().is_some() {
+            println!("{}:", test_file_name.into_string().unwrap());
```
WillLillis, line 76: Any reason for the `unwrap()` instead of `test_file_name.to_string_lossy()`?

amaanq: copied from https://github.com/tree-sitter/tree-sitter/blob/07aaf2322e0d5d7568d37e65cc7a951da252416b/cli/src/test_highlight.rs#L77

I don't really expect anyone to use non-utf8 filenames for a test file, though it doesn't hurt to use `to_string_lossy` (I'll update the other callsite)

WillLillis: Makes sense! I don't really expect it to be an issue either, but it can't really hurt to use the lossy form here.

WillLillis (approved):

PR #4015, opened 2024-12-17
feat: add xtask to bump `emscripten-version`

### LINKED ISSUE #1223: Reliable releases distribution for tree-sitter CLI and other components
From https://github.com/tree-sitter/tree-sitter/issues/930#issuecomment-801519577:

> It would be awesome to automate release process for all official tree-sitter tools, especially for tree-sitter-cli, for all official bindings Wasm, Rust, Node.js, Python, Haskell, Ruby and the Playground with its separately living parsers and keep all in sync with the core tree-sitter library releases. This would help to reduce misunderstanding and situations that some things work somewhere and somewhere don't.
> - For now **tree-sitter-cli** installation from the [crate](https://crates.io/crates/tree-sitter-cli) seems the bad idea, the crate is stuck in 2 years old version.
> - #1122 - _tree-sitter-highlight 0.19.2 does not compile with tree-sitter 0.19.5_ - demonstrates an issue that changes in tree-sitter's Rust binding requires bumping version in all dependencies that use changed parts. Otherwise there need to be a CI check that would test that the last dependent can be built against all equal or higher versions of the dependence.
> - I can't say about all bindings but Node and Python bindings use **static** linking to tree-sitter core library and this means that these are lag behind the core library and don't receiving core fixes and logic improvements synchronously. IMO that's the important reason why such updates need to be automated. This doesn't solves problem with the core lib features covering but at least bug fixes would be delivered in time.

_I've previously posted a set of actual live badges there https://github.com/tree-sitter/tree-sitter/issues/930#issuecomment-801519577_
_But it makes sense to duplicate these here in this issue (all badges are clickable)._
### Versions
[NPM](https://www.npmjs.org/package/tree-sitter-cli) [crates.io](https://crates.io/crates/tree-sitter-cli)

[crates.io](https://crates.io/crates/tree-sitter-loader) [crates.io](https://crates.io/crates/tree-sitter-config)

[crates.io](https://crates.io/crates/tree-sitter-highlight) [crates.io](https://crates.io/crates/tree-sitter-tags)

#### Bindings
[binding:wasm:npm](https://www.npmjs.org/package/web-tree-sitter) [binding:rust](https://crates.io/crates/tree-sitter) [binding:node](https://www.npmjs.org/package/tree-sitter)

[binding:python](https://pypi.org/project/tree-sitter) [binding:haskell](http://hackage.haskell.org/package/tree-sitter) [binding:ruby](https://rubygems.org/gems/tree-sitter)

The [0.20.0](https://github.com/tree-sitter/tree-sitter/releases/tag/v0.20.0) release happened 2 days ago and on above badges it's seen that it doesn't exist in **any** distribution channel except directly download from Github artifacts, use OS package managers (if these have it) or build from the source. CLI on the [crates.io](https://crates.io/crates/tree-sitter-cli) stuck in the past for 2 years.
As I understand current release process, there are a lot of manual actions that need to be happen.

Also I'd suggest to place such badges into the [README.md](https://github.com/tree-sitter/tree-sitter#readme) to make it easy control if something wrong [happened](https://github.com/tree-sitter/tree-sitter/issues/1216#issuecomment-871811627) with a release and to always have actual information about all components version in the well known place.

---
### TODO:
- Improvements:
  - [x] Migrate the whole `script/` folder on [xtasks](https://github.com/matklad/cargo-xtask) to make all scripts portable.
 
- CI [xtasks](https://github.com/matklad/cargo-xtask):
  - [x] Update [`pkg-config` ](https://github.com/tree-sitter/tree-sitter/issues/1608) [version](https://github.com/tree-sitter/tree-sitter/issues/1158) in Makefile.
  - [x] `cargo update` + `git commit -m "Update cargo deps"`
  - [x] Rust bindings regeneration on every C code changes.
    - [x] Use latest `bindgen` version: https://github.com/tree-sitter/tree-sitter/pull/2171
  - [ ] Check latest Emscripten version and bump its version for TS.
  - [x] Update WASM Languages for docs' playground: #1997
  - [x] Release packages on NPM:
    - [x] [`tree-sitter-cli`](https://www.npmjs.com/package/tree-sitter-cli)
    - [x] [`web-tree-sitter`](https://www.npmjs.com/package/web-tree-sitter) https://github.com/tree-sitter/tree-sitter/pull/2169 _(look to a manual task list that needs to be automated)_

- CLI [xtasks](https://github.com/matklad/cargo-xtask):
  - [x] Leverage Git for scanning TS for changed parts from the last release and inform what components require a new version.
  - [x] Automatically update deps versions for related components.
  - [ ] A task for periodically bump fixture dependency versions as part of the release process in the main repo.
         _This means to get away from using main branches all the time for fixtures._
         _**Reasons:** I saw and fixed many CI fails due to fixture changes in the last months and it's counter productive when. e.g., working on a PR  but it fails for external reasons or when merges a good PR but it unexpectedly fails on the merged commit on the master because someone merged something in some fixtures repo just several minutes ago._
         _**Fail example:** https://github.com/tree-sitter/tree-sitter-cpp/pull/202#issuecomment-1546279646_ 

- Issues:
  - [x] Check: https://github.com/tree-sitter/tree-sitter/issues/352#issuecomment-499642718
  - [x] Release latest CLI to the **crates.io**: https://github.com/tree-sitter/tree-sitter/issues/963#issuecomment-811894568
  - [x] #610 
  - [x] #820
  - [x] #1158
  - [x] #2167
  - [x] #2125
  - [x] #2237

- Extra:
  - [x] #949
  - [ ] https://github.com/tree-sitter/tree-sitter/pull/1884#issuecomment-1387393173 - check comment: https://github.com/tree-sitter/tree-sitter/issues/1880#issuecomment-1254114863

amaanq:
- Closes #1223

### Problem

Manual intervention is required to update the file located at https://github.com/tree-sitter/tree-sitter/blob/master/cli/loader/emscripten-version, and this can easily be overlooked. It'd be nice to automate updating this file.

### Solution

An xtask is added that bumps this file by calling the GitHub API to fetch the git tags info. Then, the first (latest) tag is selected, and the file is updated w/ this new version. Additionally, a CI job was added that runs as a "post-script" (taken from [here](https://github.com/dependabot/dependabot-core/issues/5962#issuecomment-1303781931)) whenever a dependabot PR is opened or synchronized, which happens on a weekly cadence currently. This way, the emscripten version is updated automatically.

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -1,0 +1,32 @@
+name: Update Emscripten
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  update-emscripten:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Set up stable Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Run emscripten update xtask
+        env:
+          GIT_AUTHOR_NAME: dependabot[bot]
+          GIT_AUTHOR_EMAIL: 49699333+dependabot[bot]@users.noreply.github.com
+          GIT_COMMITTER_NAME: dependabot[bot]
+          GIT_COMMITTER_EMAIL: 49699333+dependabot[bot]@users.noreply.github.com
+        run: cargo xtask update-emscripten
+
+      - name: Push updated version
+        run: git push origin HEAD:$GITHUB_HEAD_REF
--- a/xtask/src/upgrade_emscripten.rs
+++ b/xtask/src/upgrade_emscripten.rs
@@ -1,0 +1,40 @@
+use std::{fs, path::Path};
+
+use anyhow::{anyhow, Result};
+use git2::Repository;
+use serde_json::Value;
+
+use crate::create_commit;
+
+pub fn run() -> Result<()> {
+    let response = ureq::get("https://api.github.com/repos/emscripten-core/emsdk/tags")
+        .call()?
+        .into_string()?;
+
+    let json = serde_json::from_str::<Value>(&response)?;
+    let version = json
+        .as_array()
+        .and_then(|arr| arr.first())
+        .and_then(|tag| tag["name"].as_str())
+        .ok_or(anyhow!("No tags found"))?;
+
+    let version_file = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .unwrap()
+        .join("cli")
+        .join("loader")
+        .join("emscripten-version");
+
+    fs::write(version_file, version)?;
+
+    println!("Upgraded emscripten version to {version}");
+
+    let repo = Repository::open(".")?;
+    create_commit(
+        &repo,
+        "build(deps): bump emscripten version",
+        &["cli/loader/emscripten-version"],
+    )?;
+
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
 name = "ahash"
 version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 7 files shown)


.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Commit updated version
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git diff --exit-code cli/loader/emscripten-version || {
+            git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY
+            git add cli/loader/emscripten-version
+            git commit -m "build(deps): bump emscripten version"
+            git push origin HEAD:${{ github.event.pull_request.head.ref }}
+          }
```
ObserverOfTime, line 35: Perhaps have the xtask create the commit like the wasmtime one does?
```suggestion
        env:
          GIT_AUTHOR_NAME: dependabot[bot]
          GIT_AUTHOR_EMAIL: 49699333+dependabot[bot]@users.noreply.github.com
          GIT_COMMITTER_NAME: dependabot[bot]
          GIT_COMMITTER_EMAIL: 49699333+dependabot[bot]@users.noreply.github.com
```

xtask/src/upgrade_emscripten.rs
```diff
@@ -0,0 +1,40 @@
...
+        .join("emscripten-version");
+
+    fs::write(version_file, version)?;
+
+    println!("Upgraded emscripten version to {version}");
+
+    let repo = Repository::open(".")?;
+    create_commit(
+        &repo,
+        "build(deps): bump emscripten version",
```
WillLillis, line 35: Would it be helpful to include the version being bumped to in the commit message as well?

ObserverOfTime (approved):

xtask/src/upgrade_emscripten.rs
amaanq: whoops, thought I had addressed this but I hadn't, sorry 😅

PR #4016, opened 2024-12-18
feat(cli): display parse and edit times separately for `parse --time`

### LINKED ISSUE #840: adding time measurement for source code edit on the fly
When using `tree-sitter parse` from CLI we have option to use -t flag to obtain time elapsed to parse source code. But this gives total time elapsed when --edit flag is used. it starts time, parses source code, makes edits and updates tree. then measure elapsed time. It would be nicer to have an option to obtain time elapsed by tree update due to source code.

WillLillis:
Currently, using the `--time` flag for the `parse` subcommand displays a duration for both parse and edit time. This provides a correction by recording and displaying these two times separately. This also has the added benefit of correctly calculating parse times when edits are applied.

Before/After Screenshot:

image

Closes #840

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -221,7 +221,7 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
         })));
     }
 
-    let time = Instant::now();
+    let parse_time = Instant::now();
 
     #[inline(always)]
     fn is_utf16_le_bom(bom_bytes: &[u8]) -> bool {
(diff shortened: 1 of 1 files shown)


cli/src/parse.rs
```diff
@@ -548,11 +551,12 @@ pub fn parse_file_at_path(parser: &mut Parser, opts: &ParseFileOptions) -> Resul
         }
 
         if first_error.is_some() || opts.print_time {
+            let path = opts.path.to_string_lossy();
             write!(
                 &mut stdout,
-                "{:width$}\t{duration_ms:>7.2} ms\t{:>6} bytes/ms",
-                opts.path.to_str().unwrap(),
-                (source_code.len() as u128 * 1_000_000) / duration.as_nanos(),
+                "{:width$}\tParse: {parse_duration_ms:>7.2} ms\t{:>6} bytes/ms",
```
WillLillis, line 557: If no edits are provided, this "Parse: " marker is redundant. Should it be omitted in this case?

amaanq: I'm indifferent to that

PR #4021, opened 2024-12-19
fix(cli): rename .editorconfig to fix vendoring

### LINKED ISSUE #4018: tree-sitter-cli cannot be vendored
### Problem

I'm still working on the same project as #3442. The specific problem I described in that issue has been fixed, but `tree-sitter-cli` still fails to be vendored.

In this case, it's because of `cli/src/templates/.editorconfig`. I suspect Cargo refuses to vendor any files whose names begin with a `.`. Would it be possible to rename this file to `editorconfig`? This would also required updating this line:

https://github.com/tree-sitter/tree-sitter/blob/7d3dbc062d8be6575f8e6e4ada18af989ba4e6ef/cli/src/init.rs#L63

### Steps to reproduce

1. `git clone https://github.com/andrewtbiehl/kramdown-syntax_tree_sitter`
2. `cd kramdown-syntax_tree_sitter/ext/tree_sitter_adapter`
3. `cargo update`
4. `RUBY=/path/to/bin/ruby cargo build` (observe that this succeeds)
5. `cargo vendor` (observe that this appears to succeed)
6. Modify .cargo/config.toml as directed by the output of cargo vendor
7. `cargo build` (observe that this fails)
8. There will also be no file `vendor/tree-sitter-cli/src/templates/.editorconfig`, although it does exist in the this repo

### Expected behavior

I expected that vendoring this crate would succeed.

N/A

### Operating system/version

macOS 14.7.1, NixOS 25.05

Samasaur1:
Closes #4018

I can't test this, because when I vendor this package as a git dependency it already works, so this will only be testable when a version with this fix makes it to crates.io. That said, I am pretty confident this will solve the problem, especially so given that the templates for `.gitignore` and other files that start with a `.` don't have a leading `.` in their template file, only when they are written into place.


ObserverOfTime:
The file is

Issue #4023, opened 2024-12-22
Suggestion: follow XDG on macOS for tree-sitter CLI


thecaralice:
### Problem

Currently, `xdg_config_file` uses `dirs::config_dir()`, which gives a rather unusable `$HOME/Library/Application Support/tree-sitter`. Granted, it is a standardized path on macOS, but this standard was intended for GUI apps and this convention is not well-suited for files which are supposed to be accessed by the user directly rather than through a certain GUI panel. tree-sitter is the only CLI following that convention I know of.

### Expected behavior

A solution would be to use XDG convention on both Linux and macOS; there are crates like `etcetera` that take a better approach at choosing config/data/etc paths compared to `dirs`/`directories`. Namely, `etcetera::base_strategy::choose_base_strategy` and `etcetera::app_strategy::choose_app_strategy` use Windows convention on Windows and XDG otherwise.

WillLillis:
Simply swapping out the current implementation for XDG would lead to unnecessary breakage for any macOS user with an existing config. Maybe instead the XDG path could be used first and `$HOME/Library/Application Support/tree-sitter` could be checked afterwards if that directory doesn't contain the expected file? This would match with how the `$HOME/.tree-sitter/` config path was replaced in https://github.com/tree-sitter/tree-sitter/pull/1157.

This function should either be renamed to better reflect its behavior, or updated:

https://github.com/tree-sitter/tree-sitter/blob/90666c951d53c13cc6cf5002d971a6debed74244/cli/config/src/lib.rs#L52-L58

thecaralice:
Yep, I'd expect it to respect legacy paths (note: the `etcetera` crate support all these conventions, the `.tree-sitter` thing is called `Unix` and the macOS thing is called `Apple`)

PR #4026, opened 2024-12-23
fix(cli): use xdg config directory on macOS

WillLillis:
Currently on macOS, the tree-sitter cli program will look inside `$HOME/Library/Application Support/tree-sitter` for a `config.json` file rather than the XDG `$HOME/.config/tree-sitter` location. This is due to our usage of the [`dirs` crate](https://crates.io/crates/dirs), which [doesn't plan on](https://github.com/dirs-dev/directories-rs/issues/47) changing this behavior or adding any additional APIs to access the XDG location. As suggested in #4023, the [`etcetera` crate](https://docs.rs/etcetera/latest/etcetera) provides this functionality as a reasonable drop-in replacement. 

This PR uses `etcetera` to point to `$HOME/.config/tree-sitter` on macOS as the default config location. If a config file is found in the old `$HOME/Library/Application Support/tree-sitter` location, it is moved to the new location and a warning message is printed (similar to the package.json's `tree-sitter` field in https://github.com/tree-sitter/tree-sitter/pull/3700.

Unfortunately I don't have access to an apple computer, so I'll have to wait for someone who does (cc @amaanq maybe?) to confirm that this actually implements the intended behavior on macOS. 

Closes https://github.com/tree-sitter/tree-sitter/issues/4023

### DIFF
--- a/cli/config/src/lib.rs
+++ b/cli/config/src/lib.rs
@@ -2,7 +2,8 @@
 
 use std::{env, fs, path::PathBuf};
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
+use etcetera::BaseStrategy as _;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
@@ -38,8 +39,24 @@ impl Config {
             return Ok(Some(xdg_path));
         }
 
-        let legacy_path = dirs::home_dir()
-            .ok_or_else(|| anyhow!("Cannot determine home directory"))?
+        if cfg!(target_os = "macos") {
+            let legacy_apple_path = etcetera::base_strategy::Apple::new()?
+                .data_dir() // `$HOME/Library/Application Support/`
+                .join("tree-sitter")
+                .join("config.json");
+            if legacy_apple_path.is_file() {
+                fs::create_dir_all(xdg_path.parent().unwrap())?;
+                fs::rename(&legacy_apple_path, &xdg_path)?;
+                println!(
+                    "Warning: your config.json file has been automatically migrated from \"{}\" to \"{}\"",
+                    legacy_apple_path.display(),
+                    xdg_path.display()
+                );
+                return Ok(Some(xdg_path));
+            }
+        }
+
+        let legacy_path = etcetera::home_dir()?
             .join(".tree-sitter")
             .join("config.json");
         if legacy_path.is_file() {
@@ -50,8 +67,8 @@ impl Config {
     }
 
     fn xdg_config_file() -> Result<PathBuf> {
-        let xdg_path = dirs::config_dir()
-            .ok_or_else(|| anyhow!("Cannot determine config directory"))?
+        let xdg_path = etcetera::choose_base_strategy()?
+            .config_dir()
             .join("tree-sitter")
             .join("config.json");
         Ok(xdg_path)
@@ -63,7 +80,7 @@ impl Config {
     ///   - Location specified by the path parameter if provided
     ///   - `$TREE_SITTER_DIR/config.json`, if the `TREE_SITTER_DIR` environment variable is set
     ///   - `tree-sitter/config.json` in your default user configuration directory, as determined by
-    ///     [`dirs::config_dir`](https://docs.rs/dirs/*/dirs/fn.config_dir.html)
+    ///     [`etcetera::choose_base_strategy`](https://docs.rs/etcetera/*/etcetera/#basestrategy)
     ///   - `$HOME/.tree-sitter/config.json` as a fallback from where tree-sitter _used_ to store
     ///     its configuration
     pub fn load(path: Option<PathBuf>) -> Result<Self> {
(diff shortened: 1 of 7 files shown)


cli/config/src/lib.rs
```diff
@@ -38,8 +39,23 @@ impl Config {
...
-        let legacy_path = dirs::home_dir()
-            .ok_or_else(|| anyhow!("Cannot determine home directory"))?
+        if cfg!(target_os = "macos") {
+            let legacy_apple_path = etcetera::base_strategy::Apple::new()?
+                .data_dir() // `$HOME/Library/Application Support/`
+                .join("tree-sitter")
+                .join("config.json");
+            if legacy_apple_path.is_file() {
+                fs::copy(&legacy_apple_path, &xdg_path)?;
+                fs::remove_file(&legacy_apple_path)?;
```
amaanq, line 49: A [`rename`](https://linux.die.net/man/2/rename) would be faster than a [`copy`](https://man7.org/linux/man-pages/man2/copy_file_range.2.html) + [`unlink`](https://linux.die.net/man/2/unlink)

```suggestion
                fs::rename(&legacy_apple_path, &xdg_path)?;
```

cli/config/src/lib.rs
```diff
@@ -38,8 +39,23 @@ impl Config {
...
+                println!(
+                    "Warning: your config.json file has been automatically migrated to \"{}\"",
+                    xdg_path.display()
```
amaanq, line 52:
```suggestion
                    "Warning: your `config.json` file has been automatically migrated from \"{}\" to \"{}\"",
                    legacy_apple_path.display(),
                    xdg_path.display()
```

cli/config/src/lib.rs
```diff
@@ -50,8 +66,8 @@ impl Config {
     }
 
     fn xdg_config_file() -> Result<PathBuf> {
-        let xdg_path = dirs::config_dir()
-            .ok_or_else(|| anyhow!("Cannot determine config directory"))?
+        let xdg_path = etcetera::base_strategy::choose_base_strategy()?
```
amaanq, line 69: Seems like we don't always have to use the full qualifier, some items are re-exported.

```suggestion
        let xdg_path = etcetera::choose_base_strategy()?
```

cli/config/src/lib.rs
```diff
@@ -63,7 +79,7 @@ impl Config {
     ///   - Location specified by the path parameter if provided
     ///   - `$TREE_SITTER_DIR/config.json`, if the `TREE_SITTER_DIR` environment variable is set
     ///   - `tree-sitter/config.json` in your default user configuration directory, as determined by
-    ///     [`dirs::config_dir`](https://docs.rs/dirs/*/dirs/fn.config_dir.html)
+    ///     [`etcetera::base_strategy::choose_base_strategy`](https://docs.rs/etcetera/latest/etcetera/#basestrategy)
```
amaanq, line 82:
```suggestion
    ///     [`etcetera::choose_base_strategy`](https://docs.rs/etcetera/*/etcetera/#basestrategy)
```

cli/loader/src/lib.rs
```diff
@@ -379,8 +380,8 @@ impl Loader {
     pub fn new() -> Result<Self> {
         let parser_lib_path = match env::var("TREE_SITTER_LIBDIR") {
             Ok(path) => PathBuf::from(path),
-            _ => dirs::cache_dir()
-                .ok_or_else(|| anyhow!("Cannot determine cache directory"))?
+            _ => etcetera::base_strategy::choose_base_strategy()?
```
amaanq, line 383:
```suggestion
            _ => etcetera::choose_base_strategy()?
```

Also, let's clean up the old cache files if this is running on macOS

cli/loader/src/lib.rs
```diff
@@ -733,8 +734,8 @@ impl Loader {
                 .join("lock")
                 .join(format!("{}.lock", config.name))
         } else {
-            dirs::cache_dir()
-                .ok_or_else(|| anyhow!("Cannot determine cache directory"))?
+            etcetera::base_strategy::choose_base_strategy()?
```
amaanq, line 737:
```suggestion
            etcetera::choose_base_strategy()?
```

amaanq:
> Unfortunately I don't have access to an apple computer, so I'll have to wait for someone who does (cc @amaanq maybe?) to confirm that this actually implements the intended behavior on macOS.

Yeah I have an Apple computer (unfortunately), and the main issue is that if the parent dir of `xdg_path` doesn't exist, the CLI will fail because writing to the file fails. If my comments are address, it works great (tested locally :wink:)

cli/loader/src/lib.rs
WillLillis: Do you think a warning message is helpful in this case?

amaanq: I think in this case it's fine to omit, as it's just cache files and nothing the user touched

amaanq (approved):

PR #4034, opened 2024-12-25
fix: re-export `StreamingIterator` and `StreamingIteratorMut`

WillLillis:
This PR aims to address the issues brought up in the followup discussion from #3504. Namely, our use of [`streaming_iterator`](https://crates.io/crates/streaming-iterator) leads to some confusion, as users of the rust lib have to import the `streaming_iterator` crate in order to use things like query captures and matches. This change simply re-exports `StreamingIterator` and `StreamingIteratorMut` for the Rust binding. A user can then import them directly from tree-sitter. 

I tested this locally with the following program (note the import of `tree_sitter::StreamingIterator`):

```rust
use anyhow::Result;
use tree_sitter::{Parser, Query, QueryCursor, StreamingIterator};

fn main() -> Result<()> {
    parser.set_language(&tree_sitter_c::LANGUAGE.into())?;
    let src = "int x = 0;";
    let tree = parser.parse(src, None).unwrap();
    let querytext = "(identifier)";
    let query = Query::new(&tree_sitter_c::LANGUAGE.into(), querytext).unwrap();
    let mut query_cursor = QueryCursor::new();

    let mut matches = query_cursor.matches(&query, tree.root_node(), src.as_bytes());
    while let Some(m) = matches.next() {
        println!("{m:#?}");
    }
    Ok(())
}
```

This compiles and gives the expected output:

```sh
» cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/stream_test`
QueryMatch { id: 0, pattern_index: 0, captures: [] }
```

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -28,7 +28,7 @@ use std::os::fd::AsRawFd;
 #[cfg(all(windows, feature = "std"))]
 use std::os::windows::io::AsRawHandle;
 
-use streaming_iterator::{StreamingIterator, StreamingIteratorMut};
+pub use streaming_iterator::{StreamingIterator, StreamingIteratorMut};
 use tree_sitter_language::LanguageFn;
 
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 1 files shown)


BobbieCarroll:
Sent to computer prior.

woodruffw:
Thanks a ton!

amaanq (approved):

PR #4036, opened 2024-12-25
update ci paths

amaanq:
### Problem

When the docs are updated, or files like the README or LICENSE, the full CI workflow is run no matter what, which is totally unnecessary.

### Solution

For the main CI, we exclude files in `docs/`, `LICENSE`, any `README.md`, `CONTRIBUTING.md`, and any of the template files.

For the docs CI, we only run a deployment if any file in `docs/` has changed.

For the bindgen CI, we only check the output of bindgen when `api.h` has changed

### DIFF
--- a/.github/workflows/bindgen.yml
+++ b/.github/workflows/bindgen.yml
@@ -2,8 +2,14 @@ name: Check Bindgen Output
 
 on:
   pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


WillLillis (approved):

jorgeferreir2k21 (commented):
V

PR #4054, opened 2024-12-31
feat(cli): parse, highlight, query, and tag from stdin

### LINKED ISSUE #1511: Tree Sitter query from stdin instead of a file
It would be very useful to be able to use the Tree Sitter CLI `query` command on chunks of code provided on standard input.

~~It would also be useful to be able to manually select a language, instead of attempting to detect it automatically from the input file.~~ Done! See https://github.com/tree-sitter/tree-sitter/issues/1511#issuecomment-1040733058

The ability to manually select a language would also facilitate reading from stdin, because your only failure mode is "parse error", rather than accidentally detecting the wrong language.

I am not a Rust developer at all, but the code doesn't look that difficult to modify. I can try to put together a PR if the Tree Sitter authors are interested (or maybe this would be a "good first issue" for someone in general).

Apologies if this has been discussed elsewhere; I didn't see it in the issue tracker.

WillLillis:
This PR adds the ability to parse, highlight, and query from stdin. If no `path` argument is provided for these subcommands, the CLI will then collect input from the user. This works pretty seamlessly with parse and query, but requires the `--scope` option for the highlight subcommand. Some quick demos with tree-sitter-c:

Parse:

stdin_parse

Highlight:

stdin_highlight

Query:

stdin_query

Closes #1511


cli/src/query.rs
```diff
@@ -19,14 +19,15 @@ use crate::{
 #[allow(clippy::too_many_arguments)]
 pub fn query_files_at_paths(
     language: &Language,
-    paths: Vec<String>,
+    paths: &Vec<String>,
```
amaanq, line 22: If we're not mutating or consuming the Vec, we can just pass in a slice instead

```suggestion
    paths: &[String],
```

cli/src/query.rs
```diff
@@ -19,14 +19,15 @@ use crate::{
...
     query_path: &Path,
     ordered_captures: bool,
     byte_range: Option<Range<usize>>,
     point_range: Option<Range<Point>>,
     should_test: bool,
     quiet: bool,
     print_time: bool,
+    stdin: bool,
```
amaanq, line 30: this is getting a bit unruly, can we refactor this to move these parameters into a `QueryFileOptions` struct, similar to the `ParseFileOptions` for the parse subcommand?

cli/src/main.rs
```diff
@@ -853,17 +853,26 @@ impl Parse {
...
-            (collect_paths(self.paths_file.as_deref(), self.paths)?, None)
+            (
+                if let Some(collected) = collect_paths(self.paths_file.as_deref(), self.paths)? {
+                    collected
+                } else {
+                    vec![get_tmp_parse_file()?.to_str().unwrap().to_owned()]
+                },
+                None,
+                true,
+            )
```
amaanq, line 875: can you hoist out the paths into its own variable above the tuple construction? This feels a bit unruly and messy...also, there's a repeated habit of calling `get_tmp_parse_file()?.to_str().unwrap().to_owned()`, and `get_tmp_parse_file()` isn't using its path besides to convert it as a string, so we should move that logic into `get_tmp_parse_file` and have it return a String, so in the end we have something like this:

```suggestion
            let paths =
                if let Some(collected) = collect_paths(self.paths_file.as_deref(), self.paths)? {
                    collected
                } else {
                    vec![get_tmp_parse_file()?]
                };
            (paths, None, true)
```

cli/src/main.rs
```diff
@@ -875,6 +884,9 @@ impl Parse {
         let mut stats = parse::Stats::default();
 
         for path in &paths {
+            if parse_stdin {
+                println!(); // place user input and parser output on separate lines
+            }
```
amaanq, line 889: Conceptually, I wouldn't think this should be in a loop, even though the length of `paths` is guaranteed to be one with the current code - it'd make more sense for this to happen outside the loop, unless we handle "multiple" stdin files

amaanq (approved):

PR #4059, opened 2025-01-02
chore(generate): remove unused fields

414owen:
As far as I can tell, these aren't used.
The real question is, why isn't dead code reported as a warning? I took a look, and couldn't figure it out.


Subtle-max:
• Good catch! 👍 • Agree, this looks like dead code.

Subtle-max:
Thanks for cleaning this up! • I'm curious about that too – why no dead
code warning? 🤔

amaanq:
that's a good question, maybe because of the derives, or the fact that it's public, but not sure. thanks for the PR!

PR #4067, opened 2025-01-04
fix(lib): propagate `last_child` status to pattern alternatives in queries

### LINKED ISSUE #1811: Query with alternation and anchor changes behavior when alternatives are swapped
The behavior of this query depends on the order of the alternatives in a `[...]` subquery. That can't be right.

test.scm:

```scheme
(block [
  (type_cast_expression)
  (expression_statement (if_expression))
]? @last .) @block
```

My intent is that this query matches every `(block)` exactly once.

test.rs:

```rust
fn f() {
    if 0 == 0 {}
    if 1 == 1 {}
}
```

Expected behavior: The query should match the function block (selecting the last if-statement as `@last`) and each nested if-block separately. 3 matches in all.

Actual behavior:

```console
$ tree-sitter query test.scm test.rs
test.rs
  pattern: 0
    capture: block, start: (0, 7), end: (3, 1)
    capture: 0 - last, start: (1, 4), end: (1, 16), text: `if 0 == 0 {}`
  pattern: 0
    capture: 1 - block, start: (1, 14), end: (1, 16), text: `{}`
  pattern: 0
    capture: block, start: (0, 7), end: (3, 1)
    capture: 0 - last, start: (2, 4), end: (2, 16), text: `if 1 == 1 {}`
  pattern: 0
    capture: 1 - block, start: (2, 14), end: (2, 16), text: `{}`
```

It matches the function-block twice. It's as if the anchor is ignored.

However, if I swap lines 2 and 3 of the query, it works as expected:

```console
$ cat test.scm
(block [
  (expression_statement (if_expression))
  (type_cast_expression)
]? @last .) @block

$ tree-sitter query test.scm test.rs
test.rs
  pattern: 0
    capture: 1 - block, start: (1, 14), end: (1, 16), text: `{}`
  pattern: 0
    capture: block, start: (0, 7), end: (3, 1)
    capture: 0 - last, start: (2, 4), end: (2, 16), text: `if 1 == 1 {}`
  pattern: 0
    capture: 1 - block, start: (2, 14), end: (2, 16), text: `{}`
```

amaanq:
- Closes #1811

### Problem

When a query pattern is marked as the last child (using the `.` anchor), any alternative patterns specified in a `[...]` block didn't inherit this constraint. This caused inconsistent query matching behavior where the order of alternatives could affect the results, as seen in #1811 where swapping alternatives in a query led to different match outcomes.

### Solution

The query parsing code now properly propagates the `last_child` status to all alternative patterns, so when a pattern is marked as the last child, we walk through its alternatives and mark them as being the last child as well. This gives us the expected matching behavior regardless of the order of alternatives in the query.

For testing, I used the example in the linked issue, but slightly tweaked, and verify that flipping the order of the alternative patterns does not change the expected matches.

### DIFF
--- a/cli/src/tests/query_test.rs
+++ b/cli/src/tests/query_test.rs
@@ -5442,3 +5442,81 @@ fn test_wildcard_behavior_before_anchor() {
         ],
     );
 }
+
+#[test]
+fn test_pattern_alternatives_follow_last_child_constraint() {
+    let language = get_language("rust");
+    let mut parser = Parser::new();
+    parser.set_language(&language).unwrap();
+
+    let code = "
+fn f() {
+    if a {} // <- should NOT match
+    if b {}
+}";
+
+    let tree = parser.parse(code, None).unwrap();
+    let mut cursor = QueryCursor::new();
+
+    let query = Query::new(
+        &language,
+        "(block
+        [
+            (type_cast_expression)
+            (expression_statement)
+        ] @last
+        .
+        )",
+    )
+    .unwrap();
+
+    let matches = {
+        let root_node = tree.root_node();
+        let matches = cursor.matches(&query, root_node, code.as_bytes());
+        collect_matches(matches, &query, code)
+            .into_iter()
+            .map(|(i, m)| {
+                (
+                    i,
+                    m.into_iter()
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Nice, thanks for fixing this!

PR #4068, opened 2025-01-04
 Improve Tree-Sitter Build Compatibility with POSIX Source Mode on Apple Platforms and make SPM aware of all source files

0xLucasMarcal:
# Improve Tree-Sitter Build Compatibility with POSIX Source Mode on Apple Platforms

## Problem
When building Tree-Sitter with `_POSIX_C_SOURCE` defined, we encounter compilation issues on Apple platforms. This occurs because:
1. The current implementation relies on `libkern/OSByteOrder.h`, which requires Darwin libraries
2. Darwin and DarwinFoundation libraries are not POSIX-compliant
3. These libraries become unavailable when `_POSIX_C_SOURCE` is defined

## Solution
Modified `portable/endian.h` to:
- Use POSIX-compliant `<machine/endian.h>` when `_POSIX_C_SOURCE` is defined
- Maintain the existing `libkern/OSByteOrder.h` implementation for non-POSIX builds
- Implement proper byte-swapping functions using `__DARWIN_OSSwapInt*` for POSIX mode

## Impact
- Ensures compatibility when building in strict POSIX environments
- Maintains existing functionality for standard builds
- Improves cross-platform compatibility
- Prevents unexpected behavior from missing source files in SPM (Swift Package Manager. The build system must be aware of all source files)

## Testing
- Verified builds work correctly with and without `_POSIX_C_SOURCE` defined
- Tested endianness conversions in both modes


Package.swift
```diff
@@ -14,12 +14,21 @@ let package = Package(
...
-                        .define("_DEFAULT_SOURCE"),
-                ]),
+                  .headerSearchPath("src"),
+                  .define("_POSIX_C_SOURCE", to: "200112L"),
+                  .define("_DEFAULT_SOURCE"),
+                ])
     ],
     cLanguageStandard: .c11
 )
+
```
ObserverOfTime, line 34:
```suggestion
```

lib/src/portable/endian.h
```diff
@@ -19,27 +19,70 @@
...
+
+#       define htobe64(x) OSSwapHostToBigInt64(x)
+#       define htole64(x) OSSwapHostToLittleInt64(x)
+#       define be64toh(x) OSSwapBigToHostInt64(x)
+#       define le64toh(x) OSSwapLittleToHostInt64(x)
+
+#       define __BYTE_ORDER    BYTE_ORDER
+#       define __BIG_ENDIAN    BIG_ENDIAN
+#       define __LITTLE_ENDIAN LITTLE_ENDIAN
+#       define __PDP_ENDIAN    PDP_ENDIAN
```
ObserverOfTime, line 43: This can be extracted outside the inner `#if`.

Package.swift
```diff
@@ -14,12 +14,21 @@ let package = Package(
     targets: [
         .target(name: "TreeSitter",
                 path: "lib",
-                sources: ["src/lib.c"],
+                exclude: [
+                  "src/wasm",
+                  "src/unicode/ICU_SHA",
+                  "src/unicode/README.md",
+                  "src/unicode/LICENSE",
```
ObserverOfTime, line 21: Don't exclude the wasm folder, in case wasm is supported in the future.
And don't exclude the vendored unicode files (particularly the license).

ObserverOfTime (commented):
Please squash into two commits:
- `fix(endian): support POSIX mode on Apple platforms`
- `build(swift): include all source files`

Package.swift
lucasmarcal-faire: Done.
We have to exclude ICU_SHA, README.md and LICENSE are not source code files, including these generates `No rule to process file` warning for client builds:

Screenshot 2025-01-04 at 8 31 47 AM

The exclude flag just omits those files from the compilation phase, but they are still preserved in the downloaded repo:

Screenshot 2025-01-04 at 8 36 10 AM

lucasmarcal-faire:
> Please squash into two commits:
> 
> * `fix(endian): support POSIX mode on Apple platforms`
> * `build(swift): include all source files`

Done!

lib/src/wasm/stdlib.c
```diff
@@ -3,6 +3,8 @@
 // as needed, and freeing is mostly a noop. But in the special case of freeing
 // the last-allocated pointer, we'll reuse that pointer again.
 
+#ifdef TREE_SITTER_FEATURE_WASM
```
lucasmarcal-faire, line 6: @ObserverOfTime Copied from `lib/src/wasm_store.c` to keep the same behavior.
Also this file requires some other changes in order to build successfully, but it's outside this PRs scope, it's related directly to wasm-enabled builds

ObserverOfTime: This shouldn't be necessary.

lucasmarcal-faire: The compiler complains about not finding `__builtin_wasm_memory_size` and `__builtin_wasm_memory_grow`. I'm not super familiar with wasm, but since all the other code that do wasm stuff are wrapped with this same condition shouldn't cause any harm

0xLucasMarcal:
Let me know if you noticed anything else. This is currently blocking us from using tools that relies on tree-sitter

lib/src/wasm/stdlib.c
amaanq: Can you add this to wasm-stdlib.h then for consistency? Also, please add a newline character at eof

Package.swift
```diff
@@ -14,12 +14,20 @@ let package = Package(
...
+                publicHeadersPath: "include",
                 cSettings: [
-                        .headerSearchPath("src"),
-                        .define("_POSIX_C_SOURCE", to: "200112L"),
-                        .define("_DEFAULT_SOURCE"),
-                ]),
+                  .headerSearchPath("src"),
+                  .define("_POSIX_C_SOURCE", to: "200112L"),
+                  .define("_DEFAULT_SOURCE"),
+                ])
```
amaanq, line 30: Retain the original formatting for spacing

lib/src/wasm/stdlib.c
0xLucasMarcal: :check:

amaanq (approved):

PR #4073, opened 2025-01-05
docs: clarify data returned by `QueryCaptures` during iteration

### LINKED ISSUE #4072: QueryCaptures iterator returns QueryMatch instead of QueryCapture
### Problem

[`QueryCaptures`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.QueryCaptures.html) is documented as being: 
> A sequence of [QueryCapture](https://docs.rs/tree-sitter/latest/tree_sitter/struct.QueryCapture.html)s associated with a given [QueryCursor](https://docs.rs/tree-sitter/latest/tree_sitter/struct.QueryCursor.html)

The documentation for [`QueryCursor::captures`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.QueryCursor.html#method.captures) seems to agree with this.

> Iterate over all of the individual captures in the order that they appear.
>
> This is useful if you don’t care about which pattern matched, and just want a single, ordered sequence of captures.

However, QueryCaptures is actually an iterator of `(QueryMatch<'query, 'tree>, usize)`.

### Steps to reproduce

Look at the generated Rust documentation for tree-sitter.

### Expected behavior

I would expect QueryCaptures to act as a sequence of QueryCapture, or the docs to be updated to better explain the current behavior.

0.24.5

### Operating system/version

Linux

WillLillis:
Closes #4072

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -304,6 +304,9 @@ pub struct QueryMatches<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]
 }
 
 /// A sequence of [`QueryCapture`]s associated with a given [`QueryCursor`].
+///
+/// During iteration, each element contains a [`QueryMatch`] and index. The index can
+/// be used to access the new capture inside of the [`QueryMatch::captures`]'s [`captures`].
 pub struct QueryCaptures<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]>> {
     ptr: *mut ffi::TSQueryCursor,
     query: &'query Query,


lib/binding_rust/lib.rs
```diff
@@ -304,6 +304,9 @@ pub struct QueryMatches<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]
 }
 
 /// A sequence of [`QueryCapture`]s associated with a given [`QueryCursor`].
+///
+/// During iteration, each element contains a [`QueryMatch`] and index. The index can
+/// be used to access the new capture inside of the [`QueryMatch`]'s `captures` field.
```
amaanq, line 309: Link to the captures field

```suggestion
/// be used to access the new capture inside of the [`QueryMatch`]'s [`captures`][QueryMatch::captures] field.
```

amaanq (approved):

maxbrunsfeld:
Thank you!

PR #4082, opened 2025-01-08
feat(xtask): add `--watch` option for `build-wasm` and `check-wasm-exports` xtask commands

WillLillis:
While working on #4074, amaanq had the idea for a "watch" feature would be helpful when working with some of the wasm parts of the project. This PR adds a `--watch` flag for the `build-wasm` and `check-wasm-exports` xtask commands, allowing for an automatic rebuild whenever a relevant file is written to. Quick demo with `check-wasm-exports`:

watch

I wasn't sure about adding was some sort of "success" print when a rebuild succeeds, as it can a little unclear whether a build/check is still running or if it passed.


xtask/src/main.rs
```diff
@@ -289,3 +299,50 @@ pub fn create_commit(repo: &Repository, msg: &str, paths: &[&str]) -> Result<Oid
...
 }
+
+#[macro_export]
+macro_rules! watch_wasm {
+    ($watch_fn:expr) => {
+        if let Err(e) = $watch_fn() {
+            eprintln!("{e}");
+        }
+
+        let watch_files = ["binding.c", "binding.js", "exports.txt", "imports.js"]
```
WillLillis, line 310: I made a best guess as to which files should be included here, but I could be missing some.

amaanq: Add prefix.js and suffix.js

amaanq (approved):
Nice, looks good but just update the list of files with prefix.js/suffix.js

Issue #4062, opened 2025-01-03
Regression - `next_named_sibling` broken on some grammars


julienvincent:
### Problem

I maintain a neovim plugin at https://github.com/julienvincent/nvim-paredit which started failing on neovim nightly. I isolated the issue down to calls to `TSNode:next_named_sibling` which now returns `nil` even when there is a node sibling (at least for the clojure grammar).

I traced it down to https://github.com/neovim/neovim/commit/28bf25d012778d911ea5757ae0e587b0bdd7c3ec in the neovim project which is a change that pulled in https://github.com/tree-sitter/tree-sitter/commit/e3c82633389256ccc2c5ab2e509046cbf20453d3 from tree-sitter.

So it seems to me that https://github.com/tree-sitter/tree-sitter/commit/5d1be545c439eba4810f34a14fef17e5f76df6c0 (https://github.com/tree-sitter/tree-sitter/commit/e3c82633389256ccc2c5ab2e509046cbf20453d3) introduced a regression in this API.

This is reproduced using the clojure grammar and the following text:

```clojure
|(a) b c
```

Which produces the following tree:

```clojure
(source ; [0, 0] - [1, 0]
  (list_lit ; [0, 0] - [0, 3]
    value: (sym_lit ; [0, 1] - [0, 2]
      name: (sym_name))) ; [0, 1] - [0, 2]
  (sym_lit ; [0, 4] - [0, 5]
    name: (sym_name)) ; [0, 4] - [0, 5]
  (sym_lit ; [0, 6] - [0, 7]
    name: (sym_name))) ; [0, 6] - [0, 7]
```

Calling `next_named_sibling` on the `list_lit` node returns nil instead of `sym_lit`.

For full context, here is the neovim issue:

https://github.com/neovim/neovim/issues/31425

### Steps to reproduce

I'm not sure how to setup an isolated reproduce-case in tree-sitter, but there is an indirect reproduce script in the linked neovim issue here: https://github.com/neovim/neovim/issues/31425#issuecomment-2562578843

If I can get some pointers I'd be happy to set something that doesn't have a neovim dependency.

### Expected behavior

I expect this API to return the next named sibling when there is one.

### Tree-sitter version (tree-sitter --version)

e3c82633389256ccc2c5ab2e509046cbf20453d3

### Operating system/version

macOS 14.3

julienvincent:
[...]

PR #4085, opened 2025-01-08
Add failing test showing empty next named sibling

julienvincent:
This adds a test to reproduce the regression described in https://github.com/tree-sitter/tree-sitter/issues/4062.

This extracts the relevant parts of the clojure grammar to demonstrate the issue.

This test passes on commits prior to 5d1be54, for example this was tested on v0.24.4.

### DIFF
--- a/cli/src/tests/node_test.rs
+++ b/cli/src/tests/node_test.rs
@@ -334,6 +334,33 @@ fn test_next_sibling_of_zero_width_node() {
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


amaanq:
There's no need to open a PR demonstrating a bug, it'd be better to add this as a reproducer in your linked issue (either by pasting the code in the description, or by linking to a fork/branch of yours that includes these changes), though I do appreciate the time taken to create a very minimal reproducer.

PR #4087, opened 2025-01-08
fix(wasm): check docker presence without arguments

guilhas07:
Problem:
When the Docker daemon is not running, the wasm build process
incorrectly reports that Docker is not found in PATH. This happens
because the `docker info` command fails when the daemon is not active,
even though the `docker` executable is present.

Solution:
Replace the use of `docker info` with a simple execution of `docker`
(without arguments) to reliably check for its presence, regardless of
the daemon's state.


amaanq:
I'm fine with the change, but I think you meant to also change this spot:
https://github.com/tree-sitter/tree-sitter/blob/5a854e603021608bf27a10140d5a82317d9686b7/xtask/src/build_wasm.rs#L36

PR #4089, opened 2025-01-09
feat(playground): provide colored highlight for captures in code editor

### LINKED ISSUE #1714: Color blind mode for playground?
Heyo! I was trying to use the tree-sitter playground "query" option, and legitimately 100% thought it might be broken in Firefox, or something like that. I would enter a valid query, but there wasn't any output anywhere, it didn't highlight the matching text, or anything like that:

image

After chatting with coworkers, apparently it *does* work and `package` and `golang` are highlighted.. I just can't see it 😅 The light-text on light-background is unnoticeable for me (granted, I have quite bad color-blindness compared to most people with it!)

I have found that the playground uses CodeMirror for the rendering, I am wondering how hard it'd be to add a "colorblind" checkbox to the page & change the `color: foo` CSS to be `background-color: foo; color: white;` so that it ends up like this:

image

No worries if there's no time to do this, I can workaround it by playing with queries in Go instead. Just thought it was worth filing.

WillLillis:
This slightly alters how captures how captures are differentiated in the playground's editor, making it more accessible to those with color blindness as described in #1714. 

Quick demo:

accessibility_captures

While the original issue just mentioned captures in the code editor, should this change also be applied to the captures in the query editor as well?

Closes #1714

### DIFF
--- a/cli/src/playground.html
+++ b/cli/src/playground.html
@@ -32,6 +32,11 @@
       </div>
 
       <div class="header-item">
+        <input id="accessibility-checkbox" type="checkbox">
+        <label for="accessibility-checkbox">accessibility</label>
+      </div>
+
+      <div class="header-item">
         <label for="update-time">parse time: </label>
         <span id="update-time"></span>
       </div>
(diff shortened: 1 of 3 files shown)


0958436455sarayut (approved):
[](url)

amaanq (approved):

PR #4092, opened 2025-01-10
docs: add mdbook-admonish for better tooltips

### LINKED ISSUE #867: Sharing definition of the `word` rule produces confusing error message
I've found an unexpected error when the rule that's chosen for keyword extraction has the same definition as another rule. Here's a minimal grammar that demonstrates:

```
module.exports = grammar({
    name: "word_identifier_error",

    word: $ => $.identifier,

    rules: {
        source_file: $ => repeat($.identifier),

        identifier: $ => /[_a-zA-Z]\w*/,

        type: $ => /[_a-zA-Z]\w*/,
    }
});
```

With this definition of `type`, `tree-sitter generate` produces an error:

> Non-terminal symbol 'identifier' cannot be used as the word token

which I found quite confusing since `identifier` does appear to be terminal, being that it's a single regex literal. (Any tweak to either of the regexes avoids the error, so I guess something is being interned and shared somehow?)

The [rust grammar's `type_identifier` and `field_identifier` rules](https://github.com/tree-sitter/tree-sitter-rust/blob/8746bd4b584b8063ee8e445bf31015e887417d33/grammar.js#L1432-L1433) use `alias` for this situation, so I guess that's the right way to avoid this problem, and I will adopt that for my grammar.

But I wonder if tree-sitter could recognize this situation and produce a more specific error message in the future.

amaanq:
- Closes #867

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,6 +17,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Set up Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
       - name: Install mdbook
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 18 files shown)


WillLillis (approved):

jabarihadi332 (commented):
importante

PR #4097, opened 2025-01-11
fix(lib): use inclusive range check for non-empty nodes in next sibling computation

### LINKED ISSUE #4062: Regression - `next_named_sibling` broken on some grammars

amaanq:
- Closes #4062

### DIFF
--- a/lib/src/node.c
+++ b/lib/src/node.c
@@ -264,7 +264,15 @@ static inline TSNode ts_node__next_sibling(TSNode self, bool include_anonymous) 
     NodeChildIterator iterator = ts_node_iterate_children(&node);
     while (ts_node_child_iterator_next(&iterator, &child)) {
       if (iterator.position.bytes <= target_end_byte) continue;
-      if (ts_node_start_byte(child) < ts_node_start_byte(self)) {
+      uint32_t start_byte = ts_node_start_byte(self);
+      uint32_t child_start_byte = ts_node_start_byte(child);
+
+      bool is_empty = start_byte == target_end_byte;
+      bool contains_target = is_empty ?
+        child_start_byte < start_byte :
+        child_start_byte <= start_byte;
+
+      if (contains_target) {
         if (ts_node__subtree(child).ptr != ts_node__subtree(self).ptr) {
           child_containing_target = child;
         }


ribru17:
Maybe could also add the posted test case to the suite?

amaanq:
Yeah, we should have a test for this, but I found the linked test case a bit too much - for one, I don't want to tie downstream grammars (like clojure) to our test suite, and two, I found the pr with the "minified" grammar a bit too much still, for a few of reasons:

- if we're going to create custom grammars in `test_grammars` for every single core issue, we're going to end up with a ton of grammars that'll kinda dilute the point of having those grammars - they're more like integration tests, testing E2E behavior of parsers that's not trivial to replicate with a small unit test...
- ... so for issues that should in theory be handled via a unit test (in this case, testing the behavior of getting the next sibling), we should be able to replicate this with one of the ~20 upstream grammars, but I didn't want to waste too much time trying to come up with a test case
- the minified grammar was a bit convoluted, and could've probably been simplified even more

It _would_ be nice to have a test for this, if you manage to come up with a case using an upstream grammar that'd be awesome :grin:

ribru17:
Ah that makes sense. I'll make a PR if I can come up with a smaller repro👍

clason:
Can that fix be backported?

amaanq:
> Can that fix be backported?

Yeah, just released 0.24.7 with the fix

PR #4099, opened 2025-01-11
fix(rust): adapt to a few new nightly lints

WillLillis:
Looks like redundant `continue`'s are on the chopping block now.


amaanq (approved):
neat

PR #4104, opened 2025-01-11
build(bindings): move header to tree_sitter subdirectory 

Scorg:
This patch allows users to include the parser by the same path from local build as well as installed location. Previously it was not possible to include the header prior to installing the built parser.

EDIT: typo.

### DIFF
--- a/cli/src/templates/cmakelists.cmake
+++ b/cli/src/templates/cmakelists.cmake
@@ -28,7 +28,10 @@ add_library(tree-sitter-PARSER_NAME src/parser.c)
 if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/src/scanner.c)
   target_sources(tree-sitter-PARSER_NAME PRIVATE src/scanner.c)
 endif()
-target_include_directories(tree-sitter-PARSER_NAME PRIVATE src)
+target_include_directories(tree-sitter-PARSER_NAME
+                           PRIVATE src
+                           INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/bindings/c>
+                                     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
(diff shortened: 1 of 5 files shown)


cli/src/templates/cmakelists.cmake
```diff
@@ -28,7 +28,9 @@ add_library(tree-sitter-PARSER_NAME src/parser.c)
 if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/src/scanner.c)
   target_sources(tree-sitter-PARSER_NAME PRIVATE src/scanner.c)
 endif()
-target_include_directories(tree-sitter-PARSER_NAME PRIVATE src)
+target_include_directories(tree-sitter-PARSER_NAME 
+                           PRIVATE src
+                           INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/bindings/c> $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}> )
```
ObserverOfTime, line 33:
```suggestion
                           INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/bindings/c>
                                     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}> )
```

ObserverOfTime (commented):
This will need a migration.

Scorg:
Added the migration. Mind you this is my first time doing something in Rust, so there very well may be a better way to do this.

ObserverOfTime:
You should also update the file contents.

Scorg:
Sorry, maybe I'm missing something, but I kinda don't see much there to update. Currently the header is written only if it's missing, it's contents are not checked, and besides the language name suffix, there is little that can change. 

We can overwrite the file, but as I understood it not something that the tool does freely.

ObserverOfTime:
The files _referencing_ the header.

Scorg:
Done

ObserverOfTime (approved):

PR #4107, opened 2025-01-11
feat!: remove `ts_node_child_containing_descendant` and migration code for `tree-sitter.json`

amaanq:
Just some cleanup before 0.25

### DIFF
--- a/lib/include/tree_sitter/api.h
+++ b/lib/include/tree_sitter/api.h
@@ -605,25 +605,15 @@ TSStateId ts_node_next_parse_state(TSNode self);
 
 /**
  * Get the node's immediate parent.
- * Prefer [`ts_node_child_containing_descendant`] for
+ * Prefer [`ts_node_child_with_descendant`] for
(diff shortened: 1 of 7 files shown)


lib/include/tree_sitter/api.h
```diff
@@ -605,25 +605,15 @@ TSStateId ts_node_next_parse_state(TSNode self);
...
 /**
  * Get the node's immediate parent.
- * Prefer [`ts_node_child_containing_descendant`] for
+ * Prefer [`ts_node_child_with_descendant`] for
  * iterating over the node's ancestors.
  */
 TSNode ts_node_parent(TSNode self);
 
-/**
- * @deprecated use [`ts_node_contains_descendant`] instead, this will be removed in 0.25
```
clason, line 614: Shouldn't that be `ts_node_child_with_descendant`?

amaanq: oops, yeah it should've been

PR #4121, opened 2025-01-16
Rewrite the web bindings in TypeScript (& more)

### LINKED ISSUE #1566: Create flag for installing web-tree-sitter built with —debug
Can we ship web-tree-sitter with two builds of the library, i.e., with and without —debug, and provide an install flag, e.g., —web-tree-sitter-debug, which chooses which of the two versions to keep? That would enable folks to switch to the —debug version without having to build from source.

amaanq:
- Closes #1566

### Problem

The web bindings feel hard to maintain, and also hard to debug. Additionally, there's a few missing components for the web bindings that would aid downstream users immensely - that being bundling the sourcemaps, and having an option to install a "debug" build of web-tree-sitter.

### Solution

We've implemented the missing components outlined above, as well as rewriting the library to TypeScript, as it feels much easier to navigate than before with the types being laid out pretty nicely (including the FFI functions!). However, with this change comes some (hopefully) positive changes to the build process.

Because we're using TypeScript, we have a new compilation step added relies on esbuild to bundle our TypeScript files into one JavaScript file with a sourcemap. 

~~However, sourcemaps were a particular pain to implement for the final JS file, because we have two "phases" of compilation - one from TS -> JS (esbuild), and one from JS -> Emscripten-JS. Emscripten doesn't provide a way to generate sourcemaps for the generated JS code, as it just copy-pastes our `post-js` as is. In theory we *should* be able to use our TS -> JS source map since our source code wasn't modified, but that's not the case because the line numbers have shifted. As such, we have to run a post-processing step to "relocate" the sourcemap line number information. This ends up working pretty well, but was not fun to figure out.~~

The above is no longer true, since I've rewritten the way the higher-level API interacts with the WASM module. Instead of gluing it all together inside the module with `pre-js` and `post-js` hacks, we're instead consuming the WASM module like a normal module, and bundling that with the final JS code. This is *way* easier, and doesn't require hacks to the sourcemaps. 

#### Debug Info

For debug information, we build the wasm files with features such as assertions and debug symbols. We store these in the `debug` folder, and optionally allow a user to use them with a sub-import, `web-tree-sitter/debug`. Effectively, tacking on `/debug` to your import will let you automatically use the debug build to see what's going on. 

In GH releases, these debug files would show up as `tree-sitter-debug.{js|wasm}(.map)?`.

#### TypeScript, CommonJS, and ESM

Rewriting the library in TypeScript also helped me improve the public type definitions, as we can autogenerate them now, and writing the tests in TypeScript was just a much smoother process overall. Hopefully others feel the same way as they peruse and navigate the codebase (and potentially contribute back, as I've added a nice little contributing.md file).

Some notes for readers - we now actually support both CommonJS and ESM modules, by invoking `emcc` with `EXPORT_ES6=1`, and `esbuild` with the format set to `cjs` *and* `esm`. 

**The default JS file is now an ES6 module**, and will be the file that's uploaded to GH releases. For users using the NPM registry, this is automatically taken care of for you with [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) (including for the debug sub-export).

I'll let this sit for a bit so people can try it out locally.

cc @verhovsky @wenkokke I think you'd both be interested in these changes (sorry for the ping if that's no longer true)


verhovsky:
> we have to run a post-processing step to "relocate" the sourcemap line number information. This ends up working pretty well, but was not fun to figure out.

my condolences lol. If there's an easy way to do it, it would be nice to sanity check the source map in a post processing lint step. Something like after it's generated do a jump to definition (if there's a way to do a jump to definition without spinning up an entire IDE) on some string that shouldn't change and check the line you jump to actually contains that string.

WillLillis:
> If there's an easy way to do it, it would be nice to sanity check the source map in a post processing lint step. Something like after it's generated do a jump to definition (if there's a way to do a jump to definition without spinning up an entire IDE) on some function and check the line you jump to actually contains that function name.

This could be done with Neovim running in headless mode with a few autocommands in its `init.lua`. It's a little hacky, but I'm using a similar approach to write an LSP integration test library currently. :)

lib/binding_web/script/postinstall.cjs
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
+const path = require('path');
```
verhovsky, line 2: why is this cjs?

Makefile
```diff
@@ -109,6 +109,9 @@ lint:
 	cargo +nightly fmt --all --check
 	cargo +nightly clippy --workspace --all-targets -- -D warnings
 
+lint-web:
+	cd lib/binding_web && npm ci && npm run lint
```
ObserverOfTime, line 113:
```suggestion
	npm --prefix lib/binding_web ci
	npm --prefix lib/binding_web run lint
```

.github/workflows/wasm_exports.yml
```diff
@@ -32,5 +32,11 @@ jobs:
         env:
           CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
 
+      - name: Build WASM Library
+        run: |
+          cd lib/binding_web
+          npm ci
+          npm run build:debug
```
ObserverOfTime, line 39:
```suggestion
        working-directory: lib/binding_web
        run: npm ci && npm run build:debug
```

xtask/src/check_wasm_exports.rs
```diff
@@ -85,7 +77,7 @@ fn check_wasm_exports() -> Result<()> {
         ])
         .output()
         .expect("Failed to run wasm-objdump");
-    bail_on_err(&wasm_objdump, "Failed to run wasm-objdump")?;
+    bail_on_err(&wasm_objdump, "Failed to run wasm-objdump").unwrap();
```
WillLillis, line 80: Was this `?` -> `unwrap()` change (and the one below) intentional or just for debugging?

lib/binding_web/script/postinstall.cjs
amaanq: yeah fair point

lib/binding_web/script/build-sourcemap.ts
```diff
@@ -0,0 +1,62 @@
...
+  const finalJs = readFileSync('tree-sitter.js', 'utf8').split('\n');
+
+  const lineMap = new Map<number, number>();
+
+  for (let distLine = 0; distLine < distJs.length; distLine++) {
+    const line = distJs[distLine].trim();
+    if (!line) continue;
+
+    for (let finalLine = 0; finalLine < finalJs.length; finalLine++) {
+      if (finalJs[finalLine].trim() === line) {
```
maxbrunsfeld, line 16: This seems reliant on every line of source code being unique, but I don't think that'll be the case. Could we do better by enforcing that `lineMap.get(row)` is monotonically-increasing function of `row`? In other words, rather than taking the *first* matching line, take the first matching line that is *after* any previously-matched line?

lib/binding_web/tree-sitter-web.d.ts
```diff
@@ -8,7 +8,7 @@ declare module "web-tree-sitter" {
     delete(): void;
     parse(
       input: string | Parser.Input,
-      oldTree?: Parser.Tree,
+      oldTree?: Parser.Tree | null,
```
maxbrunsfeld, line 11: Is this entire file redundant now? Can it be generated from the typescript source?

lib/binding_web/test/node.test.ts
```diff
@@ -0,0 +1,591 @@
+import { describe, it, expect, beforeAll, beforeEach, afterEach } from 'vitest';
+import TSParser, { type Language, type Tree, type SyntaxNode } from 'web-tree-sitter';
```
maxbrunsfeld, line 2: Why is the import called `TSParser`? I think it should just be `Parser`.

maxbrunsfeld (commented):
This seems like a good change.

I worry a little bit about the `+-` count here: the codebase has grown by more than 5000 lines, and I can't tell exactly why.

I left a couple of questions.

lib/binding_web/test/node.test.ts
amaanq: Sure, but we have a variable called "Parser", which is of type "Parser" (but there is a conflict of names there, and we can call the default import whatever we want)

amaanq:
> I worry a little bit about the +- count here: the codebase has grown by more than 5000 lines, and I can't tell exactly why.

A lot of that is from package.json (about 4600), and from the tests being moved over to ts files using vitest (hence the large minus diff as well, but overall *not* a large gain)

lib/binding_web/test/node.test.ts
maxbrunsfeld: Didn’t the default export used to be the Parser class? Seems like the the type and the class should be the same thing, not two different imports, similar to Node, Tree, etc.

amaanq: The default export *is* the parser class, but when you import the default export, you can name it whatever you like. I did this because we have a declaration below, that looks like so:

```ts
import TSParser from 'web-tree-sitter';
const Parser: typeof TSParser = await import('..').then(m => m.default);
```

Because the variable, which is of type `Parser` from the `.d.ts` file, is *also* called `Parser`, we get a naming conflict. So I decided to just name the import `TSParser`. I'll swap them so it's a bit clearer.

ObserverOfTime: It might be better to ditch the default export.

lib/binding_web/script/build-sourcemap.ts
amaanq: Oh wow, that is a really good catch - fixing that resulted in the sourcemap being totally accurate now, and coverage data is a lot more richer than before...

lib/binding_web/test/node.test.ts
maxbrunsfeld: I still don’t get what Parser is in this code. In that code you wrote, is Parser not identical to TSParser, since the helper just te-exports Parser?

maxbrunsfeld: Reading that, it looks like Parser and TSParser are just two different names for the same class. Maybe I’m missing something very subtle though?

amaanq: one's used for type hinting/info (TSParser), and one's used as the type (Parser, declared later)

maxbrunsfeld: I think the default export should just be the Parser class, which can be used as a type and also at runtime. I think that’s usually how classes are in Typescript.

maxbrunsfeld:
Ah ok, makes sense about the package lock making up a large portion of the added line count.

lib/binding_web/test/node.test.ts
amaanq: Yeah the default export is the Parser class currently, the only reason we import 'web-tree-sitter' (which pulls in type info from `tree-sitter-web.d.ts`, is so we can annotate the "Parser" used in the tests with the type, since otherwise it's just given `any` when we import `..` (because we're importing the final compiled JS file). I'll push up some changes that should make it a lot clearer imo.

amaanq: Also, I'd like to get rid of the type declarations file, but it's sorta hard. If we could restructure it a bit that'd work out, but the problem is, because every class/type is in its own file, tsc's generation of the types file throws each class in its own module like:

```ts
declare module "language" {
    import { Internal } from "constants";
    import { LookaheadIterator } from "lookahead_iterator";
    import { Query } from "query";
    export class Language {
        private [0];
        types: string[];
        fields: (string | null)[];
        constructor(internal: Internal, address: number);
        get name(): string | null;
        get version(): number;
        get fieldCount(): number;
        get stateCount(): number;
        fieldIdForName(fieldName: string): number | null;
        fieldNameForId(fieldId: number): string | null;
        idForNodeType(type: string, named: boolean): number | null;
        get nodeTypeCount(): number;
        nodeTypeForId(typeId: number): string | null;
        nodeTypeIsNamed(typeId: number): boolean;
        nodeTypeIsVisible(typeId: number): boolean;
        get supertypes(): number[];
        subtypes(supertype: number): number[];
        nextState(stateId: number, typeId: number): number;
        lookaheadIterator(stateId: number): LookaheadIterator | null;
        query(source: string): Query;
        static load(input: string | Uint8Array): Promise<Language>;
    }
}
```

Which is not what we want. I haven't been successful in wrangling this to throw everything under the Parser namespace, and remove "internal" stuff

maxbrunsfeld: I think it’s fine if it’s not all part of the Parser namespace anymore. Maybe we stop having a default export, and just export all of the classes from the top-level module?

maxbrunsfeld: Sorry, I didn’t mean to convey that I was attached to the default export in particular, I care more about the classes (including Parser) having a single way to import them, both for types and for runtime.

amaanq: Oh yeah, I totally agree with that, and removing the default export + namespace mix would be a lot easier to manage!

amaanq: Ok, this should be solved now, with some help from @savetheclocktower, [`dts-buddy`](https://github.com/Rich-Harris/dts-buddy) seems to be the best way to reliably generate a `.d.ts` file, and it also includes a sourcemap file, to enable rich goto def for users, going to the actual TypeScript source files and not just the `.d.ts` file.

lib/binding_web/README.md
```diff
@@ -50,6 +50,14 @@ await Parser.init();
 // the library is ready
```
savetheclocktower, line 50: The code examples earlier in `README.md` need updating — they use

```js
import Parser from 'web-tree-sitter';
```

and need to be changed to read

```js
import { Parser } from 'web-tree-sitter';
```

since there is now no default export.

savetheclocktower: It's also worth mentioning in the README that there are CJS and ESM entry points, and that you can opt into the debug build with (e.g.):

```js
import { Parser} from 'web-tree-sitter/debug';
```

lib/binding_web/test/node.test.ts
```diff
@@ -0,0 +1,591 @@
+import { describe, it, expect, beforeAll, beforeEach, afterEach } from 'vitest';
+import type { Parser as ParserType, Language, Tree, Node } from '../src';
```
maxbrunsfeld, line 2: Why can't `Parser` just be imported from `../src`, as opposed to returning it from the `helper`? I would think that the helper would just provide the languages.

lib/binding_web/test/helper.ts
```diff
@@ -0,0 +1,29 @@
+import { type Parser as ParserType, type Language as LanguageType } from '../src';
+import path from 'path';
+
+// @ts-expect-error We're intentionally importing ../tree-sitter.js
+import { Parser as ParserImpl, Language as LanguageImpl } from '..';
```
maxbrunsfeld, line 5: Why not just this:

```ts
import { Parser, Language } from "../src"

export type LanguageName = 'bash' | 'c' | 'cpp' | 'embedded-template' | 'go' | 'html' | 'java' | 'javascript' | 'jsdoc' | 'json' | 'php' | 'python' | 'ruby' | 'rust' | 'typescript' | 'tsx';

function languageURL(name: LanguageName): string {
  // ...
}

export default Parser.init().then(async () => ({
  languageURL,
  C: await Language.load(languageURL('c')),
  EmbeddedTemplate: await Language.load(languageURL('embedded-template')),
  HTML: await Language.load(languageURL('html')),
  JavaScript: await Language.load(languageURL('javascript')),
  JSON: await Language.load(languageURL('json')),
  Python: await Language.load(languageURL('python')),
  Rust: await Language.load(languageURL('rust')),
}));
 ```

The helper is just about asynchronously calling `init`, and then loading some languages.

The impl and the type should be able to be one and the same.

amaanq: Yeah, this wasn't possible before when we wrapped the higher level code inside the wasm module (since we'd *have* to import the root tree-sitter.js), but now it is, updated

lib/binding_web/test/node.test.ts
amaanq: We can't, because we have to ensure that the module's been initialized. Importing `Parser` directly and using that here doesn't guarantee that, which is why we use the one exported from the helper. We could ensure it's initialized in all the tests, though I'm not sure if that's necessary.

maxbrunsfeld: I agree that we need to await the promise returned by the helper, but I don't think there's any advantage of importing the `Parser` class itself from `helper`. You can just import `Parser` from `src`  at the top.

As written, this test makes it look like `web-tree-sitter` has a super weird API where you have to import the Parser type separately from the Parser class itself, and do a manual cast to `ParserType`.

savetheclocktower:
OK, the CJS build works great for me without any further tweaking.

I'm curious to see how the imports will work once the module is published, but all that is done via CI right now, so I'll have to wait and see.

lib/binding_web/test/node.test.ts
amaanq: Yeah you're right, I had sworn I'd tried that and it had errored out for me, but it seems to work now.

maxbrunsfeld (approved):

maxbrunsfeld (commented):
Is `web-tree-sitter.d.ts` a generated file now?

amaanq:
Yes, we use `dts-buddy` to autogenerate it from our TypeScript files - surprisingly it's not built-in to the typescript compiler to generate public types easily.

PR #4122, opened 2025-01-17
feat: allow parser balancing to be cancellable

### LINKED ISSUE #4019: `ts_parser_parse()` goes over the time limit
### Problem

Tree balancing, performed at the end of parsing, doesn't seem to check the timeout duration. 

For a sample lua file:
```lua
local _ = 
vim.api.nvim_set_hl(0, 'Test2', { bg = 'green' })
vim.api.nvim_set_hl(0, 'Test3', { bg = 'blue' })
... repeates for 50'000 lines
```

The `ts_subtree_balance()` function processes 1'450'005 subtrees sequentially and takes 83ms (with timeout being 3ms).

This becomes a problem when parsing is performed in the main thread of a text editor. While asynchronous parsing keeps the editor responsive, uninterrupted balancing blocks the editor, which is perceived as a lag spike.

From https://github.com/neovim/neovim/pull/31631 

### Steps to reproduce

**This is different from the example above.**
Make a directory somewhere, place the `patch.diff` file there, and run the commands below (for linux).

```shell
echo "vim.api.nvim_set_hl(0, 'Test2', { bg = 'green' })" | awk '{for(i=0;i<50000;i++) print $0}' > file.lua
git clone https://github.com/tree-sitter/tree-sitter.git
cd tree-sitter
git apply ../patch.diff
cargo build
cd ..
git clone https://github.com/MunifTanjim/tree-sitter-lua
cd tree-sitter-lua
../tree-sitter/target/debug/tree-sitter generate
../tree-sitter/target/debug/tree-sitter p ../file.lua -qst
```

<details>
<summary>patch.diff</summary>

```diff
diff --git a/lib/src/clock.h b/lib/src/clock.h
index 7a13185e..1b607132 100644
--- a/lib/src/clock.h
+++ b/lib/src/clock.h
@@ -100,6 +100,13 @@ static inline bool clock_is_gt(TSClock self, TSClock other) {
   return self.tv_nsec > other.tv_nsec;
 }
 
+static inline bool print_diff(const char* msg, TSClock start, TSClock end) {
+  uint64_t diff = (end.tv_sec - start.tv_sec) * 1000 + (end.tv_nsec - start.tv_nsec) / 1000000;
+  printf("%s: %lu ms\n", msg, diff);
+  return true;
+}
+
+
 #else
 
 // POSIX without monotonic clock support
@@ -141,6 +148,13 @@ static inline bool clock_is_gt(TSClock self, TSClock other) {
   return self > other;
 }
 
+static inline bool print_diff(const char* msg, TSClock start, TSClock end) {
+  uint64_t diff = (end - start) / (CLOCKS_PER_SEC / 1000.0);
+  printf("%s: %lu ms\n", msg, diff);
+  return true;
+}
+
+
 #endif
 
 #endif  // TREE_SITTER_CLOCK_H_
diff --git a/lib/src/parser.c b/lib/src/parser.c
index 4c2976cd..5d8afcf4 100644
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -2051,6 +2051,8 @@ TSTree *ts_parser_parse(
     self->end_clock = clock_null();
   }
 
+  TSClock before_parse = clock_now();
+
   uint32_t position = 0, last_position = 0, version_count = 0;
   do {
     for (
@@ -2108,8 +2110,15 @@ TSTree *ts_parser_parse(
     }
   } while (version_count != 0);
 
+  TSClock after_parse = clock_now();
+
+  print_diff("parse", before_parse, after_parse);
+
   ts_assert(self->finished_tree.ptr);
+  TSClock before_balance = clock_now();
   ts_subtree_balance(self->finished_tree, &self->tree_pool, self->language);
+  TSClock after_balance = clock_now();
+  print_diff("balance", before_balance, after_balance);
   LOG("done");
   LOG_TREE(self->finished_tree);
 ```

</details>

The output is:
```console
parse: 3284 ms
balance: 128 ms
../file.lua     3412.09 ms         732 bytes/ms

Total parses: 1; successful parses: 1; failed parses: 0; success percentage: 100.00%; average speed: 732 bytes/ms
```

Current tree-sitter: https://github.com/tree-sitter/tree-sitter/commit/c132f1928f447727062c0a9cda70a4a13fdb8c20
Current tree-sitter-lua: https://github.com/tree-sitter-grammars/tree-sitter-lua/commit/34e60e7f45fc313463c68090d88d742a55d1bd7a

### Expected behavior

Subtree balancing should be interruptible

0.24.5, latest (https://github.com/tree-sitter/tree-sitter/commit/c132f1928f447727062c0a9cda70a4a13fdb8c20)

### Operating system/version

Ububtu 24.04

amaanq:
- Closes #4019

### Problem

Tree-sitter's timeout functionality does not allow tree-balancing to be canceled. This is problematic, because if a tree is quite large, balancing can take dozens to hundreds of milliseconds, especially on older hardware, which causes noticeable delays in editors. Ideally, balancing should be an operation that's cancelable and resumable.

### Solution

In this PR, I've moved the logic for balancing trees into `parser.c`, so we can pass in the parser struct to access the data needed to determine if we should cancel balancing. As such, to avoid duplicating the logic for detecting cancellation, I've moved that to its own internal function, `ts_parser__check_progress`, that returns false if we should cancel. 

In the progress callback, if we are in the middle of balancing, the current byte offset should *not* change, however, one issue I noticed with this is that when *resuming* parsing from canceled balancing, the byte offset got reset to 0. This is because every time we call `ts_parser_parse_with_options`, the parse state got reset to a default struct with the given payload field, when we should only be setting the payload field.

The test added verifies parsing is canceled during balancing by inspecting the current byte offset - during parsing, this offset is advanced, but during balancing it's unchanged - we can use this to know if we *are* in balancing by storing this in a temporary variable. When we resume parsing, we also assert that the current byte offset is always the stored temporary byte offset, as it should be unchanged. At the end, we have an assertion verifying the parsed tree looks like what we expect (no errors, with 100 function nodes).


lib/src/parser.c
```diff
@@ -1837,8 +1845,48 @@ static unsigned ts_parser__condense_stack(TSParser *self) {
...
 }
 
+static bool ts_parser__balance_subtree(TSParser *self) {
+  Subtree finished_tree = self->finished_tree;
+
+  array_clear(&self->tree_pool.tree_stack);
+
+  if (ts_subtree_child_count(finished_tree) > 0 && finished_tree.ptr->ref_count == 1) {
+    array_push(&self->tree_pool.tree_stack, ts_subtree_to_mut_unsafe(finished_tree));
+  }
```
ribru17, line 1855: Do these lines undo previous work done by other iterations of the function? Ie would it make sense to take them out of the function and just run them once, before the `balance` label?

amaanq: Oh yes, you're right. I had thought about that before but forgot when I was writing this up.

WillLillis (approved):

ribru17:
Was it intentional to keep the `  array_clear(&self->tree_pool.tree_stack); ...` logic in the `balance_subtree` function?

amaanq:
ah crap, thought I had pushed that up when I posted the patch :facepalm: will follow-up

ribru17:
No worries :+1:

maxbrunsfeld:
Very cool that you've managed to make this interruptible and found a way to test it as well. Nice work @amaanq

PR #4123, opened 2025-01-18
ci: add arm ubuntu runner, & use latest runners

amaanq:
Arm64 runners are in preview, so this will be nice to leverage for build time, but for some reason, the suffix is arm, not arm64 or aarch64 :laughing: [link](https://github.com/orgs/community/discussions/148648#discussion-7793082)

Ok, so even just checking out the repo [fails](https://github.com/tree-sitter/tree-sitter/actions/runs/12839301758/job/35806355649?pr=4123)....I'll keep this as draft for now


amaanq:
Looks like GH fixed their runners permission issue.

PR #4128, opened 2025-01-20
ci: also use the arm runners for `linux-arm`

amaanq:
Arm32 might be also buildable on arm64 runners


amaanq:
Ok, guess not. Cross's containers don't support linux arm64, and wasmtime fails with "no supported isa found for arch `arm`"

PR #4130, opened 2025-01-21
feat: remove `lazy_static` in favor of `LazyLock`

nyurik:
Switch to the built-in `std::sync::LazyLock`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1017,12 +1017,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "lazy_static"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


amaanq:
Nice, I was planning to do this eventually, but you beat me to it :grin:

nyurik:
hmm, seems like there are still some issues with it being non-sync... examining...

amaanq:
I think it's ok to implement Sync for `TagsConfiguration`, because the raw pointer is pointing to data owned by the tags config, just in a c-compatible layout. If you want, I can push this up, otherwise feel free to update it

nyurik:
@amaanq want to push it to my branch?

amaanq (approved):

PR #4135, opened 2025-01-21
feat: add the semantic version to `TSLanguage`, and expose an API for retrieving it

### LINKED ISSUE #2611: Grammar versions and dealing with breaking changes
Just about any time a change is made to a grammar, the change is somehow breaking to consumers of the grammar. Right now in Emacs we don't have a great way to tell what version of a grammar is loaded without resorting to writing a simple query and seeing if it throws an error.

For example, in the javascript grammar a change was made that renamed a node. In Emacs we have a special check when writing indentation rules for this https://github.com/emacs-mirror/emacs/blob/0cf8a16928841ad120c697455f28106a5cca4a09/lisp/progmodes/js.el#L3434-L3440

As grammars evolve more and more of these changes accumulate. Since tools may use grammars that are packaged by a linux distro, there is a need to support older grammar versions.

It would be nice if there was somehow a way to detect what "version" of a grammar is loaded without executing trial-by-error queries to narrow down what version of the grammar we might be working with.

Emacs has the ability to help the user download grammars using a specific git tag, but nothing compels grammar authors to even use git tags with version numbers in them. Some kind of version number baked into the generated C code I believe would be a nice solution. Even if it's just strictly increasing number every time the grammar is generated would go a long way. This would allow us to call into the dynamic library and see if it is one of the grammar versions we tested with and know works.

I am very interested in other solutions to this. How do other tools deal with these breaking changes? Is it a matter of always downloading and compiling the exact grammar version needed?

amaanq:
- Closes #2611

### Problem

When updating tree-sitter grammars, downstream consumers have no idea of knowing what version of the parser they've compiled actually is. This is problematic, because consumers cannot automatically figure out if a parser is supposedly compatible with a prior version, meaning that users don't know if the queries targeting that parser version is compatible with the future version. This makes the process of updating grammar version for consumers very much trial-and-error (and akin to whack-a-mole). Ideally, parsers should embed their version, so that the tree-sitter API can be used to get this data.

### Solution

In this PR, I've added a `semantic_version` field to `TSLanguage`, and added a `ts_language_semantic_version` function which gets the version of the parser. This version is the semver encoded in a 32-bit integer, with the patch, minor, and major version occupying the lowest 3 bytes, in the respective order. In the Rust side, `Language::semantic_version` will return a tuple of 3 `u8`s.

We're keeping the encoded version quite simple, we don't want to be bogged down by supporting additional metadata such as pre-release info, and others, we're *only* supporting encoding a major, minor, and patch version. This requires the use of the `tree-sitter.json` manifest file, and having its version be kept updated by the author, which is easily manageable in 0.25 with `tree-sitter version`.

Because having two functions like `ts_language_version` and `ts_language_semantic_version` would be a bit confusing as to what the former is, I've gone ahead and deprecated it in favor of `ts_language_abi_version`, which clearly lets users know this will get the **ABI** version of the parser.

### TODO

- [ ] Document how to properly update a grammar with `tree-sitter version` & adhere to semver, so that grammar authors have some guidelines.

> [!NOTE]
> In our tests, I've changed them to use an internal wrapper around `tree_sitter_generate::generate_parser_for_grammar`, because that function now takes in a version tuple, which our tests do not need to care about.


lib/include/tree_sitter/api.h
```diff
@@ -1255,6 +1257,35 @@ TSSymbolType ts_language_symbol_type(const TSLanguage *self, TSSymbol symbol);
...
+ * Layout of the returned integer:
+ *
+ * MSB                               LSB
+ * +--------+--------+--------+--------+
+ * |00000000|  Major |  Minor |  Patch |
+ * +--------+--------+--------+--------+
+ * 31      24 23    16 15     8 7      0
+ *
+ */
+uint32_t ts_language_semantic_version(const TSLanguage *self);
```
ObserverOfTime, line 1287: Should we use `uint64_t` to support calendar versioning (`major > UINT8_MAX`)?

amaanq: We could, but then we can't really validate the version at generate time. I guess we should just store a string then, and let downstream figure it out.

clason: I'm sorry, but when was calver ever discussed? You set the API, you can mandate semver.

amaanq: Yeah, I, personally, am really for semver.

amaanq:
Ok, there's 3 out of 356 grammars with no tree-sitter.json file used in nvim-treesitter that must be generated from the grammar. The config file is sort of mandatory for this feature, as we use the version defined in *this* file. That means <1% of grammars are impacted, and I've gone ahead and put up PRs to migrate them 2 days ago now - if they don't bother to update, then that's on them, and they *cannot* use ABI 15 then (but nothing is broken with ABI 14).

clason:
Can you list them here so I know what will happen when users start being upgraded to ts 0.25 by their distro?

WillLillis:
> Can you list them here so I know what will happen when users start being upgraded to ts 0.25 by their distro?

Looks like it's down to 2 now, since the [pr](https://github.com/rockorager/tree-sitter-scfg/pull/1) to [scfg](https://github.com/rockorager/tree-sitter-scfg) got merged. The remaining are [mlir](https://github.com/artagnon/tree-sitter-mlir) and [latex](https://github.com/latex-lsp/tree-sitter-latex/pull/192).

clason:
LaTeX devs are pretty active (though this is not their main project) so I expect that to be fixed in time. (It is a widely used parser, though, so I would appreciate giving them the chance before pulling the rug under them.)

Be aware that once you drop ABI 13 compat _from the lib_, a lot more parsers will have to be generated from grammar so the number will skyrocket. Your project is now at the point where you need to take backward compatibility seriously if you are interested in a working ecosystem.

clason:
> > Can you list them here so I know what will happen when users start being upgraded to ts 0.25 by their distro?
> 
> Looks like it's down to 2 now, since the [pr](https://github.com/rockorager/tree-sitter-scfg/pull/1) to [scfg](https://github.com/rockorager/tree-sitter-scfg) got merged. The remaining are [mlir](https://github.com/artagnon/tree-sitter-mlir) and [latex](https://github.com/latex-lsp/tree-sitter-latex/pull/192).

Actually, three: https://github.com/314eter/tree-sitter-ocamllex

clason:
Also, this needs to be implemented for wasm parsers as well.

clason:
> Be aware that once you drop ABI 13 compat from the lib, a lot more parsers will have to be generated from grammar so the number will skyrocket. Your project is now at the point where you need to take backward compatibility seriously if you are interested in a working ecosystem.

To reiterate, because that is a serious issue even now: Once I bump tree-sitter to 0.25 in Neovim, your `generate` workflow will light up like Christmas lights (i.e., fail _hard_). Instead of hard erroring out if there's no `treesitter.json`, why not simply lock the ABI to 14 and proceed with generating from that? This way parsers can opt-in to the goodies, and old parsers remain working while you still support ABI 14.

PR #4138, opened 2025-01-22
chore: a few more minor lints

nyurik:
apply a few pedantic lints that seemed relevant

### DIFF
--- a/cli/generate/src/render.rs
+++ b/cli/generate/src/render.rs
@@ -873,7 +873,7 @@ impl Generator {
                 && chars.ranges().all(|r| {
                     let start = *r.start() as u32;
                     let end = *r.end() as u32;
-                    end <= start + 1 && end <= u16::MAX as u32
+                    end <= start + 1 && u16::try_from(end).is_ok()
                 })
             {
                 leading_simple_transition_count += 1;
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -2203,7 +2203,7 @@ impl<'cursor> TreeCursor<'cursor> {
     pub fn goto_first_child_for_byte(&mut self, index: usize) -> Option<usize> {
         let result =
             unsafe { ffi::ts_tree_cursor_goto_first_child_for_byte(&mut self.0, index as u32) };
-        (result >= 0).then_some(result as usize)
+        result.try_into().ok()
     }
 
     /// Move this cursor to the first child of its current node that contains or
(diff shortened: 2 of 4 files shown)


cli/generate/src/rules.rs
```diff
@@ -441,8 +441,8 @@ impl TokenSet {
     }
 
     pub fn len(&self) -> usize {
-        self.eof as usize
-            + self.end_of_nonterminal_extra as usize
+        usize::from(self.eof)
+            + usize::from(self.end_of_nonterminal_extra)
```
amaanq, line 445: not a fan of these, `as _` reads a bit nicer

cli/src/highlight.rs
```diff
@@ -325,9 +325,9 @@ fn closest_xterm_color(red: u8, green: u8, blue: u8) -> Color {
     // Get the xterm color with the minimum Euclidean distance to the target color
     // i.e.  distance = √ (r2 - r1)² + (g2 - g1)² + (b2 - b1)²
     let distances = colors.map(|(color_id, (r, g, b))| {
-        let r_delta = (max(r, red) - min(r, red)) as u32;
-        let g_delta = (max(g, green) - min(g, green)) as u32;
-        let b_delta = (max(b, blue) - min(b, blue)) as u32;
+        let r_delta = u32::from(max(r, red) - min(r, red));
+        let g_delta = u32::from(max(g, green) - min(g, green));
+        let b_delta = u32::from(max(b, blue) - min(b, blue));
```
amaanq, line 330: same as above, not a fan

cli/src/parse.rs
```diff
@@ -350,7 +350,7 @@ pub fn parse_file_at_path(
             }
         }
 
-        if opts.timeout > 0 && start_time.elapsed().as_micros() > opts.timeout as u128 {
+        if opts.timeout > 0 && start_time.elapsed().as_micros() > u128::from(opts.timeout) {
```
amaanq, line 353: & here

cli/src/tests/parser_test.rs
```diff
@@ -1786,7 +1786,7 @@ fn test_decode_cp1252() {
         fn decode(bytes: &[u8]) -> (i32, u32) {
             if !bytes.is_empty() {
                 let byte = bytes[0];
-                (byte as i32, 1)
+                (i32::from(byte), 1)
```
amaanq, line 1789: and this

cli/src/tests/parser_test.rs
```diff
@@ -1822,7 +1822,7 @@ fn test_decode_macintosh() {
...
```
amaanq, line 1825: and this

amaanq (commented):
looks good, except I don't really like the utype::from changes, it's not giving us any benefit besides being more explicit about the conversion, but I find `as _` nicer

cli/generate/src/rules.rs
nyurik: sure, will revert, but do note that here the order of precedence starts to play a role, i.e. `as` vs `+` -- so parens might be more explicit. `T::from(...)` imo is more explicit and simpler in that sense, but you are the maintainer :)

amaanq: It is more explicit, but I find the cognitive load of reading it a bit higher, I take an extra second to look at it, vs an `as _` being, to me, "oh yeah we're just casting a number type"

PR #4141, opened 2025-01-22
Misc web-ts improvements

amaanq:
- Followup to #4121


lib/binding_web/src/query.ts
```diff
@@ -198,19 +211,322 @@ export class Query {
...
+   * Create a new query from a string containing one or more S-expression
+   * patterns.
+   *
+   * The query is associated with a particular language, and can only be run
+   * on syntax nodes parsed with that language. References to Queries can be
+   * shared between multiple threads.
+   *
+   * @link {@see https://tree-sitter.github.io/tree-sitter/using-parsers/queries}
+   */
+  constructor(language: Language, source: string) {
```
savetheclocktower, line 224: Moving all this to `query.ts` is a good change, and I hope we can iteratively improve on this code in the future by _incrementally_ breaking all the work in this constructor into smaller, more digestible bits. It doesn't have to happen in this PR.

lib/binding_web/src/query.ts
```diff
@@ -198,19 +211,322 @@ export class Query {
...
+        address,
+        i,
+        TRANSFER_BUFFER
+      );
+      const stepCount = C.getValue(TRANSFER_BUFFER, 'i32');
+
+      predicates[i] = [];
+      textPredicates[i] = [];
+
+      const steps = new Array<PredicateStep>();
```
savetheclocktower, line 325:
```suggestion
      const steps: PredicateStep[] = [];
```

A bit more idiomatic.

lib/binding_web/src/query.ts
ObserverOfTime: You also don't _have_ to deprecate `Language.query`.

lib/binding_web/src/query.ts
```diff
@@ -198,19 +211,322 @@ export class Query {
...
+      const isStringStep = (step: PredicateStep): step is { type: 'string', value: string } => {
+        return step.type === 'string';
+      }
```
savetheclocktower, line 328: Two non-blocking nitpicks.

First, when a function like this needs no access to any variables in its lexical scope, I tend to define it near the top of the file outside a closure:

```ts
function isStringStep(step: PredicateStep): step is ({ type: 'string', value: string }) {
  return step.type === 'string';
}
```

In a modern JS engine I doubt this makes much of a difference, but tooling also tends to be better at locating functions when they're defined with `function foo` syntax rather than `const foo = function` syntax.

Second: it might also be worth defining named types like `StringPredicateStep` and `CapturePredicateStep`, but that's a much smaller quibble.

lib/binding_web/src/query.ts
```diff
@@ -198,19 +211,322 @@ export class Query {
...
+      );
+      const nameLength = C.getValue(TRANSFER_BUFFER, 'i32');
+      stringValues[i] = C.UTF8ToString(valueAddress, nameLength);
+    }
+
+    const setProperties = new Array<QueryProperties>(patternCount);
+    const assertedProperties = new Array<QueryProperties>(patternCount);
+    const refutedProperties = new Array<QueryProperties>(patternCount);
+    const predicates = new Array<QueryPredicate[]>(patternCount);
+    const textPredicates = new Array<TextPredicate[]>(patternCount);
```
savetheclocktower, line 312: Same feedback for these as the other one.

In general, `new Array()` in JS was almost extinct because it offered zero benefits over `[]`. TypeScript makes bare `let foo = []` an anti-pattern, but you can annotate the variable instead to cut down on typing without harming readability.

Even in a type annotation, the only time I'd prefer `Array<T>` over `T[]` would be when `T` is an ad-hoc type — e.g., `Array<{ type: 'capture', name: string }>` — because then it actually probably _is_ more readable than `({ type: 'capture', name: string })[]` (which leads the user down a twisty garden path and then throws a `[]` in at the end in a Shyamalan-esque twist).

Even `QueryPredicate[][]` is a bit more up-front about the two-dimensional nature of the list than `Array<QueryPredicate[]>` (which mixes two styles of array notation).

(Despite all the first-person pronouns above, I'm trying to give feedback based on the common observed TypeScript style I see in the wild, and am not just trying to encode my personal preferences.)

savetheclocktower (commented):
A few suggestions. I think they're all in code that you didn't even touch in this PR (except to move it from one location to another), so I apologize in advance for tossing nitpicks at code that was just approved in a previous PR. None of it is crucial.

In general, I think the code in `Query#constructor` should be annotated with more comments, but that's a contribution I'd be happy to make myself.

Looks like the tests aren't updated yet, so when that happens, please remember to check for the presence of both `pattern` and `patternIndex` on `QueryMatch` so nothing breaks.

lib/binding_web/src/query.ts
amaanq: > You also don't have to deprecate Language.query.

Why? It doesn't make sense to have two different ways to do the same thing, coming from two different classes. No other bindings do this

lib/binding_web/src/query.ts
```diff
@@ -134,6 +135,18 @@ export interface QueryMatch {
...
+export const CaptureQuantifier = {
+  Zero: 0,
+  ZeroOrOne: 1,
+  ZeroOrMore: 2,
+  One: 3,
+  OneOrMore: 4
+} as const;
+
+/** A quantifier for captures */
+export type CaptureQuantifier = typeof CaptureQuantifier[keyof typeof CaptureQuantifier];
```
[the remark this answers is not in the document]
amaanq: Enums in TS suck.

lib/binding_web/src/query.ts
```diff
@@ -198,19 +211,322 @@ export class Query {
...
+              predicates[i].push({ operator, operands: steps.slice(1) });
+          }
+
+          steps.length = 0;
+        }
+      }
+
+      Object.freeze(setProperties[i]);
+      Object.freeze(assertedProperties[i]);
+      Object.freeze(refutedProperties[i]);
```
savetheclocktower, line 524: Another example of feedback that should've gone in the last PR, but would it make sense to freeze `predicates` here as well? I don't think it needs to remain open to later mutation.

WillLillis (approved):

lib/binding_web/src/query.ts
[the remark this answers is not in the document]
ObserverOfTime: TS sucks.

PR #4142, opened 2025-01-23
feat: improve the init command

### LINKED ISSUE #3960: Fix/improve init command
### Problem

Issues that should be fixed and backported to 0.24:

- [ ] The grammar name is in `snake_case` but some files expect `kebab-case`
  * `Makefile`
  * `CMakeLists.txt`
  * `bindings/rust/build.rs`
  * `bindings/node/index.js`
  * `bindings/c/*.h` (filename)
  * `bindings/c/*.pc.in` (contents+filename)
- [ ] The `scope` should allow dashes (`[\\w-]`)
- [ ] The default `file-types` should not start with a dot
- [ ] Comma-separated `file-types` do _not_ get separated

### Expected behavior

Enhancements that can be made for 0.25:
- [ ] Add a `verbose` name for the `description` and error messages
- [ ] Take `file-types` into account for `injection-regex`
- [ ] Add a funding link for `package.json` & `pyproject.toml`
  * `"funding": "{}",`
  * `"Funding" = "{}"`
- [ ] Don't write the default `bindings` object
- [ ] Allow specifying a custom `class-name`
```json
"class-name": {
  "type": "string",
  "pattern": "^TreeSitter\\w+$",
  "description": "The class name for the Swift, Java & Kotlin bindings"
}
```

amaanq:
- Closes #3960

### Problem

The init command is lacking some features that could be beneficial to users.

### Solution

Changes:
- allow dashes in scopes (only changes the schema)
- don't prepend default filetype with a period, as that's incorrect
- add an optional funding field
- add a class-name field for high level languages

What I didn't address in the linked issue for reasons explained below:

-  Don't write the default bindings object
    I don't get the point of this, it's a handful of extra lines and clearly lets users know they can configure bindings, without having to dig into the init command's docs
    
- Take file types into account in injection-regex
   Not a fan of this, because we'd have to do some "regex building" of multiple alternations - plus, the language name lowercased should suffice for a lot of these cases.
       
- Add a verbose `name` for the `description` and error messages
   I don't even get this, especially the error messages part
   
- Comma-separated file-types do not get separated
   We had agreed on using space as the separator, using two separators makes no sense. 
   
- The grammar name is in snake_case but some files expect kebab-case
   I don't know of a language that has this problem (uses underscores in its name).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
           # When adding a new `target`:
           # 1. Define a new platform alias above
           # 2. Add a new record to the matrix map in `cli/npm/install.js`
-          - { platform: linux-arm64       , target: aarch64-unknown-linux-gnu   , os: ubuntu-24.04-arm , features: wasm  }
+          - { platform: linux-arm64       , target: aarch64-unknown-linux-gnu   , os: ubuntu-latest    , use-cross: true }
           - { platform: linux-arm         , target: arm-unknown-linux-gnueabi   , os: ubuntu-latest    , use-cross: true }
           - { platform: linux-x64         , target: x86_64-unknown-linux-gnu    , os: ubuntu-latest    , features: wasm  }
           - { platform: linux-x86         , target: i686-unknown-linux-gnu      , os: ubuntu-latest    , use-cross: true }
--- a/cli/loader/src/lib.rs
+++ b/cli/loader/src/lib.rs
@@ -217,6 +217,8 @@ pub struct Author {
 pub struct Links {
     pub repository: Url,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub funding: Option<Url>,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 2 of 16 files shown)


WillLillis (approved):

ObserverOfTime (changes requested):
> I don't even get this, especially the error messages part

Error messages such as these:
https://github.com/tree-sitter/tree-sitter/blob/8e5af306a0b4c2f0bd6f6e1719d027cb0f420370/cli/src/templates/lib.rs#L51

> I don't know of a language that has this problem (uses underscores in its name).

There's `ssh_config` for one.

amaanq:
> > I don't even get this, especially the error messages part
> 
> Error messages such as these:
> 
> https://github.com/tree-sitter/tree-sitter/blob/8e5af306a0b4c2f0bd6f6e1719d027cb0f420370/cli/src/templates/lib.rs#L51

Ah, what do we call this in the JSON then? `title`?

ObserverOfTime:
Sure.

ObserverOfTime (approved):

PR #4144, opened 2025-01-24
feat(bindings): add opt-in zig bindings

ObserverOfTime:


amaanq (approved):

cli/src/init.rs
```diff
@@ -283,18 +286,48 @@ pub fn generate_grammar_files(
...
+            if !contents.contains("Zig artifacts") {
+                contents = contents.replace(
+                    "*.lib",
+                    indoc! {"
+                    *.lib
+
+                    # Zig artifacts
+                    .zig-cache/
+                    zig-cache/
+                    zig-out/
```
amaanq, line 304: Every grammar doesn't have *.lib, as this was added very recently, so this replacement does nothing.

amaanq (changes requested):

amaanq (approved):

PR #4152, opened 2025-01-26
fix(cli): correct traversal behavior for finding error nodes

### LINKED ISSUE #4049: `tree-sitter parse` doesn't report an error even when ERROR node is present
### Problem

Running `tree-sitter parse` on some files will produce a syntax tree with `ERROR` nodes, but the parse command finishes with a zero exit code and no error report.

Contents of `test.cr`:
```crystal
def filename=(filename)
end
end
```

```
$ tree-sitter parse test.cr
(expressions [0, 0] - [4, 0]
  (method_def [0, 0] - [1, 3]
    name: (identifier [0, 4] - [0, 12])
    body: (expressions [0, 12] - [1, 0]
      (const_assign [0, 12] - [0, 20]
        lhs: (constant [0, 12] - [0, 12])
        rhs: (expressions [0, 13] - [0, 20]
          (identifier [0, 14] - [0, 19])))))
  (ERROR [2, 0] - [4, 0]))
$ echo $?
0
```

### Steps to reproduce

```sh
git clone https://github.com/crystal-lang-tools/tree-sitter-crystal
cd tree-sitter-crystal
git checkout a07995f

echo "def filename=(filename)
end
end" > test.cr

tree-sitter parse test.cr
```

### Expected behavior

The `parse` command should have a non-zero exit code and show a error report like
```
test.cr         0.17 ms         186 bytes/ms (ERROR [2, 0] - [4, 0])
```

tree-sitter 0.24.5 (1a983b7e2c45d9b9f45b2133a4e3a829f7e961be)

### Operating system/version

Linux (Debian 12)

amaanq:
- Closes #4049

### Problem

Tree-sitter's `parse` subcommand will walk through the tree and try to find the first error, as it'll report this specific node and its ranges to the user and exit with a non-zero error code then. However, the implementation does not account for the case where the *first* error is a hidden node, but subsequent error nodes do exist **and** are visible, as observed in the linked issue.

### Solution

If we *don't* find an error node while being inside a node that has an error node, we walk back up to the child of the very first node that has an error. Then, we check this node's siblings and see if *those* nodes also contain an error that is visible.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -634,14 +634,39 @@ pub fn parse_file_at_path(
         }
 
         let mut first_error = None;
-        loop {
+        let mut earliest_node_with_error = None;
+        'outer: loop {
             let node = cursor.node();
             if node.has_error() {
+                if earliest_node_with_error.is_none() {
+                    earliest_node_with_error = Some(node);
+                }
                 if node.is_error() || node.is_missing() {
                     first_error = Some(node);
                     break;
                 }
+
+                // If there's no more children, even though some outer node has an error,
+                // then that means that the first error is hidden, but the later error could be
+                // visible. So, we walk back up to the child of the first node with an error,
+                // and then check its siblings for errors.
                 if !cursor.goto_first_child() {
+                    let earliest = earliest_node_with_error.unwrap();
+                    while cursor.goto_parent() {
+                        if cursor.node().parent().is_some_and(|p| p == earliest) {
+                            while cursor.goto_next_sibling() {
+                                let sibling = cursor.node();
+                                if sibling.is_error() || sibling.is_missing() {
+                                    first_error = Some(sibling);
+                                    break 'outer;
+                                }
+                                if sibling.has_error() && cursor.goto_first_child() {
+                                    continue 'outer;
+                                }
+                            }
+                            break;
+                        }
+                    }
                     break;
                 }
             } else if !cursor.goto_next_sibling() {


cli/src/parse.rs
```diff
@@ -634,14 +634,41 @@ pub fn parse_file_at_path(
...
                 }
+
+                // If there's no more children, even though some outer node has an error,
+                // then that means that the first error is hidden, but the later error could be
+                // visible. So, we walk back up to the child of the first node with an error,
+                // and then check its siblings for errors.
                 if !cursor.goto_first_child() {
+                    let earliest = earliest_node_with_error.unwrap();
+                    while cursor.goto_parent() {
+                        if let Some(parent) = cursor.node().parent() {
```
WillLillis, line 656: Since `cursor.node.parent()` and `earliest_node_with_error` are both optionals, would it be more straightforward to replace the `unwrap` and `if let` with a direct comparison? 

```rust
if !cursor.goto_first_child() {
    while cursor.goto_parent() {
        if cursor.node().parent() == earliest_node_with_error {
            while cursor.goto_next_sibling() {
                let sibling = cursor.node();
                if sibling.is_error() || sibling.is_missing() {
                    first_error = Some(sibling);
                    break 'outer;
                }
                if sibling.has_error() && cursor.goto_first_child() {
...
```

amaanq: Hm, in theory yes, but the `earliest_node_with_error` *should* not be none, so unwrapping here sorta guarantees that, and I'd want it to panic then. I *can* clean this up though, using `is_some_and`

WillLillis (approved):

PR #4153, opened 2025-01-26
feat(ci): build libraries on windows x64 with mingw

### LINKED ISSUE #2034: Check in CI that everything compiles with Clang + MinGW on Windows
Implement after #1889

WillLillis:
This runs the `make` and `cmake` build processes on Windows-x64 via mingw in CI.

Closes #2034

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,6 +141,62 @@ jobs:
 
     # TODO: Remove RUSTFLAGS="--cap-lints allow" once we use a wasmtime release that addresses
     # the `mismatched-lifetime-syntaxes` lint
+    - name: Build wasmtime library (Windows x64 MSYS2)
+      if: ${{ !matrix.use-cross && contains(matrix.features, 'wasm') && matrix.platform == 'windows-x64' }}
+      run: |
+        mkdir -p target
+        WASMTIME_VERSION=$(cargo metadata --format-version=1 --locked --features wasm | \
+                           jq -r '.packages[] | select(.name == "wasmtime-c-api-impl") | .version')
+        curl -LSs "$WASMTIME_REPO/archive/refs/tags/v${WASMTIME_VERSION}.tar.gz" | tar xzf - -C target
+        cd target/wasmtime-${WASMTIME_VERSION}
+        cmake -S crates/c-api -B target/c-api \
+          -DCMAKE_INSTALL_PREFIX="$PWD/artifacts" \
+          -DWASMTIME_DISABLE_ALL_FEATURES=ON \
+          -DWASMTIME_FEATURE_CRANELIFT=ON \
+          -DWASMTIME_TARGET='x86_64-pc-windows-gnu'
+        cmake --build target/c-api && cmake --install target/c-api
+        printf 'CMAKE_PREFIX_PATH=%s\n' "$PWD/artifacts" >> $GITHUB_ENV
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -157,12 +157,32 @@ jobs:
...
       if: ${{ runner.os != 'Windows' }}
       run: make.sh -j CFLAGS="$CFLAGS"
       env:
         CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
 
+    - name: Build C library (Windows x64 make)
+      if: ${{ matrix.platform == 'windows-x64' }}
+      shell: msys2 {0}
+      run: |
+        export PATH=/c/mingw64/bin:$PATH
```
WillLillis, line 180: It's not clear to me why I need to add mingw's stuff to the path here but not for the cmake job.

Makefile
```diff
@@ -1,6 +1,8 @@
 ifeq ($(OS),Windows_NT)
+ifneq ($(shell uname -o),Msys)
```
ObserverOfTime, line 2: Let's not start adding exceptions here. We already have a working cmake build.

WillLillis: In that case, should I remove the `Build C library (Windows x64 make)` job from `build.yml`?

ObserverOfTime: Yes.

amaanq (approved):

WillLillis:
There's still an issue with this and I haven't been able to fix it yet. Even though clang is specified in the build command, Microsoft's build tools are still being used. I've spent some time working on this in a [pull request](https://github.com/WillLillis/tree-sitter/pull/3) in my fork to avoid spamming people with notifications, but I haven't been able to fully fix things yet.

WillLillis:
Gentle ping @dundargoc any chance you could take a look at this?

clason:
Drive-by suggestion: we can now use native runners for linux-arm64 in place of cross compiling.

WillLillis:
> Drive-by suggestion: we can now use native runners for linux-arm64 in place of cross compiling.

This is something I'd like to look into. IIRC though, the last time amaanq looked into those runners there were a lot of issues. Maybe they're in a better state now?

clason:
Yes, the runners had issues initially (segfaulting npm) but those seem to have been fixed for a while.

ObserverOfTime:
I switched to both the Linux and the new Windows arm runner in py-tree-sitter. They work well.

lib/CMakeLists.txt
```diff
@@ -33,7 +33,8 @@ if(MSVC)
 else()
   target_compile_options(tree-sitter PRIVATE
                          -Wall -Wextra -Wshadow -Wpedantic
-                         -Werror=incompatible-pointer-types)
+                         -Werror=incompatible-pointer-types
+                         -Wno-static-in-inline)
```
ObserverOfTime: Is this something that can be fixed instead?

WillLillis: I'm not sure, I'm assuming Max made it that way for a reason. As a reference point:

```sh
[lillis@cachyos-x8664] ~/projects/tree-sitter (master)
❯ rg "static inline" | wc -l
176
```

WillLillis: I must have been seeing this warning locally before I fixed another issue in this PR. Looks like it's fine to take it out.

PR #4164, opened 2025-01-28
fix(cli): propagate error flag in parse summary

### LINKED ISSUE #4163: `tree-sitter parse` exit code is 0 even for parse errors
### Problem

When I run `tree-sitter parse` and it generates a `ERROR` or `MISSING` node, the exit code (`$?` in the shell) is zero. 

I rely on the exit code from `tree-sitter parse` for testing my grammar across a large codebase. This behavior change means tests that should fail are now passing.

I ran `git bisect` and found https://github.com/tree-sitter/tree-sitter/pull/4048 as the point when the exit code behavior changed.

### Steps to reproduce

Build tree-sitter from the `release-0.25` branch
Run `tree-sitter parse` on any file that generates an `ERROR` node
Observe that the exit code is 0

### Expected behavior

`tree-sitter parse` has a non-zero exit code for parse errors

tree-sitter 0.25.0 (9515be4fc16d3dfe6fba5ae4a7a058f32bcf535d)

### Operating system/version

Arch Linux

keidax:
I believe this is sufficient to close #4163. 

I'd like to add tests for this behavior, but I'm not familiar enough with Rust to know how :neutral_face:

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -719,7 +719,7 @@ pub fn parse_file_at_path(
 
         opts.stats.parse_summaries.push(ParseSummary {
             file: path.to_path_buf(),
-            successful: true,
+            successful: first_error.is_none(),
             start: Some(tree.root_node().start_position().into()),
             end: Some(tree.root_node().end_position().into()),
             duration: Some(parse_duration),


WillLillis (approved):

WillLillis:
Thanks for catching this!

sogaiu:
A bit late, but trying with:

```
tree-sitter 0.25.0 (304e56850ffb5fc6359b9288ace93936cb45aac3)
```

(a later commit admittedly), seems to be working as advertised :+1: 

Thanks!

The exit code behavior is important for my testing too.

Issue #1520, opened 2021-12-03
License files are missing from the published Rust crates


alebastr:
All the crates from this project are published on crates.io without the LICENSE file. As it was already noted in #979, the MIT license requires that the code is distributed with the full license text.

Please, consider adding the license files with the next update.

PR #4167, opened 2025-01-30
Include LICENSE file

michel-slm:
This is needed by the MIT license terms

Resolves: #1520

To test, from each crate directory:

```
❯ cargo package --list | grep LICENSE
LICENSE
```


clason:
Do not add hyperlinks; this will break things on Windows.

clason:
Ideally, this (copying the LICENSE to the crate) is handled by the publish workflow, not baked into the monorepo.

ObserverOfTime:
That PR is different.

MalikNadirawan:
Subject: Re: [tree-sitter/tree-sitter] Include LICENSE file (PR #4167)

Dear ObserverOfTime,

Could you please elaborate on what makes this PR different? 🤔  

Best regards,
***@***.*** 



-------- Original Email --------

clason:
That solution is still not acceptable. Simply place copies if necessary, it's not like these can diverge.

PR #4168, opened 2025-01-30
chore(bindings): include license file in crate

ObserverOfTime:
Because cargo can't behave like every sane package manager and include license files by default.


amaanq:
Let's not merge PRs with failing CI.

PR #4176, opened 2025-02-01
fix(lib): prevent finished_tree assertion failure

ribru17:
**Problem:** When resetting the parser during subtree balancing, an error is thrown:

```
parser.c:2198: ts_parser_parse: Assertion `self->finished_tree.ptr' failed.
```

**Solution:** Reset `canceled_balancing` to false in `ts_parser_reset()`.

### DIFF
--- a/cli/src/tests/parser_test.rs
+++ b/cli/src/tests/parser_test.rs
@@ -1004,6 +1004,31 @@ fn test_parsing_with_timeout_during_balancing() {
         assert!(tree.is_none());
         assert!(in_balancing);
 
+        // This should not cause an assertion failure.
+        parser.reset();
+        let tree = parser.parse_with_options(
+            &mut |offset, _| {
+                if offset >= code.len() {
+                    &[]
+                } else {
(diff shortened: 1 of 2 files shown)


ribru17:
Of course the error shows up right after release :melting_face:

clason:
That's ok, day-one patch releases are a tradition.

amaanq:
:facepalm: Thanks @ribru17

cli/src/tests/parser_test.rs
```diff
@@ -1004,6 +1004,31 @@ fn test_parsing_with_timeout_during_balancing() {
         assert!(tree.is_none());
         assert!(in_balancing);
 
+        // This should not throw an assertion.
```
amaanq, line 1007:
```suggestion
        // This should not cause an assertion failure.
```

amaanq (approved):

mattyboyy32:
[Uploading tree-sitter-0.25.1.zip…]()

Issue #4180, opened 2025-02-02
0.25.1: NetBSD: le16toh/be16toh undefined


0-wiz-0:
### Problem

tree-sitter 0.25.0 includes `src/portable/endian.h` to define some byteswap macros. These even include a special case for NetBSD, and do include the correct headers (`endian.h` -> `sys/endian.h`).
However, when a standard is defined, the [NetBSD headers](https://github.com/NetBSD/src/blob/trunk/sys/sys/endian.h) try to only provide the symbols provided by these standard.
The default compiler command lines includes `-D_POSIX_C_SOURCE=200112L`, but not the X/Open standard that provides `le16toh`/`be16toh`, leading to:
```
gcc -O2 -g -fstack-clash-protection -std=c11 -fPIC -fvisibility=hidden -D_POSIX_C_SOURCE=200112L -D_DEFAULT_SOURCE -Ilib/src -Ilib/src/wasm -Ilib/include -g  -c -o lib/src/lexer.o lib/src/lexer.c
n file included from lib/src/lexer.c:3:
lib/src/./unicode.h: In function 'ts_decode_utf16_le':
lib/src/./unicode.h:18:9: warning: implicit declaration of function 'le16toh' [-Wimplicit-function-declaration]
   18 |     (c)=le16toh((s)[(i)++]); \
      |         ^~~~~~~
lib/src/./unicode.h:57:3: note: in expansion of macro 'U16_NEXT_LE'
   57 |   U16_NEXT_LE(((uint16_t *)string), i, length, *code_point);
      |   ^~~~~~~~~~~
lib/src/./unicode.h: In function 'ts_decode_utf16_be':
lib/src/./unicode.h:29:9: warning: implicit declaration of function 'be16toh' [-Wimplicit-function-declaration]
   29 |     (c)=be16toh((s)[(i)++]); \
      |         ^~~~~~~
lib/src/./unicode.h:67:3: note: in expansion of macro 'U16_NEXT_BE'
   67 |   U16_NEXT_BE(((uint16_t *)string), i, length, *code_point);
      |   ^~~~~~~~~~~
```
and the symbol being used, but missing in the generated library.

### Steps to reproduce

Build tree-sitter-0.25.1 on NetBSD:
```
tar xzf tree-sitter-0.25.1
cd tree-sitter-0.25.1
gmake
```

### Expected behavior

Symbols are expanded from the header file.
Some possible solutions:
[...]

PR #4181, opened 2025-02-02
build: do not define _POSIX_C_SOURCE on NetBSD

0-wiz-0:
It leads to missing symbols, see #4180.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,5 @@
 ifeq ($(OS),Windows_NT)
 $(error Windows is not supported)
-else
-	detected_OS := $(shell uname -s)
 endif
 
 VERSION := 0.25.1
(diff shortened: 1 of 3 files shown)


clason:
Unfortunately, this is not the only build system in this repo... (far from it).

So modifying `endian.h` would be preferred.

ObserverOfTime:
Can we move _both_ defines to `endian.h`?

0-wiz-0:
> Can we move both defines to `endian.h`?

I'd really appreciate that since when I just add the define in `endian.h` it still fails when building `query.c` and I haven't tracked down the inclusion chain yet (where `endian.h` seems to be included some other way, perhaps via `time.h` or another system header).

amaanq:
Probably because by default, the Makefile compiles C file individually, so you'd have to define this in every source file that needs it. Would adding `#undef _POSIX_C_SOURCE` work for NetBSD?

0-wiz-0:
I pushed a different version that I think follows your suggestions. I think it's ugly, but it works.

amaanq:
Yeah, that's a little ugly but I guess it's fine. I'm a little confused though, because `lexer.c` also includes `endian.h` (through `unicode.h`) - does the library build just fine on NetBSD as this PR is? Also, if you could add a short comment above the NetBSD undef in `query.c` explaining why that's there, that'd be great

0-wiz-0:
I've added a comment about the `#undef`.
Yes, the git head of my fork builds without warnings on NetBSD.
`lexer.c` has fewer includes than `query.c`, I guess that's the reason.

amaanq (approved):

clason:
Might have wanted to squash that...

PR #4185, opened 2025-02-05
fix(web): provide type in the exports

robertohuertasm:
When using TypeScript projects using other module settings than CommonJs, the types were not correctly exposed, and the compilation failed.

This adds the types path to the exports so compilation works for `module: NodeNext` and other variants.

I've tested this with a test project using TypeScript and several `module` and `moduleResolution` combinations.


amaanq:
CI failure is unrelated (and fixed by #4192, thanks!

amaanq (approved):

PR #4191, opened 2025-02-07
Refine the documentation and command line help for "parse --edits"

danielpunkass:
Previously the format described the position as 'row, col' which would not work with a space between the comma and the col, and I didn't know about the option to pass an integer offset until I went scavenging through the project.


amaanq (approved):

amaanq:
thanks!

PR #4194, opened 2025-02-08
refactor(web): rename tree-sitter.js to web-tree-sitter.js

### LINKED ISSUE #4183: Incorrect sourceMappingURLs in web-tree-sitter
### Problem

It appears the sourceMappingURLs at the bottom of each web-tree-sitter file in releases is incorrect. None of them follow the same naming as the files present in the releases. For example, at the bottom of `web-tree-sitter.js`, the comment shows `//# sourceMappingURL=tree-sitter.js.map`.

Every file is missing "web-" at the front, and every debug file is also missing the "-debug" in the file name. This results in the source mappings not being loaded correctly unless the files are manually renamed or the url is fixed.

The missing "web-" prefix also applies to the wasm files in the releases as well. Treesitter attempts to load "tree-sitter.wasm" instead of "web-tree-sitter.wasm"

So it looks like when file names in the releases were updated, none of the source map fields were updated.

### Steps to reproduce

1. Download any of the web-tree-sitter js files in releases and its associated source map
2. Try to use the source map
3. See mismatch
```
Source map error: request failed with status 404
Resource URL: http://localhost:10003/web-tree-sitter-debug.js
Source Map URL: tree-sitter.js.map
```

### Expected behavior

The source mappings and loaded wasm match the file names in releases.

tree-sitter 0.24.7 (5e8760bf462ce7b19b3d2396d5b7860f3906a297)

### Operating system/version

`Linux arch 6.12.10-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 18 Jan 2025 02:26:57 +0000 x86_64 GNU/Linux`

amaanq:
- Closes #4183

### Problem

The sourcemapping url in the js files points to `tree-sitter.js`, but we rename these files in the GH release to `web-tree-sitter.js` and `web-tree-sitter-debug.js`. This breaks the sourcemap, which is unhelpful to users trying to debug the library.

### Solution

Instead, we rename the output file in emscripten & esbuild to `web-tree-sitter.(js|wasm)`, so the sourcemapping url will always be correct, and this doesn't require us to move the file. This is not breaking for consumers of the web bindings since we update the exports in `package.json`, nor the playground as both filenames will be kept in the .github.io repo.

In the release job, we don't rename the files and instead throw the web files into their own tarball, including the lib files (for debugging the wasm module with its sourcemap).

For verification, a sample release with the web-tree-sitter tarball can be found [here](https://github.com/amaanq/tree-sitter/releases/tag/v0.25.2), and each sourcemap's source files has the correct relative file path.


.github/workflows/build.yml
```diff
@@ -223,11 +223,6 @@ jobs:
       if: ${{ !matrix.no-run && !matrix.use-cross && inputs.run-test }}
       run: $BUILD_CMD run -p xtask -- test-wasm
 
-    - name: Run benchmarks
```
WillLillis, line 226: Why remove this? Is it just not used anymore?

WillLillis (approved):

PR #4195, opened 2025-02-09
fix: only attempt to parse supertypes if ABI version is new enough

Liamolucko:
Previously, when parsing supertypes out of a Wasm module, it was assumed that the number of them would always be correct. This wasn't the case for Wasm modules with an ABI version which doesn't include supertypes, causing random data to be interpreted as the number of supertypes and causing an out-of-bounds memory access.

This fixes that by gating the parsing of supertypes behind the ABI version being new enough.

To reproduce the issue:

    rm -rf test/fixtures/grammars/*
    cargo xtask fetch-fixtures
    # Do NOT run cargo xtask generate-fixtures, since that will update
    # their ABI version.
    cargo xtask generate-fixtures --wasm
    cargo xtask test --wasm

It seems a bit concerning to me that untrusted data from Wasm modules is being used for the lengths of copies in the first place, though: someone could still easily replace the number of supertypes with an absurdly high number and cause this crash.

Related: https://github.com/zed-industries/zed/pull/24492

### DIFF
--- a/lib/src/wasm_store.c
+++ b/lib/src/wasm_store.c
@@ -1268,7 +1268,6 @@ const TSLanguage *ts_wasm_store_load_language(
     .large_state_count = wasm_language.large_state_count,
     .production_id_count = wasm_language.production_id_count,
     .field_count = wasm_language.field_count,
-    .supertype_count = wasm_language.supertype_count,
     .max_alias_sequence_length = wasm_language.max_alias_sequence_length,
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Sorry, I didn't see this PR when I opened https://github.com/tree-sitter/tree-sitter/pull/4210. This has now been fixed.

PR #4201, opened 2025-02-11
feat: Windows support using MinGW-w64

MisterDA:
Building tree-sitter on Windows within a Cygwin or MSYS2 environment for MinGW-w64 targets should work. As there's no configure-like step, the tools have to be specified with:

```sh
make CC=x86_64-w64-mingw32-gcc \
     AR=x86_64-w64-mingw32-ar \
     STRIP=x86_64-w64-mingw32-strip
```

Exported symbols should be tagged.
Useful reference: [How to build and use DLLs on Windows](https://nullprogram.com/blog/2021/05/31/).

This PR doesn't include support to build libtree-sitter with MSVC or clang-cl, but generates a `libtree-sitter.lib` file for MSVC/clang-cl consumers.

Cf #3418 #3417
cc @ObserverOfTime

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -163,6 +163,14 @@ jobs:
       env:
         CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
 
+    - name: Build C library on Windows (make)
+      if: ${{ runner.os == 'Windows' }}
+      run: make.sh -j CC="$CC" CFLAGS="$CFLAGS" AR="$AR"
+      env:
+        CC: x86_64-w64-mingw32-gcc
+        AR: x86_64-w64-mingw32-ar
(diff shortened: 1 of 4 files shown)


ObserverOfTime:
But it does not produce a .dll

MisterDA:
> This is a bit too much.

Ok, I've removed the `dllexport`/`dllimport` machinery. This will result in the DLL exporting non-static functions visible between translation units, and the compiler generating slightly less efficient code for calling exported functions from the DLL; but now the PR only adds support for building with MinGW-w64 to the build system and doesn't change the C code.

MisterDA:
The CI timed out but should succeed. Is the PR more acceptable now?

maxbrunsfeld (approved):

maxbrunsfeld:
Thanks @MisterDA, looks good. And thanks for the reviews @ObserverOfTime.

PR #4203, opened 2025-02-13
`Makefile` fixes (general and for MinGW-w64)

MisterDA:
Hello all,

This PR is a continuation of my #4201 from 2 days ago. I've collected a few fixes that complete my first patch, I apologize for missing them. They stem from using the MinGW-w64 build of tree-sitter in a bigger project, rather than on its own.

The first three patches fix the Makefile to build with MinGW-w64. They also add install rules for the DLLs.
I've removed the CI for MinGW-w64: it turns out that only the C compiler is available, without the binutils (so `x86_64-w64-mingw32-ar` and `x86_64-w64-mingw32-strip` are missing). The CI script would also always use MSVC, but only use MinGW-w64 to build the library. I've tried using Rust's `x86_64-pc-windows-gnu` target, but the CI keeps on using MSVC.
The final commits are general fixes to the Makefile.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -163,14 +163,6 @@ jobs:
       env:
         CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
 
-    - name: Build C library on Windows (make)
-      if: ${{ runner.os == 'Windows' }}
-      run: make.sh -j CC="$CC" CFLAGS="$CFLAGS" AR="$AR"
-      env:
-        CC: x86_64-w64-mingw32-gcc
-        AR: x86_64-w64-mingw32-ar
-        CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
-
     - name: Build C library (CMake)
       if: ${{ !matrix.use-cross }}
       run: |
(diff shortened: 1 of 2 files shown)


MisterDA:
Note that as the `install` rule depends on `all` and `$(SOEXTVER)`, `$CC` needs to be set (or correctly discovered) when running `make install`, which is a bit unusual. Also, I don't think that `make uninstall` should depend on `DESTDIR`, but rather on `PREFIX`: users don't want to remove files from their temporary, fakeroot directory, but from the fileystem…

MisterDA:
gentle ping @ObserverOfTime @maxbrunsfeld

ognevny:
hello. I maintain tree-sitter package on MSYS2 distribution (which is cygwin+mingw) and we use CMake to compile tree-sitter library. does it worth to change Makefile actually? CMake is always faster way to build than Makefiles

MisterDA:
I agree that CMake is superior here, but I'm contributing to a lengthy build process where it's not used and I'm reluctant to introduce it. I think these are overall simple fixes to the Makefile.

MisterDA:
Please consider merging this, or revert my #4201, which is incomplete and fixed by this PR.

shonfeder:
Hello! I thought I would check in here. Anything needed to help move this forward?

ObserverOfTime (approved):

PR #4205, opened 2025-02-14
feat(cli): Version updates (#4179)

### LINKED ISSUE #4179: `tree-sitter version` with no arguments displays current version
### Problem

I love the new `version` feature -- a massive quality-of-life improvement, thank you.

It would be quite convenient if the command prints out the current version if you don't pass in a new version argument.  It can pull it from package.json or something, or even print out the whole list in a table and warn if they don't all match...

On a similar vein -- `tree-sitter version --bump` could be useful as well: it reads the current version from `package.json`, increments the patch version, and updates all the other files.  But I'm happy to do that manually -- grab the existing using `tree-sitter version` and then update using the existing usage.  Because of the multiple sources of truth, a manual approach might be better.

It would be even more convenient if I can run `tree-sitter version --tag`, which then runs `git tag v1.2.3` for me, with the correct version

PS. I searched past issues and PRs for a duplicate, but searching for `version` results in such a long list that I might have missed it -- my apologies if this is a duplicate.

PPS. I was thinking of implementing this myself and then issue a PR, but I took one look at the source and released that I have a very long way to go before I understand Rust sufficiently well to be able to do such an update.

### Expected behavior

```
$  tree-sitter version

CMakeLists.txt   = 1.2.3
Cargo.toml       = 1.2.3
Makefile         = 1.2.3
package.json     = 1.2.4
pyproject.toml   = 1.2.3
tree-sitter.json = 1.2.3

warning: the versions are inconsistent

$ tree-sitter version --bump

CMakeLists.txt   = 1.2.3 => 1.2.5
Cargo.toml       = 1.2.3 => 1.2.5
Makefile         = 1.2.3 => 1.2.5
package.json     = 1.2.4 => 1.2.5
pyproject.toml   = 1.2.3 => 1.2.5
tree-sitter.json = 1.2.3 => 1.2.5

$ tree-sitter version --tag

added a new git tag v1.2.5

$
```

jpt13653903:
This closes #4179 

The CLI `version` feature can now perform automated version bumping.

```
$  tree-sitter version
Current version: 1.2.3

$ tree-sitter version --bump
Bumping version 1.2.3 to 1.2.4

$ tree-sitter version --bump-minor
Bumping version 1.2.4 to 1.3.0

$ tree-sitter version --bump-major
Bumping version 1.3.0 to 2.0.0

$ tree-sitter version 1.2.3
Bumping version 2.0.0 to 1.2.3
```

### DIFF
--- a/crates/cli/src/version.rs
+++ b/crates/cli/src/version.rs
@@ -1,28 +1,85 @@
 use std::{fs, path::PathBuf, process::Command};
 
 use anyhow::{anyhow, Context, Result};
+use clap::ValueEnum;
 use regex::Regex;
+use semver::Version as SemverVersion;
+use std::cmp::Ordering;
 use tree_sitter_loader::TreeSitterJSON;
 
+#[derive(Clone, Copy, Default, ValueEnum)]
+pub enum BumpLevel {
+    #[default]
+    Patch,
+    Minor,
+    Major,
+}
+
 pub struct Version {
-    pub version: String,
+    pub version: Option<SemverVersion>,
     pub current_dir: PathBuf,
+    pub bump: Option<BumpLevel>,
 }
 
 impl Version {
     #[must_use]
-    pub const fn new(version: String, current_dir: PathBuf) -> Self {
+    pub const fn new(
+        version: Option<SemverVersion>,
+        current_dir: PathBuf,
+        bump: Option<BumpLevel>,
+    ) -> Self {
         Self {
             version,
             current_dir,
+            bump,
         }
     }
 
-    pub fn run(self) -> Result<()> {
+    pub fn run(mut self) -> Result<()> {
         let tree_sitter_json = self.current_dir.join("tree-sitter.json");
 
         let tree_sitter_json =
             serde_json::from_str::<TreeSitterJSON>(&fs::read_to_string(tree_sitter_json)?)?;
+
+        let current_version = tree_sitter_json.metadata.version;
+        self.version = match (self.version.is_some(), self.bump) {
+            (false, None) => {
+                println!("Current version: {current_version}");
+                return Ok(());
+            }
+            (true, None) => self.version,
+            (false, Some(bump)) => {
+                let mut v = current_version.clone();
+                match bump {
+                    BumpLevel::Patch => v.patch += 1,
+                    BumpLevel::Minor => {
+                        v.minor += 1;
+                        v.patch = 0;
+                    }
+                    BumpLevel::Major => {
+                        v.major += 1;
+                        v.minor = 0;
+                        v.patch = 0;
+                    }
+                }
+                Some(v)
+            }
+            (true, Some(_)) => unreachable!(),
+        };
+
+        let new_version = self.version.as_ref().unwrap();
+        match new_version.cmp(&current_version) {
+            Ordering::Less => {
+                eprintln!("Warning: new version is lower than current!");
+                println!("Reverting version {current_version} to {new_version}");
+            }
+            Ordering::Greater => {
+                println!("Bumping version {current_version} to {new_version}");
+            }
+            Ordering::Equal => {
+                println!("Keeping version {current_version}");
+            }
+        }
 
         let is_multigrammar = tree_sitter_json.grammars.len() > 1;
 
@@ -80,7 +137,7 @@ impl Version {
                     format!(
                         "{}{}{}",
                         &line[..start_quote],
-                        self.version,
+                        self.version.as_ref().unwrap(),
                         &line[end_quote..]
                     )
                 } else {
@@ -107,7 +164,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("version =") {
-                    format!("version = \"{}\"", self.version)
+                    format!("version = \"{}\"", self.version.as_ref().unwrap())
                 } else {
                     line.to_string()
                 }
@@ -157,7 +214,7 @@ impl Version {
                     format!(
                         "{}{}{}",
                         &line[..start_quote],
-                        self.version,
+                        self.version.as_ref().unwrap(),
                         &line[end_quote..]
                     )
                 } else {
@@ -208,7 +265,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("VERSION") {
-                    format!("VERSION := {}", self.version)
+                    format!("VERSION := {}", self.version.as_ref().unwrap())
                 } else {
                     line.to_string()
                 }
@@ -230,7 +287,7 @@ impl Version {
         let cmake = fs::read_to_string(self.current_dir.join("CMakeLists.txt"))?;
 
         let re = Regex::new(r#"(\s*VERSION\s+)"[0-9]+\.[0-9]+\.[0-9]+""#)?;
-        let cmake = re.replace(&cmake, format!(r#"$1"{}""#, self.version));
+        let cmake = re.replace(&cmake, format!(r#"$1"{}""#, self.version.as_ref().unwrap()));
 
         fs::write(self.current_dir.join("CMakeLists.txt"), cmake.as_bytes())?;
 
@@ -248,7 +305,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("version =") {
-                    format!("version = \"{}\"", self.version)
+                    format!("version = \"{}\"", self.version.as_ref().unwrap())
                 } else {
                     line.to_string()
                 }
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -26,7 +26,9 @@ use tree_sitter_cli::{
     playground, query,
     tags::{self, TagsOptions},
     test::{self, TestOptions, TestStats},
-    test_highlight, test_tags, util, version, wasm,
+    test_highlight, test_tags, util, version,
+    version::BumpLevel,
+    wasm,
 };
 use tree_sitter_config::Config;
 use tree_sitter_highlight::Highlighter;
@@ -53,7 +55,7 @@ enum Commands {
     Parse(Parse),
     /// Run a parser's tests
     Test(Test),
-    /// Increment the version of a grammar
+    /// Display or increment the version of a grammar
     Version(Version),
     /// Fuzz a parser
     Fuzz(Fuzz),
@@ -327,14 +329,26 @@ struct Test {
 
 #[derive(Args)]
 #[command(alias = "publish")]
-/// Increment the version of a grammar
+/// Display or increment the version of a grammar
 struct Version {
-    #[arg(num_args = 1)]
     /// The version to bump to
-    pub version: SemverVersion,
+    #[arg(
+        conflicts_with = "bump",
+        long_help = "\
+        The version to bump to\n\
+        \n\
+        Examples:\n    \
+            tree-sitter version: display the current version\n    \
+            tree-sitter version <version>: bump to specified version\n    \
+            tree-sitter version --bump <level>: automatic bump"
+    )]
+    pub version: Option<SemverVersion>,
     /// The path to the tree-sitter grammar directory
     #[arg(long, short = 'p')]
     pub grammar_path: Option<PathBuf>,
+    /// Automatically bump from the current version
+    #[arg(long, value_enum, conflicts_with = "version")]
+    pub bump: Option<BumpLevel>,
 }
 
 #[derive(Args)]
@@ -1314,7 +1328,7 @@ impl Test {
 
 impl Version {
     fn run(self, current_dir: PathBuf) -> Result<()> {
-        version::Version::new(self.version.to_string(), current_dir).run()
+        version::Version::new(self.version, current_dir, self.bump).run()
     }
 }
 


jpt13653903:
I see I have linting errors...  I'll fix those tomorrow.  Anything else I should fix while I'm at it?

jpt13653903:
I'm trying to fix the linting errors, but I don't understand this one:

```
error: boolean to int conversion using if
Error:     --> cli/src/main.rs:1203:16
     |
1203 |           } else if self.bump {
     |  ________________^
1204 | |             1
1205 | |         } else {
1206 | |             0
1207 | |         };
     | |_________^ help: replace with from: `{ i32::from(self.bump) }`
     |
     = note: `self.bump as i32` or `self.bump.into()` can also be valid options
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bool_to_int_with_if
     = note: `-D clippy::bool-to-int-with-if` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::bool_to_int_with_if)]`
```

The if-statement is returning `1` or `0`, which are both integers, and `self.bump` is a boolean, which is given to the `if` condition, which should expect a boolean.  What is it complaining about?

jpt13653903:
> I'm trying to fix the linting errors, but I don't understand this one:

Oh -- Never mind.  I get it.

jpt13653903:
My apologies -- I was an idiot.  I wanted to update my VHDL queries in nvim-treesitter, and force-pushed to this master instead...

I'll re-open the PR

jpt13653903:
> ...and that is why you should never open PRs from your `master` branch.

Indeed -- good advice

cli/src/main.rs
```diff
@@ -289,11 +289,25 @@ struct Test {
...
 struct Version {
-    #[arg(num_args = 1)]
     /// The version to bump to
+    #[arg(
+        default_value = "0.0.0",
+        long_help = "When present, the version to bump to\n\
+                     When absent with no flags present, prints the current version\n\
+                     When absent with flags present, bumps the current version"
+    )]
     pub version: SemverVersion,
```
amaanq, line 301: Make this an Option, and remove the default

cli/src/main.rs
```diff
@@ -289,11 +289,25 @@ struct Test {
...
+    /// Bump the patch version
+    #[arg(long)]
+    pub bump: bool,
+    /// Bump the minor version
+    #[arg(long)]
+    pub bump_minor: bool,
+    /// Bump the major version
+    #[arg(long)]
+    pub bump_major: bool,
```
amaanq, line 310: Instead of 3 bool flags, this should be one argument, `bump`, that takes an enum, of which the enum has the major/minor/patch variants. Then, you can make patch the default variant.

cli/src/version.rs
```diff
@@ -7,23 +7,54 @@ use tree_sitter_loader::TreeSitterJSON;
...
-    pub fn run(self) -> Result<()> {
+    pub fn run(mut self) -> Result<()> {
         let tree_sitter_json = self.current_dir.join("tree-sitter.json");
 
         let tree_sitter_json =
             serde_json::from_str::<TreeSitterJSON>(&fs::read_to_string(tree_sitter_json)?)?;
 
+        let current_version = tree_sitter_json.metadata.version.to_string();
+
+        if self.version == "0.0.0" {
```
amaanq, line 31: This would be a check for version being `None` once you update the type

cli/src/version.rs
```diff
@@ -7,23 +7,54 @@ use tree_sitter_loader::TreeSitterJSON;
...
+            // No argument given
+            let mut parts = current_version.split('.');
+            let mut major = parts.next().unwrap().parse::<i32>().unwrap();
+            let mut minor = parts.next().unwrap().parse::<i32>().unwrap();
+            let mut patch = parts.next().unwrap().parse::<i32>().unwrap();
```
amaanq, line 36: `semver::Version::parse`

cli/src/main.rs
```diff
@@ -53,7 +53,7 @@ enum Commands {
     Parse(Parse),
     /// Run a parser's tests
     Test(Test),
-    /// Increment the version of a grammar
+    /// Print or increment the version of a grammar
```
amaanq, line 56: Nit, `display` over `print`

```suggestion
    /// Display or increment the version of a grammar
```

cli/src/main.rs
```diff
@@ -289,11 +289,25 @@ struct Test {
 
 #[derive(Args)]
 #[command(alias = "publish")]
-/// Increment the version of a grammar
+/// Print or increment the version of a grammar
```
amaanq, line 292:
```suggestion
/// Display or increment the version of a grammar
```

cli/src/main.rs
```diff
@@ -1176,7 +1190,14 @@ impl Test {
...
     fn run(self, current_dir: PathBuf) -> Result<()> {
-        version::Version::new(self.version.to_string(), current_dir).run()
+        let bump_level = if self.bump_major {
+            3
+        } else if self.bump_minor {
+            2
+        } else {
+            i32::from(self.bump)
+        };
+        version::Version::new(self.version.to_string(), current_dir, bump_level).run()
```
amaanq, line 1200: since self.bump will be an enum, the extra argument is not needed.

amaanq (changes requested):

jpt13653903:
Awesome - thanks for the review!

I'll get right on it.

clason:
> tree-sitter version --bump-minor

I would prefer a single `--bump` flag with `--bump=patch`,  `--bump=minor` or `--bump=major` (or `--bump=dev`, as default?) options. (I think that's similar to `npm`?)

jpt13653903:
> > tree-sitter version --bump-minor
> 
> I would prefer a single `--bump` flag with `--bump=patch`, `--bump=minor` or `--bump=major` (or `--bump=dev`, as default?) options. (I think that's similar to `npm`?)

I think that is what Amaan has in mind...  I'm busy wrangling Rust to try and get it to play along...  Will push my update some time later today the moment I actually get it working.

cli/src/main.rs
jpt13653903: I cannot get Rust to make "patch" the default variant.  The "Optional" wrapper is not playing along.

I'm happy to leave it as explicit though.  You then say

- `tree-sitter version` to display the current version
- `tree-sitter version 1.2.3` to set it directly (current behaviour, which I do not want to change)
- `tree-sitter version --bump patch` to bump the patch.
- `tree-sitter version --bump minor` to bump the minor; etc.

clason: Works for me; the added default is only very minor QoL.

cli/src/main.rs
```diff
@@ -290,11 +292,16 @@ struct Test {
...
 #[command(alias = "publish")]
-/// Increment the version of a grammar
+/// Display or increment the version of a grammar
 struct Version {
-    #[arg(num_args = 1)]
     /// The version to bump to
-    pub version: SemverVersion,
+    #[arg(long_help = "When present, the version to bump to\n\
+                       When absent with no flags present, displays the current version\n\
+                       When absent with the --bump flag present, bumps the current version")]
```
clason, line 300: Is that still correct? If so, should specify _how_ the version is bumped.

(Also, the description is a bit confusing, since the option makes no sense without the `--bump` flag. So the second line is impossible?)

jpt13653903: Would it help to put examples into this help string?  Something like:

```
tree-sitter version: display the current version
tree-sitter version <version>: bump to specified version
tree-sitter version --bump <level>: automatic bump, see below for options
```

jpt13653903: Or maybe with less verbosity...

```
No parameters: display the current version
<version> only: bump to specified version
--bump <level>: automatic bump, see below for options
```
?

jpt13653903: I pushed an update...  It now renders the long help as:

image

jpt13653903: I'm going to drop the "see below for options", and change "Bump the version" to "Automatically bump from the current version"

jpt13653903:
I'm busy re-basing onto the latest master, and I'm a little confused about what to do with the new [`grammar_path`](https://github.com/tree-sitter/tree-sitter/blob/876d5a96bf05037bd7fd6fa38024c2b64fcc94c2/crates/cli/src/main.rs#L308) option...

It's not being used when calling the `Version` implementation [here](https://github.com/tree-sitter/tree-sitter/blob/876d5a96bf05037bd7fd6fa38024c2b64fcc94c2/crates/cli/src/main.rs#L1206).

I think I'm simply going to leave the `grammar_path` hanging and rebase normally...

jpt13653903:
GitHub is still complaining about "1 change requested by reviewers with write access.", but I cannot figure out what that change request is...

Is this simply GitHub being confused...?  Or am I being blind again...?

jpt13653903:
> I'm busy re-basing onto the latest master, and I'm a little confused about what to do with the new [`grammar_path`](https://github.com/tree-sitter/tree-sitter/blob/876d5a96bf05037bd7fd6fa38024c2b64fcc94c2/crates/cli/src/main.rs#L308) option...
> 
> It's not being used when calling the `Version` implementation [here](https://github.com/tree-sitter/tree-sitter/blob/876d5a96bf05037bd7fd6fa38024c2b64fcc94c2/crates/cli/src/main.rs#L1206).
> 
> I think I'm simply going to leave the `grammar_path` hanging and rebase normally...

Ah -- Never-mind

It passes it in [here](https://github.com/tree-sitter/tree-sitter/blob/876d5a96bf05037bd7fd6fa38024c2b64fcc94c2/crates/cli/src/main.rs#L1736)

jpt13653903:
Will do

jpt13653903:
> Please move the docs change to a separate PR that will be merged before release and squash the other commits.

Done...

Or did you mean to have this one based on top of #4726?  Let me know, then I can run the rebase...

ObserverOfTime:
I meant before the next release but it's fine.

WillLillis:
Open question (not to block this PR): should this command display an error/warning if the version being bumped to is less than or equal to the current version?

jpt13653903:
> I'd like to rework the logic for ...

No problem -- I'll apply your patch.  I'm still learning (first time using Rust after a C/C++ background).  My goodness the syntax is weird...  Matching case on tuples is an awesome feature though -- didn't know about that one.

> ... should this command display an error/warning ...

Quite the philosophical question...  My take is that you don't want to error out.  The idea of the [print](https://github.com/jpt13653903/tree-sitter/blob/4ebb54990033d5b5308619c3b7630706fe42e6c4/crates/cli/src/version.rs#L66) is so that the user can visually inspect whether that is what he / she wanted, and maybe a backwards bump is exactly what they wanted.

How about making the [print](https://github.com/jpt13653903/tree-sitter/blob/4ebb54990033d5b5308619c3b7630706fe42e6c4/crates/cli/src/version.rs#L66) display in "info" style (i.e. with "Info: " prefix and maybe a different colour, like cyan) when bumping normally, and in "warning" style (i.e. with "Warning: " prefix and some other colour, like yellow or orange) when bumping backwards?

ObserverOfTime:
> I'm still learning (first time using Rust after a C/C++ background). My goodness the syntax is weird... Matching case on tuples is an awesome feature though -- didn't know about that one.

Rust is awesome and awful at the same time. ~~Mostly awful though.~~

jpt13653903:
I also included a suggestion related to warning about backwards bumps in the latest push...  Let me know what you think

crates/cli/src/version.rs
```diff
@@ -1,29 +1,77 @@
...
+                    }
+                }
+                Some(v)
+            }
+            (true, Some(_)) => unreachable!(),
+        };
+
+        let new_version = self.version.as_ref().unwrap();
+        if new_version <= &current_version {
+            print!("\x1b[93m{}\x1b[0m: ", "Warning"); // Bright yellow
```
ObserverOfTime:
```suggestion
            eprintln!("Warning: new version is lower than current!");
```

jpt13653903: What about:

```diff
--- a/crates/cli/src/version.rs
+++ b/crates/cli/src/version.rs
@@ -67,10 +67,14 @@ impl Version {
         };

         let new_version = self.version.as_ref().unwrap();
-        if new_version <= &current_version {
-            print!("\x1b[93m{}\x1b[0m: ", "Warning"); // Bright yellow
+        if new_version < &current_version {
+            eprintln!("Warning: new version is lower than current!");
+            println!("Reverting version {current_version} to {new_version}");
+        } else if new_version > &current_version {
+            println!("Bumping version {current_version} to {new_version}");
+        } else {
+            println!("Keeping version {current_version}");
         }
-        println!("Bumping version {current_version} to {new_version}");

         let is_multigrammar = tree_sitter_json.grammars.len() > 1;
```

instead?

ObserverOfTime: You can use `match` with [`cmp`](https://docs.rs/semver/1.0.25/semver/struct.Version.html#method.cmp) to avoid the double check.

jpt13653903: As in this?

```rust
        let new_version = self.version.as_ref().unwrap();
        match new_version.cmp(&current_version) {
            Ordering::Less => {
                eprintln!("Warning: new version is lower than current!");
                println!("Reverting version {current_version} to {new_version}");
            }
            Ordering::Greater => {
                println!("Bumping version {current_version} to {new_version}");
            }
            Ordering::Equal => {
                println!("Keeping version {current_version}");
            }
        }
```

WillLillis:
> No problem -- I'll apply your patch. I'm still learning (first time using Rust after a C/C++ background). My goodness the syntax is weird... Matching case on tuples is an awesome feature though -- didn't know about that one.

Thanks! Yeah, it's definitely very weird 😆. I'm currently learning C++ after starting out with Rust so I can definitely relate.

crates/cli/src/version.rs
```diff
@@ -80,7 +137,7 @@ impl Version {
                     format!(
                         "{}{}{}",
                         &line[..start_quote],
-                        self.version,
+                        self.version.clone().unwrap(),
```
WillLillis:
```suggestion
                        self.version.as_ref().unwrap(),
```

crates/cli/src/version.rs
```diff
@@ -107,7 +164,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("version =") {
-                    format!("version = \"{}\"", self.version)
+                    format!("version = \"{}\"", self.version.clone().unwrap())
```
WillLillis:
```suggestion
                    format!("version = \"{}\"", self.version.as_ref().unwrap())
```

crates/cli/src/version.rs
```diff
@@ -157,7 +214,7 @@ impl Version {
                     format!(
                         "{}{}{}",
                         &line[..start_quote],
-                        self.version,
+                        self.version.clone().unwrap(),
```
WillLillis:
```suggestion
                        self.version.as_ref().unwrap(),
```

crates/cli/src/version.rs
```diff
@@ -208,7 +265,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("VERSION") {
-                    format!("VERSION := {}", self.version)
+                    format!("VERSION := {}", self.version.clone().unwrap())
```
WillLillis:
```suggestion
                    format!("VERSION := {}", self.version.as_ref().unwrap())
```

crates/cli/src/version.rs
```diff
@@ -230,7 +287,7 @@ impl Version {
         let cmake = fs::read_to_string(self.current_dir.join("CMakeLists.txt"))?;
 
         let re = Regex::new(r#"(\s*VERSION\s+)"[0-9]+\.[0-9]+\.[0-9]+""#)?;
-        let cmake = re.replace(&cmake, format!(r#"$1"{}""#, self.version));
+        let cmake = re.replace(&cmake, format!(r#"$1"{}""#, self.version.clone().unwrap()));
```
WillLillis:
```suggestion
        let cmake = re.replace(&cmake, format!(r#"$1"{}""#, self.version.as_ref().unwrap()));
```

crates/cli/src/version.rs
```diff
@@ -248,7 +305,7 @@ impl Version {
             .lines()
             .map(|line| {
                 if line.starts_with("version =") {
-                    format!("version = \"{}\"", self.version)
+                    format!("version = \"{}\"", self.version.clone().unwrap())
```
WillLillis:
```suggestion
                    format!("version = \"{}\"", self.version.as_ref().unwrap())
```

jpt13653903:
> LGTM! Once you replace those `clone` calls with `as_ref` I think we're good to merge. Thanks so much!

Done 😃

PR #4207, opened 2025-02-16
fix typescript typechecking errors

vemoo:
- Update some type definitions
- Use interface instead of class for `LanguageMetadata`
- Fix assignment to wrong field
- Silence errors on internal fields
- Run typescript build on `build:ts`
- Enable composite for faster typescript builds
- Enable isolatedModules to ensure code is transpiler friendly
- Move `@types/emscripten` to peerDependencies for `EmscriptenModule`

### DIFF
--- a/lib/binding_web/lib/web-tree-sitter.d.ts
+++ b/lib/binding_web/lib/web-tree-sitter.d.ts
@@ -50,7 +50,7 @@ declare namespace RuntimeExports {
     function setValue(ptr: number, value: number, type?: string): void;
     let currentParseCallback: ((index: number, position: {row: number, column: number}) => string | undefined) | null;
     let currentLogCallback: ((message: string, isLex: boolean) => void) | null;
-    let currentProgressCallback: ((state: {currentOffset: number}) => void) | null;
+    let currentProgressCallback: ((state: {currentOffset: number, hasError: boolean}) => void) | null;
     let currentQueryProgressCallback: ((state: {currentOffset: number}) => void) | null;
     let HEAPF32: Float32Array;
     let HEAPF64: Float64Array;
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -71,7 +71,6 @@
   ],
   "devDependencies": {
     "@eslint/js": "^9.20.0",
-    "@types/emscripten": "^1.40.0",
     "@types/node": "^22.13.1",
     "@vitest/coverage-v8": "^3.0.5",
     "dts-buddy": "^0.5.4",
@@ -83,8 +82,16 @@
     "typescript-eslint": "^8.23.0",
     "vitest": "^3.0.5"
   },
+  "peerDependencies": {
+    "@types/emscripten": "^1.40.0"
+  },
+  "peerDependenciesMeta": {
+    "@types/emscripten": {
+      "optional": true
+    }
+  },
   "scripts": {
-    "build:ts": "node script/build.js",
+    "build:ts": "tsc -b . && node script/build.js",
     "build:wasm": "cd ../../ && cargo xtask build-wasm",
     "build:wasm:debug": "cd ../../ && cargo xtask build-wasm --debug",
     "build": "npm run build:wasm && npm run build:ts",
--- a/lib/binding_web/src/marshal.ts
+++ b/lib/binding_web/src/marshal.ts
@@ -168,10 +168,9 @@ export function marshalEdit(edit: Edit, address = TRANSFER_BUFFER) {
  *
  * Unmarshals a {@link LanguageMetadata} from the transfer buffer.
  */
-export function unmarshalLanguageMetadata(address: number): LanguageMetadata {
-  const result = {} as LanguageMetadata;
-  result.major_version = C.getValue(address, 'i32'); address += SIZE_OF_INT;
-  result.minor_version = C.getValue(address, 'i32'); address += SIZE_OF_INT;
-  result.field_count = C.getValue(address, 'i32');
-  return result;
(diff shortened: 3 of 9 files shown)


lib/binding_web/src/marshal.ts
```diff
@@ -168,10 +168,13 @@ export function marshalEdit(edit: Edit, address = TRANSFER_BUFFER) {
  *
  * Unmarshals a {@link LanguageMetadata} from the transfer buffer.
  */
-export function unmarshalLanguageMetadata(address: number): LanguageMetadata {
-  const result = {} as LanguageMetadata;
-  result.major_version = C.getValue(address, 'i32'); address += SIZE_OF_INT;
-  result.minor_version = C.getValue(address, 'i32'); address += SIZE_OF_INT;
-  result.field_count = C.getValue(address, 'i32');
```
vemoo, line 175: I think this was an error because `field_count` doesn't exist in `LanguageMetadata`, I have changed it to `patch_version`.

lib/binding_web/lib/web-tree-sitter.d.ts
```diff
@@ -50,7 +50,7 @@ declare namespace RuntimeExports {
     function setValue(ptr: number, value: number, type?: string): void;
     let currentParseCallback: ((index: number, position: {row: number, column: number}) => string | undefined) | null;
     let currentLogCallback: ((message: string, isLex: boolean) => void) | null;
-    let currentProgressCallback: ((state: {currentOffset: number}) => void) | null;
+    let currentProgressCallback: ((state: {currentOffset: number, hasError: boolean}) => void) | null;
```
vemoo, line 53: I just noticed that this file has a comment saying that it's automatically generated.
This change was needed to avoid an error here: https://github.com/tree-sitter/tree-sitter/blob/f0571b1e33cec3d5c1ec972a7a2001f5495089ab/lib/binding_web/src/parser.ts#L202

No sure which the proper place to fix this is.

vemoo: I have found that I can run `cargo xtask build-wasm --emit-tsd` to generate that file, but it doesn't quite match, which leads me to believe that despite the comment at the beginning this file has been edited by hand.

This line 
https://github.com/tree-sitter/tree-sitter/blob/8b864877dd1acbd2d238d97a28f4e06e1d72610e/lib/binding_web/lib/web-tree-sitter.d.ts#L220

in the code that gets generated for me appears as:
```typescript
export default function MainModuleFactory (options?: unknown): Promise<MainModule>;
```

I don't think that using `EmscriptenModule` is correct because that means that when calling `MainModuleFactory` with `options` all properties and methods of `EmscriptenModule` should be provided. I think that it should be `Partial<EmscriptenModule>`. This seems to match that:
https://github.com/amaanq/DefinitelyTyped/blob/b46ebd365a2091aa3f077e409bc9ff622a64f602/types/emscripten/index.d.ts#L99-L101

Another issue is that because `EmscriptenModule` comes from `@types/emscripten` and appears in the published types of `web-tree-sitter`, that package should be in `dependencies` or `peerDependencies` not in `devDependencies`.

lib/binding_web/tsconfig.json
```diff
@@ -18,18 +18,20 @@
...
     "alwaysStrict": true,
-    "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
     "moduleResolution": "node",
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "skipLibCheck": true,
+    "composite": true,
+    "isolatedModules": true,
```
amaanq, line 28: Please explain in your PR description or in a comment as to why this change was made

lib/binding_web/package.json
```diff
@@ -70,7 +70,7 @@
     "vitest": "^3.0.5"
   },
   "scripts": {
-    "build:ts": "node script/build.js",
+    "build:ts": "tsc -b . && node script/build.js",
```
amaanq, line 73: Why?

lib/binding_web/lib/web-tree-sitter.d.ts
amaanq: Yes, it's auto-generated but I've tweaked it iirc because some definitions loosely used `any`, or were wrong. We also add some methods at runtime, in [imports.js](https://github.com/tree-sitter/tree-sitter/blob/853ca46899e41ce23ab8297bbebd26851482258a/lib/binding_web/lib/imports.js), which is why I manually added those methods here. I'd rather it not be touched, unless necessary, though your proposed changed would be good to add. I would be okay moving `@types/emscripten` to `peerDependencies`, and I agree with the change to make `options` `Partial<EmscriptenModule>`. I appreciate the investigation here :slightly_smiling_face:

lib/binding_web/tsconfig.json
```diff
@@ -18,18 +18,20 @@
...
   },
   "include": [
     "src/**/*",
     "script/**/*",
     "test/**/*",
+    "lib/**/*.ts"
```
amaanq, line 34: and explain this, is it needed? I don't mind, but I'd like to understand what the benefit is

lib/binding_web/src/marshal.ts
amaanq: :facepalm: thank you

amaanq (changes requested):

lib/binding_web/tsconfig.json
vemoo: I added it so that `lib/web-tree-sitter.d.ts` it is type-checked also when running `tsc --build`

amaanq (approved):

PR #4218, opened 2025-02-18
Reset `result_symbol` field of lexer in wasm memory in between invocations

maxbrunsfeld:
### Background

Normally, the external scanners are expected to assign to the `result_symbol` field of the `TSLexer` before returning true. But if they don't assign it, it is initialized with a default value of zero (the first external token, whatever that is).

Currently, the [tree-sitter-yaml](https://github.com/tree-sitter-grammars/tree-sitter-yaml) external scanner (and possibly others) rely on this zero value, and sometimes return true without writing to it.

### Problem

When running external scanners loaded from WASM, we did not reset `result_symbol` in between invocations of the lexer, so lexer's `result_symbol` value would depend on what was returned from the previous call.

This PR fixes that.

### DIFF
--- a/lib/src/wasm_store.c
+++ b/lib/src/wasm_store.c
@@ -1613,13 +1613,22 @@ static void ts_wasm_store__call(
   }
 }
 
+// The data fields of TSLexer, without the function pointers.
+//
+// This portion of the struct needs to be copied in and out
+// of wasm memory before and after calling a scan function.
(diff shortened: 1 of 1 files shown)


jabari-max (commented):
- name: Setup Swift
  uses: swift-actions/setup-swift@v2.3.0

PR #4229, opened 2025-02-20
[web-tree-sitter] Fix type definitions in `exports`…

### LINKED ISSUE #4228: web-tree-sitter does not expose type information
### Problem

web-tree-sitter does not properly expose type information, making typescript tooling fail.

Following up on this discord discussion https://discord.com/channels/1063097320771698699/1063097321648312354/1341486311760072744

### Steps to reproduce

See here - https://arethetypeswrong.github.io/?p=web-tree-sitter%400.25.2

### Expected behavior

Types are properly exposed and available to tooling

0.25.2

### Operating system/version

Windows 11

savetheclocktower:
…and generate `.d.cts` files for CommonJS exports.

Fixes #4228, though someone might want to verify that with actual testing in TypeScript.

The current structure of `"exports"` in `package.json` is incorrect; [this TypeScript blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package.json-exports-imports-and-self-referencing) details how it ought to look, and reminds us that the `"types"` field must come before the `"default"` field.

[Arethetypeswrong](https://arethetypeswrong.github.io/) gives us a good grade here, if not an A+:

Screenshot 2025-02-20 at 12 37 20 AM

The “Resolution failed” there is because `node10` is the oldest TypeScript resolution algorithm, one which predates the invention of the `"exports"` field in `package.json`. [If we absolutely wanted](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NoResolution.md#true-positive-node-10-doesnt-support-packagejson-exports) the extra credit here, we could copy the `.d.ts` and `.d.cts` files into the `debug` folder, but I'm not sure it's even worth doing.

Once I did that, arethetypeswrong chided me for having one `.d.ts` file — an ESM module's type definition, since we do `"type": "module"` in `package.json` — pulling double duty as also documenting the `.cjs` file. This is supposedly bad [for the reasons explained here](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md); and if someone writes decent prose in an authoritative voice, I tend to believe it.

To fix this, I'm having the `build:dts` task generate both `.d.ts` and `.d.cts` files at once. This felt more elegant than just copying one file to the other destination on a build, especially since the corresponding `.map` files reference specific file names.

To test, I just recapitulated the steps [in the CI workflow](https://github.com/tree-sitter/tree-sitter/blob/master/.github/workflows/build.yml#L194-L204) locally, ran `npm pack`, and gave arethetypeswrong the resulting `.tgz`. I may or may not get around to testing in an actual dummy TypeScript project.


sqs:
Confirming I experienced the issue in #4228 and that this PR fixes it for me. In the meantime as workaround I just used `pnpm patch`.

amaanq (approved):

ObserverOfTime:
Looks like some previous related change didn't make it in 0.25 so this can't make it either.

PR #4231, opened 2025-02-21
Mark all format-like macros for Clippy

nyurik:
See https://doc.rust-lang.org/nightly/clippy/attribs.html#clippyformat_args

Ensure that usage of the macros with `format!`-like parameters will be validated by Clippy


WillLillis (approved):

WillLillis:
Thanks!

PR #4232, opened 2025-02-21
Place APIs for generating grammars from a file behind a feature flag

shadaj:
The `url` dependency pulls in a lot of transitive ICU crates, which impacts compile times when `tree-sitter-generate` is being used from a crate like `rust-sitter` which passes in the grammar definition in-memory. This shifts the APIs that depend on `url` behind a feature flag for loading grammars from a file. Furthermore, `url` is only needed on Windows, so we avoid the dependency altogether on non-Windows platforms.

### DIFF
--- a/crates/generate/Cargo.toml
+++ b/crates/generate/Cargo.toml
@@ -19,6 +19,10 @@ path = "src/generate.rs"
 [lints]
 workspace = true
 
+[features]
+default = ["load"]
+load = ["dep:semver", "dep:url"]
+
 [dependencies]
 anyhow.workspace = true
 heck.workspace = true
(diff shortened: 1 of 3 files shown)


ObserverOfTime:
> Furthermore, `url` is only needed on Windows, so we avoid the dependency altogether on non-Windows platforms.

That was the case before but ea3846a inadvertently changed it.

shadaj:
> > Furthermore, `url` is only needed on Windows, so we avoid the dependency altogether on non-Windows platforms.
> 
> That was the case before but [ea3846a](https://github.com/tree-sitter/tree-sitter/commit/ea3846a2c592fa4e4c141b27919eef41e6fd79f1) inadvertently changed it.

Just to make sure I parsed this correctly; my change to shift it back to being a target-specific dependency should be okay, right?

shadaj:
Looks like CI failures were due to being out-of-date, rebased.

shadaj:
Will need another re-run 😢, CI flaked in a curl.

MalikNadirawan:
***@***.***


-------- Original Email --------

amaanq (approved):

clason:
Please squash the commits (`feat(cli): ...`) before merging.

PR #4249, opened 2025-02-28
adapt to new rust lints and tweak ci

WillLillis:
Fixes the CI failure discussed in https://github.com/tree-sitter/tree-sitter/pull/4244

CC @clason

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: stable
+          components: clippy, rustfmt
 
       - name: Set up nightly Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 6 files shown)


WillLillis:
Oh dear, looks like that's a bug in cargo itself.

WillLillis:
Looks like https://github.com/rust-lang/rust/issues/137640

clason:
Again, I'm wondering why you're YOLOing nightly Rust in CI...

WillLillis:
> Again, I'm wondering why you're YOLOing nightly Rust in CI...

¯\_(ツ)_/¯

I think amaanq likes to stay on top of the nightly lints. Maybe it would be better to run the nightly lints as a separate job and use stable for everything else? That way any new lint failures are visible, but issues in the nightly toolchain won't be  a blocker like here.

clason:
Yes, that would be reasonable (and not mark it as required).

WillLillis:
Something something first try?

Changes now include fixes for two new lints, test updates to reflect changes in tree-sitter-rust, as well as moving the nightly clippy invocation to a separate job that doesn't cause the action to fail. I'm not sure why `+stable` is necessary for the regular clippy invocation. Without it, clippy nightly 1.86 is used, leading to the crash mentioned above. 

https://github.com/rust-lang/rust/issues/137640 is closed as completed, but I'm still seeing the crash n nightly. It might be worth opening a separate issue.

ObserverOfTime (approved):

PR #4253, opened 2025-03-03
Bump wasmtime version to support iOS build

zhuhaow:
wasmtime 30 add support for building for iOS, see https://github.com/bytecodealliance/wasmtime/releases/tag/v30.0.0 and https://github.com/bytecodealliance/wasmtime/pull/9888

This enables tree-sitter to be built for iphone and iPad.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,22 +3,19 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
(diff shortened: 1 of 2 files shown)


amaanq:
Can you explain the second commit? I'd think if you're pulling it in from crates.io it should use the published version.

clason:
superseded by #4465

PR #4255, opened 2025-03-03
fix symbol leak for character sets

### LINKED ISSUE #4209: Unable to link multiple grammars with same rule since 0.25.1
### Problem

I'm updating my tree-sitter parsers to version 0.25.1, but I'm having problems linking the updated parsers into my Rust application.

The issue seems to occur because I am using a subset of one parser in another. This makes it that both parsers include
```c
const TSCharacterRange sym_identifier_character_set_1[] = {
  {0, 0x08}, {0x0e, 0x1f}, {'#', '#'}, {'0', '9'}, {'?', 'Z'}, {'\\', '\\'}, {'_', 'z'}, {0x7f, 0x218f},
  {0x2191, 0x2191}, {0x2193, 0x21cf}, {0x21d1, 0x21d1}, {0x21d3, 0x21d3}, {0x21d5, 0x2207}, {0x2209, 0x2215}, {0x2217, 0x221d}, {0x221f, 0x2226},
  {0x222b, 0x225f}, {0x2261, 0x2263}, {0x2266, 0x2285}, {0x2288, 0x22ba}, {0x22bc, 0x27f6}, {0x27f8, 0x10ffff},
};
```
When compiling on Ubuntu, the linker complains about this conflict since it find the symbol `sym_identifier_character_set_1` twice.

Although I'm using the one parser as a dependency of the other, I believe the same issue would occur if you just have any two grammars that have rules with the same name that generate one of these `character_set`'s.

The difference between the new generated parser seems to be that `sym_identifier_character_set_1` went from a `static` to a `const`.

**Some additional context**
- [Error in GitHub actions](https://github.com/shackle-rs/shackle/actions/runs/13363896308/job/37317994965)
- [Changes to first `parser.c`](https://github.com/shackle-rs/shackle/pull/245/files#diff-b72f6871848787c48d686d082d163ebc71a91fad896460fbd5a919d1a95a46de)
- [Changes to second `parser.c`](https://github.com/shackle-rs/shackle/pull/245/files#diff-9accbf313d7ba190137557bbc502dcdb49b2c83399bdb5934a8c1e28fbaaf25a)
- [Any other information (Update PR)](https://github.com/shackle-rs/shackle/pull/245)

### Steps to reproduce

1. Create two grammar packages that both include a rule with the same name, that is similar to:
```
identifier: ($) => {
	return new RegExp(`[^"'\\s\\.\\-\\[\\]\\^\\/${OPERATOR_CHARACTERS}]+`)
},
```
2. Create a Rust application that loads both grammars.
3. Compile on Ubuntu, and find a linker error.

### Expected behavior

Since the likelihood of two parsers having some shared rule is rather large, and since `tree-sitter` is expected to be used with multiple languages together, I would expect that the generated parser would not contain any conflict when linked together.

tree-sitter 0.25.1 (f5afe475deb7c0bae6407fb776c76824f717bb61)

### Operating system/version

`ubuntu-latest` on GitHub Actions

polazarus:
fix #4209

### DIFF
--- a/cli/generate/src/render.rs
+++ b/cli/generate/src/render.rs
@@ -1113,7 +1113,11 @@ impl Generator {
             return;
         }
 
-        add_line!(self, "const TSCharacterRange {}[] = {{", info.constant_name);
+        add_line!(
+            self,
+            "static const TSCharacterRange {}[] = {{",
+            info.constant_name
+        );
 
         indent!(self);
         for (ix, range) in characters.ranges().enumerate() {


maxbrunsfeld:
Looks like we need to run `cargo fmt`

PR #4258, opened 2025-03-04
Update build.zig.zon for zig master/0.14

nihklas:
In a recently landed change (https://github.com/ziglang/zig/pull/22994), the structure of the `build.zig.zon` file changed. This PR fixes the structure to be compatible with the newest master build and coming 0.14 release.


clason:
Note that `master` should report as `0.26.0`; that should be fixed before other build file changes.

PR #4259, opened 2025-03-04
fix(build): bump version to 0.26.0

clason:
After a release, `master` branch should be bumped to the next _minor_ version.

(`cargo xtask bump-version` seems broken since it doesn't respect the `-v` flag for the `Workspace` entry in `Cargo.toml`.)


clason:
Looks like `cargo xtask bump-version` is broken, period. Not going to waste time on this.

PR #4263, opened 2025-03-04
fix(docs): update test command flag from -f to -i

jonshea:
The `-f` flag on `tree-sitter test` seems to have been removed or replaced with a `-i, --include` flag. This PR updates the "Writing Tests" page to reflect this change.

### DIFF
--- a/docs/src/creating-parsers/5-writing-tests.md
+++ b/docs/src/creating-parsers/5-writing-tests.md
@@ -80,7 +80,7 @@ By default, the `tree-sitter test` command runs all the tests in your `test/corp
 can use the `-f` flag:
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/src/creating-parsers/5-writing-tests.md
```diff
@@ -80,7 +80,7 @@ By default, the `tree-sitter test` command runs all the tests in your `test/corp
 can use the `-f` flag:
```
WillLillis, line 80:
```suggestion
can use the `-i` flag:
```

Could you also update the flag here?

jonshea: Done. Thank you for catching that.

WillLillis (approved):

PR #4276, opened 2025-03-11
fix(cli): properly escape invisible characters in parse error output

WillLillis:
This just applies the existing `escape_invisible` logic to the error output of the `parse` command. 

Issue originally raised in the community chat.

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -686,19 +686,23 @@ pub fn parse_file_at_path(
             if let Some(node) = first_error {
                 let start = node.start_position();
                 let end = node.end_position();
+                let mut node_text = String::new();
+                for c in node.kind().chars() {
+                    if let Some(escaped) = escape_invisible(c) {
+                        node_text += escaped;
+                    } else {
+                        node_text.push(c);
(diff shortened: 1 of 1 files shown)


cli/src/parse.rs
```diff
@@ -686,19 +686,23 @@ pub fn parse_file_at_path(
...
                     if node.is_named() {
-                        write!(&mut stdout, "MISSING {}", node.kind())?;
+                        write!(&mut stdout, "MISSING {node_text}")?;
                     } else {
-                        write!(
-                            &mut stdout,
-                            "MISSING \"{}\"",
-                            node.kind().replace('\n', "\\n")
-                        )?;
+                        write!(&mut stdout, "MISSING \"{node_text}\"",)?;
```
amaanq, line 702: small nit

```suggestion
                        write!(&mut stdout, "MISSING \"{node_text}\"")?;
```

amaanq (approved):

cli/src/parse.rs
WillLillis: oops, nice catch

PR #4283, opened 2025-03-12
Change `(Node | null)[]` return types to `Node[]`

tmr232:
In the rewrite for web-tree-sitter 0.25, some functions that used to return `Node[]` were changed to `(Node | null)[]`.
This makes using them tricky, as it requires filtering out the `null` values.

But from what I gather - they don't actually return `null` values now either.

Given the way the [C part is written](https://github.com/tree-sitter/tree-sitter/blob/07a1b7fc003f8a896f74a03805498a5ca0a8d4d0/lib/binding_web/lib/tree-sitter.c#L680), if I understand correctly, those functions should always return a `Node[]`, and have no `null` elements.

This PR adds non-null assertions to change those type signatures and make using the functions the same as with previous versions.

### DIFF
--- a/lib/binding_web/src/node.ts
+++ b/lib/binding_web/src/node.ts
@@ -12,10 +12,10 @@ export class Node {
   private [0] = 0; // Internal handle for WASM
 
   /** @internal */
-  private _children?: (Node | null)[];
+  private _children?: Node[];
 
   /** @internal */
-  private _namedChildren?: (Node | null)[];
+  private _namedChildren?: Node[];
 
   /** @internal */
   constructor(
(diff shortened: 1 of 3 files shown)


maxbrunsfeld (approved):
Thank you!

maxbrunsfeld:
It looks like some type assertions in the tests are no longer necessary. Could you remove those, so that the lint passes for the tests?

tmr232:
@maxbrunsfeld Done. `npm run lint` now passes on my machine.

amaanq:
thanks @tmr232!

PR #4284, opened 2025-03-12
fix(make): `make install` shouldn’t fail when a parser bundles no queries

mavit:
Relates to https://github.com/uyha/tree-sitter-cmake/issues/34#issuecomment-2714724883


uyha:
> Arguably, a parser _should_ bundle queries, so this _should_ fail...

eh, is there a standard for the queries yet? I know that some editors adopt neovim’s query convention but is there anything from tree-sitter itself?

clason:
The thing that is (not) standardized are capture names, but those are purely cosmetic and easy to adapt. The important thing is to have queries that 1. can be adapted instead of everyone having to write from scratch and 2. more importantly, can be tested in CI to make sure breaking grammar changes don't slip through without being marked as such (and demonstrated how to adapt to).

uyha:
> The thing that is (not) standardized are capture names, but those are purely cosmetic and easy to adapt. The important thing is to have queries that 1. can be adapted instead of everyone having to write from scratch and 2. more importantly, can be tested in CI to make sure breaking grammar changes don't slip through without being marked as such (and demonstrated how to adapt to).

makes sense, I still steal the queries from neovim then.

clason:
Doesn't mean we shouldn't fix this, too.

ObserverOfTime (approved):

PR #4285, opened 2025-03-12
feat(lib): expose large state count for languages

clason:
Problem: Unable to check large state count for languages (e.g., for ~~shaming~~ regression tests).

Solution: Add `ts_language_large_state_count` to API and bindings.


lib/include/tree_sitter/api.h
```diff
@@ -1207,6 +1207,15 @@ uint32_t ts_language_symbol_count(const TSLanguage *self);
 */
 uint32_t ts_language_state_count(const TSLanguage *self);
 
+/**
+ * Get the number of large states in this language.
+ *
+ * ("Large states" are states with a number of terminal states above a certain
+ * threshold, which don't allow an optimized representation. Thus this number
+ * correlates with parser complexity and performance.)
+ */
```
clason, line 1216: Is that explanation correct? There's no real explanation of this stat anywhere in the documentation.

Sab010401: Hallo

lib/binding_rust/lib.rs
```diff
@@ -474,6 +474,13 @@ impl Language {
         unsafe { ffi::ts_language_state_count(self.0) as usize }
     }
 
+    /// Get the number of large states in this language.
+    #[doc(alias = "ts_language_large_state_count")]
+    #[must_use]
+    pub fn parse_large_state_count(&self) -> usize {
```
WillLillis, line 480: Super picky nit but should this be parser_large_state_count? Parse is used as a verb a lot in the rest of the lib.

clason: I mirrored the `parse_state_count` above. I agree with you, but I think consistency here is more important (and changing the other function would be a breaking change without sufficient reason...)

WillLillis: Ah gotcha, that makes sense.

WillLillis (approved):

clason:
@amaanq could you take a look at whether this is desirable (and the explanation correct)?

amaanq:
My only concern with this is that it's an internal number and isn't meant for public consumption..if it's *only* used for regression tests why not just grep `parser.c`?

clason:
Well, my use case is from nvim-treesitter (or ts_query_ls), where I only have the compiled libraries available.

I am with you regarding the iffyness of exposing internals -- but then, doesn't the same apply to the regular `state_count`? What's the purpose of exposing that?

(Not trying to be argumentative; I genuinely want to understand the thinking behind this.)

amaanq:
The actual state count is useful, you can use the parse state (or next parse state) of a given node to create a lookahead iterator, and know what symbols are valid for that given state. That state count represents something meaningful in the generated parser, vs the large state count being a number used for determining which state table to perform look-ups in.

My main concern is a user might see this function in the API and wonder "what is this used for", and the answer is, really, nothing, unless you want to perform regression tests.

For neovim/nvim-treesitter, you could just directly access the language struct's `large_state_count` field (offset of 24 bytes), that won't change.

maxbrunsfeld:
I agree, this is kinda exposing implementation details.

PR #4299, opened 2025-03-19
fix(xtask): using reqwest instead of ureq to fix #4295

### LINKED ISSUE #4295: Consider changing dependency on `ureq`
### Problem

### Expected behavior

xiaoma20082008:
fix #4295 by using [reqwest](https://github.com/seanmonstar/reqwest) that which is based on `system-native TLS`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.0"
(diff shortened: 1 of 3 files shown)


clason:
Thank you, but we don't want to pull in dozens of new dependencies for a simple task that updates a single number. A better fix would be to rewrite the task itself to use simple `git` calls.

ObserverOfTime:
That would be inefficient as it would require cloning the repository.

clason:
As `ureq` is now an internal dependency, this has become relevant. But I'm still not happy with the inflated dependency chain -- isn't there are more light-weight alternative? We don't use it for much...

clason:
Hmm, `reqwest` looks like the go-to HTTPS crate for Rust; everything else seems to be implemented on top of that. So I guess this is fine **but** can you please
1. rebase on master
2. remove the xtask
3. and switch the _new_ `ureq` call to `reqwest`?

clason:
Superseded by https://github.com/tree-sitter/tree-sitter/pull/4396, but thanks!

PR #4301, opened 2025-03-21
build.zig.zon: Fix package hashes for Zig 0.14

JacobCrabill:
## Summary

Zig 0.14 changed how package hashes are computed and used, and if the old package hashes are left, every call to `zig build` will re-download every dependency every time (this also occurs with downstream packages that include tree-sitter in _their_ `build.zig.zon`).  Updating to the new hash format solves this.

##  Verification

To verify each package hash for yourself (as once a package is downloaded only the hash will be used to reference it), you can use the `zig fetch` command:

```
$ zig fetch https://github.com/bytecodealliance/wasmtime/releases/download/v29.0.1/wasmtime-v29.0.1-aarch64-android-c-api.tar.xz
N-V-__8AAC3KCQZMd5ea2CkcbjldaVqCT7BT_9_rLMId6V__
```

After downloading the tarball it will compute and print the package hash.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -13,57 +13,57 @@
   .dependencies = .{
     .wasmtime_c_api_aarch64_android = .{
       .url = "https://github.com/bytecodealliance/wasmtime/releases/download/v29.0.1/wasmtime-v29.0.1-aarch64-android-c-api.tar.xz",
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
Thanks!

PR #4303, opened 2025-03-21
fix(rust): use core crates for no_std

### LINKED ISSUE #4302: Tree sitter 0.25 does not compile with no_std ("std" feature disabled)
### Problem

It seems that when bumping dependencies to 0.25 there are references to std in the lib.rs:
```
Compiling tree-sitter v0.25.3
error[E0433]: failed to resolve: use of undeclared crate or module `std`
    --> /home/simon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tree-sitter-0.25.3/binding_rust/lib.rs:1225:38
     |
1225 |             let (c, len) = D::decode(std::slice::from_raw_parts(data, len as usize));
     |                                      ^^^ use of undeclared crate or module `std`
     |
help: consider importing one of these modules
     |
10   + use crate::slice;
     |
10   + use core::slice;
     |
10   + use alloc::slice;
     |
help: if you import `slice`, refer to it directly
     |
1225 -             let (c, len) = D::decode(std::slice::from_raw_parts(data, len as usize));
1225 +             let (c, len) = D::decode(slice::from_raw_parts(data, len as usize));
     |

error[E0433]: failed to resolve: use of undeclared crate or module `std`
    --> /home/simon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tree-sitter-0.25.3/binding_rust/lib.rs:1425:17
     |
1425 |                 std::ptr::from_ref::<AtomicUsize>(flag).cast::<usize>(),
     |                 ^^^ use of undeclared crate or module `std`
     |
help: consider importing this module
     |
10   + use core::ptr;
     |
help: if you import `ptr`, refer to it directly
     |
1425 -                 std::ptr::from_ref::<AtomicUsize>(flag).cast::<usize>(),
1425 +                 ptr::from_ref::<AtomicUsize>(flag).cast::<usize>(),
     |

error[E0599]: no method named `stop_printing_dot_graphs` found for mutable reference `&mut Parser` in the current scope
    --> /home/simon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tree-sitter-0.25.3/binding_rust/lib.rs:1435:14
     |
1435 |         self.stop_printing_dot_graphs();
     |              ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `&mut Parser`

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `tree-sitter` (lib) due to 3 previous errors
```

### Steps to reproduce

Update tree_sitter to 0.25 in Cargo.toml in a no_std project:
```
[dependencies]
tree-sitter = { version = "0.25", default-features = false, features = [] }
```

### Expected behavior

Builds without std

0.25.3

### Operating system/version

Arch Linux

Tiggilyboo:
Fixes #4302 

Adds a feature flag to the binding lib, used core where possible. Maybe worth a no_std test case?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -238,7 +238,6 @@ jobs:
       if: ${{ !matrix.no-run && !matrix.use-cross && inputs.run-test }}
       run: $BUILD_CMD run -p xtask -- test-wasm
 
-
     - name: Upload CLI artifact
       uses: actions/upload-artifact@v4
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -203,6 +203,13 @@ jobs:
         npm run build
         npm run build:debug
 
+    - name: Check no_std builds
+      if: ${{ !matrix.no-run && inputs.run-test }}
+      shell: bash
+      run: |
+        cd lib
+        $BUILD_CMD check --no-default-features
+
     - name: Build target
       run: $BUILD_CMD build --release --target=${{ matrix.target }} --features=${{ matrix.features }}
 
@@ -230,6 +237,7 @@ jobs:
     - name: Run wasm tests
       if: ${{ !matrix.no-run && !matrix.use-cross && inputs.run-test }}
       run: $BUILD_CMD run -p xtask -- test-wasm
+
 
     - name: Upload CLI artifact
       uses: actions/upload-artifact@v4
--- a/lib/binding_rust/ffi.rs
+++ b/lib/binding_rust/ffi.rs
@@ -16,6 +16,7 @@ extern "C" {
 }
 
 #[cfg(windows)]
+#[cfg(feature = "std")]
 extern "C" {
     pub(crate) fn _ts_dup(handle: *mut std::os::raw::c_void) -> std::os::raw::c_int;
 }
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -1222,7 +1222,7 @@ impl Parser {
             len: u32,
             code_point: *mut i32,
         ) -> u32 {
-            let (c, len) = D::decode(std::slice::from_raw_parts(data, len as usize));
+            let (c, len) = D::decode(core::slice::from_raw_parts(data, len as usize));
             if let Some(code_point) = code_point.as_mut() {
                 *code_point = c;
             }
@@ -1422,7 +1422,7 @@ impl Parser {
         if let Some(flag) = flag {
             ffi::ts_parser_set_cancellation_flag(
                 self.0.as_ptr(),
-                std::ptr::from_ref::<AtomicUsize>(flag).cast::<usize>(),
+                core::ptr::from_ref::<AtomicUsize>(flag).cast::<usize>(),
             );
         } else {
             ffi::ts_parser_set_cancellation_flag(self.0.as_ptr(), ptr::null());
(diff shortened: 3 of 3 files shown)


WillLillis:
I tried applying your patch locally but still got a compile error related to the `Drop` implementation for `Parser`:

```
error[E0599]: no method named `stop_printing_dot_graphs` found
    --> /home/lillis/projects/tree-sitter/lib/binding_rust/lib
.rs:1435:14
     |
1435 |         self.stop_printing_dot_graphs();
     |              ^^^^^^^^^^^^^^^^^^^^^^^^ method not found 
in `&mut Parser`

For more information about this error, try `rustc --explain E0
599`.
error: could not compile `tree-sitter` (lib) due to 1 previous
 error
```

It looks like locking the call to `stop_printing_dot_graphs` behind the `std` feature fixes things:

```diff
diff --git a/lib/binding_rust/lib.rs b/lib/binding_rust/lib.rs
index 62839d9b..8577fb1d 100644
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -1432,7 +1432,10 @@ impl Parser {

 impl Drop for Parser {
     fn drop(&mut self) {
-        self.stop_printing_dot_graphs();
+        #[cfg(feature = "std")]
+        {
+            self.stop_printing_dot_graphs();
+        }
         self.set_logger(None);
         unsafe { ffi::ts_parser_delete(self.0.as_ptr()) }
     }
```

> Maybe worth a no_std test case?

A new test case would be welcome, but could also be taken care of in a followup.

clason:
> A new test case would be welcome, but could also be taken care of in a followup.

I would strongly encourage adding a test _now_. Otherwise it'll likely never happen and things can easily backslide. (In fact, I would _start_ with the test so you know which changes to make to have it work.)

Tiggilyboo:
I've pushed the above fix for the call to stop_printing_dot_graphs, thanks for that - however I'm unsure of how go about setting up the no_std test case. The repo has a Dockerfile set up to test with `CMD cargo test --all-features`, could it be as easy as just running `cargo test --no-default-features` afterward?

Just looking for some direction on how this should be acheived, cheers!

WillLillis:
`cargo test --no-default-features` still works for me locally without this patch, so I don't think that's the way to go. What about something like this added to `.github/workflows/build.yml`?

```yml
    - name: Check no_std
      if: ${{ !matrix.no-run && inputs.run-test }}
      run: |
        cd ..
        cargo new no_std_check
        cd no_std_check
        cat "tree-sitter = { path = \"../tree-sitter/lib/\", default-features = false, features = [] }" >> Cargo.toml
        cargo build
```

CC @ObserverOfTime?

ObserverOfTime:
What about `cargo build --no-default-features` ?

WillLillis:
> What about `cargo build --no-default-features` ?

That succeeds when it's run from the root directory, but fails with the expected errors when I run it inside `lib/`. I'm not sure why it succeeds in the former case, but that should work regardless, thanks!

@Tiggilyboo do you mind adding the CI changes, or would you prefer for me to push to your branch?

Tiggilyboo:
Sorry for the delay, I've added the build check in the workflow. Let me know if anything looks off

.github/workflows/build.yml
```diff
@@ -231,6 +238,7 @@ jobs:
       if: ${{ !matrix.no-run && !matrix.use-cross && inputs.run-test }}
       run: $BUILD_CMD run -p xtask -- test-wasm
 
+
```
WillLillis, line 241:
```suggestion
```

WillLillis (approved):
Besides the extra empty line, everything LGTM. Thanks!

PR #4305, opened 2025-03-23
fix(rust): address new clippy lint for pointer comparisons

WillLillis:
Uses ` core::ptr::eq` rather than the suggested  `std::ptr::eq` to maintain `no_std` compatibility.


clason (approved):

clason:
Prefer `fix` or `refactor` over `chore`.

clason:
CodeQL complains about two more instances of raw pointer comparison -- do you want to include those here as well?

clason:
Nevermind, probably just needs a rebase to remove those.

PR #4319, opened 2025-03-29
Fix pkgconfig and use of GNUInstallDirs

dg0yt:
Ensure initialization of `CMAKE_INSTALL_INCLUDEDIR` by including `GNUInstallDirs` before pc file generation.  
Install the pc file to `CMAKE_INSTALL_LIBDIR/pkgconfig`. It contains the *architecture-dependent* path to the library.


dg0yt:
- Fixes includedir being just prefix.
- Same patterns can be found in grammar repos.

amaanq (approved):

dg0yt:
So there is a new release, and this is still missing.

dg0yt:
> Can you also change `cli/src/templates/cmakelists.cmake` accordingly?

Done. Moving `GNUInstallDirs` further up because the variables is also used for the CMake install interface.

PR #4323, opened 2025-03-31
fix(rust): clarify error message for non-token reserved words

jonshea:
Improve the `NonTokenReservedWord` error message by including the specific reserved word that was not used as a token. The Updated message also clarifies that reserved words must be defined as tokens appearing in the grammar rules.

### DIFF
--- a/cli/generate/src/prepare_grammar/extract_tokens.rs
+++ b/cli/generate/src/prepare_grammar/extract_tokens.rs
@@ -28,8 +28,8 @@ unless they are used only as the grammar's start rule.
     NonSymbolExternalToken,
     #[error("Non-terminal symbol '{0}' cannot be used as the word token, because its rule is duplicated in '{1}'")]
     NonTerminalWordToken(String, String),
-    #[error("Reserved words must be tokens")]
-    NonTokenReservedWord,
+    #[error("Reserved word '{0}' must be a token")]
+    NonTokenReservedWord(String),
 }
 
 pub(super) fn extract_tokens(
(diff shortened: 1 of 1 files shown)


jonshea:
I’d also be happy to change this error to a warning. I can’t think of any reason it needs to be an error instead of a warning, but maybe I am missing something.

cli/generate/src/prepare_grammar/extract_tokens.rs
```diff
@@ -28,8 +28,8 @@ unless they are used only as the grammar's start rule.
     NonSymbolExternalToken,
     #[error("Non-terminal symbol '{0}' cannot be used as the word token, because its rule is duplicated in '{1}'")]
     NonTerminalWordToken(String, String),
-    #[error("Reserved words must be tokens")]
-    NonTokenReservedWord,
+    #[error("Reserved word '{0}' must be also be a token in the grammar rules")]
```
WillLillis, line 31:
```suggestion
    #[error("Reserved word '{0}' must be a token")]
```

WillLillis (approved):

WillLillis:
This is a great improvement to the existing error message. For a future improvement, it might be nice to cover the other rule types. I came up with this first pass, but it needs some more testing/cleanup and I don't want that to hold up this change :)

```rust
...
let rule = match &reserved_rule {
    Rule::String(s) | Rule::Pattern(s, _) | Rule::NamedSymbol(s) => {
        format!("'{s}'")
    }
    Rule::Blank => "`blank`".to_string(),
    Rule::Choice(c) => {
        if c.len() == 2 && matches!((&c[0], &c[1]), (Rule::Repeat(_), Rule::Blank))
        {
            "`repeat`".to_string()
        } else if c.len() == 2 && matches!(&c[1], Rule::Blank) {
            "`optional`".to_string()
        } else {
            "`choice`".to_string()
        }
    }
    Rule::Repeat(_) => "`repeat1`".to_string(),
    Rule::Seq(_) => "`seq`".to_string(),
    Rule::Metadata { params, .. } => {
        if params.alias.is_some() {
            "`alias`".to_string()
        } else if params.field_name.is_some() {
            "`field`".to_string()
        } else if params.associativity.is_some()
            || params.precedence != Precedence::None
            || params.dynamic_precedence != 0
        {
            "`prec`".to_string()
        } else if params.is_token {
            "`token`".to_string()
        } else {
            "<unexpected rule>".to_string() // unreachable?
        }
    }
    // unreachable as well?
    Rule::Symbol(_) | Rule::Reserved { .. } => "<unexpected rule>".to_string(),
};
Err(ExtractTokensError::NonTokenReservedWord(rule))?;
...
```

PR #4326, opened 2025-04-02
reference additional github tag query kwds

sielicki:


sielicki:
I mean, all tag queries are useless unless you have some additional tooling which is going to consume them. By what criteria are you claiming that the current table of arbitrary keywords is relevant to tree-sitter itself, but that this expanded set isn't?

clason:
Because that set is documented as a "standard set" that _every_ implementation is expected to support? So the audience here is writers of tooling.

sielicki:
Sure -- and this just documents an expanded set of tags that are relevant to github, relevant to the same exact audience.

clason:
Why would any other consumer care about what Github implements?

sielicki:
They live in the same tags.scm file for the grammar -- anyone using that grammar can benefit from the expanded vocabulary. If tree-sitter-c starts defining `@definition.macro` in its queries/tags.scm file, I can go write a quick program with [tree-sitter-tags](https://crates.io/crates/tree-sitter-tags) that iterates over all the macro definitions in a given project.

how does it make the project better to refuse to document that this expanded vocabulary of types exist?

clason:
> But they are, because they live in the same grammar!

No they don't. Downstream rarely uses bundled queries; they are mostly used for testing and as a blueprint for the _actual_ queries used in projects.

sielicki:
maybe because the people maintaining the upstream grammar are difficult to deal with? ;)

clason:
You might very well think that; I couldn't possibly comment. (I _am_ a downstream maintainer.)

PR #4327, opened 2025-04-03
Make highlighting more deterministic when themes are ambiguous

### LINKED ISSUE #1758: Non deterministic highlighting
Hey hey, I've been working on improving the tree-sitter highlighting queries for solidityhttps://github.com/JoranHonig/tree-sitter-solidity.

However, I'm encountering some weird behavior. The highlighting seems to be non deterministic, picking one or another color between every execution of `tree-sitter highlight`.

In addition to that it seems like not all configuration parameters are being used. For example, the underlining option does not seem to be working for me.

I tried and the behavior seems consistent between my ubuntu and mac machines.

narpfel:
Resolves #1758 (see that issue for a reproducer).

### DIFF
--- a/cli/src/highlight.rs
+++ b/cli/src/highlight.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashSet,
     fmt::Write,
     fs,
     io::{self, Write as _},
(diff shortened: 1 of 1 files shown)


WillLillis (approved):

WillLillis:
Thanks for the reproducer and fix!

PR #4328, opened 2025-04-04
docs: Fix link to named-vs-anonymous-nodes section

ZedThree:
Link in this admonition is broken: https://tree-sitter.github.io/tree-sitter/creating-parsers/5-writing-tests.html#admonition-tip

> The S-expression does not show syntax nodes like func, ( and ;, which are expressed as strings and regexes in the grammar. It only shows the named nodes, as described in [this section][named-vs-anonymous-nodes] of the page on parser usage.


WillLillis (approved):

WillLillis:
Thanks!

PR #4333, opened 2025-04-06
fix(highlight): account for multiple rows in highlight testing assertions

### LINKED ISSUE #4330: test regression with arrows
### Problem

The PR https://github.com/tree-sitter/tree-sitter/pull/3927 introduced a [regression in a tree-sitter-vim test](https://github.com/tree-sitter-grammars/tree-sitter-vim/issues/32#issuecomment-2780692207); specifically
```vim
echo
"\ lorem ipsum
" <- @comment
```
fails with
```
Failure - row: 1, column: 0, expected highlight 'comment', actual highlights: none.
```
But `tree-sitter hi` highlights this correctly with the given queries (as does Neovim).

Similar issue in https://github.com/tree-sitter-grammars/tree-sitter-lua/issues/53 (although that test is actually wrong, nevertheless the failure is new).

### Steps to reproduce

```sh
git clone https://github.com/tree-sitter-grammars/tree-sitter-vim
cd tree-sitter-vim
tree-sitter t
```

### Expected behavior

Test passes, as it did prior to the linked PR.

tree-sitter 0.25.3 (and HEAD)

### Operating system/version

macOS 15.5

WillLillis:
This highlight test case is currently broken:

```vim
echo
"\ lorem ipsum
" <- @comment
```

I believe the issue is that the `line_continuation_comment` node starts at position [0,4], rather than [1,0]. This starting position causes [this](https://github.com/tree-sitter/tree-sitter/blob/521da2b0a7e8815efbcf1e4b98e5ac16113c5763/cli/src/test_highlight.rs#L175) if block to incorrectly enter. Even though the highlight's *column* is past the position's end column, the highlight starts a row earlier. The fix is to just make sure the highlight's row is considered in addition to the column.

Bonus commit: Noticed this while debugging this issue. When zero test cases are parsed, "NaN%" is displayed for the  success percentage and "0 bytes/ms" is shown for the parse rate. When no (measured) parsing took place, we don't have a valid success or parse rates to display. N/A seems more correct in this case.

cc @clason 

- Closes #4330

### DIFF
--- a/cli/src/parse.rs
+++ b/cli/src/parse.rs
@@ -29,18 +29,28 @@ pub struct Stats {
 impl fmt::Display for Stats {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let duration_us = self.total_duration.as_micros();
+        let success_rate = if self.total_parses > 0 {
+            format!(
+                "{:.2}%",
+                ((self.successful_parses as f64) / (self.total_parses as f64)) * 100.0,
+            )
+        } else {
+            "N/A".to_string()
+        };
+        let duration_str = match (self.total_parses, duration_us) {
+            (0, _) => "N/A".to_string(),
+            (_, 0) => "0 bytes/ms".to_string(),
+            (_, _) => format!(
+                "{} bytes/ms",
+                ((self.total_bytes as u128) * 1_000) / duration_us
+            ),
+        };
         writeln!(
             f,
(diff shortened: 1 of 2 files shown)


clason:
Yeah, that seems like a different error? (Possibly the arrow tests are more broadly bugged.)

clason:
Ah, no, I see now why the test should pass -- it's explicitly testing that the node extends across lines. (Bit of a weird test, tho.)

PR #4355, opened 2025-04-12
fix(lib): reset parser options after use

ribru17:
**Problem:** After `ts_parser_parse_with_options()`, the parser options are still stored in the parser object, meaning that a successive call to `ts_parser_parse()` will actually behave like
`ts_parser_parse_with_options()`, which is not obvious and can have unintended consequences.

**Solution:** Reset to empty options state after
`ts_parser_parse_with_options()`.

See https://github.com/neovim/neovim/pull/33437 for real life example

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1540,7 +1540,7 @@ static bool ts_parser__check_progress(TSParser *self, Subtree *lookahead, const 
   if (self->operation_count >= OP_COUNT_PER_PARSER_TIMEOUT_CHECK) {
     self->operation_count = 0;
   }
-  if (self->parse_options.progress_callback && position != NULL) {
+  if (position != NULL) {
     self->parse_state.current_byte_offset = *position;
     self->parse_state.has_error = self->has_error;
   }
@@ -2238,6 +2238,8 @@ TSTree *ts_parser_parse_with_options(
   self->parse_options = parse_options;
   self->parse_state.payload = parse_options.payload;
   TSTree *result = ts_parser_parse(self, old_tree, input);
+  // Reset parser options before further parse calls.
+  self->parse_options = (TSParseOptions) {0};
   return result;
 }
 


ribru17:
A question based on the way the failing tests are set up; is this behavior actually intended?

WillLillis:
> A question based on the way the failing tests are set up; is this behavior actually intended?

Reading through the failing test, I believe it is intended for state to persist between calls to `ts_parse_with_options`:
- An initial parse is done with a callback that ensures it exits parsing while balancing is in progress
- The parser is reset.
- A fresh parse is started with the same callback as the original (intending for parsing to stop during balancing) 
- `parser.parse_with_options()` is called *again*, with the assumption that the parsing state was preserved during calls.

That being said, as far as I can tell there is no intention (or reason to) preserve the progress callback between calls. That *does* seem like a footgun.  I tried the following change to this patch with the tree-sitter test suite as well as the neovim [reproducer](https://github.com/neovim/neovim/issues/33277#issuecomment-2774470950) and both are running clean:

```diff
diff --git a/lib/src/parser.c b/lib/src/parser.c
index 71cf7bdb..49c4d440 100644
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -2238,9 +2238,8 @@ TSTree *ts_parser_parse_with_options(
   self->parse_options = parse_options;
   self->parse_state.payload = parse_options.payload;
   TSTree *result = ts_parser_parse(self, old_tree, input);
-  // Reset parser state & options before further parse calls.
-  self->parse_options = (TSParseOptions) {0};
-  self->parse_state = (TSParseState) {0};
+  // Reset callback before further parse calls
+  self->parse_options.progress_callback = NULL;
   return result;
 }
```

(Please double check me though! It's easy to make mistakes this late at night :sweat_smile:)

Reading through the [original PR](https://github.com/tree-sitter/tree-sitter/pull/3843) that added `ts_parser_parse_with_options`, I think it probably makes sense to reset the callback in `ts_query_cursor_exec_with_options` as well. However it's currently marked as `const`, preventing such a reassignment directly:

https://github.com/tree-sitter/tree-sitter/blob/b3410731929d57d2eb8e20dbfae90ef3deb234b5/lib/src/query.c#L314-L336

I'm not sure what the original motivation behind this is. Probably safe to remove?

ribru17:
Nice! Thanks, that makes a lot of sense. So the patch you gave does work, or it still fails the neovim reproducer? Also, if we clear callbacks, should the patch be changed to the following?

```patch
diff --git a/lib/src/parser.c b/lib/src/parser.c
index adcdf8d0..0429ce9a 100644
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1540,7 +1540,7 @@ static bool ts_parser__check_progress(TSParser *self, Subtree *lookahead, const
   if (self->operation_count >= OP_COUNT_PER_PARSER_TIMEOUT_CHECK) {
     self->operation_count = 0;
   }
-  if (self->parse_options.progress_callback && position != NULL) {
+  if (position != NULL) {
     self->parse_state.current_byte_offset = *position;
     self->parse_state.has_error = self->has_error;
   }
@@ -2238,6 +2238,7 @@ TSTree *ts_parser_parse_with_options(
   self->parse_options = parse_options;
   self->parse_state.payload = parse_options.payload;
   TSTree *result = ts_parser_parse(self, old_tree, input);
+  self->parse_options = (TSParseOptions) {0};
   return result;
 }
 ```

that way the parser state is kept consistent even when parsing without a callback (not sure if that is strictly necessary)

WillLillis:
> So the patch you gave does work, or it still fails the neovim reproducer?

I believe it still fails the neovim reproducer. I was swapping between testing changes to neovim and tree-sitter locally, and I thought the reproducer passed (running `make clean` in between neovim builds), but I had a suspicion something was getting cached incorrectly. After re-cloning neovim to ensure a fully clean build with the changes the reproducer failed.

> Also, if we clear callbacks, should the patch be changed to the following?

That seems correct, but I'm not 100% sure. I'll do some more testing locally to check it out :)

clason:
You need to `make distclean` (or manually delete the `stamp` file) to rebuild the dependencies in Neovim.

WillLillis:
@ribru17 your latest patch is working for me locally! Mind pushing them up?

WillLillis (approved):

PR #4358, opened 2025-04-13
feat(cli): add argument `grammar-path` to `test` subcommand

WillLillis:
This allows users to run a grammar's test suite without being inside the directory for said grammar.


cli/src/main.rs
```diff
@@ -273,6 +273,9 @@ struct Test {
     /// The path to an alternative config.json file
     #[arg(long)]
     pub config_path: Option<PathBuf>,
+    /// The path to the tree-sitter grammar directory
+    #[arg(long)]
```
ObserverOfTime, line 277:
```suggestion
    #[arg(long, short = 'p')]
```

WillLillis:
> This is a late request, but is it doable to get this working for the `parse` subcommand as well?

Good point!

image

amaanq (approved):

sogaiu:
It may be pilot error on my end (or perhaps a misunderstanding as to how things should work), but I'm not having much luck using `--grammar-path` with the `parse` subcommand using ac13c86675c3ad131270dc135d1effac01ae09f9.  At least, when I try to switch between two different paths that should yield different results, it doesn't seem to be working.

My expectation is that by using different values for `--grammar-path` I ought to be able to get different parsers to be used (e.g. via two grammar directories at different commits, say).

Is that an appropriate interpretation of what should be possible with this PR?

I looked at [these lines](https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L874-L1063) and I don't see any mention of `grammar_path` (unlike for say for `generate` [here](https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L789)).

Not sure if I'm reading things correctly (or even if I'm looking in the right places (^^;).

Any hints?

WillLillis:
> It may be pilot error on my end (or perhaps a misunderstanding as to how things should work), but I'm not having much luck using `--grammar-path` with the `parse` subcommand using [ac13c86](https://github.com/tree-sitter/tree-sitter/commit/ac13c86675c3ad131270dc135d1effac01ae09f9). At least, when I try to switch between two different paths that should yield different results, it doesn't seem to be working.
> 
> My expectation is that by using different values for `--grammar-path` I ought to be able to get different parsers to be used (e.g. via two grammar directories at different commits, say).
> 
> Is that an appropriate interpretation of what should be possible with this PR?

That sounds reasonable to me. I'm not sure why that wouldn't be working, taking a look now.

> I looked at [these lines](https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L874-L1063) and I don't see any mention of `grammar_path` (unlike for say for `generate` [here](https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L789)).
> 
> Not sure if I'm reading things correctly (or even if I'm looking in the right places (^^;).

The `grammar-path` argument to parse should get picked up here:

https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L1732-L1746

which is then passed the the `parse` command's `run` function a little below that [here](https://github.com/tree-sitter/tree-sitter/blob/ac13c86675c3ad131270dc135d1effac01ae09f9/cli/src/main.rs#L1758)

WillLillis:
Update: I was able to repro. 

If I move both copies of the repo out of the paths in ` ~/.config/tree-sitter/config.json` and delete the `.so` file in between `parse` invocations, I'm able to switch between the two using `--grammar-path`. I believe there's something going on with the loader's search order. This is definitely a use case that would be nice to support, so I'll see what I can do if it's not too invasive of a change.

sogaiu:
Thanks a lot for taking a look!

Good luck and please let me know if I can help with testing, etc.

PR #4366, opened 2025-04-17
fix(web): remove reference to node apis

### LINKED ISSUE #4320: web-tree-sitter uses a dynamic require inside an ES module
### Problem

web-tree-sitter uses a dynamic `require` (see https://github.com/tree-sitter/tree-sitter/blob/521da2b0a7e8815efbcf1e4b98e5ac16113c5763/lib/binding_web/src/language.ts#L265).
Since it is published as an ES module, dynamic requires are not allowed.

I noticed this when I was trying to test my code, that uses web-tree-sitter, using vitest.
It failed on `Language.load` and reported:

```
Error: Dynamic require of "fs/promises" is not supported
```

I created #4304 to address this, and then realized I should probably give more context.

### Steps to reproduce

To reproduce it in as neutral environment as I could think, to convince myself it was not an issue with my code, I took the tests from this repo and tried to run them against the published package. I got the same error.

I put the tests in a separate repo with the changes needed to run them: https://github.com/tmr232/testing-web-tree-sitter

The changes are:
- Replacing `../src` with `web-tree-sitter` in all imports
- Changing the package name in `package.json` so that it doesn't claim to be `web-tree-sitter`
- Adding `web-tree-sitter` as a dependency

After cloning the repo and running `npm install`, run `npm test` to reproduce the issue.
Note that fixing it won't make the tests pass, as the WASM files required are not there, but the `Error: Dynamic require of "fs/promises" is not supported` error should disappear.

### Expected behavior

Importing web-tree-sitter using `import` works and testing it in vitest works as well.

0.25.3

### Operating system/version

Windows 11

bennypowers:
Closes #4320

### DIFF
--- a/lib/binding_web/src/language.ts
+++ b/lib/binding_web/src/language.ts
@@ -259,23 +259,16 @@ export class Language {
     if (input instanceof Uint8Array) {
       bytes = Promise.resolve(input);
     } else {
-      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-      if (globalThis.process?.versions.node) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports
-        const fs: typeof import('fs/promises') = require('fs/promises');
-        bytes = fs.readFile(input);
-      } else {
-        bytes = fetch(input)
-          .then((response) => response.arrayBuffer()
-            .then((buffer) => {
-              if (response.ok) {
-                return new Uint8Array(buffer);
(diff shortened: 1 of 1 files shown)


amaanq:
Wouldn't this break electron consumers?

bennypowers:
> Wouldn't this break electron consumers?

Can we use export conditions for that?

Alternatively - publish electron-tree-sitter?

bennypowers:
This was fixed by #4380 for my use case (deno). Thanks

PR #4375, opened 2025-04-19
Highlight carriage return at EOF

thaliaarchi:
The branch which checks whether the previous character was a carriage return would not trigger at EOF, so a carriage return as the last character in the file was never highlighted. Although EOF can be considered a line ending, in a sense, CR+EOF shouldn't count as a CRLF line ending.


amaanq:
Nice, thank you!

amaanq (approved):

PR #4384, opened 2025-04-20
feat(rust): bindings for creating a raw, unchecked query pointer

ribru17:
This is useful for performing a more general query analysis using the rust bindings, without needing to worry about rust binding-specific validation.

Real world example: the query LSP needs something like this to be able to accurately validate query files which may not necessarily be consumed by the rust bindings


WillLillis (approved):

lib/binding_rust/lib.rs
```diff
@@ -2399,6 +2399,14 @@ impl Query {
...
     /// shared between multiple threads.
     pub fn new(language: &Language, source: &str) -> Result<Self, QueryError> {
+        let ptr = Self::new_ptr(language, source)?;
+        unsafe { Self::from_raw_parts(ptr, source) }
+    }
+
+    /// Constructs a raw [`TSQuery`] pointer without performing extra checks specific to the rust
+    /// bindings, like predicate validation and [`Query`] object construction. Useful for using the
+    /// rust bindings to perform more general query analysis.
+    pub fn new_ptr(language: &Language, source: &str) -> Result<*mut ffi::TSQuery, QueryError> {
```
amaanq, line 2409: I don't agree with creating a "new_ptr" function that returns a raw TSQuery - how would one clean this up? That's not immediately clear here, and returning a raw pointer here is useless on its own. I'd prefer a `validate` function to accomplish what you're looking for.

amaanq (changes requested):

lib/binding_rust/lib.rs
ribru17: > how would one clean this up

They would have to do it manually, same as if they called `Query::into_raw()`, that should probably be mentioned in the docstring

> returning a raw pointer here is useless on its own

I could be missing something, but this seems pretty useful at least with the current bindings; there isn't yet a way to construct a query using the rust bindings that, e.g., uses `#match?` with an invalid rust regex. That's usually fine unless someone wants to use to rust bindings to do a general query analysis

A validate function sounds interesting, I'm not exactly sure what that would look like here yet

amaanq: Ah I see, yeah, using `from_raw` would make sense, but I am still not a fan of returning a raw pointer. Can we instead have a `new_unchecked` function that does create a normal `Query`, but doesn't validate the predicates?

ribru17: Sure, as long as it's ok that the `general_predicates`, `capture_names`, etc. fields won't be populated (we can mention this in the docstring). Alternatively, we can graduate `TSQueryDrop` to reduce confusion

amaanq: Yeah making `TSQueryDrop` available for users would be fine. That sounds clearer too imo.

amaanq: I opted to instead rename the function `new_raw`, I think that makes the most sense. Thanks for the PR!

PR #4393, opened 2025-04-24
cli: Compile parsers to wasm using `wasi-sdk`, not emscripten

maxbrunsfeld:
The wasi-sdk is automatically downloaded if needed. It is newer and lighter-weight than emscripten. This change eliminates the need for docker/podman when building wasm parsers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2082,6 +2082,7 @@ dependencies = [
  "tree-sitter",
  "tree-sitter-highlight",
  "tree-sitter-tags",
+ "ureq",
  "url",
 ]
 
--- a/cli/loader/Cargo.toml
+++ b/cli/loader/Cargo.toml
@@ -38,6 +38,7 @@ serde.workspace = true
 serde_json.workspace = true
 tempfile.workspace = true
 url.workspace = true
+ureq = "3.0.11"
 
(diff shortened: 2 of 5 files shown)


clason:
This means we can also remove the "update emscripten" `xtask`, and with it the `ureq` dependency as requested in https://github.com/tree-sitter/tree-sitter/issues/4295.

clason:
Ah, I see that you added `ureq` as a full dependency, so the situation is now worse... Would switching to a different crate as in https://github.com/tree-sitter/tree-sitter/pull/4299 work for you? (I'm not happy about the inflated dependency chain in that PR; maybe there is a more light-weight alternative?)

maxbrunsfeld:
Oh yeah, I hadn't seen that discussion about how `ureq` is a problem for Microsoft. Thanks for pointing that out @clason; really any HTTP crate would be fine.

clason:
Maybe one could simply shell out to `curl` to grab the package?

clason:
On a related note, there's still a huge number of references to `emscripten` -- in the docs, the xtasks (I guess they can just be removed?), the bindings and the lib. What is the plan there?

clason:
That makes sense, thanks for the explanation!

PR #4396, opened 2025-04-25
Use http_req (w/ native tls) instead of ureq

maxbrunsfeld:
Fixes https://github.com/tree-sitter/tree-sitter/issues/4295


clason:
(I still think the `upgrade_emscripten` xtask is very low value -- all it does is change the number in https://github.com/tree-sitter/tree-sitter/blob/master/cli/loader/emscripten-version -- and could just be removed.)

maxbrunsfeld:
Oh, it seems like this would add a dependency on a system OpenSSL package on linux. That might not be good either. Not sure what to do there.

clason:
I think this is a reasonable requirement for any software that interacts with the internet (and in my professional experience is much preferred by distro people over rolling your own). You can mitigate this by checking for system installation of wasm32-clang before trying to download stuff (which I'm sure the Nix crowd will demand as well). 

Alternative would be to use reqwest instead, which in the PR I just closed passed CI.

alexr00:
We do indeed prefer to use the system OpenSSL package, so this PR looks like it would solve VS Code's concern with using Tree-Sitter.

clason:
I still don't understand how this concerns VS Code: The only code paths involving https requests are 
1. bumping the emscripten version (tree-sitter maintainer task, can be removed without loss);
2. building wasm parsers (optional parser maintainer task);
none of which are relevant to consumers (of the library)?

ObserverOfTime:
I think we'd rather use `curl` & `tar` than add a bunch of dependencies, considering they are available:

- On most Linux/BSD systems by default and can be declared as (optional?) package dependencies.
- On Windows 10/11 since build [17063](https://devblogs.microsoft.com/commandline/tar-and-curl-come-to-windows/) (2018) and Windows 7 is unsupported.
- On macOS by default.

_Note that the BSD/macOS variant of tar may be slightly different to the GNU variant on Linux & Windows._

PR #4402, opened 2025-04-27
fix(xtask): properly trim emscripten version at the source

WillLillis:
I noticed that `cargo xtask fetch-emscripten` was failing because of the same newline issue I fixed in https://github.com/tree-sitter/tree-sitter/pull/4365. `trim_ascii` is available in `const` contexts, so it makes more sense to just fix this at the source rather than hoping we catch all current (and future) uses of this string.

Edit: Marking this as a draft for now until the lint/wasm issue are resolved.


clason (approved):

maxbrunsfeld:
Thanks!

PR #4048, opened 2024-12-30
feat(cli): add json summary of parsing and generate errors

WillLillis:
This PR serves two purposes:

1. Provide the option to output a JSON summary of parsing results. 

This change is fairly small, but a nice addition.

<Details>

<Summary>Small example with a large and small file</Summary>

```sh
» cat small.c
int main(int argc, char *argv[])
{
    int * x = NULL;
    return EXIT_SUCCESS;
}

[lillis@LaptopWill] ~/projects/tree-sitter-c (master)
» /home/lillis/projects/tree-sitter/target/debug/tree-sitter parse -j small.c src/parser.c
src/parser.c    Parse:  991.82 ms         3931 bytes/ms (MISSING identifier [5513, 17] - [5513, 17])
{
  "parse_summaries": [
    {
      "file": "small.c",
      "successful": true,
      "start": {
        "row": 0,
        "column": 0
      },
      "end": {
        "row": 5,
        "column": 0
      },
      "duration": {
        "secs": 0,
        "nanos": 123995
      },
      "bytes": 82
    },
    {
      "file": "src/parser.c",
      "successful": false,
      "start": {
        "row": 0,
        "column": 0
      },
      "end": {
        "row": 119615,
        "column": 0
      },
      "duration": {
        "secs": 0,
        "nanos": 991820360
      },
      "bytes": 3899295
    }
  ],
  "cumulative_stats": {
    "successful_parses": 1,
    "total_parses": 2,
    "total_bytes": 3899377,
    "total_duration": {
      "secs": 0,
      "nanos": 991944355
    }
  }
}
```
</Details>

2. Aggressively refactor how errors are propagated and reported in the `generate` crate

This is a much larger change. Currently, the `generate` crate heavily leans on [anyhow](https://crates.io/crates/anyhow) to create and propagate errors up, typically as a `String`. This is fine for user-facing errors when used via `tree-sitter-cli`, but less than ideal when the `generate` crate is used as a library. Additionally, this makes it difficult to report errors in a machine-readable format (such as JSON). This PR makes heavy used of [thiserror](https://crates.io/crates/thiserror) so that errors consist of structured data. This does add a bit of boilerplate-like code, but also serves to better show how the various pieces of `generate` can fail. In some places, it also serves to significantly streamline some core logic. Since our errors are returned as structured data, we can use serde to report these in JSON if a user desires.

Purposefully breaking the tree-sitter-c grammar yields the following errors:

image

<Details>
[...]

PR #4404, opened 2025-04-27
Fix test_flatten_grammar_with_recursive_inline_variable

jamessan:
The `thiserror` refactor in #4048 dropped the period in the message for `FlattenGrammar::RecursiveInline`.

Due to this, the `test_flatten_grammar_with_recursive_inline_variable` has been failing since.

Restore the period to the error message to fix the test and add testing of tree-sitter-grammar to CI to avoid future regressions. This adds at most 15s to each platform that runs the tests.

I wasn't sure if there was a reason why the "test" xtask wasn't being used to test, so I took the conservative approach of just adding explicit testing for tree-sitter-grammar.

### DIFF
--- a/cli/generate/src/prepare_grammar/flatten_grammar.rs
+++ b/cli/generate/src/prepare_grammar/flatten_grammar.rs
@@ -533,7 +533,7 @@ mod tests {
 
         assert_eq!(
             result.unwrap_err().to_string(),
-            "Rule `test` cannot be inlined because it contains a reference to itself.",
+            "Rule `test` cannot be inlined because it contains a reference to itself",
         );
     }
 }


amaanq:
Oh wow, good catch. The generate crate's tests should be running, but isn't since we didn't pass in `--all` to `cargo` in the test xtask. I'll fix this up (I would rather remove the period in the test)

amaanq:
Let me know if my changes look good to you @jamessan, thanks!

jamessan:
> Let me know if my changes look good to you @jamessan, thanks!

CI still isn't running all tests, because it's using `cargo test` directly, not `cargo run -p xtask -- test`.

clason:
Be careful what you wish for ;)

jamessan:
Should we maybe separate the bug fix from the broader CI changes?

clason:
Yes!

PR #4408, opened 2025-04-28
Use native TLS for ureq when fetching wasi-sdk

### LINKED ISSUE #4295: Consider changing dependency on `ureq`

amaanq:
Closes #4295

### Problem

ureq uses `rustls` by default - some consumers would prefer if it used the OS-native TLS libraries (e.g. OpenSSL)

### Solution

In the ureq dependency, I've enabled the `native-tls` feature, and configure it to use native TLS as the TLS provider.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,13 +98,13 @@ jobs:
       if: ${{ matrix.use-cross }}
       run: |
         printf '%s\n' > Cross.toml \
-          '[target.${{ matrix.target }}]'                                   \
-          'image = "ghcr.io/cross-rs/${{ matrix.target }}:edge"'            \
-          '[build]'                                                         \
-          'pre-build = ['                                                   \
-          '  "dpkg --add-architecture $CROSS_DEB_ARCH",'                    \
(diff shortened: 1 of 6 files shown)


clason (approved):

amaanq:
Ppc64 might be screwed since Ubuntu doesn't ship BE packages.. :facepalm:

clason:
Can't you use the [native ARM runners](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md) that are (now) available on GHA?

For PowerPC, you might want to consider dropping support (tier) for this, or gate the new SDK download behind a feature flag that you disable for this architecture (and then skip the WASM tests).

maxbrunsfeld:
The thing I like about https://github.com/tree-sitter/tree-sitter/pull/4396 over this approach is that it removes the rustls dependency. It seems like with this feature flag, ureq keeps the dependency but just lets you turn it off. That said, its not a huge deal; whatever works for end users is fine.

wizzelman801 (commented):
Nice

PR #4430, opened 2025-05-04
fix(lib): replace raw array accesses with array_get

WillLillis:
Following up on #4422, this replaces all raw array accesses with `array_get`. `array_get` adds an assertion for OOB accesses in only for debug builds, meaning this will add 0 overhead to release builds. This should be a great help in finding other instances of UB in the library.

In adding these assertions, I found another OOB access in the query library from one of our existing Rust tests (`test_highlighting_cancellation`).

## The Problem

If a query consists of multiple patterns, and some number of patterns had predicates associated with them at the start of the query, but later patterns did *not* have predicates, an OOB access would occur in `self-predicate_steps`. This is due to the incorrect check here inside `ts_query_predicates_for_pattern`:

https://github.com/tree-sitter/tree-sitter/blob/b1d2b7cfb807410871a3679ee36e412ed2451d72/lib/src/query.c#L2981-L2983

This if condition will evaluate to true only if the query has *no* predicates at all, not if the given pattern doesn't have any predicates. This bug in theory wasn't observable from the Rust side, as the `step_count` out parameter was still correctly set to 0. However, this could pose an issue for consumers of the C library if there was a query where some patterns at the start had predicates, some number in the middle didn't, and at least one at the end did. Calls to `ts_query_predicates_for_pattern` for those middle patterns with no predicates would incorrectly return the predicate steps for later patterns that did have predicates. 

## The solution

Correct the early return to check if the *slice's* length is 0 for this particular pattern.


lib/src/stack.c
```diff
@@ -689,14 +689,15 @@ void ts_stack_renumber_version(Stack *self, StackVersion v1, StackVersion v2) {
...
-  self->heads.contents[v2] = temporary_head;
+  StackHead temporary_head = *array_get(&self->heads, v1);
+  *array_get(&self->heads, v1) = *array_get(&self->heads, v2);
+  *array_get(&self->heads, v2) = temporary_head;
 }
 
 StackVersion ts_stack_copy_version(Stack *self, StackVersion version) {
   ts_assert(version < self->heads.size);
-  array_push(&self->heads, self->heads.contents[version]);
+  StackHead version_head = *array_get(&self->heads, version);
```
WillLillis, line 699: Open to nits for this variable name. Pulled this out of `array_push` due to this warning:

```sh
In file included from lib/src/././subtree.h:12,
                 from lib/src/./language.h:8,
                 from lib/src/stack.c:2:
lib/src/stack.c: In function ‘ts_stack_copy_version’:
lib/src/./././array.h:65:33: warning: operation on ‘self->heads.size’ may be undefined [-Wsequence-point]
   65 |    (self)->contents[(self)->size++] = (element))
      |                     ~~~~~~~~~~~~^~
lib/src/stack.c:699:3: note: in expansion of macro ‘array_push’
  699 |   array_push(&self->heads, *array_get(&self->heads, version));
      |   ^~~~~~~~~~
```

clason (approved):

clason:
Maybe pull the actual fix out into a separate PR so we can get it into 0.25.4 while still waiting on resolving the refactor question?

clason:
Do we still want to backport this? Normally, one would only backport fixes, not refactors.

WillLillis:
> Do we still want to backport this? Normally, one would only backport fixes, not refactors.

I still view it as a "fix" because the library will now terminate (in debug mode) instead of entering into undefined behavior territory. I don't feel super strongly and am ok with removing, though.

clason:
If the effects are only noticeable in Debug mode, I don't think there's much to be gained from backporting. (Unless it makes future backports easier.)

WillLillis:
My main thought was that these checks would be accessible through the rust bindings, since tests are run in debug mode, and programs are typically built in debug  during development. It wouldn't make any difference for the published CLI tool, though.

clason:
Either way is fine with me; that was an honest question.

WillLillis:
Gotcha. I'm leaning towards keeping it as a backport, although probably not for this upcoming release because of pending reviews. :)

PR #4443, opened 2025-05-12
Add Erlang Tree Sitter Grammar

robertoaloi:


robertoaloi:
It actually looks like thurd-party grammars are available in the [wiki](https://github.com/tree-sitter/tree-sitter/wiki/List-of-parsers), so ignore this one.

PR #4454, opened 2025-05-16
fix(generate): re-enable default url features

### LINKED ISSUE #4453: tree-sitter-generate fails to build on windows
### Problem

Running build on tree-sitter-generate in version 0.25.4 fails with this error:

```rust
error[E0599]: no function or associated item named `from_file_path` found for struct `Url` in the current scope
   --> cli\generate\src\lib.rs:347:34
    |
347 |     let grammar_path = url::Url::from_file_path(grammar_path)
    |                                  ^^^^^^^^^^^^^^ function or associated item not found in `Url`
    |
note: if you're trying to build a new `Url` consider using one of the following associated functions:
      Url::parse
      Url::parse_with_params
   --> C:\Users\Jetoni\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\url-2.5.4\src\lib.rs:340:5
    |
340 |       pub fn parse(input: &str) -> Result<Url, crate::ParseError> {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
370 | /     pub fn parse_with_params<I, K, V>(input: &str, iter: I) -> Result<Url, crate::ParseError>
371 | |     where
372 | |         I: IntoIterator,
373 | |         I::Item: Borrow<(K, V)>,
374 | |         K: AsRef<str>,
375 | |         V: AsRef<str>,
    | |______________________^
```

I found the source of the issue, the url library has only a single feature enabled by default to set whether url should use std, this merge request: #4411 disables default-features: 
```toml
[target.'cfg(windows)'.dependencies]
url = { version = "2.5.4", default-features = false }
```
the function `from_file_path()` is gated behind the std feature: "This method is only available if the `std` Cargo feature is enabled." Disabling default features is unnecessary and prevents the crate from building. (I will be submitting a pull request to fix this issue soon)

### Steps to reproduce

git clone [tree-sitter](https://github.com/tree-sitter/tree-sitter.git)
build --package tree-sitter-generate --lib

### Expected behavior

Building tree-sitter-generate succeeds

tree-sitter 0.25.4

### Operating system/version

windows 11

holozene:
Fixes #4453 

### Problem
Disabling default features for the `url` crate causes tree_sitter_generate build to fail.

### Solution
Re-enable default features for the `url` crate.

### DIFF
--- a/cli/generate/Cargo.toml
+++ b/cli/generate/Cargo.toml
@@ -33,4 +33,4 @@ thiserror.workspace = true
 tree-sitter.workspace = true
 
 [target.'cfg(windows)'.dependencies]
-url = { version = "2.5.4", default-features = false }
+url.workspace = true


BoomchainLabs (approved):

WillLillis:
I think the original motivation behind `default-features = false` was to preserve `no-std` compatibility. That being said, I've tested this patch locally on a windows machine and it seems to still work with `no-std`. Let's see what CI has to say.

WillLillis:
@holozene can you squash those extra commits?

clason:
AFAICT, the original motivation was to make this a Windows-only dependency; the `default-features=false` was not really motivated. So  this LGTM.

PR #4465, opened 2025-05-24
build(deps): bump wasmtime to v33.0.1

clason:
- **build(deps): bump wasmtime-c-api to v33.0.1**
- **build(deps): update cargo dependencies**
- **build(cargo)!: update MSRV to 1.83** (and fix new clippy failures)
- **ci(nvim-ts): throttle parser build** (poor Windows runners can't handle the concurrency)

Note: will squash and reorder commits after review


mavit:
There appears to be a [security vulnerability affecting wasmtime 33.0.1 (and earlier versions)](https://nvd.nist.gov/vuln/detail/CVE-2025-53901), so I would suggest moving to 33.0.2 instead.

ObserverOfTime:
> This bug is in the implementation of Wasmtime's `wasmtime-wasi` crate which provides an implementation of WASIp1. 

Which we don't use so I assume we're not affected.

PR #4482, opened 2025-06-02
Fix typing for Parser.init

tmr232:
## The Change

Since we're usually only providing `locateFile`, we need the type to be
`Partial<>` to allow it.

This also matches the typing in `@types/emscripten`'s
`EmscriptenModuleFactory` type signature.

## Motivation

Over the last few days I spent several hours debugging a relatively silly bug.
I returned a `vscode.Uri` from `locateFile` instead of a `string`.
Since `tsc` did not complain, I didn't even think to check that one function.

With this PR, I hope to help others avoid the same bug.

With this PR, and installing `@types/emscripten`, users will get type errors when returning the wrong type from `locateFile`.

## Implementation

I modified the `d.ts` file for the WASM part.
Before I did that, I used the `--emit-tsd` flag to generate the file.
At least on my machine, it came out different from the committed one, so I assumed it is OK to add changes to it.

### DIFF
--- a/lib/binding_web/lib/web-tree-sitter.d.ts
+++ b/lib/binding_web/lib/web-tree-sitter.d.ts
@@ -217,4 +217,4 @@ interface WasmModule {
 }
 
 export type MainModule = WasmModule & typeof RuntimeExports;
-export default function MainModuleFactory (options?: EmscriptenModule): Promise<MainModule>;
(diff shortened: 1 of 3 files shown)


ObserverOfTime (approved):

verhovsky:
This change doesn't seem to be in the 0.25.9 NPM package.

PR #4492, opened 2025-06-06
fix(cli): require correct setuptools version

rabbiveesh:
The build process uses the module `setuptools.command.build` which was added in 62.4.0 as
per https://github.com/pypa/setuptools/blob/main/NEWS.rst#v6240

also ref https://github.com/tree-sitter-perl/tree-sitter-perl/issues/214

### DIFF
--- a/crates/cli/src/templates/pyproject.toml
+++ b/crates/cli/src/templates/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=42", "wheel"]
+requires = ["setuptools>=62.4.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]


ObserverOfTime (commented):
For backwards compatibility we could fall back to `distutils.command.build` instead.

Several distros have outdated packages according to https://pkgs.org/search/?q=setuptools

ALT Linux: 42.0.0
Amazon Linux 2023: 59.6.0
Amazon Linux 2: 49.1.3
Debian 11: 44.1.1
openSUSE Leap 15.6: 44.1.1
PCLinuxOS: 45.2.0
~~RHEL 8 & derivatives: 39.2.0~~
RHEL 9 & derivatives: 53.0.0
Slackware 15.0: 57.5.0
Ubuntu 22.04: 59.6.0

rabbiveesh:
From what I understand when you declare like I did the pip will install the build deps in a temporary env; so I'm not sure there's benefit to implementing fallback

ObserverOfTime:
Not if you use the system packages.

clason:
Can you please rebase?

ObserverOfTime:
Can we merge #4493 instead?

clason:
Why? This is a much simpler fix for the actual issue, and 62.4.0 was released three years ago.

ObserverOfTime:
The `bdist_wheel` command also needs a fallback since the setuptools version was introduced in v70.1.0 which is too recent. We might as well have a fallback for the `build` command too in that case.

maxbrunsfeld:
I agree with @clason. I would like the bindings to stay simple; I don’t think it’s worth adding the fallback scheme to support people who do not want to install recent versions of setup tools.

ObserverOfTime:
Then should we bump all the way to v70.1.0?

clason:
> Then should we bump all the way to v70.1.0?

Why? What will that buy us over 62.4.0?

ObserverOfTime:
> Adopted the `bdist_wheel` command from the `wheel` project

maxbrunsfeld:
We could also just keep using the `wheel` library directly, right? I don't have a strong opinion on that one. Wheel is not deprecated like `distutils` is. @ObserverOfTime what do you think about this PR's current approach of just upgrading `setuptools` and keeping `wheel`?

ObserverOfTime:
The package isn't deprecated but the `bdist_wheel` import is.
And switching to the `setuptools` version will allow us to drop `wheel` from `requires`.

rabbiveesh:
Do we have a conclusion here? I'm happy to pull in latest master and bump further to 70 whatever as needed.
Could I get a clarification for exactly when the install would fail, btw? I didn't quite get the case (I assume it's my ignorance)

clason:
Let's start with a simple rebase(!) on `master`. If it fixes a concrete problem, we can merge and backport quickly and then iterate further (if and when someone comes up with an _actual_ problem that can be verified to be solved).

rabbiveesh:
K, i have rebased and force-pushed and its all clean.
The version that i've bumped to does solve a concrete problem, so i'd love to see this merged

PR #4498, opened 2025-06-07
feat(cli): rebuild parsers by default when --grammar-path is passed.

WillLillis:
@sogaiu brought up a neat use case  [here](https://github.com/tree-sitter/tree-sitter/pull/4358#issuecomment-2943845395) that #4358 didn't quite cover. The CLI tool stores built parsers in between runs, which leads to a nice start time boost for non-cold invocations to commands like `parse`, `fuzz`, etc. This clashes a bit with the `--grammar-path` option, however, as a cached build of a parser from a previous invocation will be used even if a subsequent use points to a different copy of the same grammar with `--grammar-path`. This is a nice use case if, say, you have two copies of a grammar checked out at different commits. 

The work around to this issue is to simply rebuild the parser when `--grammar-path` is supplied. This seems reasonable to me -- a user passing `--grammar-path` is indicating they want to use the grammar exactly at *that* path and that path only. Any residual global state (e.g. `.so` files left behind in `~/.cache/tree-sitter/lib`) should be ignored. The extra time spent rebuilding the parser can be a bit annoying if you aren't using two copies of the same grammar, however. ~~To mitigate this pain point, I've also added `--no-rebuild` as a `--grammar-path` only argument to circumvent this behavior.~~

Also, because the `query`, `highlight`, and `tags` commands load parsers, this PR also adds the associated `--rebuild` flag.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -182,8 +182,8 @@ struct Parse {
     /// The source file(s) to use
     #[arg(num_args=1..)]
     pub paths: Option<Vec<PathBuf>>,
-    /// The path to the tree-sitter grammar directory
-    #[arg(long, short = 'p')]
+    /// The path to the tree-sitter grammar directory, implies --rebuild
+    #[arg(long, short = 'p', conflicts_with = "rebuild")]
     pub grammar_path: Option<PathBuf>,
     /// The path to the parser's dynamic library
     #[arg(long, short = 'l')]
@@ -280,8 +280,8 @@ struct Test {
     /// Only run corpus test cases from a given filename
     #[arg(long)]
     pub file_name: Option<String>,
-    /// The path to the tree-sitter grammar directory
-    #[arg(long, short = 'p')]
+    /// The path to the tree-sitter grammar directory, implies --rebuild
+    #[arg(long, short = 'p', conflicts_with = "rebuild")]
     pub grammar_path: Option<PathBuf>,
     /// The path to the parser's dynamic library
     #[arg(long, short = 'l')]
@@ -346,8 +346,8 @@ struct Fuzz {
     /// Subdirectory to the language
     #[arg(long)]
     pub subdir: Option<PathBuf>,
-    /// The path to the tree-sitter grammar directory
-    #[arg(long, short = 'p')]
+    /// The path to the tree-sitter grammar directory, implies --rebuild
+    #[arg(long, short = 'p', conflicts_with = "rebuild")]
     pub grammar_path: Option<PathBuf>,
     /// The path to the parser's dynamic library
     #[arg(long)]
@@ -385,8 +385,8 @@ struct Query {
     /// Path to a file with queries
     #[arg(index = 1, required = true)]
     query_path: PathBuf,
-    /// The path to the tree-sitter grammar directory
-    #[arg(long, short = 'p')]
+    /// The path to the tree-sitter grammar directory, implies --rebuild
+    #[arg(long, short = 'p', conflicts_with = "rebuild")]
     pub grammar_path: Option<PathBuf>,
     /// The path to the parser's dynamic library
     #[arg(long, short = 'l')]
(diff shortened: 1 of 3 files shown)


WillLillis:
The test failure seems unrelated. Will take a look Sunday if someone else hasn't gotten to it by then.

clason:
Alternative suggestion: add `--build` and `--generate` flags that you can pass to other commands (including `test`, `parse`, `build`)? It's usually better to opt-in than opt-out.

sogaiu:
Is the idea of just being able to directly specify a path to a parser shared object worth considering (e.g. for `parse`)?

The use cases I have in mind is comparing behavior of:

* two different commits from the "same" grammar
* two grammars with the same name but different repositories

No rebuilding of parsers would be needed (I think) if the parser shared objects are stored in a location other than under `~/.cache/tree-sitter/lib`.  I presume I can use `tree-sitter build --output ...` to put the built parser shared objects in desired locations.

clason:
That was also the use case I had in mind a while back (before `ts_query_ls` took that job).

> I presume I can use tree-sitter build --output ... to put the built parser shared objects in desired locations.

Yes, definitely!

WillLillis:
> Is the idea of just being able to directly specify a path to a parser shared object worth considering (e.g. for parse)?

Revisiting this, I'm not sure if this functionality should replace or be added in addition to the changes in this PR?

I've looked into adding support for loading a shared library directly, but in order to do this we need the language's name to grab the correct symbol from the lib. I think we can: 

- Attempt to infer this based on the provided the file name (e.g. `c.so` -> `tree_sitter_c`
- Add an optional argument to allow users to specify this directly (to cover cases where a user has renamed the file)

CC @clason @sogaiu What are your thoughts here?

sogaiu:
> Revisiting this, I'm not sure if this functionality should replace or be added in addition to the changes in this PR?

Speaking only for my own situation, I think I'd be most helped by (and likely to use) the ability to specify a shared object path.  I believe that capability would take care of a recurring and error-prone "juggling" I end up doing during testing of and comparison of "similar" grammars.

Regarding:

> in order to do this we need the language's name to grab the correct symbol from the lib ... we can:
>
> * Attempt to infer this based on the provided the file name (e.g. c.so -> tree_sitter_c
> * Add an optional argument to allow users to specify this directly (to cover cases where a user has renamed the file)

If it's not too costly, these both seem worth doing to me, though if I had to choose one, it would be the latter.

crates/cli/src/main.rs
```diff
@@ -170,6 +170,14 @@ struct Parse {
     /// The path to the tree-sitter grammar directory
     #[arg(long, short = 'p')]
     pub grammar_path: Option<PathBuf>,
+    /// Don't force rebuild the parser when `--grammar_path` is specified
```
clason: These comments are a bit misleading, if I understand the PR correctly. Isn't it more of a
```rust
/// `--grammar_path` implies force rebuild
```
?

crates/cli/src/main.rs
```diff
@@ -910,7 +945,7 @@ impl Parse {
         let mut parser = Parser::new();
 
         loader.debug_build(self.debug_build);
-        loader.force_rebuild(self.rebuild);
+        loader.force_rebuild(self.rebuild || !self.no_rebuild);
```
clason: I expect this to be a more involved change, but any chance these two properties can be combined? It's a bit weird to have "tea" and "no tea" at the same time...

WillLillis: Looking back over these changes, I'm not quite sure what the right approach is here. I worry that `--no-rebuild` is an option for too niche of a case, and will just lead to confusion. I'm leaning towards just removing the `--no-rebuild` flag, and adding your proposed comment about how `--grammar-path` implies `--rebuild`.

I still do think that rebuilding by default makes sense when `--grammar-path` is passed; I think the behavior allowed if we don't rebuild is very non-intuitive. It certainly took me a bit to figure out what was going on.

clason: Yes, I think it's perfectly fine to always rebuild if grammar-path is specified -- you would not specify it unless you wanted that grammar to be used, period.

WillLillis: Alright I removed the `--no-rebuild` flag and parsers should always be rebuilt now when `--grammar-path` is passed. To (hopefully) avoid confusion, I've made `--grammar-path` conflict with `--rebuild` as well.

ObserverOfTime (approved):

clason (approved):

PR #4501, opened 2025-06-08
feat(cli): support selecting bindings in init cmd

sammyne:
This PR supports selecting bindings when `tree-sitter init` find no tree-stter.json during setup.
It will prompt user to select bindings to generate as follows
image

The default enabled bindings is consistent with `tree_sitter_loader::Bindings::default()`.


ObserverOfTime (commented):
Comment out Java and Kotlin until the bindings are actually available.

sammyne:
> Comment out Java and Kotlin until the bindings are actually available.

Updated as requested. @ObserverOfTime

PR #4505, opened 2025-06-09
fix(loader): fix no-default-features build

### LINKED ISSUE #4500: tree-sitter-loader does not build without default features
### Problem

#1638 attempted to make tree-sitter-loader's dependency on tree-sitter-highlight / tags optional. However, it doesn't actually build when using `--no-default-features`.

### Steps to reproduce

```sh
$ cargo build -p tree-sitter-loader --no-default-features
...
error[E0412]: cannot find type `Error` in this scope
   --> crates/loader/src/loader.rs:813:21
    |
813 |     ) -> Result<(), Error> {
    |                     ^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
8   + use std::error::Error;
    |
8   + use std::fmt::Error;
    |
8   + use std::io::Error;
    |
8   + use core::error::Error;
    |
      and 11 other candidates

error[E0412]: cannot find type `Error` in this scope
   --> crates/loader/src/loader.rs:975:21
    |
975 |     ) -> Result<(), Error> {
    |                     ^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
8   + use std::error::Error;
    |
8   + use std::fmt::Error;
    |
8   + use std::io::Error;
    |
8   + use core::error::Error;
    |
      and 11 other candidates

error[E0412]: cannot find type `Error` in this scope
    --> crates/loader/src/loader.rs:1023:21
     |
1023 |     ) -> Result<(), Error> {
     |                     ^^^^^ not found in this scope
     |
help: consider importing one of these items
     |
8    + use std::error::Error;
     |
8    + use std::fmt::Error;
     |
8    + use std::io::Error;
     |
8    + use core::error::Error;
     |
       and 11 other candidates

error[E0412]: cannot find type `Error` in this scope
    --> crates/loader/src/loader.rs:1048:57
     |
1048 |     fn ensure_wasi_sdk_exists(&self) -> Result<PathBuf, Error> {
     |                                                         ^^^^^ not found in this scope
     |
help: consider importing one of these items
     |
8    + use std::error::Error;
     |
8    + use std::fmt::Error;
     |
8    + use std::io::Error;
     |
8    + use core::error::Error;
     |
       and 11 other candidates

error[E0392]: lifetime parameter `'a` is never used
   --> crates/loader/src/loader.rs:305:34
    |
305 | pub struct LanguageConfiguration<'a> {
    |                                  ^^ unused lifetime parameter
    |
    = help: consider removing `'a`, referring to it in a field, or using a marker such as `PhantomData`

Some errors have detailed explanations: E0392, E0412.
For more information about an error, try `rustc --explain E0392`.
error: could not compile `tree-sitter-loader` (lib) due to 5 previous errors
```

### Expected behavior

The crate builds.

97722708686023e4c21ea02d3bc6f763b3cf24bd

### Operating system/version

Linux

ObserverOfTime:
Fixes #4500

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -9,6 +9,7 @@ use std::{
     collections::HashMap,
     env, fs,
     io::{BufRead, BufReader, Write as _},
+    marker::PhantomData,
     mem,
     path::{Path, PathBuf},
     process::Command,
@@ -16,7 +17,6 @@ use std::{
     time::SystemTime,
 };
 
-#[cfg(any(feature = "tree-sitter-highlight", feature = "tree-sitter-tags"))]
 use anyhow::Error;
 use anyhow::{anyhow, Context, Result};
 use etcetera::BaseStrategy as _;
@@ -323,6 +323,7 @@ pub struct LanguageConfiguration<'a> {
     highlight_names: &'a Mutex<Vec<String>>,
     #[cfg(feature = "tree-sitter-highlight")]
     use_all_highlight_names: bool,
+    _phantom: PhantomData<&'a ()>,
 }
 
 pub struct Loader {
@@ -1231,6 +1232,7 @@ impl Loader {
                     highlight_names: &self.highlight_names,
                     #[cfg(feature = "tree-sitter-highlight")]
                     use_all_highlight_names: self.use_all_highlight_names,
+                    _phantom: PhantomData,
                 };
 
                 for file_type in &configuration.file_types {
@@ -1300,6 +1302,7 @@ impl Loader {
                 highlight_names: &self.highlight_names,
                 #[cfg(feature = "tree-sitter-highlight")]
                 use_all_highlight_names: self.use_all_highlight_names,
+                _phantom: PhantomData,
             };
             self.language_configurations.push(unsafe {
                 mem::transmute::<LanguageConfiguration<'_>, LanguageConfiguration<'static>>(


crates/loader/src/loader.rs
```diff
@@ -323,6 +324,8 @@ pub struct LanguageConfiguration<'a> {
     highlight_names: &'a Mutex<Vec<String>>,
     #[cfg(feature = "tree-sitter-highlight")]
     use_all_highlight_names: bool,
+    #[cfg(not(any(feature = "tree-sitter-highlight", feature = "tree-sitter-tags")))]
+    _phantom: PhantomData<&'a ()>,
```
maxbrunsfeld, line 326: Does this need to be conditional on the features? Can we just always have the phantom data?

ObserverOfTime: It's not needed if the highlight feature is enabled, which uses the lifetime.

maxbrunsfeld: Right, but the `PhantomData` has zero cost, and we want to avoid unnecessary code/conditionals. So we should only add the conditional compilation if it *needs* to be there. Does this crate compile cleanly with both the highlight fields *and* the `_phantom` field? If so, we should just always include that phantom field.

maxbrunsfeld:
Thanks for the fix @ObserverOfTime !

PR #4521, opened 2025-06-15
fix(generate): only display conflicting symbol name in non-terminal word token error message if available

### LINKED ISSUE #4497: Panic when word has a choice
### Problem

When word is set to a rule which contains a choice `tree-sitter generate` panics.

```
$ ./tree-sitter-linux-x64 generate

thread 'main' panicked at cli/generate/src/prepare_grammar/extract_tokens.rs:169:18:
Failed to find a variable with the same rule as the word token
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: tree_sitter_generate::prepare_grammar::extract_tokens::extract_tokens
   4: tree_sitter_generate::prepare_grammar::prepare_grammar
   5: tree_sitter_generate::generate_parser_for_grammar_with_opts
   6: tree_sitter_generate::generate_parser_in_directory
   7: tree_sitter::Generate::run
   8: tree_sitter::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### Steps to reproduce

run `tree-sitter init`
replace `grammar.js`'s content with
```js
module.exports = grammar({
  name: "spaced_binary",
  word: $ => $.bit,
  rules: {
    bit: _ => choice("0", "1"),
  }
});
```
run `tree-sitter generate`
Observe the panic.

### Expected behavior

The generation should either fail with an error message or succeed with generating the parser.

0.25.6

### Operating system/version

Ubuntu noble 24.04 x86_64

WillLillis:
This panic is caused when a non-terminal is used for a grammar's `word` token, but said rule isn't referenced anywhere else in the grammar. This PR just removes the panic when the message when said conflicting rule isn't found, and instead defaults to the error message displayed before [352393c](https://github.com/tree-sitter/tree-sitter/pull/4092/commits/352393c970b7362fa028e9809b3eed3e602cc920).

- Closes #4497

### DIFF
--- a/crates/generate/src/prepare_grammar/extract_tokens.rs
+++ b/crates/generate/src/prepare_grammar/extract_tokens.rs
@@ -26,10 +26,34 @@ unless they are used only as the grammar's start rule.
     ExternalTokenNonTerminal(String),
     #[error("Non-symbol rules cannot be used as external tokens")]
     NonSymbolExternalToken,
-    #[error("Non-terminal symbol '{0}' cannot be used as the word token, because its rule is duplicated in '{1}'")]
-    NonTerminalWordToken(String, String),
+    #[error(transparent)]
+    WordToken(NonTerminalWordTokenError),
     #[error("Reserved word '{0}' must be a token")]
     NonTokenReservedWord(String),
+}
+
+#[derive(Debug, Error, Serialize)]
+pub struct NonTerminalWordTokenError {
+    pub symbol_name: String,
+    pub conflicting_symbol_name: Option<String>,
+}
+
+impl std::fmt::Display for NonTerminalWordTokenError {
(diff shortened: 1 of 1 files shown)


clason:
Is it possible to add a _new_ error message for this? The linked change was made because people found the previous message not helpful enough. (Generation failures can be inscrutable...)

WillLillis:
> Is it possible to add a _new_ error message for this? The linked change was made because people found the previous message not helpful enough. (Generation failures can be inscrutable...)

I'd be open to adding a different error message in cases where the issue isn't the symbol being duplicated, but just a non terminal like in #4497's repro. I'm not sure what additional context we could add here, though. Any ideas?

clason:
I was thinking more of not reverting the more explicit error message for the case it was introduced for; I'd prefer not regressing on that.

PR #4523, opened 2025-06-16
feat(rust)!: use `ops::ControlFlow` as parse and query progress return value

d-e-s-o:
Instead of returning an undocumented boolean flag, use a `core::ops::ControlFlow` object. At the expense of being a bit more verbose, this is a type that should be self-explanatory in the context of a callback, as an indication of whether to continue processing or stop.

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -1,6 +1,7 @@
 use std::{
     fmt, fs,
     io::{self, Write},
+    ops::ControlFlow,
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
     time::{Duration, Instant},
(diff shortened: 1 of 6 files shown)


d-e-s-o:
Note: if accepted, we should do something similar for the query progress callback, but I wasn't sure whether this kind of change is desired and so I didn't do that right away. But happy to do it.

ObserverOfTime (commented):
Can you add an internal convenience function for converting `bool` to `ControlFlow`?

WillLillis:
> Note: if accepted, we should do something similar for the query progress callback, but I wasn't sure whether this kind of change is desired and so I didn't do that right away. But happy to do it.

Good point, just pushed up that update.

> Can you add an internal convenience function for converting bool to ControlFlow?

This would negate a lot of the explicitness/readability gains the PR is trying to add. We also can't cleanly implement  `From`/`Into` since we're dealing with a foreign trait and type.

PR #4536, opened 2025-06-25
fix(bindings): only include top level LICENSE file

ObserverOfTime:
Ref: tree-sitter/workflows#33


WillLillis:
What invocation of `tree-sitter init` can I use to test this? Trying this out locally I'm not sure what this is supposed to be doing with the LICENSE file.

ObserverOfTime:
This will exclude license files from node modules which cargo will try to include otherwise.
```shell
npm install
cargo package --dry-run # error
tree-sitter init --update
cargo package --dry-run # no error
```

WillLillis (approved):

PR #4545, opened 2025-06-27
ci(nvim-ts): update paths to refactor

clason:
Problem: After https://github.com/tree-sitter/tree-sitter/pull/4496, the crater tests from nvim-treesitter no longer run on PRs with relevant changes.

Solution: Update the paths in the workflows to the new structure.

Also throttle the build step to 10 parallel tasks (reduces flakiness on Windows) ~~and raise the limit for the generate step to 3 (reduces the chance of timeouts)~~ (macOS runners have only 7GB RAM 😭 ).

Also raise the timeout limit to 30 minutes (it's more expensive to have to restart flaky tests than to let them run).

(Actually, I would suggest just removing this timeout completely -- it's not that this workflow is unbounded or can freeze.)

### DIFF
--- a/.github/workflows/nvim_ts.yml
+++ b/.github/workflows/nvim_ts.yml
@@ -3,7 +3,10 @@ name: nvim-treesitter parser tests
 on:
   pull_request:
     paths:
-      - 'cli/**'
+      - 'crates/cli/**'
+      - 'crates/config/**'
+      - 'crates/generate/**'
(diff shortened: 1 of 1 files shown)


clason:
For later reference: generate step takes 11-12m on macOS and Ubuntu and 16m on Windows.

WillLillis (approved):

PR #4555, opened 2025-07-03
test(python): improve bindings test to detect ABI incompatibilities

rmuir:
Previously, python bindings test would not detect ABI incompatibilities. At this time, the latest pypi release of py-tree-sitter does not support 15, which is the default, and it creates the problem.

Improve the test to create a Parser, which will detect it:
```
======================================================================
ERROR: test_can_load_grammar
----------------------------------------------------------------------
Traceback (most recent call last):
  File "../tests/test_binding.py", line 10, in test_can_load_grammar
    _ = Parser(language)
ValueError: Incompatible Language version 15. Must be between 13 and 14

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (errors=1)
```

### DIFF
--- a/crates/cli/src/templates/test_binding.py
+++ b/crates/cli/src/templates/test_binding.py
@@ -1,12 +1,12 @@
 from unittest import TestCase
 
-import tree_sitter
+from tree_sitter import Language, Parser
 import tree_sitter_LOWER_PARSER_NAME
 
 
 class TestLanguage(TestCase):
     def test_can_load_grammar(self):
         try:
-            tree_sitter.Language(tree_sitter_LOWER_PARSER_NAME.language())
+            Parser(Language(tree_sitter_LOWER_PARSER_NAME.language()))
         except Exception:
             self.fail("Error loading TITLE_PARSER_NAME grammar")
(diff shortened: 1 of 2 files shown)


crates/cli/src/templates/test_binding.py
```diff
@@ -1,12 +1,10 @@
...
 
 
 class TestLanguage(TestCase):
     def test_can_load_grammar(self):
-        try:
-            tree_sitter.Language(tree_sitter_LOWER_PARSER_NAME.language())
-        except Exception:
-            self.fail("Error loading TITLE_PARSER_NAME grammar")
+        language = Language(tree_sitter_LOWER_PARSER_NAME.language())
+        _ = Parser(language)
```
ObserverOfTime:
```suggestion
        try:
            Parser(Language(tree_sitter_LOWER_PARSER_NAME.language()))
        except Exception:
            self.fail("Error loading TITLE_PARSER_NAME grammar")
```

rmuir: Thanks: agreed it ensures a good error always.

ObserverOfTime (approved):

PR #4557, opened 2025-07-03
Fix population of `extra` field in `node-types.json`

### LINKED ISSUE #4556: Really expose `extras` in node-types.json
### Problem

As a grammar reuser, it would be useful to know whether a given node type belongs to the `extras` or not.
At the moment, this information doesn't seem to be exposed, neither on the concrete tree-sitter nodes themselves, nor in `node-types.json`.

To determine if a particular node is an extra, the only solution I have though of so far is to look up the possible children of its parents via `node-types.json`, and check if it is included in those. But it's not straightforward since I need to expand the set of children mentioned there via subtypes.

### Expected behavior

Since `node-types.json` is an array at its root, it doesn't seem feasible to add an `"extras": ["comment", …]` field there. Instead, we could add an `"extra": true` field on each node entry in this list, which shouldn't cause compatibility issues as most systems processing JSON data should be happy with new keys on the objects they process.

Another option would be to introduce a new file `extras.json`, but it would require updating the bindings (and their generation) to expose that properly.

Another option would be to expose a boolean field on the tree-sitter node data structure itself. This seems wasteful as it would increase memory consumption without a good reason.

wetneb:
Fixes #4556.

The existing feature failed to populate the `extra` field in certain cases. I've added a minimal example of a grammar where the issue appears, as the first commit of this PR. You can check out this commit to see that the test fails before the second commit.

The underlying reason is that some entries of `node-types.json` were generated before the extra fields were listed and resolved, without their `extra` field being updated later. I've re-organized the generation so that all node entries are generated with their correct `extra` value from the start, removing the need for them to be updated later.

### DIFF
--- a/crates/generate/src/node_types.rs
+++ b/crates/generate/src/node_types.rs
@@ -507,6 +507,31 @@ pub fn generate_node_types_json(
 
     let aliases_by_symbol = get_aliases_by_symbol(syntax_grammar, default_aliases);
 
+    let empty = HashSet::new();
+    let extra_names = syntax_grammar
+        .extra_symbols
+        .iter()
+        .flat_map(|symbol| {
+            aliases_by_symbol
+                .get(symbol)
+                .unwrap_or(&empty)
+                .iter()
+                .map(|alias| {
+                    alias.as_ref().map_or(
+                        match symbol.kind {
+                            SymbolType::NonTerminal => &syntax_grammar.variables[symbol.index].name,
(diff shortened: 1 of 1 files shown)


WillLillis (approved):
Changes LGTM, thanks for the patch! This should be rebased on top of https://github.com/tree-sitter/tree-sitter/pull/4563 once it's merged.

WillLillis:
@wetneb mind rebasing off of latest master?

PR #4567, opened 2025-07-07
feat: Warn about unreachable variables

### LINKED ISSUE #4565: feat: Warn about unreachable variables when generating the parser
### Problem

Variable names (i.e. node types) can be unreachable when they aren't linked to the root rule (generally named `source_file`) and aren't part of `extras`.

When such variable names appear, it might be a symptom of a problem: either the dead rule should be removed, or it should be linked to reachable rules in other ways.

Even though dead rules don't make it into the generated parser at all (as far as I can see), it could be nice to have `tree-sitter generate` warn grammar authors about such cases, possibly with a CLI flag to enable or disable this warning. When using meta-programming to generate the rules dynamically, it can be that such unreachable variables are expected and should just be ignored. This might be a reason for not enabling this check by default.

### Expected behavior

Output a warning on stderr when this feature is enabled and some rules are unreachable.

This warning would be formatted in a similar way to those for unnecessary conflict declarations, for instance.

wetneb:
Closes #4565.

I added this warning as a `println!`, similarly to the existing warnings for unnecessary conflict declarations. Ideally, those should be using `log::warn`, but the logging system is only initialized when the user has passed `-l` / `--log` on the CLI.

Turning such warnings into logging statements would also make them easier to unit test, for instance via the [`testing_logger`](https://crates.io/crates/testing_logger) crate.

### DIFF
--- a/crates/generate/src/parse_grammar.rs
+++ b/crates/generate/src/parse_grammar.rs
@@ -246,6 +246,7 @@ pub(crate) fn parse_grammar(input: &str) -> ParseGrammarResult<InputGrammar> {
             &mut in_progress,
         ) && grammar_json.word.as_ref().is_none_or(|w| w != name)
         {
+            println!("Warning: unused variable {name:?}");
             grammar_json.conflicts.retain(|r| !r.contains(name));
             grammar_json.supertypes.retain(|r| r != name);
             grammar_json.inline.retain(|r| r != name);
--- a/crates/generate/src/parse_grammar.rs
+++ b/crates/generate/src/parse_grammar.rs
@@ -246,7 +246,7 @@ pub(crate) fn parse_grammar(input: &str) -> ParseGrammarResult<InputGrammar> {
             &mut in_progress,
         ) && grammar_json.word.as_ref().is_none_or(|w| w != name)
         {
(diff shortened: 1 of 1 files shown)


clason:
That sounds like a general refactor worth doing (as long as the warnings are still printed to stdout/stderr by default).

crates/generate/src/parse_grammar.rs
```diff
@@ -246,6 +246,7 @@ pub(crate) fn parse_grammar(input: &str) -> ParseGrammarResult<InputGrammar> {
             &mut in_progress,
         ) && grammar_json.word.as_ref().is_none_or(|w| w != name)
         {
+            println!("Warning: unused variable {name:?}");
```
ObserverOfTime:
```suggestion
            eprintln!("Warning: unused variable {name:?}");
```

clason:
Hmm, this triggers false positives in tree-sitter-markdown, where a rule is defined in `common.js` but only used in `tree-sitter-markdown_inline`, not `tree-sitter-markdown`. (I expect similar issues arise in other split parsers.)

I think we may have to revert that until we have better (Info/Hint) output.

PR #4571, opened 2025-07-08
perf: reserve `Vec` capacities where appropriate

WillLillis:
I replaced all usages `Vec::new()` with `Vec::with_capacity()` if a reasonable initial capacity (or lower bound) could be ascertained. This gives a slight speed boost in `generate`, for example:

**tree-sitter-c** (4.4%):
image
**tree-sitter-ocaml** (2.5%):
image
**tree-sitter-javascript** (2.2%):
image

### DIFF
--- a/crates/cli/src/fuzz.rs
+++ b/crates/cli/src/fuzz.rs
@@ -220,8 +220,9 @@ pub fn fuzz_language_corpus(
                 }
 
                 // Perform a random series of edits and reparse.
-                let mut undo_stack = Vec::new();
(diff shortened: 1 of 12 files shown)


clason:
Nim and systemverilog are the worst offenders; if they get even a few percent faster, that's a win for me.

WillLillis:
> Nim and systemverilog are the worst offenders; if they get even a few percent faster, that's a win for me.

Not much of a difference for Nim:
image
It looks like there's a boost for system verilog. Not sure why the branch misses increased, though:
image

PR #4578, opened 2025-07-09
feat(bindings): support building Rust package for Wasm

d-e-s-o:
Tree-sitter claims to support building for WebAssembly and yet on every step of the way there are road blocks if you want something more than download a pre-built WASM binary blob. That people struggle with it is obvious from the various discussions surrounding the matter, e.g.:
- https://github.com/tree-sitter/tree-sitter/discussions/1550
- https://github.com/tree-sitter/tree-sitter/discussions/2010
- https://github.com/tree-sitter/tree-sitter/discussions/1024

It's not just the core library: even languages fail to build for the target (certainly their Rust bindings). As it turns out it's actually not that hard to get going and this change boils down the various recommendations from some of the discussions into making generated language grammars build for the `wasm32-unknown-unknown` target out of the box. All that has to be done is to add and make known a minimal sysroot with header files that make certain types known and stub out core C allocation functions.
Note that the sized type definitions are 32 bit specific and will likely need adjustment once 64 bit WebAssembly is being used more broadly. Given the `wasm32-unknown-unknown` check this should be obvious and can be postponed until `wasm64-unknown-unknown` [0] is more widely used, I think.

[0] https://doc.rust-lang.org/nightly/rustc/platform-support/wasm64-unknown-unknown.html

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -69,6 +69,10 @@ const RUST_BINDING_VERSION_PLACEHOLDER: &str = "RUST_BINDING_VERSION";
 const LIB_RS_TEMPLATE: &str = include_str!("./templates/lib.rs");
 const BUILD_RS_TEMPLATE: &str = include_str!("./templates/build.rs");
 const CARGO_TOML_TEMPLATE: &str = include_str!("./templates/_cargo.toml");
+const SYSROOT_STDINT_TEMPLATE: &str = include_str!("./templates/stdint.h");
+const SYSROOT_STDLIB_TEMPLATE: &str = include_str!("./templates/stdlib.h");
+const SYSROOT_STRING_TEMPLATE: &str = include_str!("./templates/string.h");
+const SYSROOT_WCTYPE_TEMPLATE: &str = include_str!("./templates/wctype.h");
 
 const INDEX_JS_TEMPLATE: &str = include_str!("./templates/index.js");
 const INDEX_D_TS_TEMPLATE: &str = include_str!("./templates/index.d.ts");
@@ -351,9 +355,31 @@ pub fn generate_grammar_files(
                 generate_file(path, LIB_RS_TEMPLATE, language_name, &generate_opts)
             })?;
 
-            missing_path(path.join("build.rs"), |path| {
-                generate_file(path, BUILD_RS_TEMPLATE, language_name, &generate_opts)
-            })?;
+            missing_path_else(path.join("build.rs"),
+                allow_update,
+                |path| {
+                    generate_file(path, BUILD_RS_TEMPLATE, language_name, &generate_opts)
+                },
+                |path| {
+                    let mut contents = fs::read_to_string(path)?;
+                    if !contents.contains("wasm32-unknown-unknown") {
+                        contents = contents
+                            .replace(
+                                "c_config.flag(\"-utf-8\");\n",
+                                indoc!{ r#"
+                                c_config.flag("-utf-8");
+
+                                    if std::env::var("TARGET").unwrap() == "wasm32-unknown-unknown" {
+                                        let sysroot_dir = std::path::Path::new("bindings/rust/wasm-sysroot");
+                                        c_config.include(sysroot_dir);
+                                    }
+                                "#}
+                            );
+                        write_file(path, contents)?;
+                    }
+                    Ok(())
+                },
+            )?;
 
             missing_path_else(
                 repo_path.join("Cargo.toml"),
@@ -374,6 +400,22 @@ pub fn generate_grammar_files(
                     Ok(())
                 },
             )?;
+
+            missing_path(path.join("wasm-sysroot"), create_dir)?.apply(|path| {
+                missing_path(path.join("stdint.h"), |path| {
+                    generate_file(path, SYSROOT_STDINT_TEMPLATE, language_name, &generate_opts)
+                })?;
+                missing_path(path.join("stdlib.h"), |path| {
+                    generate_file(path, SYSROOT_STDLIB_TEMPLATE, language_name, &generate_opts)
+                })?;
+                missing_path(path.join("string.h"), |path| {
+                    generate_file(path, SYSROOT_STRING_TEMPLATE, language_name, &generate_opts)
+                })?;
+                missing_path(path.join("wctype.h"), |path| {
+                    generate_file(path, SYSROOT_WCTYPE_TEMPLATE, language_name, &generate_opts)
+                })?;
+                Ok(())
+            })?;
 
             Ok(())
         })?;
--- a/crates/cli/src/templates/build.rs
+++ b/crates/cli/src/templates/build.rs
@@ -7,6 +7,11 @@ fn main() {
     #[cfg(target_env = "msvc")]
     c_config.flag("-utf-8");
 
+    if std::env::var("TARGET").unwrap() == "wasm32-unknown-unknown" {
+        let sysroot_dir = std::path::Path::new("bindings/rust/wasm-sysroot");
+        c_config.include(sysroot_dir);
+    }
+
     let parser_path = src_dir.join("parser.c");
     c_config.file(&parser_path);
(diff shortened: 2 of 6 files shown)


d-e-s-o:
I am not really sure how best to test this change. Is there something in CI building an instantiated template or so that I missed?

crates/cli/src/templates/build.rs
```diff
@@ -4,6 +4,12 @@ fn main() {
...
     c_config.std("c11").include(src_dir);
 
+    // Set a minimal C sysroot if building for wasm32-unknown-unknown.
+    if std::env::var("TARGET").unwrap() == "wasm32-unknown-unknown" {
+        let sysroot_dir = std::path::Path::new("bindings/rust/wasm-sysroot");
+        c_config.include(sysroot_dir);
+    }
+
     #[cfg(target_env = "msvc")]
     c_config.flag("-utf-8");
```
ObserverOfTime:
```suggestion
    if cfg!(target_family = "wasm") {
        let sysroot = std::path::Path::new("bindings/rust/wasm-sysroot");
        c_config.include(sysroot);
    }

    if cfg!(target_env = "msvc") {
        c_config.flag("-utf-8");
    }
```

crates/cli/src/init.rs
```diff
@@ -369,9 +371,19 @@ pub fn generate_grammar_files(
...
+            missing_path_else(path.join("build.rs"),
+                allow_update,
+                |path| {
+                    generate_file(path, BUILD_RS_TEMPLATE, language_name, &generate_opts)
+                },
+                |path| {
+                    let contents = fs::read_to_string(path)?;
+                    if !contents.contains("wasm32-unknown-unknown") {
+                        generate_file(path, BUILD_RS_TEMPLATE, language_name, &generate_opts)?;
+                    }
```
```suggestion
                    let mut contents = fs::read_to_string(path)?;
                    if !contents.contains("wasm32-unknown-unknown") {
                        contents = contents
                            .replace(
                                indoc! {r#"
                                #[cfg(target_env = "msvc")]
                                c_config.flag("-utf-8");
                                "#},
                                indoc! {r#"
                                if cfg!(target_family = "wasm") {
                                    let sysroot = std::path::Path::new("bindings/rust/wasm-sysroot");
                                    c_config.include(sysroot);
                                }

                                if cfg!(target_env = "msvc") {
                                    c_config.flag("-utf-8");
                                }
                                "#}
                            );
                        write_file(path, contents)?;
                    }
```

ObserverOfTime (changes requested):

crates/cli/src/templates/stdlib.h
ObserverOfTime, line 1: You probably want to include all of [these](https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/wasm/stdlib-symbols.txt) symbols.

d-e-s-o:
Thanks for the review. The suggestions seem questionable:
- For one, the update path doesn't trigger anymore correctly (probably at least in part due to `indoc` usage causing indentation mismatch on multi-line string, but it's not all).
- Second, the suggested `cfg` usage in a build script seems wrong. The build script is built for the host, not the target. Compilation for WASM will, in the vast majority of cases, be a cross-compilation. It may work for the toolchain type in general (by virtue of it not changing), but not for the architecture (the very point of cross compuilation). Even if it were working, as I explained in the change description, the sysroot in its current form is 32 bit specific (I am not sure which `#define`s would be set etc. to make this conditional in the header [don't have `wasm64-unknown-unknown` support at hand] and shipping a separate sysroot for 64 bit seems questionable), whereas now we silently accept anything WASM.
- Lastly, it's not clear why we are touching stuff completely unrelated to the changes at hand, i.e., the `msvc` logic.

That being said, I updated the sysroot to include the symbols you hinted at. Thanks for the pointer.

ObserverOfTime (commented):
1. Don't replace the entire file.
2. Add parameter names to the declarations.
3. What about other `wasm32` targets?

d-e-s-o:
> 1. Don't replace the entire file.

Done.

> 2. Add parameter names to the declarations.

Why?

> 3. What about other `wasm32` targets?

Can be added as need. Can't test them, don't have them.

crates/cli/src/templates/wctype.h
ObserverOfTime, line 1: `wint_t` is not defined anywhere.

crates/cli/src/init.rs
```diff
@@ -369,9 +373,29 @@ pub fn generate_grammar_files(
...
+                    if !contents.contains("wasm32-unknown-unknown") {
+                        contents = contents
+                            .replace(
+                                r#"c_config.flag("-utf-8");"#,
+                                r#"c_config.flag("-utf-8");
+
+    if std::env::var("TARGET").unwrap() == "wasm32-unknown-unknown" {
+        let sysroot_dir = std::path::Path::new("bindings/rust/wasm-sysroot");
+        c_config.include(sysroot_dir);
+    }"#
```
ObserverOfTime: Use `indoc!`.

d-e-s-o: That results in an unindented replacement, which isn't what `build.rs` contains.

amaanq:
@d-e-s-o wouldn't this still be unusable without providing stub implementations for the libc functions used? I would think compiling would succeed now, but actually using the parser wouldn't work for parsers that have a scanner that make use of libc.

d-e-s-o:
> @d-e-s-o wouldn't this still be unusable without providing stub implementations for the libc functions used? I would think compiling would succeed now, but actually using the parser wouldn't work for parsers that have a scanner that make use of libc.

The change allows one to compile the language parser to a WASM object, which otherwise fails early because of missing headers included by the generated `parser.h`. My guess would be that `libc` function references would be left unresolved at that point and resolution would happen in the final link step with `tree-sitter`. I can see the various allocation functions to be provided by `tree-sitter`. For others, it's possible that they'd need to be provided then, which I'd think is fine and reasonable. It's also possible they'd never be referenced in a language parser at all (`tree-sitter` parser generator folks should know that). Looking at the language parser I tested with (for BPF C), I don't see any references to any of the functions, so it's possible it works out of the box because of that. I also checked `cpp` and `haskell` `parser.c` files, none of which seem to have any traces of these functions either.

From the looks of it, in TypeScript binding land things are similar in that declarations are present but no implementations: https://github.com/tree-sitter/tree-sitter/blob/56325d2a3b1e27cb70179734bb31ceb0e6a5f0b9/lib/binding_web/lib/web-tree-sitter.d.ts#L183-L198
(though I could be misunderstanding; not a TypeScript expert)

That it does in fact work can be seen [here](https://d-e-s-o.github.io/bpflint/) for example, which uses `tree-sitter` for parsing files.

amaanq:
Wait, you're building the *Rust* crate for a given parser to WASM and use that with web-tree-sitter? Why? You shouldn't have to do that, you can simply build a wasm file that web-tree-sitter can consume with the CLI, and these wasms are uploaded to npm.

I was meaning to ask how you were using these rust wasm parsers, because currently the core library's Rust bindings doesn't compile to wasm either due to missing libc headers + implemenetations, but it seems like you're using web-tree-sitter? (correct me if I'm wrong)

d-e-s-o:
No, I am not using `web-tree-sitter`. I am building a Rust project that uses tree-sitter for WASM.

amaanq:
How are you using the core library? I wouldn't think that compiles to wasm

d-e-s-o:
> How are you using the core library? I wouldn't think that compiles to wasm

I use the transpiled Rust version, `tree-sitter-c2rust`.

clason:
Might have wanted to lead with that.

amaanq:
Oh right, I had forgotten about that. I think it'd be great to support wasm32-unknown-unknown in general, though I think we'll have to write stubs for the blessed libc functions. I'd also like to eventually work on providing support to compile the Rust + C library to wasm. For this PR though, I think we should implement everything in `stdlib-symbols.txt`, since this is what's available to authors of external scanners. I would be willing to work on that if you'd like.

d-e-s-o:
Sure, go ahead.

PR #4591, opened 2025-07-11
feat(rust): Add direct link to documentation

d-e-s-o:
Make the URL to the documentation of the crate known so that it will be easier to click through to it from a crates.io search [0].

[0] https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -8,6 +8,7 @@ rust-version.workspace = true
 readme = "README.md"
 homepage.workspace = true
 repository.workspace = true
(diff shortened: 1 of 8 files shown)


WillLillis (changes requested):
This isn't the correct place, that link points to https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_rust. That being said, it looks like most/all of the project's crates are missing their documentation links. Care to make those changes instead?

d-e-s-o:
Good point, updated.

WillLillis (approved):

PR #4601, opened 2025-07-13
feat: Only evaluate `grammar.js` to `grammar.json`

wetneb:
This adds an `--evaluate-only` option to `tree-sitter generate` so that it only does the evaluation of `grammar.js` to `src/grammar.json`, without continuing on with the generation of `src/parser.c` and related files.

It's a follow-up to #4580.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -133,6 +133,9 @@ struct Generate {
         default_value = "node"
     )]
     pub js_runtime: Option<String>,
+    /// Only generate `grammar.json` by evaluating `grammar.js`, but do not generate `parser.c` and related files afterwards
+    #[arg(long)]
+    pub evaluate_only: bool,
 }
 
 #[derive(Args)]
@@ -807,6 +810,7 @@ impl Generate {
             abi_version,
             self.report_states_for_rule.as_deref(),
             self.js_runtime.as_deref(),
+            self.evaluate_only,
         ) {
             if self.json {
                 eprintln!("{}", serde_json::to_string_pretty(&err)?);
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -85,6 +85,17 @@ struct Init {
     pub grammar_path: Option<PathBuf>,
 }
 
+#[derive(Clone, Debug, Default, ValueEnum, PartialEq, Eq)]
+enum GenerationStage {
+    /// Generate `grammar.json` and `node-types.json`
+    Json,
+    /// Generate `parser.c` and related files
+    #[default]
+    Parser,
+    /// Compile to a library
+    Lib,
+}
+
 #[derive(Args)]
 #[command(alias = "gen", alias = "g")]
 struct Generate {
@@ -107,8 +118,12 @@ struct Generate {
                 )
     )]
     pub abi_version: Option<String>,
-    /// Compile all defined languages in the current dir
-    #[arg(long, short = 'b')]
+    /// Which generation stage to end after
+    #[arg(long)]
+    #[clap(value_enum, default_value_t=GenerationStage::Parser)]
+    pub stage: GenerationStage,
+    /// Deprecated: use --stage=lib.
+    #[arg(long, short = 'b', conflicts_with = "stage")]
     pub build: bool,
     /// Compile a parser in debug mode
     #[arg(long, short = '0')]
@@ -133,9 +148,6 @@ struct Generate {
         default_value = "node"
     )]
     pub js_runtime: Option<String>,
-    /// Only generate `grammar.json` by evaluating `grammar.js`, but do not generate `parser.c` and related files afterwards
-    #[arg(long)]
-    pub evaluate_only: bool,
 }
 
 #[derive(Args)]
@@ -803,6 +815,11 @@ impl Generate {
                         version.parse().expect("invalid abi version flag")
                     }
                 });
+        if self.build {
+            // TODO: remove the `--build` argument in 0.27
+            // TODO: migrate to `warn!` once https://github.com/tree-sitter/tree-sitter/pull/4604 is merged
+            eprintln!("Warning: --build is deprecated, use --stage=lib instead");
+        }
         if let Err(err) = tree_sitter_generate::generate_parser_in_directory(
             current_dir,
             self.output.as_deref(),
@@ -810,7 +827,7 @@ impl Generate {
             abi_version,
             self.report_states_for_rule.as_deref(),
             self.js_runtime.as_deref(),
-            self.evaluate_only,
+            self.stage != GenerationStage::Json,
         ) {
             if self.json {
                 eprintln!("{}", serde_json::to_string_pretty(&err)?);
@@ -821,7 +838,7 @@ impl Generate {
                 Err(anyhow!(err.to_string())).with_context(|| "Error when generating parser")?;
             }
         }
-        if self.build {
+        if self.stage == GenerationStage::Lib || self.build {
             if let Some(path) = self.libdir {
                 loader = loader::Loader::with_parser_lib_path(path);
             }
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -196,9 +196,8 @@ where
     let src_path = out_path.map_or_else(|| repo_path.join("src"), |p| p.into());
     let header_path = src_path.join("tree_sitter");
 
-    // Ensure that the output directories exist.
+    // Ensure that the output directory exists
     fs::create_dir_all(&src_path)?;
-    fs::create_dir_all(&header_path)?;
 
     if grammar_path.file_name().unwrap() != "grammar.json" {
         fs::write(src_path.join("grammar.json"), &grammar_json).map_err(|e| {
@@ -210,12 +209,13 @@ where
     }
 
     // If our job is only to generate `grammar.json` and not `parser.c`, stop here.
+    let input_grammar = parse_grammar(&grammar_json)?;
+
     if !generate_parser {
+        let node_types_json = generate_node_types_from_grammar(&input_grammar)?.node_types_json;
+        write_file(&src_path.join("node-types.json"), node_types_json)?;
         return Ok(());
     }
-
-    // Parse and preprocess the grammar.
-    let input_grammar = parse_grammar(&grammar_json)?;
 
     let semantic_version = read_grammar_version(&repo_path)?;
 
@@ -238,6 +238,7 @@ where
 
     write_file(&src_path.join("parser.c"), c_code)?;
     write_file(&src_path.join("node-types.json"), node_types_json)?;
+    fs::create_dir_all(&header_path)?;
     write_file(&header_path.join("alloc.h"), ALLOC_HEADER)?;
     write_file(&header_path.join("array.h"), ARRAY_HEADER)?;
     write_file(&header_path.join("parser.h"), tree_sitter::PARSER_HEADER)?;
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     env, fs,
     io::Write,
     path::{Path, PathBuf},
@@ -7,7 +8,9 @@ use std::{
 };
 
 use anyhow::Result;
+use node_types::VariableInfo;
 use regex::{Regex, RegexBuilder};
+use rules::{Alias, Symbol};
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -25,7 +28,7 @@ mod tables;
 
 use build_tables::build_tables;
 pub use build_tables::ParseTableBuilderError;
-use grammars::InputGrammar;
+use grammars::{InlinedProductionMap, InputGrammar, LexicalGrammar, SyntaxGrammar};
 pub use node_types::{SuperTypeCycleError, VariableInfoError};
 use parse_grammar::parse_grammar;
 pub use parse_grammar::ParseGrammarError;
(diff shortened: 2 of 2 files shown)


clason:
I'd prefer `--json-only`, as that makes it clearer what happens.

wetneb:
I thought it would cause confusion with `--json`, but as you like… Let's wait for others to chime in and I'll update it to whatever gathers consensus.

clason:
Yeah, that flag is a bit poorly named. The issue is that "evaluate" is a bit of a meaningless term as it doesn't express the _output_.

WillLillis:
What if we pulled the different "output" types into a single flag expressed as an enum, maybe something like

```rust
enum GenerateArtifact {
    JSON,
    Parser,
    Lib,
}
```

The `build` flag could be removed and replaced with the `Lib` variant. The default `generate` behavior could stay the same, but a user could then do something like 

```
ts g --artifact=json
```

to just create the `grammar.json` file.

clason:
Hmm. I like that, but if we rethink this, we should also include the _input_ (grammar.js, grammar.json, parser.c) into the discussion. (And I'm not excited about the name "artifact"; I feel we can come up with a better term.)

(I could simplify the nvim-treesitter pipeline if we could `ts b -o foo.so` straight from whatever is available -- optionally even skipping writing the intermediary files to disk, for hopefully another performance boost.)

wetneb:
Perhaps "stage" could be a better term than "artifact"?

ObserverOfTime:
> so that it only does the evaluation of `grammar.js` to `src/grammar.json`

And `src/node-types.json`?

wetneb:
`src/node-types.json` is generated alongside `src/parser.c` and other files (`src/tree-sitter/*`) at the moment. Would you like it to behave differently?

clason:
Ideally, yes.

wetneb:
Could you tell me how exactly? I can think of multiple ways:
* generating it at the same stage as `grammar.json`
* generating it in its own stage, between `grammar.json` and `parser.c`

clason:
I don't think it makes sense to add a separate step for this. It's closer to grammar.json, so it should be included in the same step.

wetneb:
I've changed the CLI option to `--artifact=json/parser/lib` as suggested by @WillLillis above. It required a bit more of a refactoring but I guess the diff is still not too daunting.

clason:
I like your suggestion of `stage` better -- if only because then I wouldn't have to worry about AE vs. BE spelling ;)

Have you tested if this behaves correctly when giving an argument (such as `src/grammar.json`) to `generate`?

ObserverOfTime:
Maybe `generate` shouldn't build the library at all since the `build` command exists.
Though other commands will still do so implicitly…

crates/cli/src/main.rs
```diff
@@ -107,7 +118,11 @@ struct Generate {
                 )
     )]
     pub abi_version: Option<String>,
-    /// Compile all defined languages in the current dir
+    /// Which type of artifact to generate from the grammar's source
+    #[arg(long)]
+    #[clap(value_enum, default_value_t=GenerateArtifact::Parser)]
```
WillLillis:
```suggestion
    #[clap(value_enum, default_value_t=GenerateArtifact::Parser,  conflicts_with = "build")]
```

crates/cli/src/main.rs
```diff
@@ -85,6 +85,17 @@ struct Init {
     pub grammar_path: Option<PathBuf>,
 }
 
+#[derive(Clone, Debug, Default, ValueEnum, PartialEq, Eq)]
+enum GenerateArtifact {
```
WillLillis: I'm open to other suggestions, but I think Clason's idea of "stage" is better than artifact for this.

crates/cli/src/main.rs
```diff
@@ -107,7 +118,11 @@ struct Generate {
                 )
     )]
     pub abi_version: Option<String>,
-    /// Compile all defined languages in the current dir
+    /// Which type of artifact to generate from the grammar's source
+    #[arg(long)]
+    #[clap(value_enum, default_value_t=GenerateArtifact::Parser)]
+    pub artifact: GenerateArtifact,
+    /// Deprecated: use --artifact=lib.
     #[arg(long, short = 'b')]
```
WillLillis:
```suggestion
    #[arg(long, short = 'b', conflicts_with = "artifact")]
```

crates/cli/src/main.rs
```diff
@@ -800,13 +815,21 @@ impl Generate {
                         version.parse().expect("invalid abi version flag")
                     }
                 });
+        let artifact = if self.build {
```
WillLillis: This assignment to `artifact` can be removed, but the warning print should stay.

(Just use `self.artifact` below)

crates/generate/src/generate.rs
```diff
@@ -152,6 +164,7 @@ pub fn generate_parser_in_directory<T, U, V>(
     mut abi_version: usize,
     report_symbol_name: Option<&str>,
     js_runtime: Option<&str>,
+    json_only: bool,
```
WillLillis: I think just passing in the stage here would be better for readability.

crates/generate/src/generate.rs
```diff
@@ -199,6 +211,15 @@ where
...
     let input_grammar = parse_grammar(&grammar_json)?;
 
+    if json_only {
+        let node_types_json = generate_node_types_from_grammar(&input_grammar)?.node_types_json;
+        write_file(&src_path.join("node-types.json"), node_types_json)?;
+        return Ok(());
+    }
+
+    // Ensure that the directory for tree-sitter headers exists
+    fs::create_dir_all(&header_path)?;
```
WillLillis: Is there anything preventing this from being moved down to right above the series of `write_file` calls?

WillLillis (changes requested):
Looks good, I left a few comments.

crates/generate/src/generate.rs
wetneb: I think it would imply moving the declaration of the `enum` to this crate. For it to `#[derive(ValueEnum)]` we would also need to add `clap` as a dependency to this crate, which doesn't sound so appropriate to me, does it? Or can you see a better solution?

crates/cli/src/main.rs
wetneb: So that would mean that the `--build` CLI argument won't have any effect anymore, right? In that case, the warning should perhaps be turned into an error?

clason: No, we want to keep `--build` working. It's important to have a proper deprecation cycle, as this is used in consumer projects.

You could add a `todo: remove in 0.27` comment though.

crates/generate/src/generate.rs
WillLillis: Ah good point. We definitely don't want to add the `clap` dependency to `generate`. We could make a wrapper type like this:

<details>

```diff
diff --git a/crates/cli/src/main.rs b/crates/cli/src/main.rs
index 89f254b0..94bfdea1 100644
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -85,7 +85,11 @@ struct Init {
     pub grammar_path: Option<PathBuf>,
 }
 
-#[derive(Clone, Debug, Default, ValueEnum, PartialEq, Eq)]
+/// Defines the stages of parser generation that should be performed.
+///
+/// This serves as a serializable wrapper for [`tree_sitter_generate::Stage`], because
+/// this needs to be passed in by the CLI tool.
+#[derive(Copy, Clone, Debug, Default, ValueEnum, PartialEq, Eq)]
 enum GenerationStage {
     /// Generate `grammar.json` and `node-types.json`
     Json,
@@ -96,6 +100,16 @@ enum GenerationStage {
     Lib,
 }
 
+impl Into<tree_sitter_generate::Stage> for GenerationStage {
+    fn into(self) -> tree_sitter_generate::Stage {
+        match self {
+            GenerationStage::Json => tree_sitter_generate::Stage::Json,
+            GenerationStage::Parser => tree_sitter_generate::Stage::Parser,
+            GenerationStage::Lib => tree_sitter_generate::Stage::Lib,
+        }
+    }
+}
+
 #[derive(Args)]
 #[command(alias = "gen", alias = "g")]
 struct Generate {
@@ -829,7 +843,7 @@ impl Generate {
             abi_version,
             self.report_states_for_rule.as_deref(),
             self.js_runtime.as_deref(),
-            stage == GenerationStage::Json,
+            stage.into(),
         ) {
             if self.json {
                 eprintln!("{}", serde_json::to_string_pretty(&err)?);
diff --git a/crates/generate/src/generate.rs b/crates/generate/src/generate.rs
index d2fcb258..26774d1d 100644
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -157,6 +157,18 @@ impl From<semver::Error> for JSError {
     }
 }
 
+/// Defines the stages of parser generation that should be performed.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub enum Stage {
+    /// Generate `grammar.json` and `node-types.json`
+    Json,
+    /// Generate `parser.c` and related files
+    #[default]
+    Parser,
+    /// Compile to a library
+    Lib,
+}
+
 pub fn generate_parser_in_directory<T, U, V>(
     repo_path: T,
     out_path: Option<U>,
@@ -164,7 +176,7 @@ pub fn generate_parser_in_directory<T, U, V>(
     mut abi_version: usize,
     report_symbol_name: Option<&str>,
     js_runtime: Option<&str>,
-    json_only: bool,
+    stage: Stage,
 ) -> GenerateResult<()>
 where
     T: Into<PathBuf>,
@@ -211,7 +223,7 @@ where
     // Parse and preprocess the grammar.
     let input_grammar = parse_grammar(&grammar_json)?;
 
-    if json_only {
+    if stage == Stage::Json {
         let node_types_json = generate_node_types_from_grammar(&input_grammar)?.node_types_json;
         write_file(&src_path.join("node-types.json"), node_types_json)?;
         return Ok(());
```

</details>

However after thinking on it some more I think going with a boolean parameter probably makes the most sense for this public API. I'm a still a bit unsure readability wise, since it's not super clear what `json_only` means by reading the function signature. How about instead we pass in `generate_parser` like so:

```diff
diff --git a/crates/generate/src/generate.rs b/crates/generate/src/generate.rs
index d2fcb258..7ddbb1b1 100644
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -164,7 +164,7 @@ pub fn generate_parser_in_directory<T, U, V>(
     mut abi_version: usize,
     report_symbol_name: Option<&str>,
     js_runtime: Option<&str>,
-    json_only: bool,
+    generate_parser: bool,
 ) -> GenerateResult<()>
 where
     T: Into<PathBuf>,
@@ -211,7 +211,7 @@ where
     // Parse and preprocess the grammar.
     let input_grammar = parse_grammar(&grammar_json)?;
 
-    if json_only {
+    if !generate_parser {
         let node_types_json = generate_node_types_from_grammar(&input_grammar)?.node_types_json;
         write_file(&src_path.join("node-types.json"), node_types_json)?;
         return Ok(());
```

clason: maybe simply `--parser=false` (default `true`)?

WillLillis (approved):
Changes LGTM. When you get a chance, could you squash the last 4 commits into the appropriate changes from the first 3? Happy to merge once that's cleaned up, thanks for the PR!

crates/generate/src/generate.rs
natalejoeraco1: > I think it would imply moving the declaration of the `enum` to this crate. For it to `#[derive(ValueEnum)]` we would also need to add `clap` as a dependency to this crate, which doesn't sound so appropriate to me, does it? Or can you see a better solution?

amaanq:
@wetneb looking back at this (and noticing how the lib variant makes no sense....), is this even needed? I would recommend grammars commit `grammar.json`, because then a JS runtime isn't a requirement (until 0.26 is out with the native JS runtime). Even if you decide not to, I don't really get how this helps.

wetneb:
This PR was mostly designed to enable https://github.com/tree-sitter/tree-sitter/pull/4646. The use cases I have in mind are:
* for a grammar user, make it possible to compile a grammar by just typing `make`, regardless of the files that are already included in the archive or repository (assuming they have a JS runtime)
* for a grammar maintainer, in the case where the `grammar.json` is tracked in the repository, this also makes it possible to update it easily, without generating the `parser.c` in the same go (which can take quite some time).

clason:
The makefile is its own thing, we don't really care about that (and the makefile can easily call generate then build, so no need to change the CLI). 

Do you have benchmarks that show that skipping the parser generation makes a significant difference (order of minutes)?

wetneb:
Changing the Makefile only was indeed my first approach, in https://github.com/tree-sitter/tree-sitter/pull/4580. It was requested by @WillLillis ([here](https://github.com/tree-sitter/tree-sitter/pull/4580#issuecomment-3057364156)) and @ObserverOfTime ([here](https://github.com/tree-sitter/tree-sitter/pull/4580#issuecomment-3063545935)) that we first change the CLI to avoid the double generation of artifacts, which also makes sense to me.

If you would rather do it differently (or not at all), I'm sure you can revert my changes - no hard feelings about that on my side :)

amaanq:
Yeah I'd rather update the Makefile/CMake, I'm not totally against having an early exit for `grammar.json/node-types.json`, but I think if updating the Makefile solves the use cases you outlined I'd rather do that. And since the lib variant makes no sense, I'd say a simple flag would suffice for it (if need be).

ObserverOfTime:
Some parsers take a rather long time to generate. Regardless of _where_ the bottleneck lies, running the entire generation process twice is not desirable. If we want to have separate steps in make/CMake (which I do like having), they need to be separate steps in the generate command as well.

amaanq:
Why not just unify it in the make/cmake files (why even generate twice)?

wetneb:
Can you give an example of the Makefile rules you would use? It would help me understand the solution you have in mind.

ObserverOfTime:
> Why not just unify it in the make/cmake files (why even generate twice)?

We used to generate from `src/grammar.json` but this prevented people from generating directly from `grammar.js`. 
And we can't properly support both without separating the generation steps.

clason:
I think we need to partially revert this; `--stage` was a (my) mistake. 

@amaanq Note that even with a native JS runtime, working from evaluated JSON as an intermediate state is still very valuable for grammars that extend other grammars (and hence require `npm` -- a no-go for downstream users like nvim-treesitter). So I'd still like to push for (and support) committing those files.

ObserverOfTime:
We should soft deprecate imports from `node_modules` in 0.27 and recommend submodules or vendoring.

PR #4604, opened 2025-07-13
refactor: Output warnings via the logger

wetneb:
Instead of writing warnings about grammar sources files on a mixture of the standard output / standard error streams, this makes them go through the logging infrastructure, which is made available to all CLI commands (instead of just the `generate` command). I've tried to preserve the output format of the warnings.

This PR isn't meant to be comprehensive, there are many more `(e)println!` statements that could be migrated, but I imagine this can be done in follow-up changes.

Also I noticed that the existing logging statements in the `generate` crate are written with the `info` level, but in my opinion they rather belong to `debug`. Moving them to this level would make it possible to set `info` as default logging level, which would make more sense to me.

This is a follow-up to #4567. Depending on which one is merged first, I'll update the other one accordingly.

### DIFF
--- a/crates/cli/src/logger.rs
+++ b/crates/cli/src/logger.rs
@@ -1,4 +1,4 @@
-use log::{LevelFilter, Log, Metadata, Record};
+use log::{Level, LevelFilter, Log, Metadata, Record};
 
 #[allow(dead_code)]
 struct Logger {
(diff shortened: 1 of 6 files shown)


crates/cli/src/logger.rs
```diff
@@ -11,20 +11,26 @@ impl Log for Logger {
...
     fn log(&self, record: &Record) {
-        eprintln!(
-            "[{}] {}",
-            record
-                .module_path()
-                .unwrap_or_default()
-                .trim_start_matches("rust_tree_sitter_cli::"),
-            record.args()
-        );
+        if record.level() == Level::Error {
```
ObserverOfTime: Use `match record.level()`.

clason:
> This PR isn't meant to be comprehensive, there are many more (e)println! statements that could be migrated, but I imagine this can be done in follow-up changes.

This change is welcome, but it would be more valuable if we could achieve complete coverage here -- this would allow us to ensure  consistency across the project.

For example, we might want to turn "unnecessary" warnings into `Hint`s, to match standard linter usage.

wetneb:
I find it difficult to reach complete coverage here. There are a lot of print statements out there:
```
$ grep -R println crates/ | wc -l
220
```
and more are being added by other PRs. So it would be more convenient for me to merge this PR as is, to at least make it possible to use proper logging in my other PRs.

Otherwise I propose to close this PR and let someone else tackle this.

clason:
Need not be across the whole project, but at least the `generate` crate. And we'll make sure other PRs use the logger as part of review.

wetneb:
Then let me close this PR and let someone else tackle this, as I won't have time to work on this soon.

clason:
Let's keep this open so it's tracked and someone else can pick up from it.

PR #4606, opened 2025-07-13
Remove src files from npm package

### LINKED ISSUE #4596: Bundling source files in npm package
### Problem

Why are we bundling all the source files in the web-tree-sitter npm package?
are they actually needed?

v2.24.7 is 372kB containing 6 files
https://www.npmjs.com/package/web-tree-sitter/v/0.24.7?activeTab=code

v2.25.0 is 5.78mb containing ~50 files
https://www.npmjs.com/package/web-tree-sitter/v/0.25.0?activeTab=code

a bunch of the src files even have errors
which is rather annoying if TS finds them
Image
Image
Image

### Steps to reproduce

https://www.npmjs.com/package/web-tree-sitter/v/0.25.0?activeTab=code

### Expected behavior

For the npm package to only ship necessary files

tree-sitter 0.25.7

### Operating system/version

Windows 11

RedCMD:
Fixes #4596

* removed all `.c` and `.h` files
* removed most of the useless `.map` files (that is what `/debug` is for)
* merged `web-tree-sitter.d.ts` and `web-tree-sitter.d.cts` (as they are the same)
* normalized the `repository` string to an `object`
```sh
npm warn publish npm auto-corrected some errors in your package.json when publishing.  Please run "npm pkg fix" to address these errors.
npm warn publish errors corrected:
npm warn publish "repository" was changed from a string to an object
npm warn publish "repository.url" was normalized to "git+https://github.com/tree-sitter/tree-sitter.git"
```

### DIFF
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -2,7 +2,10 @@
   "name": "web-tree-sitter",
   "version": "0.26.0",
   "description": "Tree-sitter bindings for the web",
-  "repository": "https://github.com/tree-sitter/tree-sitter",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tree-sitter/tree-sitter.git"
+  },
   "homepage": "https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_web",
   "license": "MIT",
   "author": {
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -23,7 +23,7 @@
         "default": "./web-tree-sitter.js"
       },
       "require": {
-        "types": "./web-tree-sitter.d.cts",
+        "types": "./web-tree-sitter.d.ts",
         "default": "./web-tree-sitter.cjs"
       }
     },
@@ -34,7 +34,7 @@
         "default": "./debug/web-tree-sitter.js"
       },
       "require": {
-        "types": "./web-tree-sitter.d.cts",
+        "types": "./web-tree-sitter.d.ts",
         "default": "./debug/web-tree-sitter.cjs"
       }
     },
@@ -50,11 +50,8 @@
   "files": [
     "README.md",
     "web-tree-sitter.cjs",
-    "web-tree-sitter.cjs.map",
     "web-tree-sitter.js",
-    "web-tree-sitter.js.map",
     "web-tree-sitter.wasm",
-    "web-tree-sitter.wasm.map",
     "debug/web-tree-sitter.cjs",
     "debug/web-tree-sitter.cjs.map",
     "debug/web-tree-sitter.js",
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
> merged `web-tree-sitter.d.ts` and `web-tree-sitter.d.cts` (as they are the same)

What was the reasoning behind this separation in the first place? @savetheclocktower

savetheclocktower:
I forgot, but #4229 reminded me:

> Once I changed the format of the `"exports"` field, arethetypeswrong chided me for having one `.d.ts` file — an ESM module's type definition, since we do `"type": "module"` in `package.json` — pulling double duty as also documenting the `.cjs` file. This is supposedly bad [for the reasons explained here](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md); and if someone writes decent prose in an authoritative voice, I tend to believe it.

savetheclocktower:
It sounds as though, if a package declares `"type": "module"`, a type definition file with a `.d.cts` extension is the only way to tell TypeScript that it's actually dealing with a CommonJS module. So even though the contents of the two type definition files are entirely the same, they can't be consolidated without TypeScript incorrectly assuming that the thing being imported is an ESM module.

I don't know if this is true, but that's what that URL seems to claim. And if it were true, that'd be an annoying enough consequence (needing to do `await import` on the dependency instead of `require` for no particular reason) that we'd want to avoid it.

RedCMD:
atm neither file uses require/import
so having both is not needed

tho in the future, it might not be the case
at which point `.d.cts` can just be re-added

savetheclocktower:
> atm neither file uses require/import so having both is not needed
> 
> tho in the future, it might not be the case at which point `.d.cts` can just be re-added

It's not about what they use internally; it's about how they're imported into another project. It is important to have `web-tree-sitter` be importable via `require` instead of having TypeScript think that it can only be imported via `await import`.

RedCMD:
sure
but I'm not seeing the problems you claim

VSCode extensions are commonJS
even if I explicitly set `"module": "commonjs", "target": "es5"` in `tsconfig.json`
there still is no errors when importing `web-tree-sitter`
even still works in VSCode Web (which is even more restrictive)

savetheclocktower:
> sure but I'm not seeing the problems you claim
> 
> VSCode extensions are commonJS even if I explicitly set `"module": "commonjs", "target": "es5"` in `tsconfig.json` there still is no errors when importing `web-tree-sitter` even still works in VSCode Web (which is even more restrictive)

I don't know what conditions are necessary to reproduce what that page claims, or even if it's still valid. I did it the way I did it originally because I trust [arethetypeswrong](https://arethetypeswrong.github.io/) more than I trust my own intuitions about TypeScript `.d.ts` files.

That page explains in great detail why someone in _our exact scenario_ should publish two different definition files with different extensions. I suspect that, if the status quo changes and that's no longer necessary, arethetypeswrong would update their documentation to reflect that.

Since the only possible “harm” being prevented here is two files with the same extension, I don't really see the point in changing it.

verhovsky:
Removing .map files is nonsense. Error traces are useful, you shouldn't need to rewrite import statements every time you want to find out why you're getting an error. Imagine if every npm library did that.

ObserverOfTime:
I assume removing _just_ the source files is fine now that #4671 was merged? @verhovsky

verhovsky:
@ObserverOfTime yea

PR #4607, opened 2025-07-13
Fix: add wasm32 support to portable/endian.h

Alex-Aron:
When targeting wasm32-unknown-unknown, the build currently fails with:
#error platform not supported
in src/portable/endian.h, even though WebAssembly is little-endian

This PR adds support for __wasm__/ __wasm32__ by adding the two to the sequence of defined statements

### DIFF
--- a/lib/src/portable/endian.h
+++ b/lib/src/portable/endian.h
@@ -24,7 +24,8 @@
     defined(__CYGWIN__) || \
     defined(__MSYS__) || \
     defined(__EMSCRIPTEN__) || \
-    defined(__wasi__)
+    defined(__wasi__) || \
+    defined(__wasm32__)
 
 #if defined(__NetBSD__)
 #define _NETBSD_SOURCE 1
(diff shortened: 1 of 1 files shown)


lib/src/portable/endian.h
```diff
@@ -24,7 +24,9 @@
     defined(__CYGWIN__) || \
     defined(__MSYS__) || \
     defined(__EMSCRIPTEN__) || \
-    defined(__wasi__)
+    defined(__wasi__) || \
+    defined(__wasm32__) || \
```
ObserverOfTime: Is `__wasm32__` also defined in WASI and wasm64?

Alex-Aron: oops, just __wasm__ should be enough .

ObserverOfTime: No, just `__wasm32__` should be enough for all 3 I think. But it doesn't matter either way.

Alex-Aron: Want me to replace the 3 with __wasm32__?

clason: Not unless you test ;) Otherwise just add what you know works for wasm32 and leave refactors for later. (It is important not to regress for platforms you don't know anything about.)

Alex-Aron: ok i will leave it as is then 👍

ObserverOfTime (approved):

PR #4609, opened 2025-07-14
fix(bindings): ignore go.sum

ObserverOfTime:
Alternatively we can unignore all lockfiles, or just `package-lock.json` (required for child grammars).


WillLillis (commented):
I don't have much experience in Go, but it looks like `go.sum` should be included? 

https://go.dev/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file

It makes sense to me to include the lock files.

WillLillis (approved):

PR #4611, opened 2025-07-14
Update out of date web-tree-sitter.d.ts type files

RedCMD:
_from #4600_

simply ran `npm run build:dts`
 the `lib/binding_web/web-tree-sitter.d.cts` and `lib/binding_web/web-tree-sitter.d.ts` files have been out of date for some months now
* #4283
* #4482


ObserverOfTime:
If it doesn't mess with local development, we should drop these files (and the maps) from version control and let [`npm prepare`](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts) generate them for the package.

amaanq (approved):

PR #4614, opened 2025-07-14
fix(query): prevent cycles when analyzing hidden children

### LINKED ISSUE #4042: Recursive grammar explodes query compile time
### Problem

Very simple query taking an insanely long time `249298.728ms` to compile 
`((while) (end) @end)`

### Steps to reproduce

Build grammar from [repo](https://github.com/RedCMD/TmLanguage-Syntax-Highlighter/blob/main/src/tree-sitter/tree-sitter-json/grammar.js#L318-L327)

Create test file and parse it

Query the `rootNode` with `((while) (end) @end)`
notice that the query takes upwards of 4min to compile

now locate `_injection_scopes` in the grammar
replace:
```js
fieldAlias($,
	"scope",
	/[_a-zA-Z0-9.:][_a-zA-Z0-9.:-]*/,
),
```
with
```js
alias(
	/[_a-zA-Z0-9.:][_a-zA-Z0-9.:-]*/,
	$.scope,
),
```
rebuild grammar
now notice it taking mere milliseconds to compile

sadly, attempting to simplify the grammar results in a much faster query time
so it must be a combination of many node types and `field`

### Expected behavior

For it to take mere milliseconds to compile

even extremely slow queries are faster
`(pattern [(match) (include)] [(begin) (while) (end) (beginCaptures) (whileCaptures) (endCaptures) (contentName) (applyEndPatternLast)] @beginPatterns)`: `8000ms` https://github.com/tree-sitter/tree-sitter/issues/1903

web-tree-sitter 0.24.5

### Operating system/version

Windows 11

ribru17:
**Problem:** Query structure analyzer can get stuck in a cycle when hidden nodes have children that reference themselves, causing query parsing to run until the iteration limit and take longer than it should.

**Solution:** Track the symbols we have already analyzed in a list, and don't descend into hidden node children if we've already seen their symbol.

Fixes #4042

(Can you confirm that it does fix this, @RedCMD ? I wasn't able to reproduce the original report, even on `master`)

Testing with https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/dart/textobjects.scm:

### Before:

```
time tree-sitter query ../nvim-treesitter-textobjects/queries/dart/textobjects.scm test.dart
test.dart

real    0m2.044s
user    0m2.010s
sys     0m0.023s

time tree-sitter query ../nvim-treesitter-textobjects/queries/dart/textobjects.scm test.dart
test.dart

real    0m1.848s
user    0m1.805s
sys     0m0.024s
```

### After:

```
time ../tree-sitter/target/release/tree-sitter query ../nvim-treesitter-textobjects/queries/dart/textobjects.scm test.dart
test.dart

real    0m0.134s
user    0m0.116s
sys     0m0.017s

time ../tree-sitter/target/release/tree-sitter query ../nvim-treesitter-textobjects/queries/dart/textobjects.scm test.dart
test.dart

real    0m0.133s
user    0m0.118s
sys     0m0.015s
```

~15x speed up for weird queries


RedCMD:
I had fixed it on my end back last year
I think the old problematic code is still there commented out

I'll try and re-repo again and see if this fixes it

RedCMD:
if I go back to the old grammar
and swap the comments on lines L366-L373
https://github.com/RedCMD/TmLanguage-Syntax-Highlighter/blob/dea70d9675543e3fd2ff86b8aec8e3d0af80c978/src/tree-sitter/tree-sitter-json/grammar.js#L366-L373
I am able to repo again. 137422.345ms compile time

for some reason this seemingly unrelated commit fixes the lag
https://github.com/RedCMD/TmLanguage-Syntax-Highlighter/commit/eff5954730eee92235ebb55d21d7ee5990eb1332#diff-9f7ccd848aee4b2ea085510f9f4d3a9529624cb9d0d8355576e765389bd48631

RedCMD:
@ribru17 
as long as I've built this right
sadly it doesn't fix the issue

ribru17:
Thanks for the response- that's unfortunate :/ I'll look into why it works with dart and not jsontm

clason:
Gave this a full run on nvim-treesitter, and a) it doesn't seem to break anything and b) gives about a 10% performance increase total. So I would say this is valuable to merge (and backport) in any case.

ribru17:
Sure, I'll amend to mark this as not fixing #4042 and leave a comment that the solution isn't complete

ribru17:
Hold off on merging, I think I have a solution finally...? looks like there was already some logic for reducing duplicate calculations, but it was using an incorrect comparison function. I'll push a change that fixes @RedCMD's issue and the dart issue, let me know if you see any bugs or perf gains/losses in nvim-treesitter

ribru17:
(Left some important info in the commit message)

RedCMD:
can confirm fixes my original issue
from 137422ms to 35ms

(testing web-tree-sitter within VSCode Web)

however...

when testing the larger full query
https://github.com/RedCMD/TmLanguage-Syntax-Highlighter/blob/dea70d9675543e3fd2ff86b8aec8e3d0af80c978/src/DiagnosticCollection.ts#L466-L506
it drops from 170215ms to 617ms (which is good)

but if I test with the problem fixed in my grammar (swapping the commented code)
it has increased from 335ms to that 617ms (bad)

so I am noticing that performance regression you mentioned

clason:
> Let me know if you see any bugs or perf gains/losses in nvim-treesitter

Still passes without errors, and with the same ~10% performance increase.

clason (approved):

WillLillis (approved):

PR #4615, opened 2025-07-15
perf(generate): reserve more `Vec` capacities

WillLillis:
Some more small perf wins, this just reserves a few more `Vec` capacities up front before they're `push`ed to.

**tree-sitter-c** (3.3%):
image

**tree-sitter-ocaml** (4.2%):
image

### DIFF
--- a/crates/generate/src/build_tables/build_parse_table.rs
+++ b/crates/generate/src/build_tables/build_parse_table.rs
@@ -312,6 +312,12 @@ impl<'a> ParseTableBuilder<'a> {
             }
         }
 
+        let non_terminal_sets_len = non_terminal_extra_item_sets_by_first_terminal.len();
+        self.non_terminal_extra_states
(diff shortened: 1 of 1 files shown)


crates/generate/src/build_tables/build_parse_table.rs
```diff
@@ -312,6 +312,15 @@ impl<'a> ParseTableBuilder<'a> {
             }
         }
 
+        self.non_terminal_extra_states
+            .reserve(non_terminal_extra_item_sets_by_first_terminal.len());
```
ObserverOfTime: Maybe set a variable for this length?

WillLillis: Good point, thanks! Felt a bit awkward naming it, but I think what I just pushed up should work.

wow49 (commented):
[Uploading tree-sitter.windows-x86.zip…]()

PR #4618, opened 2025-07-15
Replace deprecated function on build.zig

nvlled:
The deprecated function addStaticLibrary() has been removed entirely so it no longer builds on recent dev versions of zig.

~~Also the build.zig weirdly uses two-space indentation, but I didn't format the whole file to avoid cluttering the diff.~~ 
Edit: Well actually the whole codebase uses two-space so never mind.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -8,17 +8,22 @@ pub fn build(b: *std.Build) !void {
   const shared = b.option(bool, "build-shared", "Build a shared library") orelse false;
   const amalgamated = b.option(bool, "amalgamated", "Build using an amalgamated source") orelse false;
 
-  const lib: *std.Build.Step.Compile = if (!shared) b.addStaticLibrary(.{
+  const lib: *std.Build.Step.Compile = if (!shared) b.addLibrary(.{
     .name = "tree-sitter",
-    .target = target,
-    .optimize = optimize,
-    .link_libc = true,
-  }) else b.addSharedLibrary(.{
+    .root_module = b.createModule(.{
+        .target = target,
+        .optimize = optimize,
+        .link_libc = true,
+    }),
+  }) else b.addLibrary(.{
     .name = "tree-sitter",
-    .pic = true,
-    .target = target,
-    .optimize = optimize,
-    .link_libc = true,
+    .linkage = .dynamic,
+    .root_module = b.createModule(.{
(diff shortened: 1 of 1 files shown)


clason:
Deprecated in which version? We target releases (14.1 currently), not HEAD.

build.zig
```diff
@@ -8,17 +8,22 @@ pub fn build(b: *std.Build) !void {
...
+        .optimize = optimize,
+        .link_libc = true,
+    }),
+  }) else b.addLibrary(.{
     .name = "tree-sitter",
-    .pic = true,
-    .target = target,
-    .optimize = optimize,
-    .link_libc = true,
+    .linkage = .dynamic,
```
ObserverOfTime: You can simplify this by moving the ternary to `.linkage`.

nvlled:
> Deprecated in which version? We target releases (14.1 currently), not HEAD.

It is deprecated in 0.14.1, see https://ziglang.org/documentation/0.14.1/std/#std.Build. It would be nice to update this sooner, at least I could avoid having too many personal forks just to fix the build errors.

build.zig
nvlled: `.pic` also needs a ternary, but it's null by default not false

clason (approved):

clason:
@nvlled thank you. Can you please make the same changes to the file in `crates/cli/src/templates/`?

nvlled:
@clason okay

clason:
(best to wait for the zig fmt PR to be merged, though)

clason:
@nvlled Turns out that the template file is already formatted so no need to wait to make that change.

nvlled:
Hmm, the template files actually has a lot of problems, aside from using the deprecated build functions. I'll include those fixes in the pull request as well.

clason:
Why am I not surprised ;)

nvlled:
@clason The current zig templates files were the same tree-sitter-c files (they were added at the same time), so I opened a pull request for the tree-sitter-c as well. Ideally, zig-tree-sitter should be updated as well, but @amaanq is not active or is not accepting any pull request at the moment.

https://github.com/tree-sitter/tree-sitter/pull/4637
https://github.com/tree-sitter/tree-sitter-c/pull/275

PR #4619, opened 2025-07-16
fix(bindings): correct indices for `Node::utf16_text`

### LINKED ISSUE #4616: `utf16_text` index out of range after `parse_utf16_le` in Rust
### Problem

The correct implementation **might be**:

```rs
&source[self.start_byte() / 2..self.end_byte() / 2]
```

But the current implementation is (including Go binding):

```rs
&source[self.start_byte()..self.end_byte()]
```

Or maybe I misunderstood this API?

### Steps to reproduce

Create a new grammar and a test crate:

```sh
mkdir utf16le
cd utf16le
tree-sitter init
tree-sitter generate
cargo new --lib hello
```

Add dependencies to crate `hello`:

```toml
tree-sitter = "0.25.8"
tree-sitter-utf16le = { path = "../" }
```

Write the Rust test:

```rs
#[cfg(test)]
mod tests {
    use tree_sitter::Parser;
    use tree_sitter_utf16le::LANGUAGE;

    #[test]
    fn it_works() {
        parser.set_language(&LANGUAGE.into()).unwrap();
        let text = "你好".encode_utf16().collect::<Box<_>>();
        let tree = parser.parse_utf16_le(&text, None).unwrap();
        let root = tree.root_node().utf16_text(&text);
//                                  ^~~~~~~~~~~~~~~~^ fails here
        let hello = String::from_utf16(root).unwrap();
        assert_eq!(hello, "你好");
    }
}
```

Run the test:

```sh
cd hello
cargo test
```

Result:

```plaintext
---- tests::it_works stdout ----

thread 'tests::it_works' panicked at C:\Users\XXX\.cargo\registry\src\index.crates.io-XXX\tree-sitter-0.25.8\binding_rust\lib.rs:2065:16:
range end index 4 out of range for slice of length 2
```

### Expected behavior

Get the `"你好"` text via `String::from_utf16`.

tree-sitter 0.25.4

### Operating system/version

Windows 10 22H2

WillLillis:
`Node::utf16_text` uses *byte* offsets to index into a slice of `u16`s. These indices need to be adjusted. 

I added handling in case these adjusted indices happen to fall on a utf16 [surrogate](https://en.wikipedia.org/wiki/UTF-16#U+D800_to_U+DFFF_(surrogates)). I'm not sure how to construct such a pathological test case that has a node's text boundary split on a surrogate pair, but I think it's better to be safe than sorry in this case.

Alternatively, we could just use `String::from_utf16`, however this would return owned data (also making the requisite heap allocation), as well as break parity with `Node::utf8_text`'s return type.

- Closes #4616

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -2068,7 +2068,7 @@ impl<'tree> Node<'tree> {
 
     #[must_use]
     pub fn utf16_text<'a>(&self, source: &'a [u16]) -> &'a [u16] {
-        &source[self.start_byte()..self.end_byte()]
+        &source[self.start_byte() / 2..self.end_byte() / 2]
     }
 
     /// Create a new [`TreeCursor`] starting from this node.
--- a/crates/cli/src/tests/text_provider_test.rs
+++ b/crates/cli/src/tests/text_provider_test.rs
@@ -108,6 +108,19 @@ fn test_text_provider_for_arc_of_bytes_slice() {
 }
 
 #[test]
+fn test_text_provider_for_vec_utf16_text() {
+    let source_text = "你好".encode_utf16().collect::<Vec<_>>();
+
+    let language = get_language("c");
+    let mut parser = Parser::new();
+    parser.set_language(&language).unwrap();
+    let tree = parser.parse_utf16_le(&source_text, None).unwrap();
+
+    let tree_text = tree.root_node().utf16_text(&source_text);
+    assert_eq!(source_text, tree_text);
+}
+
+#[test]
 fn test_text_provider_callback_with_str_slice() {
     let text: &str = "// comment";
 


ObserverOfTime:
We (I?) normally use the `bindings` scope for the binding templates. I suggest using `rust` instead.
We should probably document these at some point…

lib/binding_rust/lib.rs
```diff
@@ -2068,7 +2068,25 @@ impl<'tree> Node<'tree> {
...
+        // If the start byte is a high surrogate, we need to index back one
+        let start = if start > 0 && (0xD800..=0xDBFF).contains(&source[start - 1]) {
+            start - 1
+        } else {
+            start
+        };
+
+        // If the end byte is a low surrogate, we need to index forward one
+        let end =
+            if end < source.len().saturating_sub(1) && (0xDC00..=0xDFFF).contains(&source[end]) {
```
ribru17:
```suggestion
            if end < source.len() && (0xDC00..=0xDFFF).contains(&source[end]) {
```
I think the saturating sub should be removed, otherwise surrogate handling will not work when the end byte is on a pair which is also the last character of the source string

Tested with the following suite, maybe a small version could be added as a unit test here to get coverage for the surrogate stuff? Just a thought, not a blocker

<details><summary>Tests</summary>

```rust
fn utf16_text(source: &[u16], start_byte: usize, end_byte: usize) -> &[u16] {
    let start = start_byte / 2;
    let end = end_byte / 2;

    // If the start byte is a high surrogate, we need to index back one
    let start = if start > 0 && (0xD800..=0xDBFF).contains(&source[start - 1]) {
        start - 1
    } else {
        start
    };

    // If the end byte is a low surrogate, we need to index forward one
    let end = if end < source.len().saturating_sub(1) && (0xDC00..=0xDFFF).contains(&source[end]) {
        end + 1
    } else {
        end
    };

    &source[start..end]
}

fn main() {
    let src = "🤠🤠🤠".encode_utf16().collect::<Vec<_>>();
    let new_src = match String::from_utf16(utf16_text(&src, 0, 0)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from(""), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 4)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from("🤠"), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 2)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from("🤠"), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 6)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from("🤠🤠"), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 8)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from("🤠🤠"), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 12)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"),
    };
    assert_eq!(String::from("🤠🤠🤠"), new_src);
    let new_src = match String::from_utf16(utf16_text(&src, 0, 10)) {
        Ok(src) => src,
        Err(err) => panic!("{err}"), // fails here
    };
    assert_eq!(String::from("🤠🤠🤠"), new_src);
}
```

</details>

WillLillis: Ah good catch. After sleeping on this, I'm not actually sure if we should be checking for surrogate mismatches in this function. That seems like it would be hiding underlying bugs in the parser and/or `Node` struct. I think it would still be worthwhile to add the test cases you gave above though, will work on that in the next day or so.

What are your thoughts?

ribru17: Yeah, makes sense to me :+1: If we did this, what would be the purpose of keeping the test cases though? Genuine question, not trying to say it's no longer worthwhile

WillLillis: Was mostly thinking of including them since I didn't see much coverage for utf16 in general, but don't feel super strongly about it. If you think they're redundant I'm fine leaving them out 😄

ribru17: That makes sense. I was just thinking we wouldn't need coverage for those paths if we don't worry about surrogate logic, but I could definitely be wrong on that...

WillLillis: Yeah I think we're fine leaving them out for now. Thanks so much for the help!

ribru17 (approved):

PR #4622, opened 2025-07-16
chore(zig): run zig fmt

clason:
It's mandatory(tm).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,6 +10,9 @@ insert_final_newline = true
 [*.rs]
 indent_size = 4
 
+[*.zig]
+indent_size = 4
+
 [Makefile]
 indent_style = tab
 indent_size = 8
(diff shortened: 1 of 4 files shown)


ObserverOfTime:
Please also update `.editorconfig` accordingly.

clason:
It's not (just) about the indentation, though.

.editorconfig
```diff
@@ -10,6 +10,9 @@ insert_final_newline = true
 [*.rs]
 indent_size = 4
 
+[*.zig]
```
ObserverOfTime, line 13:
```suggestion
[*.{zig,zon}]
```

clason: The `zon` should not be edited manually anyway, and is not stable yet.

ObserverOfTime:
> It's not (just) about the indentation, though.

No, but it will prevent a regression.

.editorconfig
ObserverOfTime: At the very least the version is edited manually.

Eakinter99:
> It's mandatory(tm).

pratama22wawan-max:
###

PR #4630, opened 2025-07-20
fix(bindings): fix clean makefile target

ObserverOfTime:


WillLillis:
Why remove the `.lib` pattern?

ObserverOfTime:
It doesn't exist.

WillLillis:
Ah, I still had one floating around in the repo inside git bash, but I guess that was left over from awhile ago.

PR #4631, opened 2025-07-21
Revert "feat(generate): warn about unreachable variables (#4567)"

clason:
This reverts commit 25c601bd2fd4b7bd6c4f575d6ea5867cfff7c008.

Reason: Too many false positives (especially for split parsers).

That information is useful, but unresolvable warnings for a significant class of parsers are not acceptable. We need to (re)think carefully how we report this (and similar) "lints".


WillLillis (approved):

wetneb:
Unfortunate that I didn't think about the case of split grammars, sorry about that. I don't see an easy way to keep this warning relevant, sadly.

clason:
No worries; we'll find some way!

PR #4633, opened 2025-07-22
feat(test): allow cst as expected output of test case

### LINKED ISSUE #4476: Test against CST
### Problem

I've run into some bugs in a grammar I work on where the CST has not been captured correctly but the AST was fine (e.g. due to not using `lexer->mark_end` in the correct place in an external scanner). It would be handy in these cases to be able to test against the CST, the same way it's possible to include the fields.

### Expected behavior

A strawperson syntax example:

```
==============
Number literal
==============

42

--------------

(number_literal `42`)
```

would check that the text of the `number_literal` node was actually `42`

WillLillis:
This PR allows for grammar test cases to use a CST as the expected output (with the `:cst` attribute). Tiny example with tree-sitter-c:

```
==============
Number Literal
:cst
==============

42;

--------------

0:0 - 1:0   translation_unit
0:0 - 0:3     expression_statement
0:0 - 0:2        number_literal `42`
0:2 - 0:3        ";"
```

This uses the same output as you can get with `parse --cst`. I'm opening this up early as a draft to hopefully field some testing from any interested grammar authors. Open questions/merge checklist:

- [x] Diff rendering is very unclear when there is a whitespace-only difference between the expected and actual CSTs
	- This can be improved by escaping invisibles (i.e. `\n`) in the diff and/or being more careful with surrounding whitespace in the CST
- [x] Consider exposing the `no_ranges` option for the CST and/or making this the default behavior. Would like to hear some thoughts from users on this -- Not going to do this for now, but open to hearing arguments otherwise
<details>

<Summary> no_ranges example</Summary>

```
  translation_unit
    expression_statement
      number_literal
        `42`
      ";"
```

</details>

- [x] General code cleanup -- this original implementation was a bit hacked together and (particularly the test runner code) needs another lookover
- [x] Make sure `:cst` cases interact as expected when combined with other testing features (i.e. update)

- Closes #4476

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -973,9 +973,10 @@ fn cst_render_node(
         } else {
             opts.parse_theme.node_kind
         };
-        write!(out, "{} ", paint(kind_color, node.kind()))?;
+        write!(out, "{}", paint(kind_color, node.kind()),)?;
 
         if node.child_count() == 0 {
+            write!(out, " ")?;
             // Node text from a pattern or external scanner
             write_node_text(
                 opts,


sogaiu:
I've given it a brief try.

Somehow there seems to be some kind of off-by one thing going on here (could be pilot error ofc :sweat_smile:).

I tried something simple with `tree-sitter-clojure` for the input `#()`.

The test file content living under `test/corpus` was:

```
================================================================================
Empty Anonymous Function
:cst
================================================================================

#()

--------------------------------------------------------------------------------

0:0 - 1:0   source 
0:0 - 0:3     anon_fn_lit 
0:0 - 0:1       "#"
0:1 - 0:2       "("
0:2 - 0:3       ")"
```

Output for "actual" (appears in red in `tree-sitter test` output):

```
0:0 - 2:0   source 
1:0 - 1:3     anon_fn_lit 
1:0 - 1:1       "#"
1:1 - 1:2       "("
1:2 - 1:3       ")"
```

Output for "expected" (appears in green in `tree-sitter test` output, but get the same numbers when parsing manually via `parse` subcommand [1]):

```
0:0 - 1:0   source 
0:0 - 0:3     anon_fn_lit 
0:0 - 0:1       "#"
0:1 - 0:2       "("
0:2 - 0:3       ")"
```

---

For reference, full output of `tree-sitter test`:
```
$ tree-sitter.cst test
  anon_fn_lit:
      1. ✗ Empty Anonymous Function

1 failure:

correct / expected / unexpected

  1. Empty Anonymous Function:
0:0 - 2:0   source 
1:0 - 1:3     anon_fn_lit 
1:0 - 1:1       "#"
1:1 - 1:2       "("
1:2 - 1:3       ")"

0:0 - 1:0   source 
0:0 - 0:3     anon_fn_lit 
0:0 - 0:1       "#"
0:1 - 0:2       "("
0:2 - 0:3       ")"
```

---

[1] Output corresponds to with `#()` in a file as well as manual input via standard input.

jpt13653903:
> Somehow there seems to be some kind of off-by one thing going on here (could be pilot error ofc 😅).

Should be an easy fix...
The thing that parses the test corpus must remove leading and trailing empty lines from the source...

WillLillis:
@sogaiu thanks for taking a look! With my latest changes I have your test case passing after removing the leading empty line in the input.

I think that whitespace (including empty leading/trailing lines) in the input needs to be preserved, but I also needed to clean up the expected output from the test corpus. Let me know how the latest version works!

sogaiu:
Attempting to follow:

> removing the leading empty line in the input

I modified the test file to look like:

```
================================================================================
Empty Anonymous Function
:cst
================================================================================
#()

--------------------------------------------------------------------------------

0:0 - 1:0   source
0:0 - 0:3     anon_fn_lit
0:0 - 0:1       "#"
0:1 - 0:2       "("
0:2 - 0:3       ")"
```


With this change https://github.com/tree-sitter/tree-sitter/commit/b732f4a4554ca14b22d0fab6c6236b2f871dbdb1 worked for me :+1: 

---

Should we assume going forward that we ought to remove the blank line between `=== ... =` and the source input in order to match the `tree-sitter parse --cst` output?

WillLillis:
> Should we assume going forward that we ought to remove the blank line between === ... = and the source input in order to match the tree-sitter parse --cst output?

Do you have a preference/ any thoughts? The line numbers don't matter for regular sexp tests but with cst you're opting into including the ranges as part of the expected output (unless we decide to use/support `no_ranges`).  That being said,  just trimming the surrounding whitespace of the input could be a more reasonable default (I can't think of any cases where being this "strict" this would *actually* matter, but I could be wrong).

sogaiu:
I'm not sure if it's really important, but I think I do prefer having a blank line before and after the input.  This might be partly for perception as well as aesthetics...I hesitate to claim it's very important though :sweat_smile:

I haven't looked in-depth at what other folks have done but my impression is that that's what I tend to see in other repositories.

I did a quick:

```
$ rg -A 2 ==== repos/*/*/*/test/corpus/*
```

in my local collection of tree-sitter grammar repositories and manually scanning through some of the output I think that impression doesn't seem off [1].

It looks like since the leading blank line is currently optional(?) people have been happy to choose to their liking?

> That being said, just trimming the surrounding whitespace of the input could be a more reasonable default (I can't think of any cases where being this "strict" this would actually matter, but I could be wrong).

I'm not coming up with a case where it (trimming) could matter, but since there are a lot of languages out there I can understand that caution might be in order.

I haven't looked at how the existing test code works...may be for the non-`:cst` (so the default) case, trimming is already happening?  If so, if trimming had been an issue, may be we would have seen a report already...

---

[1] [This directory](https://github.com/yutaro-sakamoto/tree-sitter-cobol/tree/e99dbdc3d800d5fa2796476efd60af91f6b43d93/test/corpus) has many cases where there is no blank line between `===...=` and the input...but it's not the only repository:

* https://github.com/urbit-pilled/tree-sitter-hoon
* https://github.com/vantreeseba/tree-sitter-haxe/blob/12e814cd9b0656b0eaa3e04b0036b3b6625cc0d7/test/corpus/assignment.txt#L17-L18
* https://github.com/uben0/tree-sitter-typst/blob/master/test/corpus/positive.scm#L3-L4

jpt13653903:
I have a thought...

## TL;DR;

> The thing that parses the test corpus must remove leading and trailing empty lines from the source...

I was wrong

> I think that whitespace (including empty leading/trailing lines) in the input needs to be preserved

You are right.

## Arguments

- When testing the external scanner, it is sometimes necessary to verify the behaviour around leading and trailing whitespace, especially checking the correct handling of the exact placement of the EOF, as well as the correct initialisation with leading whitespace with or without new-line characters.
- Some future language might have significant whitespace in the header, for adding doc-strings or something, I don't know.  It might mean that an empty header line is expected for the code to start, and if there is text at the top, that's actually a comment.
- Someone might implement a parser for [Whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language)), who knows :wink:

In other words, the lines between the `===== ... -----` sections should exactly represent the source file, by default.  This can be documented, and it's fairly easy to infer the problem (and fix) when the user firsts encounters the "off-by-one" symptom.  This requires some careful handling:

```text
================================================================================
Basic function
:cst
================================================================================

fun myFunction()
    print('Hello')

--------------------------------------------------------------------------------
```

should parse as:

```c
"\nfun myFunction()\n    print('Hello')\n"
```

Note that there is only one `\n` at the end.  One of the `\n`'s are stripped.  If there were whitespace on the last line, those should be preserved -- only the final `\n` should be stripped.

## Suggestion

I would suggest preserving whitespace by default, and adding an attribute for stripping leading and trailing whitespace.  This new attribute is independent from this PR, and applies to the "normal" "existing" test method(s) as well.

sogaiu:
> I would suggest preserving whitespace by default, and adding an attribute for stripping leading and trailing whitespace.

If this is the default most would want, wouldn't the reverse make more sense?

jpt13653903:
> If this is the default most would want, wouldn't the reverse make more sense?

Just a suggestion that I threw into the pot.  I'm happy either way.

ObserverOfTime:
I agree with keeping the default as is and adding a `:trim` attribute (in a separate PR).

sogaiu:
I'm fine with either but IIUC the suggested default behavior differs from the current default (non-cst) [1], so I think it would be good to document that it has a different default.

---

[1] Please point out if this is mistaken.

sogaiu:
On second thought, if there is no urgent need for trimming and the default is to not trim, may be it's better to just not implement it first and wait for a request down the line?

If the original test format didn't trim, I suspect people would have lived with it (just a guess).

WillLillis:
> On second thought, if there is no urgent need for trimming and the default is to not trim, may be it's better to just not implement it first and wait for a request down the line?
> 
> 
> 
> If the original test format didn't trim, I suspect people would have lived with it (just a guess).

The current implementation trims a single trailing newline. I think leaving the status quo and adjusting as needed later on sounds like a good plan.

tsujp:
Regarding the off by 1 stuff, tree-sitter seems to count lines as 0-indexed but isn't the convention 1-indexed? The first line in a file is line 1 and not line 0.

All ranges output (e.g. from `parse`) are currently like this.

tsujp:
Made a GH issue: https://github.com/tree-sitter/tree-sitter/issues/4653

docs/src/creating-parsers/5-writing-tests.md
```diff
@@ -103,12 +103,14 @@ The following attributes are available:
...
 input is invalid without displaying the whole parse tree, as such you should omit the parse tree below the `---` line.
-* `:fail-fast` — This attribute will stop the testing additional tests if the test marked with this attribute fails.
+* `:fail-fast` — This attribute will stop the testing of additional cases if the test marked with this attribute fails.
 * `:language(LANG)` — This attribute will run the tests using the parser for the specified language. This is useful for
 multi-parser repos, such as XML and DTD, or Typescript and TSX. The default parser used will always be the first entry in
 the `grammars` field in the `tree-sitter.json` config file, so having a way to pick a second or even third parser is useful.
 * `:platform(PLATFORM)` — This attribute specifies the platform on which the test should run. It is useful to test platform-specific
 behavior (e.g. Windows newlines are different from Unix). This attribute must match up with Rust's [`std::env::consts::OS`][constants].
+* `:cst` - This attribute specifies that the expected output should be in the form of a CST instead of the normal S-expression. This
+CST matches the format given by `parse --cst`.
```
ObserverOfTime: Maybe swap this and `:skip` so the list is in alphabetical order.

ObserverOfTime (commented):
The CST contains trailing whitespace in some lines. Can that be stripped?

Edit: I mean the CST generated by `--update`.

sogaiu:
> The CST contains trailing whitespace in some lines. Can that be stripped?

Regarding removal of trailing whitespace, there is some discussion starting earlier from [here](https://github.com/tree-sitter/tree-sitter/pull/4633#issuecomment-3112622174).

My impression was that some sort of agreement had been reached in [this comment](https://github.com/tree-sitter/tree-sitter/pull/4633#issuecomment-3115805903).

But may be you have some additional thoughts that the discussion missed.

ObserverOfTime:
> Regarding removal of trailing whitespace, there is some discussion starting earlier from here.

That's the input whitespace. I'm talking about the output whitespace.
image

sogaiu:
Oh I see.

Sorry, I misunderstood.

Thanks for the clarification.

WillLillis:
> The CST contains trailing whitespace in some lines. Can that be stripped?
> 
> Edit: I mean the CST generated by `--update`.

Nice catch! Should be fixed now.

ObserverOfTime (approved):
I think this is fine to ship as is and we can revisit the `no_ranges` option later if desired.

g0t4:
IIAC this will come in v0.26?

clason:
Yes.

PR #4635, opened 2025-07-22
fix(templates): Generate from grammar.js.

slg95:
Switch both the template Makefile and CMakeLists.txt to default to generating from grammar.js instead of grammar.json.

The template does not include a grammar.json, so trying to use the Makefile or CMake will fail. The docs do suggest running tree-sitter generate after tree-sitter init, but this shouldn't be necessary for the build system to work.

Since the documentation focuses on editing the grammar.js file, the build system should rebuild when changes are made to grammar.js instead of src/grammar.json

### DIFF
--- a/crates/cli/src/templates/cmakelists.cmake
+++ b/crates/cli/src/templates/cmakelists.cmake
@@ -20,8 +20,8 @@ include(GNUInstallDirs)
 find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
 
 add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/parser.c"
-                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
No. Also #4601.

wetneb:
@sig95 I sympathize with your need, but we're in the process of solving it in a different way.
Once #4601 is merged, I plan to reopen a new version of #4580 which will add a goal to generate `grammar.json` from `grammar.js` (without generating `parser.c`). I believe this should satisfy your needs of running `make` in a repo where only `grammar.js` is available.

slg95:
@wetneb I took a look at your PR and I appreciate your effort there. For me personally I just edited the Makefile since all I care about is being able to edit grammar.js then running make to get my .so file.

I get that many common use cases need the grammar.json, but it just seemed weird that the workflow would have to be:

- edit grammar.js
- run tree-sitter generate
- run make

So anyway, I think your work will make things more intuitive for workflows like mine :)

clason:
(you can -- and arguably should -- use `tree-sitter build` as step 3)

PR #4637, opened 2025-07-24
Fix zig template files

nvlled:
Fixed and updated to work with zig v0.14.1.

The changes are generated from https://github.com/tree-sitter/tree-sitter-c/pull/275 using this [script](https://gist.github.com/nvlled/65b6c1b9385f28665baf6bb486612686).

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -751,9 +751,15 @@ pub fn generate_grammar_files(
 
     // Generate Zig bindings
     if tree_sitter_config.bindings.zig {
-        missing_path(repo_path.join("build.zig"), |path| {
-            generate_file(path, BUILD_ZIG_TEMPLATE, language_name, &generate_opts)
-        })?;
+        missing_path_else(
+            repo_path.join("build.zig"),
+            allow_update,
+            |path| generate_file(path, BUILD_ZIG_TEMPLATE, language_name, &generate_opts),
+            |path| {
+                eprintln!("Replacing build.zig");
+                generate_file(path, BUILD_ZIG_TEMPLATE, language_name, &generate_opts)
+            },
+        )?;
 
         missing_path_else(
             repo_path.join("build.zig.zon"),
@@ -766,9 +772,15 @@ pub fn generate_grammar_files(
         )?;
 
         missing_path(bindings_dir.join("zig"), create_dir)?.apply(|path| {
-            missing_path(path.join("root.zig"), |path| {
-                generate_file(path, ROOT_ZIG_TEMPLATE, language_name, &generate_opts)
-            })?;
+            missing_path_else(
+                path.join("root.zig"),
+                allow_update,
+                |path| generate_file(path, ROOT_ZIG_TEMPLATE, language_name, &generate_opts),
+                |path| {
+                    eprintln!("Replacing root.zig");
+                    generate_file(path, ROOT_ZIG_TEMPLATE, language_name, &generate_opts)
+                },
+            )?;
 
             missing_path(path.join("test.zig"), |path| {
                 generate_file(path, TEST_ZIG_TEMPLATE, language_name, &generate_opts)
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -755,21 +755,23 @@ pub fn generate_grammar_files(
             generate_file(path, BUILD_ZIG_TEMPLATE, language_name, &generate_opts)
         })?;
 
-        missing_path(repo_path.join("build.zig.zon"), |path| {
-            generate_file(path, BUILD_ZIG_ZON_TEMPLATE, language_name, &generate_opts)
-        })?;
-
-        missing_path(bindings_dir.join("zig"), create_dir)?.apply(|path| {
-            missing_path(path.join("test.zig"), |path| {
-                generate_file(path, TEST_ZIG_TEMPLATE, language_name, &generate_opts)
-            })?;
-
-            Ok(())
-        })?;
+        missing_path_else(
+            repo_path.join("build.zig.zon"),
+            allow_update,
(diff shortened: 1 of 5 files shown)


crates/cli/src/init.rs
```diff
@@ -707,6 +708,14 @@ pub fn generate_grammar_files(
             generate_file(path, BUILD_ZIG_ZON_TEMPLATE, language_name, &generate_opts)
         })?;
 
+        missing_path(bindings_dir.join("zig"), create_dir)?.apply(|path| {
+            missing_path(path.join("test.zig"), |path| {
+                generate_file(path, TEST_ZIG_TEMPLATE, language_name, &generate_opts)
+            })?;
```
ObserverOfTime: Move this below `root.zig`.

crates/cli/src/templates/build.zig
```diff
@@ -42,38 +42,43 @@ pub fn build(b: *std.Build) !void {
...
-
     const test_step = b.step("test", "Run unit tests");
     test_step.dependOn(&run_tests.step);
 }
 
-inline fn hasScanner(dir: std.fs.Dir) bool {
-    dir.access("src/scanner.c", .{}) catch return false;
+inline fn fileExists(b: *std.Build, filename: []const u8) bool {
+    const dir = b.build_root.handle;
+    dir.access(filename, .{}) catch return false;
```
ObserverOfTime, line 88: This change is unnecessary. We don't need a generic function.

crates/cli/src/templates/root.zig
```diff
@@ -1,19 +1,5 @@
-const testing = @import("std").testing;
+extern fn tree_sitter_PARSER_NAME() callconv(.c) *const anyopaque;
```
ObserverOfTime, line 1: FWIW the bindings accepted `anyopaque` at first but @amaanq didn't like that.
~~It's probably better to keep it as `Language` so consumers don't have to `@ptrCast` all the time.~~

crates/cli/src/init.rs
```diff
@@ -707,6 +708,14 @@ pub fn generate_grammar_files(
             generate_file(path, BUILD_ZIG_ZON_TEMPLATE, language_name, &generate_opts)
```
ObserverOfTime: Use `missing_path_else` to replace the previous files (with a message).

crates/cli/src/templates/test.zig
```diff
@@ -0,0 +1,17 @@
...
+
+test "can load grammar" {
+    const parser = Parser.create();
+    defer parser.destroy();
+
+    const lang: *const ts.Language = @ptrCast(root.language());
+    defer lang.destroy();
+
+    try testing.expectEqual(parser.setLanguage(lang), void{});
+    try testing.expectEqual(parser.getLanguage(), lang);
```
ObserverOfTime: The parameter order should be reversed (it was incorrect initially).

crates/cli/src/templates/test.zig
ObserverOfTime, line 1: ~~I'd rather keep the tests in the root file if we can move the `testing` import inside the `test` block.~~

ObserverOfTime (changes requested):

crates/cli/src/templates/build.zig
```diff
@@ -42,38 +42,43 @@ pub fn build(b: *std.Build) !void {
...
+    if (fileExists(b, "queries")) {
+        b.installDirectory(.{
+            .source_dir = b.path("queries"),
+            .install_dir = .prefix,
+            .install_subdir = "queries",
+            .include_extensions = &.{"scm"},
+        });
+    }
+
+    const ts_dep = b.dependency("tree_sitter", .{});
```
ObserverOfTime: Is it possible to make this a lazy dependency now that it's not used in the root module?

crates/cli/src/templates/root.zig
nvlled: Does `Language` work though? I couldn't get it to build when both `zig-tree-sitter` and `tree-sitter-c` is used as a dependency on a project. Is there a sample project that I could refer to? I might just be doing something wrong.

The go binding also returns opaque pointer, so it's probably done that way for the same reasons to avoid problems.

crates/cli/src/templates/build.zig
nvlled: Yeah, but since it also needs to check if queries exist, I thought it would be better to just use one generic function for both. Would you prefer to have `hasScanner` and `hasQueries` instead? I don't mind either ways.

crates/cli/src/templates/build.zig
nvlled: Will do, I'll make it a lazy dependency like what you did here https://github.com/tree-sitter/zig-tree-sitter/blob/f0f412e43ecfac12c48f6a1e810441ee9a10a7c4/build.zig#L63

crates/cli/src/init.rs
nvlled: Should I do that for all zig files? Or just the buld.zig.zon?

crates/cli/src/init.rs
```diff
@@ -704,23 +704,25 @@ pub fn generate_grammar_files(
             generate_file(path, BUILD_ZIG_TEMPLATE, language_name, &generate_opts)
         })?;
```
ObserverOfTime: `build.zig` also needs to be replaced.

crates/cli/src/templates/build.zig
ObserverOfTime: Oh whoops, I didn't notice that. Does `installDirectory` fail if the `source_dir` doesn't exist?

nvlled: Yeah, it fails if it can't find it. Alternately, we can just create the directory when `tree-sitter init` runs.

PR #4645, opened 2025-07-27
feat(cli): allow users to specify dynamic libraries directly

WillLillis:
### The Problem

The tree-sitter CLI does a lot of "magic" behind the scenes to find and load the correct dynamic library when various subcommands are used (i.e. `parse`, `query`, etc.). For some more advances use cases (i.e. testing two different commits for the same grammar), this auto-magical selection isn't desired.

### The Solution

This PR adds the ability to specify the path to a shared library directly via `--lib_path`. Internally, once the library has been loaded, the next step is to extract the language function symbol (i.e. `tree_sitter_c` from `c.so`). In my current implementation, we attempt to derive this symbol from the name of the shared library. If this doesn't work, users can pass `--language_name`.

Sample usage:

```
$ tree-sitter parse sample.c --lib-path /path/to/first/shared/lib/c.so
...
$ tree-sitter parse sample.c --lib-path /path/to/another/shared/lib/other_c.so --lang-name c
```

#### Other considerations

I'm opening this as a draft for now, until the following items can be settled agreed upon:

- [x] Ergonomics of the new arguments. Naming, documentation, etc.
- [x] Should a warning be displayed if `--lang_name` is specified when `--lib_path` isn't?
- [x] Further testing once I have some more time

Once we've settled on these changes, I'd like to:

- [x] Update the docs
- [x] *Potentially* extend this functionality to other subcommands (i.e. `fuzz`). Language loading works a bit differently for other subcommands, so we'll have to be careful not to introduce subtle regressions.

CC @sogaiu and @clason since you've both expressed interest in something like this.

Related: https://github.com/tree-sitter/tree-sitter/pull/4498

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -849,12 +849,21 @@ impl Loader {
             }
         }
 
-        let library = unsafe { Library::new(&output_path) }
-            .with_context(|| format!("Error opening dynamic library {}", output_path.display()))?;
+        Self::load_language(&output_path, &language_fn_name)
+    }
+
+    pub fn load_language(path: &Path, function_name: &str) -> Result<Language> {
+        let library = unsafe { Library::new(path) }
+            .with_context(|| format!("Error opening dynamic library {}", path.display()))?;
         let language = unsafe {
             let language_fn = library
-                .get::<Symbol<unsafe extern "C" fn() -> Language>>(language_fn_name.as_bytes())
-                .with_context(|| format!("Failed to load symbol {language_fn_name}"))?;
+                .get::<Symbol<unsafe extern "C" fn() -> Language>>(function_name.as_bytes())
+                .with_context(|| {
+                    format!(
+                        "Failed to load symbol {function_name} from {}",
+                        path.display()
+                    )
+                })?;
             language_fn()
         };
         mem::forget(library);
@@ -1410,8 +1419,13 @@ impl Loader {
         path: &Path,
         current_dir: &Path,
         scope: Option<&str>,
+        // path to dynamic library, name of language
+        lib_info: Option<(&Path, &str)>,
     ) -> Result<Language> {
-        if let Some(scope) = scope {
+        if let Some((lib_path, language_name)) = lib_info {
+            let language_fn_name = format!("tree_sitter_{}", language_name.replace('-', "_"));
+            Self::load_language(lib_path, &language_fn_name)
+        } else if let Some(scope) = scope {
             if let Some(config) = self
                 .language_configuration_for_scope(scope)
                 .with_context(|| format!("Failed to load language for scope '{scope}'"))?
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -723,10 +723,7 @@ impl Loader {
 
     pub fn load_language_at_path_with_name(&self, mut config: CompileConfig) -> Result<Language> {
         let mut lib_name = config.name.to_string();
-        let language_fn_name = format!(
-            "tree_sitter_{}",
-            replace_dashes_with_underscores(&config.name)
-        );
+        let language_fn_name = format!("tree_sitter_{}", config.name.replace('-', "_"));
         if self.debug_build {
             lib_name.push_str(".debug._");
         }
@@ -1713,16 +1710,4 @@ fn needs_recompile(lib_path: &Path, paths_to_check: &[PathBuf]) -> Result<bool> 
 
 fn mtime(path: &Path) -> Result<SystemTime> {
     Ok(fs::metadata(path)?.modified()?)
-}
-
-fn replace_dashes_with_underscores(name: &str) -> String {
-    let mut result = String::with_capacity(name.len());
-    for c in name.chars() {
-        if c == '-' {
-            result.push('_');
-        } else {
-            result.push(c);
-        }
-    }
-    result
 }
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1791,6 +1791,7 @@ impl DumpLanguages {
         for (configuration, language_path) in loader.get_all_language_configurations() {
             println!(
                 concat!(
+                    "name: {}\n",
                     "scope: {}\n",
                     "parser: {:?}\n",
                     "highlights: {:?}\n",
@@ -1798,6 +1799,7 @@ impl DumpLanguages {
                     "content_regex: {:?}\n",
                     "injection_regex: {:?}\n",
                 ),
+                configuration.language_name,
                 configuration.scope.as_ref().unwrap_or(&String::new()),
                 language_path,
                 configuration.highlights_filenames,
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -185,6 +185,13 @@ struct Parse {
     /// The path to the tree-sitter grammar directory
     #[arg(long, short = 'p')]
     pub grammar_path: Option<PathBuf>,
+    /// The path to the parser's dynamic library
+    #[arg(long, short = 'l')]
+    pub lib_path: Option<PathBuf>,
+    /// If `--lib_path` is used, the name of the language used to extract the
+    /// library's language function
+    #[arg(long)]
+    pub lang_name: Option<String>,
     /// Select a language by the scope instead of a file extension
     #[arg(long)]
     pub scope: Option<String>,
@@ -276,6 +283,13 @@ struct Test {
     /// The path to the tree-sitter grammar directory
     #[arg(long, short = 'p')]
     pub grammar_path: Option<PathBuf>,
+    /// The path to the parser's dynamic library
+    #[arg(long, short = 'l')]
+    pub lib_path: Option<PathBuf>,
+    /// If `--lib_path` is used, the name of the language used to extract the
+    /// library's language function
+    #[arg(long)]
+    pub lang_name: Option<String>,
     /// Update all syntax trees in corpus files with current parser output
     #[arg(long, short)]
     pub update: bool,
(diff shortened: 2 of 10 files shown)


crates/loader/src/loader.rs
```diff
@@ -1410,8 +1407,27 @@ impl Loader {
         path: &Path,
         current_dir: &Path,
         scope: Option<&str>,
+        // path to dynamic library, name of language
+        lib_info: Option<&(&Path, &str)>,
     ) -> Result<Language> {
-        if let Some(scope) = scope {
+        if let Some((lib_path, language_name)) = lib_info {
```
WillLillis, line 1422: This logic is marginally adapted from the end of `Loader::load_language_at_path_with_name`. It may make sense to pull this logic into a helper for `Loader` so it can live in one place.

sogaiu:
I gave baebcea9 a brief try:

Here's a "log" of using `--lib-path`:
```
$ cd ~/src/tree-sitter-clojure

# make test file
$ echo "^:wake ^:sit ^:sleep #{}" > meta-on-set.clj

# build library
$ tree-sitter.load_shared build

# parse with --lib-path
$ tree-sitter.load_shared parse --cst --lib-path ~/src/tree-sitter-clojure/clojure.so meta-on-set.clj 
0:0  - 1:0    source 
0:0  - 0:24     set_lit 
0:0  - 0:6        meta: meta_lit 
0:0  - 0:1          "^"
0:1  - 0:6          value: kwd_lit 
0:1  - 0:2            ":"
0:2  - 0:6            name: kwd_name `wake`
0:7  - 0:12       meta: meta_lit 
0:7  - 0:8          "^"
0:8  - 0:12         value: kwd_lit 
0:8  - 0:9            ":"
0:9  - 0:12           name: kwd_name `sit`
0:13 - 0:20       meta: meta_lit 
0:13 - 0:14         "^"
0:14 - 0:20         value: kwd_lit 
0:14 - 0:15           ":"
0:15 - 0:20           name: kwd_name `sleep`
0:21 - 0:22       "#"
0:22 - 0:23       "{"
0:23 - 0:24       "}"
```

Here's a "log" in a different checkout, using `--lib-path` and then also `--language-name`:
```
$ cd /tmp

# clone grammar repository (output omitted)
$ git clone https://github.com/sogaiu/tree-sitter-clojure

$ cd tree-sitter-clojure

# checkout a different branch / commit from above
$ git checkout unstable-20250709

# make test file
$ echo "^:wake ^:sit ^:sleep #{}" > meta-on-set.clj

# build library
$ tree-sitter.load_shared build

# parse with --lib-path
$ tree-sitter.load_shared parse --cst --lib-path /tmp/tree-sitter-clojure/clojure.so meta-on-set.clj 
0:0  - 1:0    source 
0:0  - 0:6      meta: meta_lit 
0:0  - 0:1        "^"
0:1  - 0:6        value: kwd_lit 
0:1  - 0:2          ":"
0:2  - 0:6          name: kwd_name `wake`
0:7  - 0:12     meta: meta_lit 
0:7  - 0:8        "^"
0:8  - 0:12       value: kwd_lit 
0:8  - 0:9          ":"
0:9  - 0:12         name: kwd_name `sit`
0:13 - 0:20     meta: meta_lit 
0:13 - 0:14       "^"
0:14 - 0:20       value: kwd_lit 
0:14 - 0:15         ":"
0:15 - 0:20         name: kwd_name `sleep`
0:21 - 0:24     set_lit 
0:21 - 0:22       "#"
0:22 - 0:23       "{"
0:23 - 0:24       "}"

# prepare to test --language-name
$ cp clojure.so clj.so

# parse with --lib-path and --language-name
$ tree-sitter.load_shared parse --cst --lib-path /tmp/tree-sitter-clojure/clj.so --language-name clojure meta-on-set.clj 
0:0  - 1:0    source 
0:0  - 0:6      meta: meta_lit 
0:0  - 0:1        "^"
0:1  - 0:6        value: kwd_lit 
0:1  - 0:2          ":"
0:2  - 0:6          name: kwd_name `wake`
0:7  - 0:12     meta: meta_lit 
0:7  - 0:8        "^"
0:8  - 0:12       value: kwd_lit 
0:8  - 0:9          ":"
0:9  - 0:12         name: kwd_name `sit`
0:13 - 0:20     meta: meta_lit 
0:13 - 0:14       "^"
0:14 - 0:20       value: kwd_lit 
0:14 - 0:15         ":"
0:15 - 0:20         name: kwd_name `sleep`
0:21 - 0:24     set_lit 
0:21 - 0:22       "#"
0:22 - 0:23       "{"
0:23 - 0:24       "}"
```

The test file is being parsed differently as expected :)

Looks good so far :+1:

sogaiu:
> Should a warning be displayed if --language_name is specified when --lib_path isn't?

Unless the value from `--language-name` can be used in some other situation where `--lib-path` is not specified, having a warning seems sensible.

On a somewhat related note, looking at `dump-languages` output, I don't see anything corresponding to values for `--language-name` (the closest being values for scope perhaps?).  I wonder if it would be worth having the info there as well...

WillLillis:
> > Should a warning be displayed if --language_name is specified when --lib_path isn't?
> 
> Unless the value from `--language-name` can be used in some other situation where `--lib-path` is not specified, having a warning seems sensible.
> 
> On a somewhat related note, looking at `dump-languages` output, I don't see anything corresponding to values for `--language-name` (the closest being values for scope perhaps?). I wonder if it would be worth having the info there as well...

Added the warning, and good point on `dump-languages`!

sogaiu:
Warning seems to be working in f8704c6:

```
$ tree-sitter.load_shared parse --cst --language-name clojure meta-on-set.clj 
Warning: --language-name` specified without --lib-path. This argument will be ignored.
0:0  - 1:0    source 
0:0  - 0:24     set_lit 
0:0  - 0:6        meta: meta_lit 
0:0  - 0:1          "^"
0:1  - 0:6          value: kwd_lit 
0:1  - 0:2            ":"
0:2  - 0:6            name: kwd_name `wake`
0:7  - 0:12       meta: meta_lit 
0:7  - 0:8          "^"
0:8  - 0:12         value: kwd_lit 
0:8  - 0:9            ":"
0:9  - 0:12           name: kwd_name `sit`
0:13 - 0:20       meta: meta_lit 
0:13 - 0:14         "^"
0:14 - 0:20         value: kwd_lit 
0:14 - 0:15           ":"
0:15 - 0:20           name: kwd_name `sleep`
0:21 - 0:22       "#"
0:22 - 0:23       "{"
0:23 - 0:24       "}"
```

I see the new output from `dump-languages` as well (slightly edited):

```
$ tree-sitter.load_shared dump-languages
name: clojure
scope: source.clojure
parser: "$HOME/src/tree-sitter-clojure/."
highlights: None
file_types: ["bb", "clj", "cljc", "cljs", "edn"]
content_regex: None
injection_regex: None

name: clojure
scope: 
parser: "/tmp/tree-sitter-clojure"
highlights: None
file_types: []
content_regex: None
injection_regex: None
```

---

On the subject of ergnomics of the arguments...I wonder if folks would prefer `--lang-name` to `--language-name` [1].  We already seem to be abbreviating in the following cases:

* `config` for `configuration`
* `js` for `javascript` (in `--js-runtime`)
* `libdir` for `library directory` (?)
* `subdir` for `subdirectory`
* `stat` for `statistics`

This PR also currently uses `--lib-path` (which is fine by me :) ).

We're also using `grammar-path` so I wonder if there is some sense to `grammar-name`...though that's longer and we already say `dump-languages`...

---

[1] May be `lang` is not likely to be interpreted as anything other than `language`...especially in the context of tree-sitter?

WillLillis:
> I wonder if folks would prefer --lang-name to --language-name

Fine by me!

> We're also using grammar-path so I wonder if there is some sense to grammar-name...though that's longer and we already say dump-languages...

My main thought for using `grammar-path` there was because that argument points to a tree-sitter grammar directory. I'm not totally confident on the verbiage, but my sense was that the languages are something you "get out of" the grammars. Don't feel super strongly either way, but I'd lean towards keeping that one the same just for the sake of minimizing unnecessary breakage between versions.

sogaiu:
> My main thought for using grammar-path there was because that argument points to a tree-sitter grammar directory. 

Yes, I think using the word `grammar` (in `grammar-path`) is a fine choice.

Sorry, I didn't mean to imply that ought to change or anything.

There is also `name` inside of the `grammars` array in `tree-sitter.json` which I presume is meant to match what is in `grammar.js` as well as `grammar.json`.  Although I'm not highly enthusiastic about suggesting `grammar-name` over `lang-name`, I think `grammar-name` feels more consistent [1].

As someone who has worked on grammars for which there were multiple different implementations, the idea of there being multiple grammars for a single language is much more at the front of my mind so-to-speak.

Perhaps other folks have opinions though :)

---

[1] Though I won't be surprised to see mistypings of `grammar` being more common than `lang` :sweat_smile:

sogaiu:
It occurs to me that since the `--lang-name` / `--grammar-name` option is not likely to be used as much:

> If this doesn't work, users can pass --language_name.

may be it's ok if the name is `--grammar-name`?

WillLillis:
> It occurs to me that since the `--lang-name` / `--grammar-name` option is not likely to be used as much:
> 
> > If this doesn't work, users can pass --language_name.
> 
> may be it's ok if the name is `--grammar-name`?

The only objection I'd have to this is that I found "language name" to be used internally for this kind of data while figuring this feature out:

https://github.com/tree-sitter/tree-sitter/blob/8e90799e2791a322f6df5c435280433e2c381773/crates/loader/src/loader.rs#L854-L859

Beyond that, I don't have a super strong preference either way. Going to give this some more thought as I work through the rest of this PR.

sogaiu:
> The only objection I'd have to this is that I found "language name" to be used internally for this kind of data

Since it's not likely to be used that often, may be in the interest of keeping things simpler, staying with `--language-name` or `--lang-name` is better then.

WillLillis:
>  Update the docs

I actually won't do this in *this* PR. The website is served off of the master branch, so it doesn't make sense to add this until we're ready to cut a release with these changes. I've added updates to some existing options that were missing entries, though. (Maybe this should be pulled into a separate PR that can be backported?)

> *Potentially* extend this functionality to other subcommands (i.e. fuzz). Language loading works a bit differently for other subcommands, so we'll have to be careful not to introduce subtle regressions.

I've added the `lib-path` and `--lang-name` arguments to the `test` and `fuzz` subcommands. I didn't add the arguments for the following subcommands:

- `highlight`: the current highlight functionality requires a `LanguageConfiguration` that is retrieved by the passed in `scope`. Hacking around this would lead to unintuitive/wrong behavior, and isn't work the added complexity imo.
- `tags`: Similar story to `highlight`
- `playground`: Uses a wasm file rather than a dynamic library. This would require a bit of a refactor, but could be do-able as a future improvement. I don't see great utility in this, though. 
- other subcommands: N/A

PR #4646, opened 2025-07-27
cli: Adapt default Makefile/CMakeLists for repos not storing grammar.json

wetneb:
This is a new version of #4580, to make it easier not to include tree-sitter generated files in grammar repos, as outlined in #930.

This makes it possible to run `make` or `cmake --build .` in grammar repos where `grammar.json` hasn't been included, and still get the parser compilation to work. The build scripts still work if `grammar.json` is included.

When `grammar.json` is present, this gives:
```sh
$ make
tree-sitter generate --stage=parser src/grammar.json
cc -Isrc -std=c11 -fPIC   -c -o src/parser.o src/parser.c
…
```

When only `grammar.js` is present, this gives:
```sh
$ make
tree-sitter generate --stage=json grammar.js
tree-sitter generate --stage=parser src/grammar.json
cc -Isrc -std=c11 -fPIC   -c -o src/parser.o src/parser.c
…
```
(and similarly for `cmake --build .`)

I don't know if it's possible to encourage `make` to just run `tree-sitter generate grammar.js` in this case, generating all files in one go. The [rules with grouped targets](https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html) seem to be theoretically designed for that, but I couldn't get it to work without overriding the more granular rules. I think it could also work without this feature, as the overhead of starting two processes shouldn't be significant compared to the complexity of the generation work.

### DIFF
--- a/crates/cli/src/templates/cmakelists.cmake
+++ b/crates/cli/src/templates/cmakelists.cmake
@@ -19,10 +19,17 @@ include(GNUInstallDirs)
 
 find_program(TREE_SITTER_CLI tree-sitter DOC "Tree-sitter CLI")
 
+add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
+                   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/grammar.js"
+                   COMMAND "${TREE_SITTER_CLI}" generate grammar.js
+                            --stage=json
+                   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                   COMMENT "Generating grammar.json")
+
 add_custom_command(OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/src/parser.c"
                    DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/grammar.json"
                    COMMAND "${TREE_SITTER_CLI}" generate src/grammar.json
-                            --abi=${TREE_SITTER_ABI_VERSION}
+                            --stage=parser --abi=${TREE_SITTER_ABI_VERSION}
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
> I don't know if it's possible to encourage `make` to just run `tree-sitter generate grammar.js` in this case, generating all files in one go. 

Maybe we can use a variable (`TS_GENERATE_ALL`?) for it.

wetneb:
I've added migrations and tested them on a couple of grammars with `tree-sitter init -u`. They seem to work as expected.

> Maybe we can use a variable (TS_GENERATE_ALL?) for it.

I don't know how to do that, and am not sure if it's worth it if people need to enable a specific variable for it. I propose to stick with the two-stage generation. People can also edit the Makefile if they prefer a different process, I would say.

crates/cli/src/init.rs
```diff
@@ -487,6 +487,24 @@ pub fn generate_grammar_files(
                     if !contents.contains("cd '$(DESTDIR)$(LIBDIR)' && ln -sf") {
                         eprintln!("Replacing Makefile");
                         generate_file(path, MAKEFILE_TEMPLATE, language_name, &generate_opts)?;
+                    } else {
+                        // Migration from https://github.com/tree-sitter/tree-sitter/pull/4646,
+                        // to generate grammar.json from grammar.js
```
```suggestion
```

ObserverOfTime:
```make
ifdef TS_GENERATE_ALL
$(PARSER): grammar.js
	$(TS) generate $^
else
$(SRC_DIR)/grammar.json: grammar.js
	$(TS) generate --stage=json $^

$(PARSER): $(SRC_DIR)/grammar.json
	$(TS) generate --stage=parser $^
endif
```
Similarly for CMake, but as you said it's probably not worth it.

PR #4647, opened 2025-07-27
docs: update options for generate command

wetneb:
This syncs up the docs with `tree-sitter generate --help`. I've removed the section about `--build` to avoid advertising a deprecated option.
Follow-up to #4601.


clason:
Please don't do that; most people use 0.25 (or earlier) and will be confused when this is suddenly gone (because the docs are pulled from master).

Let's park documentation updates until the 0.26 release when we know the final shape of things and can push a full update in one go.

wetneb:
Okay, I expected the docs would only get published when a release is created.
Feel free to close this PR or leave it open to merge it when 0.26 is released.

Issue #4515, opened 2025-06-13
Function parameter inconsistencies


ohunter:
### Problem

The functions [`Node::child`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.child) and [`Node::named_child`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.named_child) both take `usize` as their index parameter, but the functions [`Node::field_name_for_child`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.field_name_for_child) and [`Node::field_name_for_named_child`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.field_name_for_named_child) expect `u32` as their index parameter. 

### Steps to reproduce

NA

### Expected behavior

I'd expect them to have the same expected type for the index parameter.

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.25.3

### Operating system/version

Linux 6.15.0-1

abhilov23:
Hey, I would like to work on this.

PR #4652, opened 2025-07-28
fix(rust): change index parameter types from u32 to usize

junnplus:
Fix https://github.com/tree-sitter/tree-sitter/issues/4515


ObserverOfTime:
I think the opposite makes more sense.

PR #4657, opened 2025-07-29
Expose internal error nodes as `ERROR_INTERNAL`

jaboatman:
This PR remaps `_ERROR` nodes as `ERROR_INTERNAL` and makes them visible (but not named). This is useful for identifying which parts of an `ERROR` node actually caused the error to occur.

I'm not sure what kind of broader implications these changes would have on the whole system, but it seems to work well. Let me know if you have any suggestions or if this would cause other problems.

Thanks

### DIFF
--- a/lib/src/language.c
+++ b/lib/src/language.c
@@ -129,7 +129,7 @@ TSSymbolMetadata ts_language_symbol_metadata(
   if (symbol == ts_builtin_sym_error)  {
     return (TSSymbolMetadata) {.visible = true, .named = true};
   } else if (symbol == ts_builtin_sym_error_repeat) {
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I know we have to improve the ability to diagnose errors. I think that we should do this via special API functions that you can call on `ERROR` nodes, not by directly exposing these internal error nodes.

jaboatman:
> I know we have to improve the ability to diagnose errors. I think that we should do this via special API functions that you can call on `ERROR` nodes, not by directly exposing these internal error nodes.

Yes, fair enough. Working on it now, will post a follow up PR shortly for review.

jaboatman:
Follow up PR at #4660

PR #4660, opened 2025-07-31
API additions to expose error information on `ERROR` nodes

jaboatman:
Follow up from #4657 this PR adds some APIs to access information about the error nodes.

It works by accessing the inner `_ERROR` node when it exists to implement the byte/point range functions. If it doesn't exist, it assumes the node is an `ERROR` leaf node, so it returns the whole range instead.


lib/src/parser.h
```diff
@@ -10,6 +10,7 @@ extern "C" {
 #include <stdlib.h>
 
 #define ts_builtin_sym_error ((TSSymbol)-1)
+#define ts_builtin_sym_error_repeat (ts_builtin_sym_error - 1)
```
ObserverOfTime, line 13: Is this necessary? This restricts fields to `UINT16_MAX - 2` (which should be enough anyway but still).

jaboatman: In general, I don't know. But I didn't define this field, just moved it from a different header to be able to access it in different places.

clason:
Closing as stale; if this is still desired (which needs to be argued for with a concrete use case), it needs to be opened against the current state of `master`, which has diverged quite a bit.

jaboatman:
I have a workaround that makes this unnecessary, so no need for this PR.

PR #4671, opened 2025-08-05
Inline C source code in Wasm source map

### LINKED ISSUE #4356: Put source files in .map in JS build
Emscripten now allows you to do `-gsource-map=inline` which copies the C code into the source map file instead of having to serve all of them for debug information. I don't know how or if you can pass this to esbuild but it makes debugging easier because you dont have to configure your webserver to serve C files from node_modules just to see C code when you get an error. It will make the source map a few MB though.

#3381

verhovsky:
Closes #4356

Currently, if you go to https://www.npmjs.com/package/web-tree-sitter?activeTab=code and look at tree-sitter.js.map it will have a `"sourcesContent":` field. This is where it copies the source code of the relevant JavaScript files so that when a browser loads the .map file it has everything needed to show the source code of a minified function. But if you open tree-sitter.wasm.map it doesn't have that field, to see the C source code of a given error line your browser needs to load tree-sitter.wasm.map but also the relevant C file as well, which (as I understand it) requires making sure the generated path is correct and that your server serves the C files at the resulting path. Also debug/tree-sitter.wasm.map has the same paths in `"sources":` which probably means it won't work, I would expect that all the sources should be `../<whatever` since the lib/ directory is not copied in debug/. Also lib/tree-sitter.wasm.map has `"sources":["../../src/alloc.c"` which I'm pretty sure are wrong as well, since `../..` would be outside the tree-sitter/ root folder...

Adding `-gsource-map=inline` copies the files into `sourcesContent` and makes using the source maps much easier but it requires emcc 4.0.6+ and makes the .map file size larger, which doesn't really matter.


crates/xtask/src/build_wasm.rs
```diff
@@ -150,8 +150,8 @@ pub fn run_wasm(args: &BuildWasm) -> Result<()> {
 
     #[rustfmt::skip]
     emscripten_flags.extend([
-        "-gsource-map",
-        "--source-map-base", ".",
+        "-gsource-map=inline",
+        "--source-map-base", "./",
```
verhovsky: `--source-map-base .` makes the browser request `.web-tree-sitter.wasm.map` (leading dot) instead of `web-tree-sitter.wasm.map`, which is obviously incorrect

image

amaanq (approved):

crates/xtask/src/build_wasm.rs
verhovsky: actually, `"--source-map-base", "./"` should be redundant

verhovsky: Removing that argument does make a difference

```
sourceMappingURL
./web-tree-sitter.wasm.map
```

vs

```
sourceMappingURL
web-tree-sitter.wasm.map
```

But bare filenames in the browser (like in HTML and [in this case too](https://426.ecma-international.org/1.0/index.html#linking-generated-code)) they're considered "relative to the generated code’s source origin" and other tools/bundlers just put the filename, for example for the generated javascript

```sh
grep -i "sourceMappingURL\|\.map" web-tree-sitter.js
  const values = stringSteps.map((s) => s.value);
//# sourceMappingURL=web-tree-sitter.js.map
```

so we should as well in the Wasm, so I removed this argument, it seems to have been added by mistake anyway.

amaanq: Awesome, thanks for double checking

PR #4673, opened 2025-08-05
fix(bindings): use class name + some cleanup

ObserverOfTime:
- Use custom class name in Swift bindings
- Drop unused `path_in_ignore` function
- Make conditional assignments consistent

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -465,12 +465,12 @@ pub fn generate_grammar_files(
                     generate_file(path, MAKEFILE_TEMPLATE, language_name, &generate_opts)
                 },
                 |path| {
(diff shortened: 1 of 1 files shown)


crates/cli/src/init.rs
```diff
@@ -729,22 +711,17 @@ pub fn generate_grammar_files(
     // Generate Swift bindings
     if tree_sitter_config.bindings.swift {
         missing_path(bindings_dir.join("swift"), create_dir)?.apply(|path| {
-            let lang_path = path.join(format!("TreeSitter{camel_name}"));
+            let lang_path = path.join(class_name.clone());
```
WillLillis:
```suggestion
            let lang_path = path.join(&class_name);
```

WillLillis (approved):

PR #4678, opened 2025-08-07
fix(rust): ignore new mismatched-lifetime-syntaxes lint

WillLillis:
Fixes the CI failures seen in https://github.com/tree-sitter/tree-sitter/pull/4677

Not sure about the missing headers at the moment...

EDIT: Basically the same issue but with a dependency of wasmtime


ObserverOfTime:
Lifetimes were a mistake.

sogaiu:
Perhaps it takes a lifetime to realize some things :sweat_smile:

clason (approved):

PR #4681, opened 2025-08-08
feat(bindings): add Java bindings

ObserverOfTime:
I believe the Java bindings are now stable enough to include (opt-in).

The parser namespace (by default `io.github.tree-sitter`) is converted to `io.github.treesitter.jtreesitter.parsername` for the package name and `src/main/java/io/github/treesitter/jtreesitter/parsername` for the directory.

The `pom.xml` includes the necessary configuration to publish packages to [Maven Central](https://central.sonatype.org/).
However, the user must have a Sonatype account and a GPG key and add 4 secrets to the workflow…
image

Alternatively, we can publish to [JitPack](https://jitpack.io/) which doesn't require any secrets or workflows.
Instead, it needs a `jitpack.yml` file and the namespace must start with <code><em>com</em>.github</code>.
Also, end users must add its repository to their build files alongside Maven Central.


crates/cli/src/templates/binding.java
ObserverOfTime, line 1: /cc @Marcono1234

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,62 @@
...
+
+public final class PARSER_CLASS_NAME {
+    private static final ValueLayout VOID_PTR =
+            ValueLayout.ADDRESS.withTargetLayout(MemoryLayout.sequenceLayout(Long.MAX_VALUE, ValueLayout.JAVA_BYTE));
+    private static final FunctionDescriptor FUNC_DESC = FunctionDescriptor.of(VOID_PTR);
+    private static final Linker LINKER = Linker.nativeLinker();
+    private static final TreeSitterJava INSTANCE = new PARSER_CLASS_NAME();
+
+    private final Arena arena = Arena.ofAuto();
+    private SymbolLookup lookup = null;
```
Marcono1234: Maybe the `lookup` field should be `volatile`; otherwise you could get difficult to debug issues when `language()` is called by separate threads (not necessarily concurrently).

Using `volatile` solves some of the thread-safety issues but you could still end up with a race-condition where two threads use different `lookup` values; not sure if that is safe. If not, might even need synchronization.

See related https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private SymbolLookup findLibrary() {
+        try {
+            var library = System.mapLibraryName("tree-sitter-KEBAB_PARSER_NAME");
+            return SymbolLookup.libraryLookup(library, arena);
+        } catch (IllegalArgumentException ex1) {
+            try {
+                return SymbolLookup.loaderLookup();
+            } catch (UnsatisfiedLinkError ex2) {
```
Marcono1234, line 44: Is this correct? At least based on the documentation `loaderLookup()` won't throw `UnsatisfiedLinkError`.

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+    }
+
+    private static UnsatisfiedLinkError unresolved(String name) {
+        return new UnsatisfiedLinkError("Unresolved symbol: %s".formatted(name));
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    private static MemorySegment call(SymbolLookup lookup, String name) throws UnsatisfiedLinkError {
```
Marcono1234, line 56: What about using `Language#load(SymbolLookup, String)` from jtreesitter instead? I know that you originally wanted to get rid of it again in the future, but I still think it is quite useful and could also avoid the code duplication here.

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,62 @@
+package PARSER_NS_CLEANED.jtreesitter.LOWER_PARSER_NAME;
+
+import java.lang.foreign.*;
+
+public final class PARSER_CLASS_NAME {
```
Marcono1234, line 5: Maybe add an explicit `private PARSER_CLASS_NAME() {}` constructor to override the implicit `public` default constructor? This avoids that users accidentally create instances of this class themselves.

crates/cli/src/templates/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <maven.compiler.release>22</maven.compiler.release>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.deploy.skip>true</maven.deploy.skip>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>io.github.tree-sitter</groupId>
+      <artifactId>jtreesitter</artifactId>
+      <version>0.25.4</version>
+      <optional>true</optional>
```
Marcono1234, line 44: Why `optional`? jtreesitter is required to use the bindings.

If a consuming project needs a different jtreesitter version, then it can still override the version (respectively the user can just change it in this `pom.xml` file).
Making it `optional` requires that consuming projects always have to specify a jtreesitter dependency, which might make usage more cumbersome.

crates/cli/src/templates/binding.java
Marcono1234: Thanks for the ping! I have added some comments, hopefully they are helpful.

crates/cli/src/templates/pom.xml
ObserverOfTime: The tree-sitter dependency is optional in all of the bindings.
This makes it possible to use a hypothetical fork instead.

crates/cli/src/init.rs
```diff
@@ -910,30 +972,49 @@ fn generate_file(
...
             }
+            "pom.xml" => {
+                replacement = replacement.replace(AUTHOR_EMAIL_PLACEHOLDER_JAVA, "");
+            }
             _ => {}
         }
     }
 
-    if filename == "package.json" {
+    if filename == "package.json" || filename == "pom.xml" {
```
WillLillis: Nit, but having to check for the filename twice feels a bit awkward.  I think this could be a little clearer with a match:

```rust
match (generate_opts.author_url, filename) {
    (Some(url), "package.json" | "pom.xml") => replacement.replace(AUTHOR_URL_PLACEHOLDER, url),
    (None, "package.json") => replacement.replace(AUTHOR_URL_PLACEHOLDER_JS, ""),
    (None, "pom.xml") => replacement.replace(AUTHOR_URL_PLACEHOLDER_JAVA, ""),
    _ => {},
}
```

crates/cli/src/templates/binding.java
ObserverOfTime: > you could still end up with a race-condition where two threads use different `lookup` values

We can deal with that if and when it happens.

crates/cli/src/templates/binding.java
ObserverOfTime: What do you think regarding Maven Central vs JitPack?

crates/cli/src/templates/binding.java
ObserverOfTime: I have many issues with that function.
Most importantly, I don't want a public function that can load arbitrary symbols.

crates/cli/src/templates/pom.xml
Marcono1234: Ah ok. But do you think that this case will be common? I assume it is the exception, and in that case users might rather edit the `pom.xml` directly. Otherwise they risk that their fork causes issues for the bindings, but would not notice it for the bindings themselves (since those use the official jtreesitter) and only notice it for consuming projects.

crates/cli/src/templates/binding.java
Marcono1234: I am not completely sure, though I haven't published any artifacts on Maven Central yet. The JitPack config is probably nice to have for users who don't want to publish on Maven Central (yet). I guess without the `jitpack.yml` file JitPack would pick a too old JDK by default?

Though maybe you have actually more insight into how many grammar authors publish their bindings on public repositories (e.g. JavaScript bindings on npm, ...), or how it is handled for the official tree-sitter grammars.

The following other user groups come into my mind:
- Those who only perform local development (and install the bindings to their local Maven repository with `mvn install`)
- Those who already have a publishing setup and possibly just tweak the `pom.xml` to reference the setup as `<parent>`
- Those with a custom remote repository (e.g. in companies), where neither deployment to Maven Central nor JitPack is relevant (might overlap with point 2)

So I guess unless the generated files hamper those users too much, it might be fine to try to support both Maven Central and JitPack (in the way you have currently implemented it)?

That is just my opinion on this though. Not sure if that is what you meant with your question.

ObserverOfTime: Local development doesn't need anything to be supported.
But only one remote repository will be supported by default.

ObserverOfTime:
## Maven Central

### Pros

- End users won't need to add an extra repository to their build files
- The namespace/package of core parsers will be consistent with jtreesitter

### Cons

- Parser authors will need a Sonatype account and GPG key
- Requires a new reusable workflow

## JitPack

### Pros

- Parsers authors don't need to do anything special
- Doesn't require a new workflow

### Cons

- Requires an additional generated file
- End users will need to add an extra repository to their build files
- The namespace/package of parsers will be inconsistent with jtreesitter

ObserverOfTime:
I'd like some user opinions regarding the packaging.

/cc @stackoverflow @bioball @Emmeral @maloewe-ona

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,65 @@
...
+        return language(INSTANCE.lookup);
+    }
+
+    /**
+     * Get the tree-sitter language for this grammar.
+     *
+     * <strong>The {@linkplain Arena} used in the {@code lookup}
+     * must not be closed while the language is being used.</strong>
+     */
+    public static MemorySegment language(SymbolLookup lookup) {
```
Emmeral, line 32: Does it make sense to return a jtreesitter `Language` object directly here?

Emmeral:
So if I understand this correctly, the actual binaries will not be uploaded to the remote repository but only the bindings class and the user of the bindings needs to supply the binaries themself?

Emmeral:
> I'd like some user opinions regarding the packaging.

I don't have a specific opinion on the Maven Central vs JitPack topic. For my use case, I would need to modify the pom in both cases.

crates/cli/src/templates/binding.java
ObserverOfTime: No. None of the bindings return Language objects.

sogaiu:
The cons of Maven Central seem significantly worse to me than those for JitPack.

ObserverOfTime:
Perhaps, but something like this is not possible in JitPack:

> So if I understand this correctly, the actual binaries will not be uploaded to the remote repository but only the bindings class and the user of the bindings needs to supply the binaries themself?

ObserverOfTime:
I'm leaning towards Maven Central after all since it allows us to fully control the release process.

sogaiu:
Just to clarify:

> Parser authors will need a Sonatype account and GPG key

Does this mean that individual grammar / parser authors need to setup and maintain these bits over the long term then?

ObserverOfTime:
They also need (if they want) to set up and maintain tokens for npmjs, pypi, and crates. The only extra here is the GPG key.

sogaiu:
I wonder about the practicality of this kind of arrangement over the longer term.  It seems to me that many a grammar / parser author has lost interest after a certain point.  In a nicer world, may be there would be two or more folks looking after a given grammar / parser for robustness (though OFC that's a larger attack surface as well...).

I've been ok with looking after a few grammars, but I have no interest in setting up and properly looking after additional accounts and/or tokens (and thereby becoming a slightly more juicier target).  Perhaps this is a minority position though.

ObserverOfTime:
Although, now that I think about it, those can also be installed directly from git.
That is _sort of_ possible with JitPack but not (easily) with Maven Central.

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,65 @@
+package PARSER_NS_CLEANED.jtreesitter.LOWER_PARSER_NAME;
+
+import java.lang.foreign.*;
+
+public final class PARSER_CLASS_NAME {
+    private static final ValueLayout VOID_PTR =
+            ValueLayout.ADDRESS.withTargetLayout(MemoryLayout.sequenceLayout(Long.MAX_VALUE, ValueLayout.JAVA_BYTE));
+    private static final FunctionDescriptor FUNC_DESC = FunctionDescriptor.of(VOID_PTR);
+    private static final Linker LINKER = Linker.nativeLinker();
+    private static final TreeSitterJava INSTANCE = new PARSER_CLASS_NAME();
```
amaanq: is hardcoding `TreeSitterJava` here intended

crates/cli/src/templates/binding.java
```diff
@@ -0,0 +1,65 @@
...
+        return call(lookup, "tree_sitter_PARSER_NAME");
+    }
+
+    private SymbolLookup findLibrary() {
+        try {
+            var library = System.mapLibraryName("tree-sitter-KEBAB_PARSER_NAME");
+            return SymbolLookup.libraryLookup(library, arena);
+        } catch (IllegalArgumentException ex1) {
+            try {
+                System.loadLibrary("tree-sitter-KEBAB_PARSER_NAME")
```
amaanq: does this not need a semicolon

crates/cli/src/init.rs
```diff
@@ -909,6 +924,45 @@ pub fn generate_grammar_files(
...
+        missing_path(repo_path.join("pom.xml"), |path| {
+            generate_file(path, POM_XML_TEMPLATE, language_name, &generate_opts)
+        })?;
+
+        missing_path(bindings_dir.join("java"), create_dir)?.apply(|path| {
+            missing_path(path.join("main"), create_dir)?.apply(|path| {
+                let package_path = generate_opts
+                    .namespace
+                    .unwrap_or("io.github.treesitter")
+                    .replace(['-', '_'], "")
```
amaanq, line 1077: just double checking, is replacing `-` unconditionally intended?

ObserverOfTime: Yes. Dashes are not allowed in identifiers and underscores are generally avoided.

crates/cli/src/init.rs
```diff
@@ -909,6 +924,45 @@ pub fn generate_grammar_files(
...
+                    .split('.')
+                    .fold(path.to_path_buf(), |path, dir| path.join(dir))
+                    .join("jtreesitter")
+                    .join(language_name.to_lowercase().replace('_', ""));
```
Vinzent03, line 1081: For a language like `simple_lang` this produces `simplelang` as a directory, but the template still uses `simple_lang` for the package declaration e.g. `package io.github.treesitter.jtreesitter.simple_lang;`
But those two being different is not allowed in java.

ObserverOfTime: You're right. We need to adapt `LOWER_PARSER_NAME` in this case.

WillLillis (approved):
Assuming local tests are working on your end, let's give this a spin.

clason (changes requested):
Missing documentation under https://tree-sitter.github.io/tree-sitter/cli/init.html#the-bindings-field

(which is also missing the Zig bindings!)

(and shouldn't list the Kotlin bindings until they're actually available)

PR #4686, opened 2025-08-10
fix(cli): escape delimiters '"' and '`' in cst node text

WillLillis:
CC @ObserverOfTime

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -735,6 +735,14 @@ const fn escape_invisible(c: char) -> Option<&'static str> {
     })
 }
 
+const fn escape_delimiter(c: char) -> Option<&'static str> {
+    Some(match c {
+        '`' => "\\`",
+        '\"' => "\\\"",
+        _ => return None,
+    })
+}
+
 pub fn render_cst<'a, 'b: 'a>(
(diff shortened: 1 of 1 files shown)


crates/cli/src/parse.rs
```diff
@@ -735,6 +735,14 @@ const fn escape_invisible(c: char) -> Option<&'static str> {
     })
 }
 
+const fn escape_delimiter(c: char) -> Option<&'static str> {
+    Some(match c {
+        '`' => "\\`",
+        '\"' => "\\\"",
```
ribru17, line 741: Should backslashes be escaped too in this case?

WillLillis: Do you think that would help in readability? I'm assuming you mean 
```
foo\\`bar
```

as opposed to (the behavior of the PR currently)

```
foo\`bar
```

ObserverOfTime: I think he meant `\` -> `\\` which is already handled above.

ribru17: I did mean that, sorry I didn't know it was already handled :+1:

ObserverOfTime (approved):

PR #4707, opened 2025-08-16
ci: add a spam closing workflow

ObserverOfTime:
With this we will just add the `spam` label to an issue or PR and it will automatically be cleaned, closed, and locked.


amaanq:
sucks that gh seems to not care about this issue

PR #4712, opened 2025-08-19
feat(lib): add lldb pretty-printers for some C lib types

WillLillis:
Debugging C can be difficult. Something I've run into quite a bit while stepping through the C library is the following pattern:

```C
struct Foo {
	uint32_t bar_count;
	Bar* bars;
	...
}
```

This is used directly in several struct definitions, as well as types created via the `Array` macro. There's no way for debuggers to know that `bars` is a many-item pointer, so only the first item will is displayed. For example (with nvim-dap):

image

Import the module:

```sh
(lldb) command script import /path/to/tree-sitter/lib/lldb_pretty_printers/tree_sitter_types.py
```

image

Or with lldb directly:

image

This should cover all uses of the `Array` macro, as well as several many-item pointer instances within other struct definitions. More improvements can be made in the future (covering more types, addressing a few of the TODOs, etc.), but I think this is a good starting point.

Some docs: https://lldb.llvm.org/use/variable.html#synthetic-children


lib/lldb_pretty_printers/tree_sitter_types.py
```diff
@@ -0,0 +1,59 @@
+import lldb
+
+# Even though these are "unused", we still need them in scope in order for the classes
+# to exist when we register them with the debugger
+from ts_tree import TSTreeSyntheticProvider
+from table_entry import TableEntrySyntheticProvider
+from ts_array import (ArraySyntheticProvider, anon_array_recognizer)
```
ObserverOfTime:
```suggestion
from .ts_tree import TSTreeSyntheticProvider
from .table_entry import TableEntrySyntheticProvider
from .ts_array import ArraySyntheticProvider, anon_array_recognizer
```
To avoid trying to import from a hypothetical package with that name.
May or may not need an `__init__.py` file to work properly.

lib/lldb_pretty_printers/tree_sitter_types.py
```diff
@@ -0,0 +1,59 @@
...
+                  TreeSitterType.STACK_SUMMARY,
+                  TreeSitterType.SUBTREE_ARRAY,
+                  TreeSitterType.MUTABLE_SUBTREE_ARRAY,
+                  TreeSitterType.STRING_DATA,
+                  TreeSitterType.STACK_NODE_ARRAY
+                 ]
+
+def __lldb_init_module(debugger: lldb.SBDebugger, _dict):
+    # lldb.formatters.Logger._lldb_formatters_debug_level = 2
+    # lldb.formatters.Logger._lldb_formatters_debug_filename = "/tmp/tree-sitter/lldb.log"
```
ObserverOfTime:
```suggestion
```

lib/lldb_pretty_printers/tree_sitter_types.py
WillLillis: Not having any luck with/without `__init__.py`. Going to keep looking but any help would be appreciated, I've written very little python.

```
(lldb) command script import ../tree-sitter/lib/lldb_pretty_printers/tree_sitter_types.py
error: module importing failed: Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/lillis/projects/tree-sitter/lib/lldb_pretty_printers/tree_sitter_types.py", line 12, in <module>
    from .ts_tree import TSTreeSyntheticProvider
ImportError: attempted relative import with no known parent package
```

WillLillis: This feels very hacky but it works:

```python
import sys, os;
SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
sys.path.append(os.path.dirname(SCRIPT_DIR))
if __package__ is None or __package__ == "":
    __package__ = "lldb_pretty_printers"
```

ObserverOfTime: Just keep it as it was.

PR #4716, opened 2025-08-20
build(zig)!: upgrade build.zig to zig 0.15

bfredl:
This no longer supports zig 0.14.

As an alternative, with some tomfoolery (`initWithCapacity`) it would be possible to support both versions at once. However zig 0.14 is considered end-of-life with the release of zig 0.15

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -95,7 +95,7 @@ fn wasmtimeDep(target: std.Target) []const u8 {
 }
 
 fn findSourceFiles(b: *std.Build) ![]const []const u8 {
-  var sources = std.ArrayList([]const u8).init(b.allocator);
+  var sources : std.ArrayList([]const u8) = .empty;
 
   var dir = try b.build_root.handle.openDir("lib/src", .{ .iterate = true });
   var iter = dir.iterate();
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
We _just_ upgraded to 0.14 and 0.15 isn't even released yet.

bfredl:
Neither of those statements are true. tree-sitter has worked with zig 0.14 more or less since 0.14 was released.  Zig 0.15 has a source release but it is true binaries might take one more day or two. but surely that does not prevent us from having a discussion?

ObserverOfTime:
It worked using deprecated functions and outdated metadata.

How hacky/difficult would it be to support both versions?

bfredl:
> It worked using deprecated functions 

Which was cleaned up in order to keep compatibility with master and support the imminent _0.15_ release.

It can easily be done. I can push the workaround I already proposed:p

ObserverOfTime:
You should also lower the minimum supported version.

bfredl:
calm down. this is a proposal (now with two variants). I want to hear the opinion from more maintainers, then I will adjust this to match whatever actually ends up decided.

clason:
@bfredl I trust you on this. Since this is zerover support, compatibility considerations do not apply. Anybody who is actually using this is already champing at the bit to upgrade to 0.15. I'll wait for the actual release (on GH and the binary download page), then I'll merge this (without the compat shim).

PR #4718, opened 2025-08-20
chore: remove emscripten xtask & workflow

ObserverOfTime:
The workflow won't run after dependabot PRs anymore (which is currently disabled).
Instead, it will run once per month and on demand.

I can do the same for the wasmtime xtask if we want.


clason:
Come on, this is literally one line in one file. Let people who actually need the newer version handle this manually.

clason:
(And wasmtime moves too fast to track blindly. There's a reason I haven't bumped it beyond 29.0.)

ObserverOfTime:
> Come on, this is literally one line in one file.

That's exactly why I'm removing the xtask.

> Let people who actually need the newer version handle this manually.

I can remove the schedule if you just want to run the workflow manually.

clason:
My point is that this doesn't need a workflow.

ObserverOfTime:
The workflow was already here and is less complicated than the xtask.

ObserverOfTime:
I tried the latest emscripten version (`4.0.13`) and the tests fail with this error:

```
 FAIL  test/query.test.ts > Query > Set a timeout > returns less than the expected matches
ReferenceError: LE_HEAP_STORE_I64 is not defined
 ❯ _clock_time_get lib/web-tree-sitter.mjs:1576:3
    1574|   // "now" is in ms, and wasi times are in ns.
    1575|   var nsec = Math.round(now * 1e3 * 1e3);
    1576|   LE_HEAP_STORE_I64(((ptime) >> 3) * 8, BigInt(nsec));
       |   ^
    1577|   return 0;
    1578| }
 ❯ __clock_gettime wasm:/wasm/web-tree-sitter.wasm-000c84f6:1:165095
 ❯ ts_query_cursor_exec wasm:/wasm/web-tree-sitter.wasm-000c84f6:1:79885
 ❯ ts_query_matches_wasm wasm:/wasm/web-tree-sitter.wasm-000c84f6:1:163354
 ❯ Query.matches src/query.ts:736:7
 ❯ test/query.test.ts:46
```

ObserverOfTime:
Looks like emscripten-core/emscripten#25042 is going to fix that error.
FWIW it works when compiled in debug mode which enables `SAFE_HEAP`.

amaanq:
Yeah I think I agree with clason, the workflow + xtask are unnecessary and it's probably better to manually bump emscripten once in a while.

ObserverOfTime:
And why not use the workflow to manually bump and test it?

amaanq:
It's unnecessary when I can just test it locally and push a commit. If the workflow fails CI I'd still be testing it locally, and it saves all of 5 seconds when it succeeds.

amaanq (approved):
ty!

PR #4723, opened 2025-08-22
build.zig improvements

ObserverOfTime:
- Expose `wasmtimeDep` function
- Propagate link mode to wasmtime
- Reformat files (supersedes #4622)


bfredl (approved):

ObserverOfTime:
I still seem to be getting this warning _sometimes_ even after propagating the link mode:
```
error: warning(link): unexpected LLD stderr:
ld.lld: warning: libtree-sitter.a: archive member libwasmtime.so is neither ET_REL nor LLVM bitcode
```
I guess we should only link against wasmtime when building in shared mode?

WillLillis:
> I still seem to be getting this warning _sometimes_ even after propagating the link mode:
> 
> ```
> error: warning(link): unexpected LLD stderr:
> ld.lld: warning: libtree-sitter.a: archive member libwasmtime.so is neither ET_REL nor LLVM bitcode
> ```
> 
> I guess we should only link against wasmtime when building in shared mode?

I think this is an issue on zig's side, see:
- https://github.com/ziglang/zig/issues/19341
- https://github.com/ziglang/zig/issues/20476

ObserverOfTime:
Then there's nothing else we can do about it.

PR #4725, opened 2025-08-22
feat(playground): add a button to copy the tree

ObserverOfTime:
This also fixes some paths that broke with the great relocation™.

Closes #4349


poliorcetics:
This MR broke the current (and I suspect past) release: https://github.com/tree-sitter/tree-sitter/issues/4777

PR #4728, opened 2025-08-23
fix(bindings): add tree-sitter as npm dev dependency

ObserverOfTime:
npm is supposed to automatically install peer dependencies since v7
but sometimes it's not doing it and we need this dependency for tests

¯\\_(ツ)\_/¯


ObserverOfTime:
We love random CI failures!

PR #4730, opened 2025-08-23
fix(bindings): improve zig dependency fetching logic

ObserverOfTime:
Currently, including a tree-sitter parser as a dependency in a zig project and running `zig build test` on the project will fetch the zig-tree-sitter dependency declared by the parser. This is a problem because (a) consumers may not want this dependency for whatever reason and (b) due to how often Zig breaks everything and how scarcely most tree-sitter parsers are updated, the zig-tree-sitter version pinned by the parser module will often be outdated and broken.

The workaround I used was taken from https://ziggit.dev/t/11234.

I also noticed that we were unconditionally replacing the files and fixed that while I was at it.

---

/cc @nvlled

### DIFF
--- a/crates/cli/src/templates/build.zig
+++ b/crates/cli/src/templates/build.zig
@@ -68,13 +68,16 @@ pub fn build(b: *std.Build) !void {
     });
     tests.root_module.addImport(library_name, module);
 
-    var args = try std.process.argsWithAllocator(b.allocator);
-    defer args.deinit();
-    while (args.next()) |a| {
-        if (std.mem.eql(u8, a, "test")) {
-            const ts_dep = b.lazyDependency("tree_sitter", .{}) orelse continue;
-            tests.root_module.addImport("tree-sitter", ts_dep.module("tree-sitter"));
-            break;
+    // HACK: fetch tree-sitter dependency only when testing this module
+    if (b.pkg_hash.len == 0) {
+        var args = try std.process.argsWithAllocator(b.allocator);
+        defer args.deinit();
+        while (args.next()) |a| {
+            if (std.mem.eql(u8, a, "test")) {
+                const ts_dep = b.lazyDependency("tree_sitter", .{}) orelse continue;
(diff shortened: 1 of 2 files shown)


crates/cli/src/templates/build.zig
```diff
@@ -68,13 +68,16 @@ pub fn build(b: *std.Build) !void {
...
 
-    var args = try std.process.argsWithAllocator(b.allocator);
-    defer args.deinit();
-    while (args.next()) |a| {
-        if (std.mem.eql(u8, a, "test")) {
-            const ts_dep = b.lazyDependency("tree_sitter", .{}) orelse continue;
-            tests.root_module.addImport("tree-sitter", ts_dep.module("tree-sitter"));
-            break;
+    // HACK: fetch tree-sitter dependency only when testing this module
+    if (b.pkg_hash.len == 0) {
```
nvlled, line 72: I was going to point out that this is an internal behaviour that could break anytime as castholm pointed out, but seeing how zig still frequently breaks, this isn't really an issue at all. Using it like this might actually push the core zig devs to actually make a proper API for this use case.

ObserverOfTime: Please update tree-sitter/tree-sitter-c#275 with these changes.

WillLillis (approved):

crates/cli/src/templates/build.zig
nvlled: Okay

nvlled: @ObserverOfTime The generated template runs fine on 0.14.1, but it needs to refer to an updated zig-tree-sitter for it to run for the 0.15.1.

This guy probably has something working already https://github.com/tree-sitter/zig-tree-sitter/issues/27

clason: Yes, that is deliberate. Since the switch to 0.15 is a breaking change, we cannot backport this. Migrating to 0.15 is a separate task (and probably blocked by the numerous upstream regressions that hopefully will get fixed by 0.15.2).

PR #4739, opened 2025-08-28
fix(cli): show rule names for rules with metadata

### LINKED ISSUE #4535: Allow regex to be used for reserved words
### Problem

Languages with case-insensitive or style-insensitive keywords requires regex to correctly recognize keywords. At the time of writing, reserved words does not accept anything but strings.

### Expected behavior

_No response_

amaanq:
- Closes #4535

### Problem

It's not clear that reserved rules must exist in the grammar

### Solution

Document this. Additionally, for rules wrapped in a call to `token()` in a reserved rule, the cli would say that the rule name was unknown. This has been fixed by extracting the metadata rule's underlying rule.


WillLillis (approved):

ObserverOfTime:
Welcome back! :tada:

PR #4750, opened 2025-08-29
fix(generate): warn users when extra rule can lead to parser hang

### LINKED ISSUE #4405: `tree-sitter parse` hangs parsing string literal
### Problem

I am trying to parse a simple string literal I took the rule from [tree-sitter javascript grammar](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js#L932) but `parse` command hangs until it's killed.

the bun version I am using is `1.2.8`, to install it run:
```sh
curl -fsSl https://bun.sh/install | bash -s "bun-v1.2.8"
```

### Steps to reproduce

```sh
git clone https://github.com/finelang/tree-sitter-fine.git
tree-sitter generate --js-runtime=bun
echo '"\"' >> sample.txt
tree-sitter parse sample.txt
```

### Expected behavior

I expect it to terminate with the error below.

This might shed some light, if the `module` rule is:
```js
{
  module: $ => $.string
}
```
it works: stops with the error: `sample.txt      Parse:    0.06 ms           71 bytes/ms (MISSING """ [1, 0] - [1, 0])`

tree-sitter 0.25.3

### Operating system/version

ubuntu 24.04 via WSL

WillLillis:
## The problem

When a *named* rule in extras is able to match the empty string, parsing can hang in certain situations (i.e. near EOF). This is demonstrated by #4405.

## The Solution

There isn't a great way to fix the actual hang in the core lib. An easy workaround for grammar authors is to simply inline the rule in question into extras. The CLI will now warn during generation if such a case is detected.

- Closes #4405


crates/generate/src/parse_grammar.rs
```diff
@@ -262,6 +263,27 @@ pub(crate) fn parse_grammar(input: &str) -> ParseGrammarResult<InputGrammar> {
...
+            };
+            let matches_empty = match inner_rule {
+                Rule::String(rule_str) => rule_str.is_empty(),
+                Rule::Pattern(ref value, _) => Regex::new(value)
+                    .map(|reg| reg.is_match(""))
+                    .unwrap_or(false),
+                _ => false,
+            };
+            if matches_empty {
+                eprintln!("Warning: Named extra rule `{name}` matches the empty string. This can cause infinite loops while parsing.");
```
clason:
```suggestion
                eprintln!("Warning: Named extra rule `{name}` matches the empty string. Inline this rule to avoid infinite loops while parsing.");
```

PR #4759, opened 2025-08-29
ci(spam): fix permission check

ObserverOfTime:


clason:
What even is the point of this check? If you don't have (at least) triage permission, you can't add the label in the first place.

ObserverOfTime:
Doesn't hurt to be safe. ¯\\_(ツ)\_/¯

ObserverOfTime:
I can remove the check if you want.

PR #4765, opened 2025-08-30
Web bindings cleanup

ObserverOfTime:
### Changes

- Bump emscripten version to `4.0.8` (latest _working_ version)
- Remove source files from package (closes #4606)
- Remove generated files from version control
- Add `HEAP*` symbols to emscripten exported methods
- Use `BigInt` for parser & query timeouts
- Automate the edits to `lib/web-tree-sitter.d.ts`

### Related

- #4207
- #4671
- #4718


ObserverOfTime:
I thought about removing `lib/web-tree-sitter.d.ts` from version control but since it's used in the source files and _not_ the package I think we should keep it (and make sure it's updated when necessary).

lib/binding_web/package.json
```diff
@@ -48,7 +48,6 @@
     "wasm"
   ],
   "files": [
-    "README.md",
```
amaanq, line 54: Why remove the readme?

crates/xtask/src/main.rs
```diff
@@ -87,7 +87,7 @@ struct BuildWasm {
     watch: bool,
     /// Emit TypeScript type definitions for the generated bindings,
     /// requires `tsc` to be available.
-    #[arg(long, short)]
+    #[arg(long, short, env = "EMIT_TSD")]
```
amaanq: is this env needed? it's not even used nor documented

lib/binding_web/package.json
ObserverOfTime: Always included by default.

crates/xtask/src/main.rs
ObserverOfTime: It's the only way to create the file through `npm run build`.
I was going to use it in CI but then I decided to just keep the file.

amaanq: Yeah the file should be kept, so we can remove this.

ObserverOfTime: It can still be useful though.

amaanq: It's not - if we have a use for it we can keep it (*and* document it), but we don't. It's also trivial to pass in `--emit-tsd`. This is not something anyone would use.

ObserverOfTime: It's _impossible_ to pass in `--emit-tsd` through `npm run build`.

amaanq: Nobody's doing that, though if you want it to emit the tsd by default in the build script that's fine with me.

amaanq (approved):

PR #4766, opened 2025-08-30
fix(npm): add directory to repository fields

ObserverOfTime:
and remove non-existent "main" entry point


amaanq:
Is `directory` needed/useful if there isn't a root package.json/JS monorepo structure?

ObserverOfTime:
As far as I can tell, it's only (officially) used by [`npm repo`](https://docs.npmjs.com/cli/v7/commands/npm-repo) but it doesn't hurt to include it.

amaanq (approved):

PR #4769, opened 2025-08-30
feat(test): pin fixture grammars to specific commits

WillLillis:
This attempts to address some pain points raised by downstream package maintainers. Specifically, tree-sitter's current test suite includes several "fixture" grammars (c, bash, etc.). These fixtures are always cloned from their respective repository's latest master, which can lead to unexpected breakage. This PR pins all fixture grammars to the latest tag *at this time*. New commits/releases to these grammars' repositories will not affect the version being tested in tree-sitter. In order to update the fixture grammars, `cargo xtask fetch-fixtures` command now accepts an `--update` flag.

If this update logic is too much, I can remove the majority of these changes and just hardcode the commit shas in the xtask source file. Either way, someone will have to remember to update the fixtures when appropriate.


ObserverOfTime:
I swear there was an issue about this somewhere but I cannot find it.

test/fixtures/fixtures.json
ObserverOfTime, line 1: Can you make this an object?
```json5
{
  "bash": {"ref": "..."}, // omittable "branch", defaults to "master"
  // ...
}
```

ObserverOfTime (commented):
Maybe we should also (later) add a way (feature?) to completely disable these tests.

WillLillis:
> I swear there was an issue about this somewhere but I cannot find it.

There was a discussion about it in the community channel back on August 3/4, not sure if there was ever an issue opened though.

test/fixtures/fixtures.json
WillLillis: I'll push that up in a separate commit, not sure if it's worth the extra boilerplate.

ObserverOfTime:
Yeah that was probably it.

clason:
Maybe pin to latest version instead of commit? This shouldn't make much difference but makes it easier to reason about (and the pin less arbitrary -- and reduces temptation to make a PR for every downstream commit...)

test/fixtures/fixtures.json
WillLillis: Actually digging into this a bit more I don't think it's worth it. Using an JSON object would mean a HashMap on the rust side, meaning we'd lose a consistent ordering when updating the lock fiile.

ObserverOfTime: `BTreeMap`?

clason: KISS.

(And we probably don't need to record the branch as the commit sha is sufficienct -- especially if we pin version.)

ObserverOfTime: We need the branch for updates if we use commits. We don't if we use tags since we can grab the latest.

WillLillis: Just pushed an update that uses tags (and removes all references to branches).

amaanq (approved):

PR #4775, opened 2025-08-31
Use compileStreaming when loading parser Wasm

verhovsky:
First off, this should be a bit more efficient, since it can compile Wasm while the `fetch()` is happening.

This is a follow up to #4671, I'm trying to get sourcemaps working for `Language` objects. The idea here is that `compileStreaming` would correctly edit the source map URL/preserve debug information when loading the file if the parser was compiled with source maps. With the current code that loads the binary and sends it to [here](https://github.com/emscripten-core/emscripten/blob/f78366c8dedd6f78e410090dc6989bbbaaa59cfb/src/lib/libdylink.js#L885-L902), ends up not preserving the sourceMappingUrl and requesting a `wasm://` URL and failing, but with this change it doesn't request anything at all for the source map.


crates/xtask/src/build_wasm.rs
```diff
@@ -247,7 +247,7 @@ fn build_wasm(cmd: &mut Command, edit_tsd: bool) -> Result<()> {
                     "undefined, localScope?: any | undefined, handle?: number | undefined): any"
                 ),
                 concat!(
-                    "loadWebAssemblyModule(binary: Uint8Array, flags: Record<string, boolean>,",
+                    "loadWebAssemblyModule(binary: Uint8Array | WebAssembly.Module, flags: Record<string, boolean>,",
```
verhovsky, line 250: `npm run build:dts` doesn't do anything, I edited web-tree-sitter.d.ts by hand.

verhovsky:
To generate source maps for a parser you have to modify tree-sitter cli

```patch
 crates/loader/src/loader.rs | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/crates/loader/src/loader.rs b/crates/loader/src/loader.rs
index 6f9ac1ae..483c1ed7 100644
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1040,9 +1040,10 @@ impl Loader {
         command.current_dir(src_path).args([
             "-o",
             output_name,
+            "-g",
             "-fPIC",
             "-shared",
-            "-Os",
+            "-O0",
             format!("-Wl,--export=tree_sitter_{language_name}").as_str(),
             "-Wl,--allow-undefined",
             "-Wl,--no-entry",
```

and install it like

```sh
cargo install --path ~/path/to/tree-sitter/crates/cli
```

then (from the parser's directory)

```sh
tree-sitter build --wasm

EMSCRIPTEN_ROOT="$(em-config EMSCRIPTEN_ROOT)"
DWARFDUMP="$(em-config LLVM_ROOT)/llvm-dwarfdump"

python "$EMSCRIPTEN_ROOT/tools/wasm-sourcemap.py" \
  ./tree-sitter-bash.wasm \
  -o ./tree-sitter-bash.wasm.map \
  -u "tree-sitter-bash.wasm.map" \
  -w ./tree-sitter-bash.wasm \
  --dwarfdump "$DWARFDUMP" \
  -s
```

crates/xtask/src/build_wasm.rs
ObserverOfTime: This is generated by running `npm run build:wasm -- --emit-tsd`

verhovsky: Thanks, then we need to update https://github.com/tree-sitter/tree-sitter/blob/2a721498484d32e3cbfa0239bbead888d71c7dbe/lib/binding_web/CONTRIBUTING.md?plain=1#L74

ObserverOfTime:
Compiling parsers to Wasm no longer uses emscripten, so requiring it for sourcemaps is a no-no.
Also, adding debug symbols to the Wasm binaries will make them much larger, which is also a no-no.

verhovsky:
I didn't include either of those changes in this PR, for those reasons.

`compileStreaming(fetch())` is a better way to load Wasm and we do it for the main tree-sitter.wasm already

https://github.com/emscripten-core/emscripten/blob/f78366c8dedd6f78e410090dc6989bbbaaa59cfb/src/preamble.js#L607-L608

but also

```sh
tree-sitter build --wasm --debug
```

should add debug symbols to the Wasm binaries, ~~and probably export that data into a .map file with the `wasm-sourcemap.py` command above (if emscripten is installed) because DWARF symbols in Wasm don't affect error tracebacks in my experience, you get the compiled Wasm stack frames in the error message~~ (edit: turns out I had the DWARF chrome extension disabled).

ObserverOfTime:
Adapting the flags in debug mode is fine.

verhovsky:
It turns out that I had the [DWARF debugging chrome extension](https://chromewebstore.google.com/detail/pdcpmagijalfljmkmjngeonclgbbannb) disabled, after enabling it, DWARF works for stack traces whether you load with `compileStreaming` or just loading the binary

Captura de pantalla 2025-08-31 a la(s) 22 23 52

, so the only important part of this PR I moved to #4779, but if you think it makes sense to load Language Wasm the same way as the main web-tree-sitter.wasm, then this PR can be merged. And if someone wants to use sourcemaps instead of DWARF it also makes sense to merge it.

We could ship source maps for the underlying C code for language packages on npm, but for tree-sitter-bash (for example), it's about 10 MB for the source map because of the auto-generated parser.c

amaanq:
I think this is fine, but we shouldn't ship source maps for parsers by default, 10mb is too large

lib/binding_web/src/language.ts
```diff
@@ -255,16 +255,22 @@ export class Language {
...
-      } else {
-        bytes = fetch(input)
+      try {
+        bytes = await WebAssembly.compileStreaming(fetch(input));
+      } catch (reason) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        console.error(`wasm streaming compile failed: ${reason}`);
+        console.error('falling back to ArrayBuffer instantiation');
+        // fallback, probably because of bad MIME type
+        bytes = await fetch(input)
```
amaanq: Shouldn't we check if we're in node still in the fallback?

amaanq: nvm, whoops

lib/binding_web/src/language.ts
```diff
@@ -255,16 +255,22 @@ export class Language {
...
-        const fs: typeof import('fs/promises') = await import('fs/promises');
-        bytes = fs.readFile(input);
-      } else {
-        bytes = fetch(input)
+      try {
+        bytes = await WebAssembly.compileStreaming(fetch(input));
+      } catch (reason) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        console.error(`wasm streaming compile failed: ${reason}`);
+        console.error('falling back to ArrayBuffer instantiation');
```
amaanq:
```suggestion
        console.error(`Failed to call \`WebAssembly.compileStreaming\`, using \`ArrayBuffer\` instead: ${reason}`);
```

verhovsky: This is copied from emscripten

amaanq: oh, then that's fine

lib/binding_web/src/language.ts
verhovsky: Do we actually want to have this fallback to `Uint8Array`?

It doesn't really make sense to me. It leads to the network request being repeated for any error, like if you serve bad Wasm or a 404 page on this URL it will re-request before actually erroring out. The only reason I can see for it is that `compileStreaming()` is [strict about Content-Type being `application/wasm`](https://chromium.googlesource.com/chromium/src/third_party/+/fa9ae242d338d8231894a96bf1e6bfd1dd025d89/blink/renderer/bindings/core/v8/v8_wasm_response_extensions.cc#554) and if your server doesn't add that header it will refuse to work.

we could have a more complicated implementation that doesn't retry the `fetch()` part https://chatgpt.com/share/68b5fe2a-5204-800a-b078-320b2df6445a

lib/binding_web/src/language.ts
```diff
@@ -255,29 +255,34 @@ export class Language {
...
+
+      const retryResp = response.clone();
+      try {
+        bytes = await WebAssembly.compileStreaming(response);
+      } catch (reason) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        console.error(`wasm streaming compile failed: ${reason}`);
+        console.error('falling back to ArrayBuffer instantiation');
+        // fallback, probably because of bad MIME type
+        bytes = new Uint8Array(await retryResp.arrayBuffer())
```
verhovsky: This fallback to `Uint8Array` doesn't really make sense to me. The only reason I can see for it is that `compileStreaming()` is [strict about Content-Type being `application/wasm`](https://chromium.googlesource.com/chromium/src/third_party/+/fa9ae242d338d8231894a96bf1e6bfd1dd025d89/blink/renderer/bindings/core/v8/v8_wasm_response_extensions.cc#554) and if your server doesn't add that header it will refuse to work.

lib/binding_web/src/language.ts
```diff
@@ -255,29 +255,34 @@ export class Language {
...
+        const body = await response.text();
+        throw new Error(`Language.load failed with status ${response.status}.\n\n${body}`);
+      }
+
+      const retryResp = response.clone();
+      try {
+        bytesOrModule = await WebAssembly.compileStreaming(response);
+      } catch (reason) {
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        console.error(`wasm streaming compile failed: ${reason}`);
```
ObserverOfTime:
```suggestion
        console.error('wasm streaming compile failed:', reason);
```

PR #4776, opened 2025-08-31
feat(rust)!: add Allocator trait

ObserverOfTime:
This makes overriding the library allocator more idiomatic and safer since it no longer allows users to implement _some_ of the allocation functions which may lead to incompatibilities with the defaults.

The `alloc` feature (disabled by default) enables an implementation of the trait that uses Rust's standard alloc functions.

---

This is a proof of concept. Ideally someone more familiar with Rust will pick it up and improve it.
The documentation could use some polishing and `RustAllocator` definitely needs tests.
If the edge cases (`0` size, `null` pointers) are not handled by `alloc`, we should handle them.

I considered modelling the trait after [GlobalAlloc](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html) or the experimental [Allocator](https://doc.rust-lang.org/std/alloc/trait.Allocator.html) but this is simpler.


amaanq:
I'm against partially complete/PoC PRs. This requires the author or someone else to take it over and finish it up. We can reopen the PR if/when it's *actually* ready.

PR #4779, opened 2025-09-01
Make --wasm --debug add DWARF info to language Wasm

verhovsky:
So that error stack traces contain the source code instead of raw Wasm error traces, if you have the [DWARF extension](https://chromewebstore.google.com/detail/pdcpmagijalfljmkmjngeonclgbbannb?utm_source=item-share-cb) installed.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1042,7 +1042,7 @@ impl Loader {
             output_name,
             "-fPIC",
             "-shared",
-            "-Os",
+            if self.debug_build { "-g" } else { "-Os" },
             format!("-Wl,--export=tree_sitter_{language_name}").as_str(),
             "-Wl,--allow-undefined",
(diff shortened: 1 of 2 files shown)


crates/loader/src/loader.rs
```diff
@@ -782,6 +782,7 @@ impl Loader {
                         .as_ref()
                         .and_then(|p| p.strip_prefix(config.src_path).ok()),
                     &output_path,
+                    false,
```
WillLillis: Shouldn't this be `self.debug_build`?

WillLillis (approved):

amaanq:
I think using `compileStreaming` is fine, so we could go w/ the other PR

verhovsky:
@amaanq both are needed, you still need to build with debug symbols to get debug symbols. I moved these changes into the other.

PR #4781, opened 2025-09-02
Web bindings cleanup

amaanq:
This PR entails some cleanup for the web bindings.

- Move back `@types/emscripten` to dev deps, peer deps are buggy and don't seem to get installed
- Allow users to manually specify the WASI SDK path, which makes it easier for packaging and/or isolated builds.
- Restore `sourcesContent` when building the web bindings, currently building with any optimization level higher than `-O1` causes binaryen to strip the `sourcesContent` field of the source map. This is fixed upstream, but a release hasn't been cut with it (let alone emscripten updating their version of binaryen).…module…module

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1105,21 +1105,51 @@ impl Loader {
         Ok(())
     }
 
+    /// This ensures that the wasi-sdk is available, downloading and extracting it if necessary,
+    /// and returns the path to the `clang` executable.
+    ///
+    /// If `TREE_SITTER_WASI_SDK_PATH` is set, it will use that path to look for the clang executable.
     fn ensure_wasi_sdk_exists(&self) -> Result<PathBuf, Error> {
+        let possible_executables = if cfg!(windows) {
+            vec![
+                "clang.exe",
+                "wasm32-unknown-wasi-clang.exe",
+                "wasm32-wasi-clang.exe",
+            ]
+        } else {
+            vec!["clang", "wasm32-unknown-wasi-clang", "wasm32-wasi-clang"]
+        };
+
+        if let Ok(wasi_sdk_path) = std::env::var("TREE_SITTER_WASI_SDK_PATH") {
+            let wasi_sdk_dir = PathBuf::from(wasi_sdk_path);
+
+            for exe in &possible_executables {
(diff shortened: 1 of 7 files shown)


crates/loader/src/loader.rs
```diff
@@ -1105,21 +1105,50 @@ impl Loader {
...
+            for exe in possible_executables {
+                let clang_exe = wasi_sdk_dir.join("bin").join(exe);
+                if clang_exe.exists() {
+                    return Ok(clang_exe);
+                }
+            }
+
+            return Err(anyhow!(
+                "TREE_SITTER_WASI_SDK_PATH is set to '{}', but no clang executable found in 'bin/' directory. \
+                 Looked for: clang, wasm32-unknown-wasi-clang, wasm32-wasi-clang",
```
WillLillis: Might be better to use `possible_executables.join(", ")` here.

crates/xtask/src/build_wasm.rs
```diff
@@ -287,6 +291,14 @@ fn build_wasm(cmd: &mut Command, edit_tsd: bool) -> Result<()> {
         fs::write(file, content)?;
     }
 
+    // Post-process the source map to embed source content for optimized builds
+    let map_path = std::path::Path::new("lib/binding_web/lib/web-tree-sitter.wasm.map");
```
WillLillis: Just in case we have to test this on Windows, we should use `std::path::MAIN_SEPARATOR` (or a series of `.join` calls) here instead of hard coding '/'.

crates/loader/src/loader.rs
```diff
@@ -1105,21 +1105,51 @@ impl Loader {
         Ok(())
     }
 
+    /// This ensures that the wasi-sdk is available, downloading and extracting it if necessary,
+    /// and returns the path to the `clang` executable.
+    ///
+    /// If `TREE_SITTER_WASI_SDK_PATH` is set, it will use that path to look for the clang executable.
```
WillLillis, line 1111: Should this new environment variable go anywhere in the docs?

WillLillis (approved):

crates/loader/src/loader.rs
amaanq: There really isn't a spot it fits in, since we're missing a section on building/packaging the project as a whole. We should do this sometime in the future, and mention this then.

PR #4782, opened 2025-09-02
feat(loader): switch to `minreq` for fetching wasi-sdk

### LINKED ISSUE #4295: Consider changing dependency on `ureq`

amaanq:
- Closes #4295
- Closes #4396

### Problem

`ureq` uses rustls, which is not ideal for some consumers that would rather depend on the system native tls implementation.

### Solution

This PR switches the `ureq` dependency to `minreq` with the `https-native` feature.


amaanq:
Forgot about openssl in CI :facepalm:

maxbrunsfeld:
Maybe we should just shell out to curl or wget, and fail if thats not installed.

ObserverOfTime:
I think we'd rather use `curl` & `tar` than add a bunch of dependencies, considering they are available:

* On most Linux/BSD systems by default and can be declared as (optional?) package dependencies.
* On Windows 10/11 since build [17063](https://devblogs.microsoft.com/commandline/tar-and-curl-come-to-windows/) (2018) and Windows 7 is unsupported.
* On macOS by default.

 _Note that the BSD/macOS variant of tar may be slightly different to the GNU variant on Linux & Windows._

https://github.com/tree-sitter/tree-sitter/pull/4396#issuecomment-3232360081

Cargo.toml
ObserverOfTime, line 1: We can also remove `flate2`.

ObserverOfTime:
`git2` depends on `openssl-sys` as well. Maybe we should also switch to calling `git` commands?

clason:
For simple download and uncompress of GH artifacts, platform differences of these tools don't play a role (at least that's my experience from nvim-treesitter).

Possible pain points are:
1. Broken system curl compiled without SSL support
2. Environment variables interfering with network access (system proxies)
3. Symlinks inside the archive

Since this is just a convenience feature and not core functionality, those don't need to be guarded against. 

If you want to go the extra mile (not necessarily in this PR), I would recommend the following strategy:

1. Check for wasi-sdk on path (or specified via a documented environment variable, default see step 3).
2. If that fails, check for emscripten on path;
3. If that fails, try to curl wasi-sdk to a (documented!) cache location;
4. If that fails, give an error that wasi-sdk or emscripten cannot be found and need to be installed.

maxbrunsfeld:
I think I would prefer to *not* fall back on Emscripten, and just obtain the wasi-sdk.

clason:
The idea was to not break people's existing workflow immediately (you know who you are...) but make sure that whatever works now, will work in 0.26. 

Generally, I agree that wasi-sdk is far preferable for parser compilation, and it makes sense to remove the emscripten codepath -- in 0.27, with a (non-intrusive) deprecation warning in 0.26. 

(Tree-sitter has become a load-bearing dependency for an increasing number of projects -- which is great! -- so we need to be increasingly careful about managing user experience if we want to keep it that way.)

maxbrunsfeld (approved):
Nice, love that you removed the libgit2 dependency as well.

PR #4788, opened 2025-09-03
Fix CMake Integration and Not Found Headers

kilo52:
Fixes the use case where tree-sitter is added as a dependency via the CMake [FetchContent](https://cmake.org/cmake/help/v3.28/module/FetchContent.html) machinery. Marks the scope of the public headers as public. Moves the `lib/CMakeLists.txt` file to the source root so that it is recognised as a CMake-based project and can subsequently be consumed as such. Adjusts file paths.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -175,7 +175,7 @@ jobs:
       if: ${{ matrix.platform == 'windows-x64' }}
       shell: msys2 {0}
       run: |
-        cmake -G Ninja -S lib -B build/static \
+        cmake -G Ninja -S . -B build/static \
           -DBUILD_SHARED_LIBS=OFF \
           -DCMAKE_BUILD_TYPE=Debug \
           -DCMAKE_COMPILE_WARNING_AS_ERROR=ON \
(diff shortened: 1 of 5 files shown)


kilo52:
If the `lib/CMakeLists.txt` file is moved to the source root, then the paths inside of it would have to be adjusted. Also, the setup as it is now with the root `CMakeLists.txt` adding subdirectories/modules, in this case only `lib`, is pretty common.

ObserverOfTime:
You're adding an extra file that has a single line in it for little benefit.

> If the `lib/CMakeLists.txt` file is moved to the source root, then the paths inside of it would have to be adjusted.

And the paths _outside_ of it. :)

> Also, the setup as it is now with the root `CMakeLists.txt` adding subdirectories/modules, in this case only `lib`, is pretty common.

And unnecessary since there won't be another module.

kilo52:
I've adjusted all the paths that I found referring to the CMake file and moved it to the source root. Hope I haven't missed anything.

SamuelMarks:
Related: let me know if you'd like my PR to be updated to latest tree-sitter and re-opened. Happy to contribute

#1749

PR #4792, opened 2025-09-04
fix(generate): use correct state id when adding terminal states to non terminal extras

### LINKED ISSUE #1168: Spurious ERROR node when using `choice` in `extras` definition
Grammar:
```js
module.exports = grammar({
  name: 'bugcase',

  extras: $ => [
    /[\s]/,
    $.bug_extras,
  ],

  rules: {
    source_file: $ => repeat1($.hello),
    hello: $ => 'hello',

    bug_extras: $ =>
      choice(
        $.bug_extra_1,
        $.bug_extra_2,
      ),

    bug_extra_1: $ => 'FOO',

    bug_extra_2: $ => /#.*\n/,
  }
});
```
Source file:
```
hello # comment
```

Full test case: https://gist.github.com/CyberShadow/03c7feea8892244aa7f68a6e2b06a2fc (`npm install && node_modules/.bin/tree-sitter generate && node_modules/.bin/tree-sitter parse source.bugcase`).

Expected output:

```lisp
(source_file [0, 0] - [1, 0]
  (hello [0, 0] - [0, 5])
  (bug_extras [0, 6] - [1, 0]
    (bug_extra_2 [0, 6] - [1, 0])))
```

Actual output:

```lisp
(ERROR [0, 0] - [1, 0]
  (hello [0, 0] - [0, 5])
  (bug_extra_2 [0, 6] - [1, 0]))
source.bugcase	0 ms	(ERROR [0, 0] - [1, 0])
```

Note that if you delete `bug_extra_1` from the definition of `bug_extras`, then the output is as expected.

WillLillis:
### The Problem

When adding terminal states for non-terminal extras, we unconditionally use the length of the parse table's state vector to assign an id to the newly added terminal state. This assumes we're going to push to `self.parse_table.states`, which does not happen if `add_parse_state` is called with a state has already been added.

### The Solution

Use the correct parse state id returned by `add_parse_state`.

- Closes #1168

### DIFF
--- a/crates/generate/src/build_tables/build_parse_table.rs
+++ b/crates/generate/src/build_tables/build_parse_table.rs
@@ -326,9 +326,10 @@ impl<'a> ParseTableBuilder<'a> {
                 ))?;
             }
 
-            self.non_terminal_extra_states
-                .push((terminal, self.parse_table.states.len()));
-            self.add_parse_state(&Vec::new(), &Vec::new(), item_set);
+            // Add the parse state, and *then* push the terminal and the state id into the
+            // list of nonterminal extra states
+            let state_id = self.add_parse_state(&Vec::new(), &Vec::new(), item_set);
+            self.non_terminal_extra_states.push((terminal, state_id));
         }
 
         while let Some(entry) = self.parse_state_queue.pop_front() {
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Really nice debugging @WillLillis ⚡️⚡️⚡️

PR #4802, opened 2025-09-07
Refactor the Nix flake

faukah:
Problem: current flake setup is not very good

Solution:
Refactor and improve the whole Nix codebase, dropping flake-parts and doing general cleanup.

Changes done:
- Drop flake-parts
- move all checks to flake.nix
- use lib.fileset to filter the repo src for building stuff
- Improve general nix structure and deduplicate code
- rename the flake-parts modules which only define a package to their own package.nix.

CC @amaanq


ObserverOfTime:
The only way to fix nix is to obliterate it.

RGBCube:
> The only way to fix nix is to obliterate it.

image

clason:
Please explain your changes (in the PR _and_ commit description), ideally in the form of "Problem: ... Solution: ...".

lib/binding_web/wasm-test-grammars.nix
clason, line 1: This doesn't look like a "fix"?

faukah: This was moved from lib/binding_web/flake.nix.

clason: Why?

faukah: Typically, you have a single file for one derivation in Nix. The flake.nix before that built both `wasm-test-grammars` and `web-tree-sitters` at once. Now we can build both separately in the flake outputs.

clason: On the other hand, it's useful to keep these (extraneous) files as localized as possible (most people here tolerate but don't use Nix). If I see a `flake.nix`, I know what it's for.

amaanq (approved):

PR #4803, opened 2025-09-07
Switch to npins

faukah:
Depends on #4802.


test/fixtures/fixtures.json
clason, line 1: Why are you changing this? This is not Nix specific, so please don't touch it.

(In general, I think this is way too much Nix tooling in this repo, which adds significant maintenance burden. We don't vendor this sort of code for any other package manager.)

ObserverOfTime:
Supporting nix is one thing but using it internally is unacceptable.

PR #4804, opened 2025-09-07
Fix nix (for real)

ObserverOfTime:


clason:
Amaan is using these.

ObserverOfTime:
Remember tree-sitter/tree-sitter-haskell#122?

> I'm not sure what the tools, extra stuff in test besides corpus files, and **flake stuff** are, but they're **not needed** and trying to run some of the custom "tests" didn't even work - so **simply get rid of it**.

clason:
Presumably he needs them _here_ since he added them.

PR #4808, opened 2025-09-09
Add version info to the playground

### LINKED ISSUE #2217: Add language's commit hash to playground
[We were wondering](https://github.com/tree-sitter/tree-sitter-bash/issues/148#issuecomment-1433918152) which version of tree-sitter-bash is deployed on the playground. I looked at the code and thought it was the master branch but I noticed recently that if you try to parse `$''` with the Bash grammar, it will still call it "`ansii_c_string`", when it was renamed to "`ansi_c_string`" (one "i") 5-6 months ago in https://github.com/tree-sitter/tree-sitter-bash/pull/121.

amaanq:
- Closes #2217

This PR adds version information for a given parser (if available) in the playground. Example:

image

Additionally, it fixes a few problems in the web bindings, namely, the fact that the metadata function never worked and that the xtask for updating the wasm stdlib *always* ran docker. Now, the wasm stdlib xtask will rely on the `TREE_SITTER_WASI_SDK_PATH` env var to point to the user's installed wasi sdk.

Lastly, in the playground, I've updated the button to not look like an ad-hoc untidy change, but rather look seamless.

### DIFF
--- a/crates/cli/src/playground.html
+++ b/crates/cli/src/playground.html
@@ -302,6 +302,29 @@
       box-shadow: 0 0 0 2px var(--primary-color-alpha);
     }
 
+    .toast {
+      position: fixed;
+      bottom: 20px;
+      right: 20px;
+      background-color: var(--light-text);
+      color: white;
+      padding: 12px 16px;
+      border-radius: 6px;
+      box-shadow: 0 4px 12px rgba(0, 0, 0, 0.3);
+      font-size: 14px;
+      font-weight: 500;
+      opacity: 0;
+      transform: translateY(20px);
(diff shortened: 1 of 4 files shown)


ObserverOfTime:
> the xtask for updating the wasm stdlib _always_ ran docker. Now, the wasm stdlib xtask will rely on the `TREE_SITTER_WASI_SDK_PATH` env var to point to the user's installed wasi sdk.

I wanted to work on that but never found the time.
My idea was to reuse the emscripten native/docker/podman lookup.
But native would look in an env var or the tree-sitter cache location instead of the `PATH`.

PR #4814, opened 2025-09-11
feat(lib)!: remove deprecated functions

amaanq:
Some cleanup before 0.26


WillLillis:
Are we removing the `--build` flag for the `generate command now or in the next release cycle?

EDIT: Nevermind, that's getting removed in 0.27.

PR #4816, opened 2025-09-11
chore: Add missing `<html>` tag to playground.html

wetneb:
Although the browsers I use seem to deal with this absence okay, it probably doesn't hurt to follow the standard, at least to silence validation errors when working on this file.

(The diff looks a bit ugly because of the re-indentation, but you can check that I didn't change any of the internal source, for instance by diffing using difftastic)


ObserverOfTime:
image

amaanq:
thanks! +1 for difftastic

PR #4818, opened 2025-09-12
Fix wasm bug and build issues

### LINKED ISSUE #4813: XML grammar causes wasm out of bounds memory access trap
### Problem

This is related to https://github.com/zed-industries/zed/issues/36586 and https://github.com/sweetppro/zed-xml/issues/21

Opening XML files above a certain size (~400–500KiB) with the XML extension installed in the Zed editor causes the currently released version to crash with a segfault on Linux. I think I have traced this down to tree-sitter when using the `wasm` feature. The document fails to parse due to encountering an "out of bounds memory access" trap in the wasm module generated from [tree-sitter-xml](https://github.com/tree-sitter-grammars/tree-sitter-xml):

```
     Running `target/debug/zed-ts-wasm /home/wmoore/Work/prince/ChaoYanZeCuHeiTif-1.ttx`
trap in wasm module: error while executing at wasm backtrace:
    0:  0x649 - stdlib.wasm!memcpy
    1: 0x4900 - <unknown>!tree_sitter_xml_external_scanner_serialize

Caused by:
    wasm trap: out of bounds memory access
trap in wasm module: error while executing at wasm backtrace:
    0:  0x3a2 - stdlib.wasm!free
    1: 0x49b8 - <unknown>!tree_sitter_xml_external_scanner_deserialize

Caused by:
    0: memory fault at wasm address 0x70796c43 in linear memory of size 0x430000
    1: wasm trap: out of bounds memory access

thread 'main' panicked at src/main.rs:24:50:
unable to parse
```

Smaller documents do not hit the error and parse successfully. I have also tried using the C API directly with a test binary compiled with `address` and `undefined` sanitizers enabled, but this does not encounter any issues, which is what leads me to think it's a wasm specific problem:

```c
#include <assert.h>
#include <string.h>
#include <stdio.h>
#include <tree_sitter/api.h>

const TSLanguage *tree_sitter_xml(void);

int main() {

  ts_parser_set_language(parser, tree_sitter_xml());

  FILE *f = fopen("ChaoYanZeCuHeiTif.xml", "rb");

  char *source_code = NULL;
  size_t len;
  size_t nread = getdelim(&source_code, &len, '\0', f);
  assert(nread != -1);
  fclose(f);

  TSTree *tree = ts_parser_parse_string(
    parser,
    NULL,
    source_code,
    strlen(source_code)
  );

  assert(tree != NULL);

  ts_tree_delete(tree);
  ts_parser_delete(parser);
  free(source_code);

  return 0;
}
```

### Steps to reproduce

See https://forge.wezm.net/wezm/zed-ts-wasm-xml where I have prepared a small reproducer. Getting the error message requires uncommenting the the error logs in `wasm_store.c` (see README in linked repo).

### Expected behavior

The well-formed XML document is parsed successfully.

0.25.9

### Operating system/version

Arch Linux/x86_64

amaanq:
- Closes #4813

### Problem

When a call to `ts_parser_parse` allocates and frees up a lot of memory, the WASM heap eventually gets exhausted, and further allocations or frees fail. This can be observed in the linked issue, where the sample code contains a very large XML file, causing the scanner to allocate and free a lot of memory.

### Solution

In the wasm stdlib, I've added a `Region *` that keeps track of the freed regions. When we allocate, we check if there's a free region in this list first before growing the heap, and when we free, if it's not the last allocated pointer, we add the region to the free list.

While attempting to solve this, we ran into a few problems, namely, `reset_heap` was no longer being exported by the `build-wasm-stdlib` xtask. The solution was to define `TREE_SITTER_FEATURE_WASM` and export `reset_heap`.

Additionally, we noticed that during serialization, the wasm parser does not set `has_scanner_error` if there was an error, and does not validate the length was less than or equal to the max serialization length.

### DIFF
--- a/lib/src/wasm/stdlib.c
+++ b/lib/src/wasm/stdlib.c
@@ -60,6 +60,8 @@ void reset_heap(void *new_heap_start) {
 }
 
 void *malloc(size_t size) {
+  if (size == 0) return NULL;
+
   Region *prev = NULL;
   Region *curr = free_list;
   while (curr != NULL) {


wezm:
> where the sample code contains a very large XML file

I'd suggest testing/checking with a much larger file as well. The file I included in my sample was a small fragment from a real file, which is 54 MiB.

WillLillis:
> > where the sample code contains a very large XML file
> 
> I'd suggest testing/checking with a much larger file as well. The file I included in my sample was a small fragment from a real file, which is 54 MiB.

Good point! Just checked with an 86MiB file and everything seems to be in order. 

image

maxbrunsfeld:
Nice debugging!

Maybe we should add a test case to [`wasm_language_test.rs`](https://github.com/tree-sitter/tree-sitter/blob/master/crates/cli/src/tests/wasm_language_test.rs) that exercises the same case that @wezm found.

What I don't understand is, why didn't [`test_wasm_oom`](https://github.com/tree-sitter/tree-sitter/blob/580cd9541a95854d45639ab306a73267a7c0ffe0/crates/cli/src/tests/wasm_language_test.rs#L251-L273) already catch this? It does something similar by parsing an HTML document full of custom elements (whose tag name must be stored on the heap).

@amaanq @WillLillis Do you have any idea what the difference is, such that `tree-sitter-xml` caused this crash, but that test does not?

*EDIT* - Oh nevermind, that test is catching the case of a *single* large allocation (a single custom element tag with a long name), whereas what we *need* to test is the case where there are many distinct allocations. I think it should be possible to write another test *like* `test_wasm_oom` that performs many allocations, freeing them in a non-stack-based order, such that we would exhausts the wasm heap without this free list.

amaanq:
Will and I spent some time trying to come up with a repro for this, but it's pretty damn impossible without just using the XML grammar itself, which I'd rather avoid. I think we can get this in as is, and if anyone's able to come up with a test case that reproduces the crash, that'd be great to add to the test suite.

maxbrunsfeld:
Yeah, makes sense - thanks for looking into it. Great to have this fixed.

PR #4819, opened 2025-09-13
Fix CJS build

RedCMD:
should be `".cjs"`
not `"cjs"`
https://github.com/tree-sitter/tree-sitter/commit/07986471b3f722a4a83ae83749810b484723272f#r165729739

image


RedCMD:
how did this get past ci?

amaanq:
> how did this get past ci?

doesn't return a non-zero status code

PR #4824, opened 2025-09-14
docs: explain extras in a bit more detail

### LINKED ISSUE #1653: configuration for `extras` is not well-documented
## Problem
Assuming that
```
const Comment = /\/\/[^\n]*|\/\*([^\*\/]|\*[^\/]|[^\*]\/)*\*\//
const Blank = /[ \t\r　]+/
const LF = /[\n]+/
```
, if the configuration for `extras` is
```
extras: $ => [Comment, Blank, LF]
```
, the generated C file will be too large (with an insanely complicated lexer) so that the compiled parser does not work.
To solve this problem, it is necessary to add a rule
```
comment: $ => Comment
```
and use the rule `$.comment` in `extras` instead of the token `Comment`: 
```
extras: $ => [$.comment, Blank, LF]
```
. It took 2 hours for me to figure out this.
## Proposal
In the documentation for the `extras` configuration,
we should tell users not to put a complex token into `extras`.
Also, we should tell users to avoid the configuration mentioned in #1168.

amaanq:
- Closes #1653


clason:
"optimal" is a strong word (and does not really take a comparative) -- maybe "preferable"?

docs/src/creating-parsers/3-writing-the-grammar.md
```diff
@@ -395,6 +395,87 @@ function_definition: $ =>
...
+    comment: ($) =>
+      token(
+        choice(seq("//", /.*/), seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/")),
+      ),
+  },
+});
+```
+
+```admonish warning
+When adding more complicated tokens to `extras`, it's optimal to associate the pattern
```
clason: here too

amaanq: whoops, thanks :sweat_smile:

docs/src/creating-parsers/3-writing-the-grammar.md
```diff
@@ -395,6 +395,87 @@ function_definition: $ =>
...
+  },
+});
+```
+
+```admonish note
+Tree-sitter intentionally simplifies some common regex patterns, both as a performance optimization and for simplicity,
+typically in ways that don't affect the meaning of the pattern. For example, `\w` is simplified to `[a-zA-Z0-9_]`, `\s`
+to `[ \t\n\r]`, and `\d` to `[0-9]`. If you need more complex behavior, you can always use a more explicit regex.
+```
+
```
maxbrunsfeld, line 478: The \w \d character classes match the ECMAScript standard - they’re not simplified. Only the whitespace one is simplified, I believe.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes

amaanq: Oh yeah, you're right. In Rust they match a bit more than just that, so we simplify it in the codebase, but this isn't user-facing.

PR #4833, opened 2025-09-15
refactor: use the logger

ObserverOfTime:
I may have missed some `println` calls that should be replaced or replaced some I shouldn't have. Closes #4604.

### DIFF
--- a/crates/cli/src/logger.rs
+++ b/crates/cli/src/logger.rs
@@ -1,9 +1,14 @@
-use log::{LevelFilter, Log, Metadata, Record};
+use std::io::Write;
 
-#[allow(dead_code)]
-struct Logger {
-    pub filter: Option<String>,
+use anstyle::{AnsiColor, Color, Style};
+use log::{Level, LevelFilter, Log, Metadata, Record};
+
+pub fn paint(color: Option<impl Into<Color>>, text: &str) -> String {
+    let style = Style::new().fg_color(color.map(Into::into));
+    format!("{style}{text}{style:#}")
 }
+
+struct Logger;
 
 impl Log for Logger {
     fn enabled(&self, _: &Metadata) -> bool {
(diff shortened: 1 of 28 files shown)


crates/cli/src/logger.rs
```diff
@@ -1,30 +1,46 @@
...
+                paint(Some(AnsiColor::Yellow), "Warning:"),
+                record.args()
+            ),
+            Level::Info => eprintln!("{}", record.args()),
+            _ => eprintln!(
+                "[{}] {}",
+                record
+                    .module_path()
+                    .unwrap_or_default()
+                    .trim_start_matches("rust_tree_sitter_cli::"),
```
ObserverOfTime, line 36: This might be worth having in the error level too (perhaps only in debug builds?).
Note also that we don't currently use `debug!` or `trace!` anywhere.

crates/cli/src/query.rs
```diff
@@ -117,7 +117,11 @@ pub fn query_file_at_path(
     if query_cursor.did_exceed_match_limit() {
         writeln!(
             &mut stdout,
-            "  WARNING: Query exceeded maximum number of in-progress captures!"
+            "{}",
+            paint(
+                Some(AnsiColor::Yellow),
+                " -- Warning: Query exceeded maximum number of in-progress captures!"
+            )
```
ObserverOfTime: For consistency with the slow parse rate warning in `test.rs`.

crates/cli/src/test.rs
```diff
@@ -219,78 +222,76 @@ pub fn run_tests_at_path(parser: &mut Parser, opts: &mut TestOptions) -> Result<
     parser.stop_printing_dot_graphs();
 
     if failures.is_empty() {
+        Ok(())
+    } else if opts.update && !has_parse_errors {
```
ObserverOfTime, line 226: I ended up not using the logger here but I adjusted the `println` calls to remove one level of conditions.

crates/generate/src/build_tables.rs
```diff
@@ -480,14 +480,14 @@ fn report_state_info<'a>(
         .max()
```
ObserverOfTime, line 480: I have no idea where this function is called.

crates/generate/src/build_tables/build_parse_table.rs
```diff
@@ -346,17 +347,21 @@ impl<'a> ParseTableBuilder<'a> {
...
-                        .collect::<Vec<_>>()
-                        .join(", ")
-                );
-            }
+            warn!(
+                "unnecessary conflicts:\n  {}",
+                &self
+                    .actual_conflicts
+                    .iter()
+                    .map(|conflict| {
```
ObserverOfTime, line 355: I wonder if `fold` with `format!` would be more performant than these `map`-`collect`-`join` chains.

ObserverOfTime:
```
thread 'main' panicked at crates\generate\src\build_tables\build_lex_table.rs:183:61:
index out of bounds: the len is 0 but the index is 0
```
🤔

clason:
Try rebasing on latest master.

amaanq:
you're always logging stuff that shouldn't be logged (state splitting), hence the crash.

ObserverOfTime:
I see it now. Should I change the existing `info` calls to `debug`?

amaanq:
> I see it now. Should I change the existing `info` calls to `debug`?

This, and whatever else was using the logger before, should only be logged when `--log` is passed in to `generate`. I'll do a pass through this to verify what should/shouldn't be logged and at what levels, because the behavior & output of every command should *not* change with this PR.

crates/generate/src/build_tables.rs
WillLillis: https://github.com/tree-sitter/tree-sitter/blob/0ca8fe8c12dde40008305e7925cac410952f0752/crates/generate/src/build_tables.rs#L93-L101

WillLillis (approved):

crates/generate/src/build_tables/build_parse_table.rs
amaanq: It would be a very negligible difference

PR #4837, opened 2025-09-16
feat(generate): don't embed tree-sitter CLI version in parser

clason:
Problem: embedding the CLI version used to generate a parser triggers CI
failures on all grammars for every (patch) release of tree-sitter, even
if there are no actual parser changes.

Solution: do not embed the version; instead rely on whether the update
introduces actual (presumably desirable) changes in the parser to
indicate regeneration is necessary.


clason:
enough is enough.

amaanq:
yeah this is quite annoying

PR #4838, opened 2025-09-16
fix(generate): remove warning message for CJS grammars

amaanq:


clason:
See https://github.com/tree-sitter/tree-sitter/pull/3780#issuecomment-3297365559. I would suggest not printing _any_ warning right now, instead adding this explanation to the documentation.

ObserverOfTime:
And https://github.com/tree-sitter/tree-sitter/pull/3780#issuecomment-2408863524 :)

clason:
That PR has been long merged and is in the 0.9.0 release, isn't it? And the warning is only triggered when the flag is specified, so when someone has explicitly opted in to this feature.

PR #4839, opened 2025-09-16
fix(bindings): fix ESM errors in Node bindings

ObserverOfTime:
1. The module tries to call the native binary as a function.

Only `node-gyp-build` returns a function, so the call is moved there.

2. `node-types.json` is imported with outdated syntax.

Use import attributes which require Node 18.

3. The test does not properly catch import errors.

This is solved by moving the import inside the assertion.

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -314,11 +314,10 @@ pub fn generate_grammar_files(
             if !contents.contains("module") {
                 eprintln!("Updating package.json");
                 contents = contents.replace(
-                    indoc! {r#"
-                    "repository": {"#},
+                    r#""repository":"#,
                     indoc! {r#"
                     "type": "module",
-                      "repository": {"#},
+                      "repository":"#},
                 );
             }
             write_file(path, contents)?;
--- a/crates/cli/src/templates/binding_test.js
+++ b/crates/cli/src/templates/binding_test.js
@@ -1,9 +1,11 @@
 import assert from "node:assert";
 import { test } from "node:test";
 import Parser from "tree-sitter";
-import language from "./index.js";
 
 test("can load grammar", () => {
   const parser = new Parser();
-  assert.doesNotThrow(() => parser.setLanguage(language));
+  assert.doesNotReject(async () => {
+    const { default: language } = await import("./index.js");
+    parser.setLanguage(language);
+  });
 });
--- a/crates/cli/src/templates/index.js
+++ b/crates/cli/src/templates/index.js
@@ -2,14 +2,12 @@ const root = new URL("../..", import.meta.url).pathname;
 
 const binding = typeof process.versions.bun === "string"
   // Support `bun build --compile` by being statically analyzable enough to find the .node file at build-time
-  ? await import(`../../prebuilds/${process.platform}-${process.arch}/tree-sitter-KEBAB_PARSER_NAME.node`)
-  : await import("node-gyp-build");
-
-const result = binding.default ? binding.default(root) : binding(root);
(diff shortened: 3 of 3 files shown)


ObserverOfTime:
Alternatively for (2), we can specify the minimum Node version in [`engines`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines), but it's normally just a warning.

ObserverOfTime:
We could also remove the default export from the module.

```diff
diff --git i/crates/cli/src/templates/binding_test.js w/crates/cli/src/templates/binding_test.js
index 7a91a84d..23919778 100644
--- i/crates/cli/src/templates/binding_test.js
+++ w/crates/cli/src/templates/binding_test.js
@@ -5,7 +5,6 @@ import Parser from "tree-sitter";
 test("can load grammar", () => {
   const parser = new Parser();
   assert.doesNotReject(async () => {
-    const { default: language } = await import("./index.js");
-    parser.setLanguage(language);
+    parser.setLanguage(await import("./index.js"));
   });
 });
diff --git i/crates/cli/src/templates/index.d.ts w/crates/cli/src/templates/index.d.ts
index 528e060f..84c2596c 100644
--- i/crates/cli/src/templates/index.d.ts
+++ w/crates/cli/src/templates/index.d.ts
@@ -18,10 +18,6 @@ type NodeInfo =
       children: ChildNode[];
     });
 
-type Language = {
-  language: unknown;
-  nodeTypeInfo: NodeInfo[];
-};
+export const language: unknown;
 
-declare const language: Language;
-export = language;
+export const nodeTypeInfo: NodeInfo[];
diff --git i/crates/cli/src/templates/index.js w/crates/cli/src/templates/index.js
index 7c36f515..e6b073a9 100644
--- i/crates/cli/src/templates/index.js
+++ w/crates/cli/src/templates/index.js
@@ -2,13 +2,12 @@ import { readFileSync } from "node:fs";
 
 const root = new URL("../..", import.meta.url).pathname;
 
-const binding = typeof process.versions.bun === "string"
+export const { language } = typeof process.versions.bun === "string"
   // Support `bun build --compile` by being statically analyzable enough to find the .node file at build-time
   ? await import(`${root}/prebuilds/${process.platform}-${process.arch}/tree-sitter-KEBAB_PARSER_NAME.node`)
   : (await import("node-gyp-build")).default(root);
 
+export let nodeTypeInfo;
 try {
-  binding.nodeTypeInfo = JSON.parse(readFileSync(`${root}/src/node-types.json`, "utf8"));
+  nodeTypeInfo = JSON.parse(readFileSync(`${root}/src/node-types.json`, "utf8"));
 } catch (_) {}
-
-export default binding;
```

This would simplify `await import` but import statements would become:

```js
import * as foo from "tree-sitter-foo"
```

amaanq:
> Import assertions were changed to import attributes in Node 18.

Node 18 is dead, so we could use this instead and assume people are on 20+.


Removing the default export would be breaking and I don't see the point of that, unless I'm missing something

crates/cli/src/templates/package.json
```diff
@@ -49,6 +49,9 @@
       "optional": true
     }
   },
+  "engines": {
+    "node": ">=18.20.0"
+  },
```
amaanq: Why? 18 has been EOL.

ObserverOfTime: It works. ¯\\_(ツ)\_/¯

amaanq: This is not needed.

amaanq (approved):

PR #4842, opened 2025-09-17
Revert "feat(generate): allow more characters for keywords"

amaanq:
This reverts commit 0269357.

CI is failing, because the Bash parser was regenerated with the changes reverted here, and was aggressively extracting keywords. Ideally, the candidate tokens that are allowed for keyword extraction should be configurable by the user, but this would be a pretty niche feature that most people wouldn't need, maybe something like `keyword` where users can pass in a custom pattern to override the default.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -17,10 +17,9 @@ runs:
           test/fixtures/grammars
           target/release/tree-sitter-*.wasm
         key: fixtures-${{ join(matrix.*, '_') }}-${{ hashFiles(
-          'cli/generate/src/**',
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Maybe we need to make it configurable or something. I remember it being tricky to get keyword extraction doing the right thing, and the “word characters” rule seemed to make it more predictable. Configurability initially seemed kinda complex, but maybe it can be done with a single regex.

amaanq:
I think allowing users to specify the keyword pattern with a single regex would be perfect, and the default would be `[a-z][A-Z]_+`. This could be under a `keyword` field

PR #4844, opened 2025-09-17
wasm: Fix alias map parsing + add test

nia-e:
The current code for parsing wasm grammars doesn't read the alias map correctly, as it fails to offset its internal index to account for the size of the actual `value_count` field itself. This can lead to extraneous data being copied into the language struct, and under the wrong conditions this can result in a segfault (see https://github.com/zed-industries/zed/issues/29827 - attempting to load the html-django grammar from https://github.com/interdependence/tree-sitter-htmldjango reliably segfaults).

The python grammar appears to also have an alias map set, so it was added as a test here. Note that it can still spuriously pass even without this, and on my machine it indeed does - but this should be a useful smoke test at least.

### DIFF
--- a/crates/cli/src/tests/helpers/fixtures.rs
+++ b/crates/cli/src/tests/helpers/fixtures.rs
@@ -23,6 +23,19 @@ static TEST_LOADER: LazyLock<Loader> = LazyLock::new(|| {
     loader
 });
 
+#[cfg(feature = "wasm")]
+pub static ENGINE: LazyLock<tree_sitter::wasmtime::Engine> = LazyLock::new(Default::default);
+
+#[cfg(feature = "wasm")]
+static TEST_WASM_LOADER: LazyLock<Loader> = LazyLock::new(|| {
+    let mut loader = Loader::with_parser_lib_path(SCRATCH_DIR.clone());
+    loader.use_wasm(&ENGINE);
+    if env::var("TREE_SITTER_GRAMMAR_DEBUG").is_ok() {
+        loader.debug_build(true);
+    }
+    loader
+});
+
 pub fn test_loader() -> &'static Loader {
     &TEST_LOADER
 }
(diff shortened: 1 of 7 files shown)


lib/src/wasm_store.c
```diff
@@ -1371,13 +1371,13 @@ const TSLanguage *ts_wasm_store_load_language(
...
-      TSSymbol symbol;
+      uint16_t symbol;
       memcpy(&symbol, &memory[wasm_language.alias_map + alias_map_size], sizeof(symbol));
-      alias_map_size += sizeof(TSSymbol);
+      alias_map_size += sizeof(uint16_t);
       if (symbol == 0) break;
       uint16_t value_count;
       memcpy(&value_count, &memory[wasm_language.alias_map + alias_map_size], sizeof(value_count));
-      alias_map_size += value_count * sizeof(TSSymbol);
+      alias_map_size += (value_count + 1) * sizeof(uint16_t);
```
amaanq: Why the type changes? That's not needed.

nia-e: The actual array in `parser.c` seems to be defined as `uint16_t[]`, so I was worried about the possibility that in some circumstances those might be different and introduce a bug that way. If they're always the same, then I can change it

maxbrunsfeld:
@amaanq Nia and I are pairing on improving the wasm language tests, so that we can reuse the test grammars in those.

amaanq:
> @amaanq Nia and I are pairing on improving the wasm language tests, so that we can reuse the test grammars in those.

Ah cool! That'd be great

maxbrunsfeld (approved):

amaanq (approved):

crates/cli/src/tests/helpers/fixtures.rs
```diff
@@ -136,5 +167,17 @@ pub fn get_test_language(name: &str, parser_code: &str, path: Option<&Path>) ->
...
+    if wasm {
+        #[cfg(feature = "wasm")]
+        {
+            TEST_WASM_LOADER
+                .load_language_at_path_with_name(config)
+                .unwrap()
+        }
+
+        #[cfg(not(feature = "wasm"))]
+        panic!("WASM feature is not enabled")
```
ObserverOfTime:
```suggestion
        panic!("Wasm feature is not enabled")
```

lib/src/wasm_store.c
```diff
@@ -1377,6 +1377,7 @@ const TSLanguage *ts_wasm_store_load_language(
       if (symbol == 0) break;
       uint16_t value_count;
       memcpy(&value_count, &memory[wasm_language.alias_map + alias_map_size], sizeof(value_count));
+      alias_map_size += sizeof(uint16_t);
```
WillLillis, line 1380: Not your change, but aren't we double counting the size of `alias_map` in the `copy` call below on 1384? The additions to `alias_map_size` in the `for` loop already multiply the number of elements by `sizeof(TSSymbol)` or `sizeof(uint16_t)`, so the multiplication by `sizeof(TSSymbol)` below is incorrect and copies 2x the intended length. I'm assuming this hasn't caused any issues because the calculated size doesn't leave this scope and the list is null terminated?

I'm not sure if this is related to the CI failures and I'm not able to reproduce them locally to test. :/

nia-e: Good point, you're right. Yet another possible segfault I suppose

maxbrunsfeld: Nice catch @WillLillis! 😬

amaanq:
The leak came from the fact that we were allocating two stores - `TEST_WASM_LOADER` creates one internally when we load a wasm parser, and the test itself created one, but the parser would only use the store from the test itself so the static loader would not clean up its allocations. I've decided to just forego the global wasm loader and create one every time we load a language. This can be cleaned up later

maxbrunsfeld:
Nice, makes sense. Thanks @amaanq

amaanq:
failing tests is due to a spurious network error, so this is good now

WillLillis (approved):

PR #4850, opened 2025-09-19
build(rust): bump serde

amaanq:
~This might fix the glibc errors with cross targets~

Nvm, bumping serde, downgrading serde_json, reverting my unused deps cleanup and reverting the cargo deps bump did not fix it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,9 +192,9 @@ checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
 name = "cc"
-version = "1.2.37"
+version = "1.2.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ObserverOfTime:
We should find an alternative to cross that isn't outdated.

amaanq:
Nix :trollface:

clason:
zig 🦎

amaanq:
Zig can't run/test cross-compiled binaries for CI :)

clason:
bfredl would disagree ;)

amaanq:
Well it'd work with qemu, sure

ObserverOfTime:
Can't we compile and run 32bit arm binaries on the 64bit arm runner?

amaanq:
You'd need a 32bit runtime

ObserverOfTime:
~~Only if `IA32_EMULATION` is disabled in the runner, which shouldn't be the case.~~
Apparently that only applies to x86 and not arm.

PR #4851, opened 2025-09-19
fix(generate): fix builds outside of crate workspace

amaanq:
This fixes issues with the release and when it's used as a standalone dependency, as noted in https://github.com/tree-sitter/tree-sitter/commit/9593737871708e62329fba79d695f8f0f45482ae#commitcomment-166147949.

I've tested this in a sample project that pulls in the generate crate locally, and via `cargo workspaces publish --dry-run`


maxbrunsfeld:
Is it worth all this extra code to remove the `tree-sitter` dependency from `tree-sitter-generate`? It seems like a very obscure use case that you would pull in `tree-sitter-generate` as a crate at all; I'm not sure we need to optimize its dependency graph that hard.

amaanq:
> Is it worth all this extra code to remove the `tree-sitter` dependency from `tree-sitter-generate`? It seems like a very obscure use case that you would pull in `tree-sitter-generate` as a crate at all; I'm not sure we need to optimize its dependency graph that hard.

I guess some users want to pull in the generate crate without the tree-sitter crate since that's only used to pull in 2 constants, but I'm not against removing this. It's probably easier on wasm consumers too.

At the very worst, we can copy parser.h and add a test to verify they're not out of sync.

ognevny:
thanks

PR #4852, opened 2025-09-20
refactor!: rename `stage` flag to `emit`

amaanq:
I meant to do this before the release, but....

This PR renames the new cli flag `stage` to `emit`. I think the name fits better.


WillLillis:
The docs should be updated here or https://github.com/tree-sitter/tree-sitter/pull/4647 should be updated to reflect these changes.

PR #4864, opened 2025-09-22
feat: allow the external scanner to use the internal character ranges

### LINKED ISSUE #2620: Use the internal regexp in external scanners
Hi!

Thanks for the work on this, it's fantastic!

I'm writing a grammar for [liquidsoap](https://github.com/savonet/liquidsoap) and I need to implement an external scanner that matches variable names using regular expressions that involve unicode character properties.

It seems that, by default, C and C++ do not have built-in support for matching these character properties and I would like to avoid adding an external dependency to my parser.

Is there a way to re-use the unicode-aware regular expressions from the internal parser in an external scanner?

Thanks for y'all work and help!

amaanq:
- Closes #2620

### Problem

Currently, grammars with complex tokens cannot make use of the character sets generated in the parser. This causes friction, where now users have to either copy the character set in the scanner or give up on matching the entirety of the complex token.

### Solution

Now, tree-sitter will generate helper functions that can be used in a scanner to match against these complex tokens. For every token with a "complex" token, tree-sitter will generate a function in the form of `matches_sym_{name}`. For tokens with multiple distinct groups (e.g. `[a-zA-Z][a-zA-Z0-9]*` consists of two groups), tree-sitter will now generate a function that matches each distinct group, in case you only want to match a specific part of the token. A more detailed explanation along with an example can be found in the docs changes.

### DIFF
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -62,6 +62,7 @@ struct JSONOutput {
 
 struct GeneratedParser {
     c_code: String,
+    header_code: String,
     #[cfg(feature = "load")]
     node_types_json: String,
 }
@@ -272,6 +273,7 @@ where
     // Generate the parser and related files.
     let GeneratedParser {
         c_code,
+        header_code,
         node_types_json,
     } = generate_parser_for_grammar_with_opts(
         &input_grammar,
@@ -285,7 +287,19 @@ where
     fs::create_dir_all(&header_path)?;
     write_file(&header_path.join("alloc.h"), ALLOC_HEADER)?;
     write_file(&header_path.join("array.h"), ARRAY_HEADER)?;
-    write_file(&header_path.join("parser.h"), PARSER_HEADER)?;
+    let parser_h_content = if header_code.is_empty() {
+        PARSER_HEADER.to_string()
+    } else {
+        let Some(insert_pos) = PARSER_HEADER.find("}\n\n/*\n *  Lexer Macros\n */") else {
+            panic!("Failed to find insertion point in parser.h");
+        };
+
+        let mut content = PARSER_HEADER.to_string();
+        content.insert_str(insert_pos + 1, &format!("\n\n{header_code}"));
+        content
+    };
+
+    write_file(&header_path.join("parser.h"), parser_h_content)?;
 
     Ok(())
 }
@@ -293,7 +307,7 @@ where
 pub fn generate_parser_for_grammar(
     grammar_json: &str,
     semantic_version: Option<(u8, u8, u8)>,
-) -> GenerateResult<(String, String)> {
+) -> GenerateResult<(String, String, String)> {
     let grammar_json = JSON_COMMENT_REGEX.replace_all(grammar_json, "\n");
     let input_grammar = parse_grammar(&grammar_json)?;
     let parser = generate_parser_for_grammar_with_opts(
@@ -302,7 +316,7 @@ pub fn generate_parser_for_grammar(
         semantic_version,
         None,
     )?;
-    Ok((input_grammar.name, parser.c_code))
+    Ok((input_grammar.name, parser.c_code, parser.header_code))
 }
 
 fn generate_node_types_from_grammar(input_grammar: &InputGrammar) -> GenerateResult<JSONOutput> {
@@ -354,7 +368,7 @@ fn generate_parser_for_grammar_with_opts(
         &inlines,
         report_symbol_name,
     )?;
-    let c_code = render_c_code(
+    let (c_code, header_code) = render_c_code(
         &input_grammar.name,
         tables,
         syntax_grammar,
@@ -366,6 +380,7 @@ fn generate_parser_for_grammar_with_opts(
     );
     Ok(GeneratedParser {
         c_code,
+        header_code,
         #[cfg(feature = "load")]
         node_types_json,
     })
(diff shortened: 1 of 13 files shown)


ObserverOfTime (commented):
Can the functions be added to a new header instead?

amaanq:
> Can the functions be added to a new header instead?

They can, but what for? They're defined in `parser.c`, so declaring them in `parser.h` seems correct.

crates/generate/src/generate.rs
```diff
@@ -285,15 +287,27 @@ where
...
+    write_file(&header_path.join("parser.h"), parser_h_content)?;
 
     Ok(())
 }
 
 pub fn generate_parser_for_grammar(
     grammar_json: &str,
     semantic_version: Option<(u8, u8, u8)>,
-) -> GenerateResult<(String, String)> {
+) -> GenerateResult<(String, String, String)> {
```
WillLillis, line 310: It's borderline, but I think it would be better for readability to have a struct with named fields here rather than 3 `String`s in a tuple.

crates/generate/src/render.rs
```diff
@@ -1134,6 +1149,126 @@ impl Generator {
         add_line!(self, "");
     }
 
+    fn add_character_set_helpers(&mut self) {
+        let mut symbols_with_char_sets = self
+            .symbol_to_character_sets
+            .iter()
+            .filter(|(_, sets)| !sets.is_empty())
```
WillLillis, line 1156: Is it possible for these `Vec`s to ever be empty? It looks like they'll always have at least one entry.

crates/generate/src/render.rs
```diff
@@ -1134,6 +1149,126 @@ impl Generator {
...
+            let symbol_name = &self.symbol_ids[symbol];
+
+            let used_char_sets = char_set_indices
+                .iter()
+                .copied()
+                .filter(|&ix| self.large_character_set_info[ix].is_used)
+                .collect::<Vec<_>>();
+
+            if !used_char_sets.is_empty() {
+                if !header_added {
```
WillLillis, line 1173: Couldn't this just be `if self.header_buffer.is_empty()`? I think you had that in a previous push.

maxbrunsfeld:
What is the linkage on these symbols? If they are not static, then they must be prefixed with ‘tree_sitter_$GRAMMAR_NAME’ to avoid name collisions. It would be nice to not export these unless the grammar author opts in somehow.

crates/generate/src/render.rs
```diff
@@ -1134,6 +1149,126 @@ impl Generator {
...
+                            1 => "st",
+                            2 => "nd",
+                            3 => "rd",
+                            _ => "th",
+                        };
+
+                        // Match function for this character set
+
+                        writeln!(
+                        self.header_buffer,
```
WillLillis, line 1205: rustfmt died, please save it 😆

docs/src/creating-parsers/4-external-scanners.md
```diff
@@ -294,6 +294,94 @@ bool tree_sitter_my_language_external_scanner_scan(
 
 ```
 
+### Character Set Functions
+
+If your grammar contains complex patterns, Tree-sitter automatically generates helper functions that your external scanner
+can use to check if a character matches those character sets. This avoids the need to reimplement logic for matching complex
+tokens in your scanner. These matching functions are added to the `parser.h` header file. The function names follow the pattern
+`matches_sym_<symbol_name>(int32_t lookahead)`.
```
WillLillis, line 302: Include the `bool` return type here, and the distinct character set group list below?

WillLillis (approved):

maxbrunsfeld:
Thinking about this more - I feel like this API needs a little work. The character sets that tree-sitter generates are based on some fairly arbitrary heuristics for what helps compile time. This is fine for an internal optimization, but not suitable for a public API. It needs to be a bit more specified and under the grammar author’s control, IMO.

Also, the fact that a symbol can contain multiple character sets, and they just get defined with these number suffixes, that again seems like an implementation detail, but not clear enough to be a stable API.

ObserverOfTime:
> They can, but what for? They're defined in `parser.c`, so declaring them in `parser.h` seems correct.

`parser.h` is the bread-and-butter header. I'd like to keep it minimal and identical for all grammars.

Much like the `array` API, these functions won't be needed by most authors, so they should live in a separate header.

ObserverOfTime:
> What is the linkage on these symbols? If they are not static, then they must be prefixed with ‘tree_sitter_$GRAMMAR_NAME’ to avoid name collisions. It would be nice to not export these unless the grammar author opts in somehow.

I agree. We should declare them with a prefix and either as `static inline` or `__attribute__((visibility("hidden")))` (on GNU). For ease of use, we can define a macro like:

```c
#define matches(sym, lookahead) tree_sitter_foo_matches_sym_##sym(lookahead)
```

amaanq:
> Thinking about this more - I feel like this API needs a little work. The character sets that tree-sitter generates are based on some fairly arbitrary heuristics for what helps compile time. This is fine for an internal optimization, but not suitable for a public API. It needs to be a bit more specified and under the grammar author’s control, IMO.
> 
> Also, the fact that a symbol can contain multiple character sets, and they just get defined with these number suffixes, that again seems like an implementation detail, but not clear enough to be a stable API.

Should we instead separate the internal character sets and instead expose a field in the grammar.js for users to explicitly pass in terminal rules that should be available in the scanner?

PR #4719, opened 2025-08-20
 build(deps): bump wasmtime-c-api to v33.0.2 

ObserverOfTime:
Not bumping to v34+ because it has breaking changes in C and bumps the Rust edition.

Supersedes #4465

PR #4866, opened 2025-09-22
Revert "0.26.0"

clason:
This reverts commit 9be3e2bdd8e1e62ad259e3c8ee7de7ebaed0553f.

Problem: This commit doesn't follow the release workflow (release,
branch, bump) and did not consistently update the version in the
manifests.

Solution: Revert and wait for a coordinated release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1886,7 +1886,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "bindgen 0.72.1",
  "cc",
@@ -1900,7 +1900,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-cli"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "ansi_colours",
  "anstyle",
@@ -1945,7 +1945,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-config"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "anyhow",
  "etcetera",
@@ -1956,7 +1956,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-generate"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "anyhow",
  "dunce",
@@ -1979,7 +1979,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-highlight"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "regex",
  "streaming-iterator",
@@ -1993,7 +1993,7 @@ version = "0.1.4"
 
 [[package]]
 name = "tree-sitter-loader"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "anyhow",
  "cc",
@@ -2015,7 +2015,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-tags"
-version = "0.27.0"
+version = "0.26.0"
 dependencies = [
  "memchr",
  "regex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.27.0"
+version = "0.26.0"
 authors = [
   "Max Brunsfeld <maxbrunsfeld@gmail.com>",
   "Amaan Qureshi <amaanq12@gmail.com>",
@@ -151,10 +151,10 @@ walkdir = "2.5.0"
 wasmparser = "0.229.0"
 webbrowser = "1.0.5"
 
-tree-sitter = { version = "0.27.0", path = "./lib" }
-tree-sitter-generate = { version = "0.27.0", path = "./crates/generate" }
+tree-sitter = { version = "0.26.0", path = "./lib" }
+tree-sitter-generate = { version = "0.26.0", path = "./crates/generate" }
 tree-sitter-language = { path = "./crates/language" }
-tree-sitter-loader = { version = "0.27.0", path = "./crates/loader" }
-tree-sitter-config = { version = "0.27.0", path = "./crates/config" }
-tree-sitter-highlight = { version = "0.27.0", path = "./crates/highlight" }
-tree-sitter-tags = { version = "0.27.0", path = "./crates/tags" }
+tree-sitter-loader = { version = "0.26.0", path = "./crates/loader" }
+tree-sitter-config = { version = "0.26.0", path = "./crates/config" }
+tree-sitter-highlight = { version = "0.26.0", path = "./crates/highlight" }
+tree-sitter-tags = { version = "0.26.0", path = "./crates/tags" }
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
       eachSystem = lib.genAttrs systems;
       pkgsFor = inputs.nixpkgs.legacyPackages;
 
-          version = "0.26.0";
+      version = "0.26.0";
 
       fs = lib.fileset;
       src = fs.toSource {


maxbrunsfeld:
I just used the `bump` xtask, I believe that is the newly agreed upon flow, but it did not work. I see the CI failed. https://github.com/tree-sitter/tree-sitter/actions/runs/17862371394/job/50797948240#step:4:457

I don't think we should revert, but we need to figure out why CI is failing for the release.

clason:
That was not the only problem with the release, sorry. I don't see a way to fix this without rolling back first.

The `xtask` is just a convenience script -- I don't use it myself because it is way too brittle. Much more important is to make sure you have a clean `release` branch to backport fixes to so you can make breaking changes for developing the next major version on `master`. It would also be nice to not make a release right in the middle of such breaking changes...

maxbrunsfeld:
What do you do?

clason:
For bumping the version? I just grep for the old version string and manually replace it with the new (Neovim quickfix list 4tw). But that's the smallest part of a (non-patch) release.

maxbrunsfeld:
I think the branch can be created at any time - before or after the bump. That's not related to the problems.

There seem to be two independent bugs that are both breaking the CI. One has to do with the `tree-sitter-language` crate's version number, and one is a bug in the job for publishing the NPM package.

clason:
> I think the branch can be created at any time - before or after the bump.

No, that's not correct. It is important for downstream users to be able to distinguish releases from pre-release tip builds (by the version number). This means that you need to release, branch, bump _in this order_. In particular, the release commit cannot be the same as the post-release bump.

Independent from that is the fact that HEAD was not in the state to push out to people as a release.

maxbrunsfeld:
> Independent from that is the fact that HEAD was not in the state to push out to people as a release.

Why?

clason:
There was follow-up work planned for the new CLI command lines (better names, actual docs), for one thing.

maxbrunsfeld:
That's fine. There is absolutely no reason we can't release like that.

maxbrunsfeld:
Also, rest assured that you can create any git branch, with *any* content, and *any* predecessor commit, at any time. Let me know why you think that is not possible.

clason:
Well, it's your project. As a maintainer of several downstream projects, I can assure you that such a development strategy adds significant friction. Making the effort to have updates be as smooth as possible for users very much pays off in my experience.

>  Let me know why you think that is not possible.

You cannot branch on "half a commit".

maxbrunsfeld:
Can you elaborate?

I would like to understand what your favorite process would be, so that we can figure out if there's some part we want to agree to as a group and then automate. If it is not automated, it is currently not the official process.

maxbrunsfeld:
How would you want to split this commit into multiple commits? https://github.com/tree-sitter/tree-sitter/commit/9be3e2bdd8e1e62ad259e3c8ee7de7ebaed0553f

clason:
I was under the impression that @amaanq had written all this down in a document somewhere that he shared with you and other admins.

> How would you want to split this commit into multiple commits? https://github.com/tree-sitter/tree-sitter/commit/9be3e2bdd8e1e62ad259e3c8ee7de7ebaed0553f

I would make a `release v0.26.0` commit that (in other projects) updates the changelog that is then tagged and released. From this commit, I would create a `release-0.26` branch. As the first commit on `master` after that, I would bump all versions to `0.27.0` ("prerelease").

(On patch releases, we just bump the version to the new release and tag that; people are not meant to follow the `release` branches.)

maxbrunsfeld:
There is not a changelog file in the repo. Your desired branch would simply point to the commit before the version bump. That can be created now.

clason:
> There is not a changelog file in the repo.

Yes, people know that ;)

> Your desired branch would simply point to the commit before the version bump. That can be created now.

Which would put "v0.26.0" the start of the 0.27 dev cycle and not the start of the `release-0.26` branch -- a bit confusing, wouldn't you say?

maxbrunsfeld:
I genuinely do not understand. Are you now just taking issue with the commit message of the version bump commit?

maxbrunsfeld:
If so, I actually agree; we should change the automated commit message to say "Prepare for 0.27.0 release".

clason:
I am describing the release workflow that I had thought we had agreed on, back around 0.23. If the `xtask` doesn't follow that, that is of course a problem.

maxbrunsfeld:
In case this needs to be said, I want your help @clason. I am having trouble parsing out concrete improvements we can make based on your feedback.

In https://github.com/tree-sitter/tree-sitter/pull/4867, I'm trying to fix what *I* think are the issues with the automation flow. I am not clear on what concrete changes you want to see.

clason:
> In case this needs to be said, I want your help @clason. I am having trouble parsing out concrete improvements we can make based on your feedback.

I think @amaanq putting the release process documentation somewhere public would be a good first start (which as I recall includes a bit of discussion and a public feature freeze period for "big" releases").

Then we should obviously fix the `xtask` (which seems to have other issues besides the outdated commit structure).

> In https://github.com/tree-sitter/tree-sitter/pull/4867, I'm trying to fix what I think are the issues with the automation flow. I am not clear on what concrete changes you want to see.

Those issues are entirely separate (and good to fix regardless), so don't let this discussion block them.

maxbrunsfeld:
I think the issues are fixed.

I need to get a release out today, because there are crashes in Zed that need to be fixed. I don't particularly care what the version number is. I just bumped the minor to 0.26 because that is what our official release workflow (the `bump` task) did.

The current situation is that we have a GitHub release called "0.26.0", but there are no published packages for that version. What is the most correct way to proceed in your opinion? We can delete the GitHub release, and proceed to re-try publishing 0.26.0. Or we can do something else.

clason:
> I need to get a release out today, because there are crashes in Zed that need to be fixed. 

Are there open issues on that? Any change that fixes a crash should be backported to the release-0.25 branch -- are you sure they are not fixed with 0.25.9 already? (Or the tip of `release-0.25` -- there's one commit on top of 0.25.9.) I notice that Zed is still on v0.25.6.

> The current situation is that we have a GitHub release called "0.26.0", but there are no published packages for that version. 

It's marked as a "pre-release", and I don't think any packager has touched it (last time I looked; I know homebrew skipped it).

My preference would be to 
1. (if necessary) backport more fixes to the `release-0.25` branch 
2. (if necessary) make a `v0.25.10` release
3. backout this change and wait for @amaanq and @WillLillis to give the green light on a 0.26 release (with all their breaking changes)
4. follow the above-described release process for 0.26.0
5. if necessary, immediately tag and publish a 0.26.1 fixup release from the release branch (wouldn't be the first time ;))

ObserverOfTime:
> The current situation is that we have a GitHub release called "0.26.0", but there are no published packages for that version.

There is a package for web-tree-sitter so pretending this release never happened is not an option.

clason:
>  need to get a release out today, because there are crashes in Zed that need to be fixed.

Are you talking about https://github.com/tree-sitter/tree-sitter/pull/4818?

I just backported the changes manually (https://github.com/tree-sitter/tree-sitter/pull/4868), let's see if that worked. If so, we can quickly push a 0.25.10.

ObserverOfTime:
#4865 might also be worth (merging and) backporting.

maxbrunsfeld:
> There is a package for web-tree-sitter so pretending this release never happened is not an option

Ok, good point. I didn't realize that one went through. Let's just publish 0.26.1 then, rather than deleting this one.

maxbrunsfeld:
That's one PR, but we also need 
* https://github.com/tree-sitter/tree-sitter/pull/4844
* https://github.com/tree-sitter/tree-sitter/pull/4719

At this point, I think we should go ahead and publish 0.26.1.

ObserverOfTime:
#4844 can be backported manually after #4868 is merged to avoid unnecessary conflicts.
#4719 cannot be backported since a patch release shouldn't bump the MSRV.

maxbrunsfeld:
Ok, I'm fine with *also* doing a patch release on the 0.25 branch, but I want to get a release out with the wasmtime upgrade, and also solve the problem that we have 0.26 *partially* out, so I think we should do a 0.26.1 release.

clason:
No point in putting in the work if you've already decided we are in 0.26 now.

ObserverOfTime:
If we do a 0.26.1 release _right now_ **it's still going to be partial.**

amaanq:
> If we do a 0.26.1 release _right now_ **it's still going to be partial.**

Yeah I agree. I consider 0.26.0 to be "broken", but we can't exactly undo it since web-tree-sitter was published, so we sort of have to ignore it and have 0.26.1 be the "real" release - which is just *not* ready yet. I'll backport the wasm issues and push out 0.25.10 so Zed can leverage that, does that work @maxbrunsfeld?

maxbrunsfeld:
Yeah, I'm fine w/ doing a 0.25.10 release.

The only thing is that
* I want to get the wasmtime upgrade out in a released version
* Now that 0.26.0 is out on NPM, it seems like we should get a working 0.26.1 out.

What are your concerns about publishing 0.26.1 after landing https://github.com/tree-sitter/tree-sitter/pull/4870 ? If there are more things we need to fix, we can do another patch release.

amaanq:
My concerns are that there were a couple of features I wanted to get out before 0.26.0, but since that release is a bit scuffed, I'm hoping we can just sweep it under the rug, get these features out for 0.26.1, and let users just jump to that.

amaanq:
Ok, 0.25.10 is out now and can be used with the wasm patches.

PR #4867, opened 2025-09-22
Fix issues preventing releases from successfully publishing

maxbrunsfeld:
See https://github.com/tree-sitter/tree-sitter/pull/4866

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,7 +78,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        directory: [cli/npm, lib/binding_web]
+        directory: [crates/cli/npm, lib/binding_web]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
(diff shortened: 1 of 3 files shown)


ObserverOfTime:
FWIW the release will still fail if the `tree-sitter-language` version is not bumped.
And it should be bumped to 0.2.0.

maxbrunsfeld:
Can we just bump to 0.1.5 since the API has not changed at all? Minor version bumps are incredibly disruptive in that crate, since every language crate depends on the range `0.1`.

maxbrunsfeld:
@amaanq Just want to loop you in on this, since it's about https://github.com/tree-sitter/tree-sitter/pull/4820. I'd love it if we could just bump the patch version number, because otherwise it'll break compatibility with all the grammar crates. Want to make sure I'm not missing anything though.

maxbrunsfeld:
https://github.com/tree-sitter/tree-sitter/pull/4870

amaanq:
> @amaanq Just want to loop you in on this, since it's about #4820. I'd love it if we could just bump the patch version number, because otherwise it'll break compatibility with all the grammar crates. Want to make sure I'm not missing anything though.

Yeah, I agree we should bump the patch number.

PR #4868, opened 2025-09-22
backport 4818

clason:
- **fix(lib): improve wasm scanner serialization error handling**
- **fix(lib/wasm): keep track of freed blocks that are not the last allocated pointer**
- **build(lib): regenerate wasm stdlib**
- **fix(lib/wasm): return NULL for 0-sized allocations**


clason:
@WillLillis I might need help with appeasing the linter, and I would appreciate if you could repeat your tests on this PR.

clason:
Python parser had significant breaking changes, so don't track HEAD (which we should stop doing regardless) -- v0.25.0 is right before these changes.

clason:
Never mind.

PR #4870, opened 2025-09-22
Bump tree-sitter-language version to 0.1.5

maxbrunsfeld:


ObserverOfTime:
Again, a MSRV bump should not be released as a patch. Not to mention this is a _significant_ change.

maxbrunsfeld:
Oh, I didn't realize the language crate's MSRV changed as well. But it still supports a *very* old version of Rust.

~@amaanq What was the Rust feature that prompted that upgrade?~ Nevermind, I see your comment about setting metadata in the build script.

amaanq:
1.77 is quite old, so this is fine.

PR #4873, opened 2025-09-22
fix(lib): improve wasm scanner serialization error handling

amaanq:
- Backports #4719, #4818, #4844

Yes this is technically breaking for 0.25.10, but 0.26.0 is currently broken and we want to get out a release today to fix wasm issues. I'd rather do this than publish 0.26.1 today.


maxbrunsfeld:
I'm ok delaying on the wasmtime upgrade if you want to be more semver-correct. I really do need to get the wasm crash fixes out though in *some* release.

maxbrunsfeld:
Thank you for all that you do @amaanq and @ObserverOfTime

amaanq:
> I'm ok delaying on the wasmtime upgrade if you want to be more semver-correct. I really do need to get the wasm crash fixes out though in _some_ release.

awesome, that helps a lot

PR #4876, opened 2025-09-23
Clean up tests

amaanq:
This PR has no user-facing changes, it's just some minor cleanup

- Clean up hacky `hang` test
- Prefer assert over panic in tests
- Avoid panicking in user-code where possible
- Update flake

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
       eachSystem = lib.genAttrs systems;
       pkgsFor = inputs.nixpkgs.legacyPackages;
 
-          version = "0.26.0";
+      version = "0.26.0";
 
       fs = lib.fileset;
       src = fs.toSource {
@@ -326,11 +326,12 @@
               clippy
               rust-analyzer
               rustfmt
-              cargo-cross
+              cargo-llvm-cov
 
               cmake
               gnumake
               pkg-config
+              llvm
               clang
               libclang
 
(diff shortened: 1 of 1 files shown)


crates/cli/src/tests/query_test.rs
```diff
@@ -4186,12 +4186,9 @@ fn test_query_random() {
             let pattern = pattern_ast.to_string();
             let expected_matches = pattern_ast.matches_in_tree(&test_tree);
 
-            let query = match Query::new(&language, &pattern) {
-                Ok(query) => query,
-                Err(e) => {
-                    panic!("failed to build query for pattern {pattern} - {e}. seed: {seed}");
-                }
-            };
+            let query = Query::new(&language, &pattern).unwrap_or_else(|_| {
```
WillLillis: Couldn't hurt to include the query error in the panic message here.

crates/cli/src/tests/tags_test.rs
```diff
@@ -262,34 +263,41 @@ fn test_tags_ruby() {
 
 #[test]
 fn test_tags_cancellation() {
-    use std::sync::atomic::{AtomicUsize, Ordering};
-
     allocations::record(|| {
         // Large javascript document
         let source = (0..500)
```
WillLillis: Not your change, but while we're looking at the test code I think it would be a little more straightforward to just write this as 

```rust
let source = "{}", "/* hi */ class A { /* ok */ b() {} }\n".repeat(500);
```

crates/cli/src/tests/tags_test.rs
```diff
@@ -262,34 +263,41 @@ fn test_tags_ruby() {
...
+            match tag {
+                Ok(_) => false,
+                Err(Error::Cancelled) => true,
+                Err(
+                    Error::Query(_)
+                    | Error::Regex(_)
+                    | Error::InvalidLanguage
+                    | Error::InvalidCapture(_),
+                ) => {
+                    unreachable!("Unexpected error type while iterating tags")
```
WillLillis: Could include the error in this message here. Hopefully just to save some time if we have to debug this later on.

PR #4889, opened 2025-09-25
feat: add option to disable parse state optimizations

### LINKED ISSUE #4711: LookaheadIterator returns all valid next symbols
### Problem

If you have a grammar with a rule that is reused in a few places, then the LookaheadIterator will return all of the possible next symbols for all contexts where the rule is used, because they are all in the parse table.

I am trying to use it on an `ERROR` node, however, and I want only the symbols which could have appeared in the context of the error.

I have tried a few ways to implement this, but been unable to come up with a solution (perhaps only due to my lack of understanding of the implementation).

### Expected behavior

Have a function which can filter the LookaheadIterator symbols by the context of another node.

amaanq:
- Closes #4711

### Problem

Users of the lookahead iterator that want to provide accurate information regarding possible completions for a given parse state can run into cases where there's more valid lookahead tokens than expected. This is because of the merging of compatible parse states that tree-sitter does during generation, where compatible states have their terminal and non-terminal entries extended.

### Solution

This PR adds a cli flag, `--disable-optimizations`, that allows users to opt out of this. I thought about making this more granular in the form of an enum or a number, but I think that's very premature and likely not even needed. Using an enum would be very weird for one member, and it's probable that we won't need to add anything else. I'd rather keep the flag simpler now, and then if we do need to change it in the future, we can just deprecate it and *then* use an enum.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -32,6 +32,7 @@ use tree_sitter_cli::{
     wasm,
 };
 use tree_sitter_config::Config;
+use tree_sitter_generate::OptLevel;
 use tree_sitter_highlight::Highlighter;
 use tree_sitter_loader::{self as loader, Bindings, TreeSitterJSON};
 use tree_sitter_tags::TagsContext;
@@ -162,6 +163,11 @@ struct Generate {
     /// The name or path of the JavaScript runtime to use for generating parsers, specify `native`
     /// to use the native `QuickJS` runtime
     pub js_runtime: Option<String>,
+
+    /// Disable optimizations when generating the parser. Currently, this only affects
+    /// the merging of compatible parse states.
+    #[arg(long)]
+    pub disable_optimizations: bool,
 }
 
 #[derive(Args)]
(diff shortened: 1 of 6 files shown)


crates/cli/src/main.rs
```diff
@@ -162,6 +163,11 @@ struct Generate {
     /// The name or path of the JavaScript runtime to use for generating parsers, specify `native`
     /// to use the native `QuickJS` runtime
     pub js_runtime: Option<String>,
+
+    /// Disable optimizations when generating the parser. Currently, this only affects
+    /// the merging of compatible parse states.
+    #[arg(long)]
+    pub disable_optimizations: bool,
```
ObserverOfTime, line 170: An integer `opt-level` might be better.

amaanq: a level wouldn't mean anything here, we're just disabling *one* optimization performed and this likely won't change for a while. if we did add something else, an enum would be better rather than a level, but if an enum were added in this PR, that'd be very premature

PR #4894, opened 2025-09-26
fix(lib): allow anonymous nodes in the supertype query syntax

### LINKED ISSUE #4297: Anonymous nodes don't work correctly in supertypes
### Problem

With the current version of my perl parser at https://github.com/tree-sitter-perl/tree-sitter-perl, the query
`(subscripted "[" @thingy)` will highlight EVERY `[` in the document with `@thingy`, not just ones that are in the 5 node types which are in `subscripted`.
Is this an intentional limitation, or an oversight?

### Steps to reproduce

```sh
git clone --depth=1 https://github.com/tree-sitter-perl/tree-sitter-perl.git -b veesh/supertypes-bug
tree-sitter generate
tree-sitter test
```
There will be test failures with overly zealously applied `@this.is.borked`. They should not be so.

### Expected behavior

All tests should pass, and only instances of the `[` token in nodes which belong to the supertype `subscripted` would get the particular highlighting

tree-sitter 0.24.7 (also confirmed w/ 0.25.3)

### Operating system/version

Ubuntu Linux

amaanq:
- Closes #4297

### Problem

The supertype syntax has 2 issues:

- It does not validate that the child is a valid subtype if not using the `/` syntax (e.g. `(supertype (subtype))`, and instead captures every instance of subtype regardless of its parent.
- It does not work with anonymous nodes (e.g. `(supertype/"subtype")`

### Solution

In this PR, I've added support for anonymous node subtypes in the supertype syntax, and added validation for subtype children to ensure invalid patterns do not successfully compile.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -238,6 +238,20 @@ fn test_query_errors_on_invalid_syntax() {
             ]
             .join("\n")
         );
+        assert_eq!(
+            Query::new(&language, "(statement / export_statement)").unwrap_err(),
+            QueryError {
+                row: 0,
+                offset: 11,
+                column: 11,
+                kind: QueryErrorKind::Syntax,
+                message: [
+                    "(statement / export_statement)", //
+                    "           ^"
+                ]
+                .join("\n")
+            }
+        );
     });
 }
 
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -416,11 +416,11 @@ fn test_query_errors_on_impossible_patterns() {
             Err(QueryError {
                 kind: QueryErrorKind::Structure,
                 row: 0,
-                offset: 51,
-                column: 51,
+                offset: 37,
+                column: 37,
                 message: [
                     "(binary_expression left: (expression (identifier)) left: (expression (identifier)))",
-                    "                                                   ^",
+                    "                                     ^",
                 ]
                 .join("\n"),
             })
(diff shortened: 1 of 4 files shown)


WillLillis (approved):

ribru17:
Does this go into 0.27 or 0.26.1 (if that is considered the canonical 0.26.0)?

ribru17:
Also, while we're at it, can we consume whitespace after / token? Iirc whitespace is consumed before it but not after

amaanq:
> Does this go into 0.27 or 0.26.1 (if that is considered the canonical 0.26.0)?

0.26.1, yes it is

> Also, while we're at it, can we consume whitespace after / token? Iirc whitespace is consumed before it but not after

Huh I hadn't thought of that, I would think it'd be better to ban whitespace before and after, but if you'd prefer allowing whitespace in general I wouldn't mind that.

ribru17:
I also think either is fine conceptually but banning prior whitespace now seems more breaking than allowing whitespace after

Plus we have a CLI formatter now that will remove spaces anyway

amaanq:
I think it'd be fine, I couldn't find any usages with prior whitespace on GH

https://github.com/search?q=%2F%5C%28%5Cw%2B%5Cs*%5C%2F%2F+path%3A**%2Fqueries%2F*.scm+language%3A%22Tree-sitter+Query%22&type=code&ref=advsearch

ribru17:
Sounds good, but then at least the commit should be marked as breaking, imo

amaanq:
Oh yeah, good point

clason:
Actually, the previous commit was breaking as well: The following pattern (in the [`fsharp`](https://github.com/ionide/tree-sitter-fsharp/blob/main/fsharp/grammar.js) queries) is now flagged as "impossible":
```lisp
((_type
  (long_identifier
    (identifier) @type.builtin))
  (#any-of? @type.builtin
    "bool" "byte" "sbyte" "int16" "uint16" "int" "uint" "int64" "uint64" "nativeint" "unativeint"
    "decimal" "float" "double" "float32" "single" "char" "string" "unit"))
```

(Bisected to 341665824c52b3c9259cfaea1fc179f615dccf3a)

PR #4899, opened 2025-09-28
ci(release): enable trusted publishing & attestations

ObserverOfTime:
In light of the recent (and older) supply chain attacks through compromised tokens, enabling trusted publishing allows us to get rid of our tokens and restrict releases to this particular workflow.

Provided the upcoming release is published successfully, the next steps will be to remove the tokens from the secrets (assuming they are backed up somewhere), and to [disallow tokens](https://docs.npmjs.com/trusted-publishers#how-to-configure-maximum-security) in general for the npm packages.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     permissions:
-      contents: write
+      id-token: write
+      attestations: write
+      contents: read
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


clason:
Have you run this new workflow on your fork and verified that it can successfully create a release (without uploading, of course)?

ObserverOfTime:
Unfortunately trusted publishing can't be tested without _actually_ publishing because most of the configuration is on the registry.

clason:
But you can at least test that the changes did not break the _release_ part. We want to make sure that we don't need another round of fixup releases.

maxbrunsfeld:
Great idea, thanks for setting this up @ObserverOfTime.

I can remove the tokens once we see this works.

ObserverOfTime:
@maxbrunsfeld Could you also set up tree-sitter on PyPI for tree-sitter/py-tree-sitter#421?
https://docs.pypi.org/trusted-publishers/adding-a-publisher/

PR #4910, opened 2025-09-29
fix(loader): allow parallel compilation on windows

amaanq:
### Problem

Windows users cannot run the test suite in parallel, meaning compilation cannot be done in the same dir in parallel.

### Solution

This is due to the Windows compiler writing to the same intermediate files (.exp, .lib, .obj) during compilation, which breaks in multithreaded scenarios. Now, we'll use a temp dir that's based on the pid and tid for these files, and clean them up when compilation is done.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,22 +70,23 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v5
 
-    - name: Set up environment
+    - name: Set up cross-compilation
+      if: matrix.cross
       run: |
-        printf 'EMSCRIPTEN_VERSION=%s\n' "$(<crates/loader/emscripten-version)" >> $GITHUB_ENV
+        for target in armv7-unknown-linux-gnueabihf i686-unknown-linux-gnu powerpc64-unknown-linux-gnu; do
+          camel_target=${target//-/_}; target_cc=${target/-unknown/}
+          printf 'CC_%s=%s\n' "$camel_target" "${target_cc/v7/}-gcc"
+          printf 'AR_%s=%s\n' "$camel_target" "${target_cc/v7/}-ar"
+          printf 'CARGO_TARGET_%s_LINKER=%s\n' "${camel_target^^}" "${target_cc/v7/}-gcc"
+        done >> $GITHUB_ENV
+        {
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -74,10 +74,7 @@ jobs:
       run: |
         printf 'EMSCRIPTEN_VERSION=%s\n' "$(<crates/loader/emscripten-version)" >> $GITHUB_ENV
```
ObserverOfTime: We can replace this job with two separate jobs:

```yaml
    - name: Set up cross-compilation
      if: matrix.cross
      run: |
        for target in armv7-unknown-linux-gnueabihf i686-unknown-linux-gnu powerpc64-unknown-linux-gnu; do
          camel_target=${target//-/_}; target_cc=${target/-unknown/}
          printf 'CC_%s=%s\n' "$camel_target" "${target_cc/v7/}-gcc"
          printf 'AR_%s=%s\n' "$camel_target" "${target_cc/v7/}-ar"
          printf 'CARGO_TARGET_%s_LINKER=%s\n' "${camel_target^^}" "${target_cc/v7/}-gcc"
        done >> $GITHUB_ENV
        {
          printf 'CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_RUNNER=qemu-arm -L /usr/arm-linux-gnueabihf\n'
          printf 'CARGO_TARGET_POWERPC64_UNKNOWN_LINUX_GNU_RUNNER=qemu-ppc64 -L /usr/powerpc64-linux-gnu\n'
        } >> $GITHUB_ENV

    - name: Get emscripten version
      if: contains(matrix.features, 'wasm')
      run: printf 'EMSCRIPTEN_VERSION=%s\n' "$(<crates/loader/emscripten-version)" >> $GITHUB_ENV
```

crates/loader/src/loader.rs
```diff
@@ -904,12 +904,24 @@ impl Loader {
 
         let output_path = config.output_path.as_ref().unwrap();
 
-        if compiler.is_like_msvc() {
+        let temp_dir = if compiler.is_like_msvc() {
             let out = format!("-out:{}", output_path.to_str().unwrap());
             command.arg(if self.debug_build { "-LDd" } else { "-LD" });
             command.arg("-utf-8");
+
+            let temp_dir = output_path.parent().unwrap().join(format!(
```
WillLillis, line 915: Might be helpful to leave a comment about the temporary build artifacts here, explaining why we're doing all of this for msvc.

PR #4914, opened 2025-10-01
fix(cli): prevent crash when parsing stdin

ObserverOfTime:
When we are parsing stdin via a pipe or heredoc, the source count is 0 (unsigned) so the XML output crashes while trying to subtract from it.

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -230,11 +230,21 @@ impl ParseSummary {
     }
 }
 
-#[derive(Serialize, Debug, Default)]
+#[derive(Serialize, Debug)]
 pub struct ParseStats {
     pub parse_summaries: Vec<ParseSummary>,
     pub cumulative_stats: Stats,
     pub source_count: usize,
+}
+
+impl Default for ParseStats {
+    fn default() -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
Are macos-15 runners dead? :thinking:

WillLillis:
Looking at this again, shouldn't we be accumulating parse stats if we're displaying a json summary with those stats?

```diff
diff --git a/crates/cli/src/main.rs b/crates/cli/src/main.rs
index 6f64c801..7d5e39a5 100644
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1034,7 +1034,7 @@ impl Parse {

         let mut update_stats = |stats: &mut parse::ParseStats| {
             let parse_result = stats.parse_summaries.last().unwrap();
-            if should_track_stats {
+            if should_track_stats || self.json {
                 stats.cumulative_stats.total_parses += 1;
                 if parse_result.successful {
                     stats.cumulative_stats.successful_parses += 1;
```

And separately, should we mark the `--json` and `--stat` flags as conflicting? It seems feels a bit odd to allow both at the same time:

image

ObserverOfTime:
I think those changes should be made in your PR.

PR #4955, opened 2025-10-12
feat: dynamically expose queries in bindings

### LINKED ISSUE #4954: Remove the need to manually edit Rust bindings when adding queries
### Problem

The Rust bindings generated by default by `tree-sitter init` include the following lines in `lib.rs`:

```rust
// NOTE: uncomment these to include any queries that this grammar contains:

// pub const HIGHLIGHTS_QUERY: &str = include_str!("../../queries/highlights.scm");
// pub const INJECTIONS_QUERY: &str = include_str!("../../queries/injections.scm");
// pub const LOCALS_QUERY: &str = include_str!("../../queries/locals.scm");
// pub const TAGS_QUERY: &str = include_str!("../../queries/tags.scm");
```

Unfortunately, grammar maintainers often forget to uncomment those lines to expose the queries to Rust crate reusers. Even if they think about doing it when first generating the bindings, they might not think about it when introducing a new type of query, and their edits might get overwritten when they re-generate the bindings.

This has led me to opening quite a few PRs to have Rust crates expose them: https://github.com/yuja/tree-sitter-qmljs/pull/21, https://github.com/elves/tree-sitter-elvish/pull/5, https://github.com/joelspadin/tree-sitter-devicetree/pull/21, https://github.com/slackhq/tree-sitter-hack/pull/75, https://github.com/briot/tree-sitter-ada/pull/16 … but again, even if those get merged and released, they might get overwritten at the next update.

Overall, it is good to avoid relying on workflows where people edit auto-generated files.

### Expected behavior

It would be better if the Rust bindings could automatically expose those queries whenever they are available in the parser's repository.

wetneb:
Closes #4954.

This uses the existing `build.rs` to detect the presence of each query file at compile time and define a corresponding configuration variable. An alternative (to avoid relying on `build.rs`) would be to use a dedicated macro, such as https://github.com/eliasholzmann/include_optional or https://crates.io/crates/file-exists-macro, but that would add a dependency to the crate.

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -268,15 +268,6 @@ pub fn generate_grammar_files(
         .clone()
         .unwrap_or_else(|| format!("TreeSitter{}", language_name.to_upper_camel_case()));
 
-    fn pathsjson_to_variable<'a>(paths_json: &'a PathsJSON, default: &'a PathBuf) -> &'a str {
-        match paths_json {
-            PathsJSON::Empty => Some(default),
-            PathsJSON::Single(path_buf) => Some(path_buf),
-            PathsJSON::Multiple(paths) => paths.first(),
-        }
-        .map_or("", |path| path.as_os_str().to_str().unwrap_or(""))
-    }
-
     let default_highlights_path = Path::new("queries").join(DEFAULT_HIGHLIGHTS_QUERY_FILE_NAME);
     let default_injections_path = Path::new("queries").join(DEFAULT_INJECTIONS_QUERY_FILE_NAME);
     let default_locals_path = Path::new("queries").join(DEFAULT_LOCALS_QUERY_FILE_NAME);
@@ -308,22 +299,18 @@ pub fn generate_grammar_files(
         camel_parser_name: &camel_name,
         title_parser_name: &title_name,
         class_name: &class_name,
-        highlights_query_path: pathsjson_to_variable(
-            &tree_sitter_config.grammars[0].highlights,
-            &default_highlights_path,
-        ),
-        injections_query_path: pathsjson_to_variable(
-            &tree_sitter_config.grammars[0].injections,
-            &default_injections_path,
-        ),
-        locals_query_path: pathsjson_to_variable(
-            &tree_sitter_config.grammars[0].locals,
-            &default_locals_path,
-        ),
-        tags_query_path: pathsjson_to_variable(
-            &tree_sitter_config.grammars[0].tags,
-            &default_tags_path,
-        ),
+        highlights_query_path: tree_sitter_config.grammars[0]
+            .highlights
+            .to_variable_value(&default_highlights_path),
+        injections_query_path: tree_sitter_config.grammars[0]
+            .injections
+            .to_variable_value(&default_injections_path),
+        locals_query_path: tree_sitter_config.grammars[0]
+            .locals
+            .to_variable_value(&default_locals_path),
+        tags_query_path: tree_sitter_config.grammars[0]
+            .tags
+            .to_variable_value(&default_tags_path),
     };
 
     // Create package.json
@@ -503,11 +490,11 @@ pub fn generate_grammar_files(
                             }
                         "#};
 
-                    let indented_replacement = replacement
-                        .lines()
-                        .map(|line| if line.is_empty() { line.to_string() } else { format!("    {line}") })
-                        .collect::<Vec<_>>()
-                        .join("\n");
+                        let indented_replacement = replacement
+                            .lines()
+                            .map(|line| if line.is_empty() { line.to_string() } else { format!("    {line}") })
+                            .collect::<Vec<_>>()
+                            .join("\n");
 
                         contents = contents.replace(r#"    c_config.flag("-utf-8");"#, &indented_replacement);
                     }
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -284,6 +284,17 @@ impl PathsJSON {
     const fn is_empty(&self) -> bool {
         matches!(self, Self::Empty)
     }
+
+    /// Represent this set of paths as a string that can be included in templates
+    #[must_use]
+    pub fn to_variable_value<'a>(&'a self, default: &'a PathBuf) -> &'a str {
+        match self {
+            Self::Empty => Some(default),
+            Self::Single(path_buf) => Some(path_buf),
+            Self::Multiple(paths) => paths.first(),
+        }
+        .map_or("", |path| path.as_os_str().to_str().unwrap_or(""))
+    }
 }
 
 #[derive(Serialize, Deserialize, Clone)]


clason:
You cannot rely on a fixed location of the query files, though, for a variety of reasons.

wetneb:
Sure. With this change, grammar maintainers would still be free to store queries under the location of their choice, and edit the Rust bindings to expose them if they want to.

Given that the default locations suggested by tree-sitter's template are followed by a large number of grammar repositories, I think this change would still be useful for those grammars, as it would remove the need for them to edit those Rust bindings manually.

In other words, I think this change would be useful to grammar maintainers which follow the norm, without hurting those who decide to deviate from it.

clason:
> In other words, I think this change would be useful to grammar maintainers which follow the norm

The issue is that you're now _introducing_ a norm that was not there before -- and that is something that is not to be done lightly; certainly not without discussion, documentation, and (respectful) tooling. We also need to be very careful about diverging experience between different bindings (and naked lib/CLI users).

(Personally, I would love to make queries _mandatory_. And enforce them to be Neovim compatible 😤 )

clason:
That being said, the `tree-sitter.json` config file already [includes information about bundled queries](https://tree-sitter.github.io/tree-sitter/cli/init.html#query-paths) -- why not simply rely on that instead of hard-coding? That would also remove the need for scanning the filesystem.

wetneb:
Oh, I didn't realize `tree-sitter.json` had that information. That sounds good, I'll work on pulling the paths from there.

ObserverOfTime:
I don't dislike this idea, but it should be done in _all_ bindings or none at all.

clason:
Or we should have fewer bindings. Feel free to open a tracking issue for the other bindings (all or nothing means nothing, who can give a toss about all these languages).

wetneb:
So, it looks like it should be possible to define template variables that `tree-sitter init` would replace by the paths to each query file - except for the case where multiple files are supplied for a single query type (meaning that the files should be concatenated). It seems difficult to me to handle this case, since:
* it would be hacky to encode lists of paths in a single template variable (they would need to be concatenated with a certain separator and then parsed back in `build.rs` or `lib.rs`…)
* then, in the case of Rust, we'd need some clever macro incantations mixing `include_str!` and `concat!` to make sure the end result is a `&'static str`

In addition, the case of having multiple query files concatenated together seems often used to pull queries from "grammar dependencies" (for instance [in tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/tree-sitter.json)) where the queries being pulled are stored in `node_modules/`, which isn't included in the crate according to the default `Cargo.toml`, so the compilation of the crate would fail if those paths were attempted to be fetched at compile time. This makes me doubt about the usefulness of handling this case altogether: given that the grammar maintainer will most likely need to make manual edits, it's perhaps better to let them handle this case.

Would you be interested in a version of this feature which would only support the case where each query is defined by a single file (be it at the default location or not), and bail out (not exposing the query) if multiple files are used?

clason:
I think that would be a reasonable first step. Strict improvement without shutting the door on follow up PRs.

ObserverOfTime:
### Python

Already lazy-loading the queries but commented out by default.
If we just return `None` on missing that should be good enough.

### Node

Should be fairly simple to load them if they exist:

```node
import { readFile } from "node:fs/promises";

binding.HIGHLIGHTS_QUERY = await readFile(/* ... */, "utf8").catch(_ => undefined);
// ...
```

(We can also make `binding` a `Proxy` to lazy-load them when accessed.)

### Swift

Maybe possible with proper Swift bindings and not just a C wrapper.

### Zig

Maybe possible but not worth changing with every Zig update.

### Go

No idea.

### Java

Probably not possible.

ObserverOfTime:
> So, it looks like it should be possible to define template variables that `tree-sitter init` would replace by the paths to each query file

This would get rather complicated and wouldn't solve the issue since queries are usually configured _after_ `init`.

> In addition, the case of having multiple query files concatenated together seems often used to pull queries from "grammar dependencies" (for instance [in tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/tree-sitter.json)) where the queries being pulled are stored in `node_modules/`

Yeah we definitely need a way to handle dependencies that works on more than just Node.

wetneb:
> This would get rather complicated and wouldn't solve the issue since queries are usually configured after init.

I definitely share your concern, but I would guess it would be acceptable to expect that users re-generate bindings if they change the paths to queries in `tree-sitter.json`, no? After all, that is also required for changes to many other fields of `tree-sitter.json`.

An alternative would be to attempt to parse `tree-sitter.json` at runtime in `build.rs`, but that would require adding a JSON parsing crate as a dependency to the generated crate. Not ideal as it would inflate the build time for no usability gain to end users.

@ObserverOfTime do you see another solution? Or just going with the approach that I have here, only handling the case of query files placed at their default location?

clason:
`tree-sitter.json` is the record of truth; if we can't use that, we should not presume. Guessing wrong is worse than not guessing at all.

Concretely, I agree with

> expect that users re-generate bindings if they change the paths to queries in tree-sitter.json

wetneb:
I've added the logic to customize the query locations via `tree-sitter.json` and updated the Rust and Python bindings to use that. I tested the resulting bindings on a sample grammar and they seem to work. I propose to leave out the other bindings for now as they don't expose queries yet as far as I can tell. I think it would be worth following up with other PRs to expose the queries there too if possible.

crates/cli/src/templates/__init__.py
ObserverOfTime, line 1: You also need to adapt `__init__.pyi` to return `Final[str] | None`.

crates/cli/src/init.rs
```diff
@@ -255,6 +268,28 @@ pub fn generate_grammar_files(
...
+            PathsJSON::Multiple(paths) => {
+                if paths.len() == 1 {
+                    paths.first()
+                } else {
+                    // If multiple paths are not supported for now, as we would need a more expressive template system
+                    // to be able to represent faithfully in a variable.
+                    None
+                }
+            }
+        }
```
ObserverOfTime: We should always use the first path. It's unlikely that we'll ever support multiple.
```suggestion
            PathsJSON::Multiple(paths) => paths.first(),
```

crates/cli/src/init.rs
ObserverOfTime, line 1: Can you add migrations for the affected files?

crates/cli/src/templates/lib.rs
```diff
@@ -32,12 +32,18 @@ pub const LANGUAGE: LanguageFn = unsafe { LanguageFn::from_raw(tree_sitter_PARSE
...
-// pub const LOCALS_QUERY: &str = include_str!("../../queries/locals.scm");
-// pub const TAGS_QUERY: &str = include_str!("../../queries/tags.scm");
+#[cfg(with_highlights_query)]
+pub const HIGHLIGHTS_QUERY: &str = include_str!("../../HIGHLIGHTS_QUERY_PATH");
+#[cfg(with_injections_query)]
+pub const INJECTIONS_QUERY: &str = include_str!("../../INJECTIONS_QUERY_PATH");
+#[cfg(with_locals_query)]
+pub const LOCALS_QUERY: &str = include_str!("../../LOCALS_QUERY_PATH");
+#[cfg(with_tags_query)]
+pub const TAGS_QUERY: &str = include_str!("../../TAGS_QUERY_PATH");
```
ObserverOfTime, line 49:
```suggestion
#[cfg(with_highlights_query)]
/// The syntax highlighting query for this grammar.
pub const HIGHLIGHTS_QUERY: &str = include_str!("../../HIGHLIGHTS_QUERY_PATH");

#[cfg(with_injections_query)]
/// The language injection query for this grammar.
pub const INJECTIONS_QUERY: &str = include_str!("../../INJECTIONS_QUERY_PATH");

#[cfg(with_locals_query)]
/// The local variable query for this grammar.
pub const LOCALS_QUERY: &str = include_str!("../../LOCALS_QUERY_PATH");

#[cfg(with_tags_query)]
/// The symbol tagging query for this grammar.
pub const TAGS_QUERY: &str = include_str!("../../TAGS_QUERY_PATH");
```

crates/cli/src/templates/lib.rs
```diff
@@ -32,12 +32,18 @@ pub const LANGUAGE: LanguageFn = unsafe { LanguageFn::from_raw(tree_sitter_PARSE
 /// [`node-types.json`]: https://tree-sitter.github.io/tree-sitter/using-parsers/6-static-node-types
 pub const NODE_TYPES: &str = include_str!("../../src/node-types.json");
 
-// NOTE: uncomment these to include any queries that this grammar contains:
+// NOTE: if queries are drawn from multiple files, remove the #[cfg] attribute and load them with
+// concat!(include_str!("first_file"), include_str!("second_file"))
+// Make sure that each file is included in the crate by adding necessary any paths to `Cargo.toml`
```
ObserverOfTime: Can we add a relevant notice to the docs instead?

```suggestion
```

crates/cli/src/templates/__init__.py
```diff
@@ -6,32 +6,35 @@
 
 
 def _get_query(name, file):
-    query = _files(f"{__package__}.queries") / file
-    globals()[name] = query.read_text()
+    if not file:
+        return None
```
ObserverOfTime:
```suggestion
```

crates/cli/src/init.rs
wetneb: I've added one for the Rust bindings. One could potentially make the replacement more granular to handle the case where specific lines have been uncommented (replacing each commented line independently from the others), it's a bit more work but it could be done.

For Python, I'm not too sure what a migration look like given that the edits are all over the place. IMHO it amounts to re-generating the whole `__init__.py(i)` files, I'm not sure how we'd preserve manual edits.

ObserverOfTime: You can replace the entire files (with a warning like in other such cases).

crates/cli/src/init.rs
```diff
@@ -431,8 +431,45 @@ pub fn generate_grammar_files(
     // Generate Rust bindings
     if tree_sitter_config.bindings.rust {
         missing_path(bindings_dir.join("rust"), create_dir)?.apply(|path| {
-            missing_path(path.join("lib.rs"), |path| {
+            missing_path_else(path.join("lib.rs"), allow_update, |path| {
                 generate_file(path, LIB_RS_TEMPLATE, language_name, &generate_opts)
+            }, |path| {
+                let mut contents = fs::read_to_string(path)?;
```
ObserverOfTime, line 424: Maybe check for `#[cfg(with_highlights_query)]` before running the replacement.

crates/cli/src/init.rs
```diff
@@ -473,6 +510,43 @@ pub fn generate_grammar_files(
                         contents = contents.replace(r#"    c_config.flag("-utf-8");"#, &indented_replacement);
                     }
 
+                    // Introduce configuration variables for dynamic query inclusion
```
ObserverOfTime, line 502: Another check here.

And while you're at it, move the above replacement inside the wasm32 check.

wetneb:
I think I have followed all review suggestions, let me know if you are waiting for me to do anything else.
In other projects, I would normally update the documentation as well, but [I have been told not to do so here](https://github.com/tree-sitter/tree-sitter/pull/4647#issuecomment-3124276110).

crates/cli/src/init.rs
```diff
@@ -717,17 +838,32 @@ pub fn generate_grammar_files(
...
-            })?;
+            missing_path_else(
+                lang_path.join("__init__.py"),
+                allow_update,
+                |path| {
+                    generate_file(path, INIT_PY_TEMPLATE, language_name, &generate_opts)
+                },
+                |path| {
+                    let contents = fs::read_to_string(path)?;
+                    if !contents.contains("FileNotFoundError") {
```
```suggestion
                    if contents.contains("uncomment these to include any queries") {
```

ObserverOfTime:
I squashed the commits and added support to the Node bindings.

crates/cli/src/init.rs
```diff
@@ -255,6 +268,20 @@ pub fn generate_grammar_files(
         .clone()
         .unwrap_or_else(|| format!("TreeSitter{}", language_name.to_upper_camel_case()));
 
+    fn pathsjson_to_variable<'a>(paths_json: &'a PathsJSON, default: &'a PathBuf) -> &'a str {
```
WillLillis: I think it makes more sense to move this into the `impl` for `PathsJSON`. Also not a huge fan of the name `pathsjson_to_variable`, maybe `resolve_path` or something similar instead?

crates/cli/src/init.rs
```diff
@@ -388,48 +431,126 @@ pub fn generate_grammar_files(
...
+                                c_config.include(&wasm_headers);
+                                c_config.files([
+                                    wasm_src.join("stdio.c"),
+                                    wasm_src.join("stdlib.c"),
+                                    wasm_src.join("string.c"),
+                                ]);
+                            }
+                        "#};
 
                     let indented_replacement = replacement
```
WillLillis: rustfmt kind of died here, but this `let indented_replacement = ...` block should be indented another level.

WillLillis (changes requested):
A few nits, but overall looks good! Beyond those, testing locally everything seems to be functioning correctly. As long as @ObserverOfTime gives the go ahead I think this should be good to land.

crates/cli/src/init.rs
wetneb: thanks for catching that, I'm curious why rustfmt gave up on this…

WillLillis: It just problems and silently gives up a lot of the time :laughing:

wetneb:
Thanks for the comments @WillLillis, I addressed them.

WillLillis:
> Thanks for the comments @WillLillis, I addressed them.

Saw that, thanks! I haven't had the chance to sit down and fully test this out yet, but I should be able to sometime this week. :slightly_smiling_face:

wetneb:
Rebased to solve the merge conflicts.

PR #4971, opened 2025-10-19
docs: more detailed description of `build -w`

giuliano-macedo:
Improved the docs of `build -w` because I had to look at the source code of tree-sitter-cli to undertand why it was pulling docker image of emscripten even after I've installed the [arch /extra package of emscripten](https://archlinux.org/packages/extra/x86_64/emscripten/), the issue was that by default in arch, the `emcc` command is not added to the PATH after the installation.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -173,7 +173,9 @@ struct Generate {
 #[derive(Args)]
 #[command(alias = "b")]
 struct Build {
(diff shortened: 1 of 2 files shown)


clason:
Note that does not happen on `master`/0.26 anyway, since `build -w` now pulls in wasi-sdk instead (which is much smaller). The docs should be changed to reflect that.

PR #4976, opened 2025-10-20
docs: Update npm instruction to install globally

OpenSaned:


sogaiu:
I don't think it makes sense to encourage people to install things globally via `npm`.

clason:
Agreed; `npm` and friends are for package-local dependencies; if you want it to be available globally, use your OS's package manager.

PR #4983, opened 2025-10-21
Swift bindings: scanner detection, unneeded dependency

mattmassicotte:
Hello!

I've noticed a serious issue with the Swift binding generation process. The original code was checking for the presence of a path relative to the current working directory. This had the unfortunate effect of working for tests executed within a grammar directory itself, but failing for all practical uses of the package.

The fix was straightforward. It requires using a new package API, which involved bumping the minimum required tools version. This is no big deal, as anything < 6.0 is virtually impossible to support for real projects anyways.

While I'm here I also removed the dependency on SwiftTreeSitter. It's totally unnecessary to validate the grammar module creation.

However, I need a little help to get this over the finish line.

I was trying to test on the YAML grammar, which is impacted by this bug. I cannot get `tree-sitter generate` to actually create any bindings for me locally. I wasn't able to find any documentation around how or even if this should work and my rust is... well let's just say it is not strong.

Speaking of which, I also see some code in `crates/cli/src/init.rs` that seems very related. But I'm not 100% sure what it was doing and how to best modify it.

Some pointers would be very appreciated!

### DIFF
--- a/crates/cli/src/templates/tests.swift
+++ b/crates/cli/src/templates/tests.swift
@@ -1,12 +1,9 @@
 import XCTest
-import SwiftTreeSitter
 import PARSER_CLASS_NAME
 
 final class PARSER_CLASS_NAMETests: XCTestCase {
     func testCanLoadGrammar() throws {
-        let parser = Parser()
-        let language = Language(language: tree_sitter_LOWER_PARSER_NAME())
-        XCTAssertNoThrow(try parser.setLanguage(language),
-                         "Error loading TITLE_PARSER_NAME grammar")
+        // just building validates the module creation, and linking validates the binary
+        XCTAssertNotNil(tree_sitter_LOWER_PARSER_NAME())
     }
 }
--- a/crates/cli/src/templates/package.swift
+++ b/crates/cli/src/templates/package.swift
@@ -1,10 +1,10 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.6
 
 import Foundation
 import PackageDescription
 
 var sources = ["src/parser.c"]
-if FileManager.default.fileExists(atPath: "src/scanner.c") {
+if FileManager.default.fileExists(atPath: Context.packageDirectory + "/src/scanner.c") {
     sources.append("src/scanner.c")
 }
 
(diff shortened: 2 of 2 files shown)


clason:
On current versions, bindings are no longer touched with `generate` (too noisy); you need to `init --update` to regenerate the bindings.

mattmassicotte:
Thank you @clason that's great!

But doing that appears to not actually apply any of my template modifications. I've done something wrong. Could it be related to the stuff in `init.rs` I wonder...

crates/cli/src/templates/tests.swift
```diff
@@ -1,12 +1,9 @@
...
 import PARSER_CLASS_NAME
 
 final class PARSER_CLASS_NAMETests: XCTestCase {
     func testCanLoadGrammar() throws {
-        let parser = Parser()
-        let language = Language(language: tree_sitter_LOWER_PARSER_NAME())
-        XCTAssertNoThrow(try parser.setLanguage(language),
-                         "Error loading TITLE_PARSER_NAME grammar")
+        // just building validates the module creation, and linking validates the binary
+        XCTAssertNotNil(tree_sitter_LOWER_PARSER_NAME())
```
ObserverOfTime, line 7: But it doesn't validate that the language version can be loaded.

mattmassicotte: Yes that's right. The purpose of a grammar package is to make a *loadable* language (build and link with success), but it not really the grammar package's concern with how that is being used.

So, for example, this test fails with a *linker* error for the YAML package (when used outside of its own source directory) without the Package.swift changes as well.

ObserverOfTime: The purpose of the tests in all bindings is to ensure the grammar can be _used_, not just built and linked.

mattmassicotte: Right ok that's probably a good idea. It should be done by actually reference the runtime and not SwiftTreeSitter though. Which should be fairly straightforward now that the runtime itself has a Swift package.

mattmassicotte:
Sorry about the noise all!

ObserverOfTime:
The scanner detection fix is perfectly fine.

PR #5000, opened 2025-10-26
docs: New page about ABI versions for parser users

### LINKED ISSUE #374: Documentation is missing information on ABI stability
I'm missing some information regarding ABI stability that I think it would be helpful to add to the docs.

I'm creating an Atom language package which depends on an npm package with a tree-sitter based parser. I'm unsure which tree-sitter version to use in order to make my grammar package work in as many Atom versions as possible.

Atom 1.37.0 uses the tree-sitter runtime 0.13.23, while Atom 1.38.2 uses tree-sitter runtime 0.15.4.

If I prebuild with 0.13.23, my grammar will crash Atom 1.38.2 (at least on MacOS). If I prebuild with 0.15.7, my grammar will not be recognized by Atom 1.37.0 (ie. I get no syntax highlighting, incompatible language version message in console).

Would it be possible to document how ABI stability is handled in tree-sitter?

Am I right to assume that LANGUAGE_VERSION defines an ABI version?

Which versions support which LANGUAGE_VERSIONs? Also, which Atom versions would support grammars created with which tree-sitter version? Is there a sweet spot that would allow me to support most Atom versions?

Thanks for clarifying!

wetneb:
Closes #374.

The statement about the intended backwards compatibility is purely speculative and provided as a "straw man" to help reviewers come up with a better description of the intended backwards compatibility.

### DIFF
--- a/docs/src/using-parsers/7-abi-versions.md
+++ b/docs/src/using-parsers/7-abi-versions.md
@@ -2,7 +2,7 @@
 
 Parsers generated with tree-sitter have an associated ABI version. This version establishes hard compatibility boundaries between the generated parser and the tree-sitter library.
 
-A given version of tree-sitter is only able to load parsers which have certain ABI versions:
+A given version of the tree-sitter library is only able to load parsers which have certain ABI versions:
 
 | tree-sitter version | Min parser ABI version | Max parser ABI version |
 |---------------------|------------------------|------------------------|
@@ -14,7 +14,7 @@ A given version of tree-sitter is only able to load parsers which have certain A
 | >=0.20.3, <=0.24    | 13                     | 14                     |
 | >=0.25              | 13                     | 15                     |
 
-By default, parsers are generated using the latest available ABI. Grammar authors can specify an older ABI via the `--abi` option to the `generate` command:
+By default, the tree-sitter CLI will generate parser using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
 ```
 tree-sitter generate --abi=<DESIRED-ABI>
 ```


clason:
> The statement about the intended backwards compatibility is purely speculative and provided as a "straw man" to help reviewers come up with a better description of the intended backwards compatibility.

For a variety of reasons, I would not be comfortable making _any_ claim here. Best to remove this.

(It's also important to distinguish the CLI and C library; while these have joint releases and will obviously introduce a new ABI in tandem, deprecation/removal is _not_ necessarily linked -- the CLI can stop generating deprecated ABI while the lib still supports it for backward compatibility. In fact, try generating ABI 13 now...)

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -0,0 +1,19 @@
+# ABI versions
+
+Parsers generated with tree-sitter have an associated ABI version. This version specifies the shape of the C code that they contain, which they expose to applications.
```
WillLillis: I'd reword the second bit about the ABI here. It's not the responsibility of the docs to explain *what* ABI is, the important thing here is to document how/why ABI affects tree-sitter. Instead, we could do something more along the lines of:

```
Parsers generated with the tree-sitter CLI tool have an associated ABI version. This version establishes hard compatibility boundaries between the generated parser and the tree-sitter library.
```

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -0,0 +1,19 @@
...
+| >=0.15.0, <=0.15.7  | 9                      | 10                     |
+| >=0.15.8, <=0.16    | 9                      | 11                     |
+| 0.17, 0.18          | 9                      | 12                     |
+| >=0.19, <=0.20.2    | 13                     | 13                     |
+| >=0.20.3, <=0.24    | 13                     | 14                     |
+| >=0.25              | 13                     | 15                     |
+
+Tree-sitter developers aim to maintain compatibility with the previous ABI version at least.
+
+Parser authors are able to specify the ABI version they want to use by specifying it with the `--abi` option of the `tree-sitter generate` command.
```
WillLillis: I think this would fit better as an admonish note. Also, something closer to:

```suggestion
By default, parsers are generated using the latest available ABI. Grammar authors can specify an older ABI via the `--abi` option to the `generate` command:

tree-sitter generate --abi=<DESIRED-ABI>
```

wetneb:
I have integrated the suggestions with the suggested wording, which looks good to me. I'm not sure about the admonition, it doesn't seem really necessary for a page that has so little content. The possibility of specifying a custom ABI already stands out enough thanks to the different formatting of the code block, in my opinion.

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -0,0 +1,20 @@
+# ABI versions
+
+Parsers generated with tree-sitter have an associated ABI version. This version establishes hard compatibility boundaries between the generated parser and the tree-sitter library.
+
+A given version of tree-sitter is only able to load parsers which have certain ABI versions:
```
clason:
```suggestion
A given version of the tree-sitter library is only able to load parsers which have certain ABI versions:
```

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -0,0 +1,20 @@
...
+|---------------------|------------------------|------------------------|
+| 0.14                | 9                      | 9                      |
+| >=0.15.0, <=0.15.7  | 9                      | 10                     |
+| >=0.15.8, <=0.16    | 9                      | 11                     |
+| 0.17, 0.18          | 9                      | 12                     |
+| >=0.19, <=0.20.2    | 13                     | 13                     |
+| >=0.20.3, <=0.24    | 13                     | 14                     |
+| >=0.25              | 13                     | 15                     |
+
+By default, parsers are generated using the latest available ABI. Grammar authors can specify an older ABI via the `--abi` option to the `generate` command:
```
clason:
```suggestion
By default, the tree-sitter CLI will generate parser using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
```

WillLillis (approved):

PR #5012, opened 2025-10-30
fix(npm): improve DSL function definitions

ObserverOfTime:
Problem: The choice and seq functions allow calls with 0 arguments.

Solution: Define a new array type to require at least one argument.


ObserverOfTime:
Turns out this interferes with `seq(...mappedRules)` and `choice(...mappedRules)`

Issue #5013, opened 2025-10-31
Tree-sitter 0.26.x release checklist


clason:
### Problem

Tree-sitter 0.26.0 was released in a partially unfinished state.

Stan4402:
/

PR #5017, opened 2025-11-01
fix(cli): remove `--emit=lib` generate option

WillLillis:
This also replaces the `--emit` option with an `--emit-json` flag. The default value is false, meaning a parser is still generated by default.

See https://github.com/tree-sitter/tree-sitter/pull/4601#issuecomment-3472303711

Partially addresses #5013

CC @clason

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -572,14 +572,14 @@ pub fn generate_grammar_files(
                             .replace(
                                 indoc! {r"
                                 $(PARSER): $(SRC_DIR)/grammar.json
-                                	$(TS) generate $^
+                                        $(TS) generate $^
                                 "},
                                 indoc! {r"
                                 $(SRC_DIR)/grammar.json: grammar.js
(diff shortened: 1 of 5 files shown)


ObserverOfTime:
Wouldn't it make more sense to have an `--emit-parser` flag that is true by default, since the JSON is always emitted?

clason:
Maybe, but that would complicate the code and be a bit more tedious to use (forcing to specify a boolean value instead of using a plain flag).

WillLillis:
Agreed with Clason, I think it would be less intuitive to have a boolean flag that defaults to `true` if unspecified. We don't have any other flags like that in thei CLI.

ObserverOfTime:
`--emit-json` isn't a very intuitive name though. Maybe `--json-only` or `--no-parser`?

WillLillis:
> `--emit-json` isn't a very intuitive name though. Maybe `--json-only` or `--no-parser`?

`--no-parser` sounds better to me, I can push that up later tonight.

PR #4908, opened 2025-09-29
Test Refactor

WillLillis:
This PR aims to refactor a good chunk of the CLI tool's internal testing logic. The only user-facing change should be the addition of the `--json-summary` flag, which outputs all test results in a JSON format. Internally, there are a few changes:

- cleanup the filtering logic a bit for parse tests. Also, only update a test case's expected output if the `--update` flag has been passed.
- Collect all test results as structured data, rather than printing results as we go. Once all tests have been run (or, if an unrecoverable error occurrs), we then display the results all at once. All of the mutable state which previously lived in `opts` now belongs to `test_summary`.
- Greatly declutter the logic displaying parse rates alongside individual test case's pass/fail status. Parse rates are now logically associated along with their associated test case, rather than "zipped" up at the end in the display logic.
- Generalize the display of test diffs a bit. This was mostly so the diff key and test case diffs could be used inside of `TestSummary`'s `Display` implementation.

This PR also deprecates the `--json` flags in favor of `--json-summary`.

<Details>

<Summary>tree-sitter-c json test report</Summary>

```json
{
  "parse_results": [
    {
      "name": "ambiguities",
      "children": [
        {
          "name": "pointer declarations vs expressions",
          "outcome": "Passed",
          "parse_rate": 2.772817603707442,
          "test_num": 1
        },
        {
          "name": "casts vs multiplications",
          "outcome": "Passed",
          "parse_rate": 1706.0606562536232,
          "test_num": 2
        },
        {
          "name": "function-like type macros vs function calls",
          "outcome": "Passed",
          "parse_rate": 3052.5877618981544,
          "test_num": 3
        },
        {
          "name": "function calls vs parenthesized declarators vs macro types",
          "outcome": "Passed",
          "parse_rate": 254.71817733995675,
          "test_num": 4
        },
        {
          "name": "Call expressions vs empty declarations w/ macros as types",
          "outcome": "Passed",
          "parse_rate": 1783.2222049930222,
          "test_num": 5
        },
        {
          "name": "Comments after for loops with ambiguities",
          "outcome": "Passed",
          "parse_rate": 33.21285186802364,
          "test_num": 6
        },
        {
          "name": "Top-level macro invocations",
          "outcome": "Passed",
          "parse_rate": 1488.5549782333942,
          "test_num": 7
        }
      ]
    },
    {
      "name": "crlf",
      "children": [
        {
          "name": "Line comments with escaped CRLF line endings",
          "outcome": "Passed",
[...]

PR #5018, opened 2025-11-01
fix(docs): update cli docs to reflect changes to various subcommand arguments

WillLillis:
Completely open to bikesheds, nitpicks, or other related docs improvements we can make here. This should be re-checked and merged after #4908 and #5017.

Partially addresses #5013

### DIFF
--- a/docs/src/cli/generate.md
+++ b/docs/src/cli/generate.md
@@ -44,7 +44,7 @@ Print the overview of states from the given rule. This is useful for debugging a
 item sets for all given states in a given rule. To solely view state count numbers for rules, pass in `-` for the rule argument.
 To view the overview of states for every rule, pass in `*` for the rule argument.
 
-### `--json`
+### `--json-summary`
 
 Report conflicts in a JSON format.
 
@@ -54,3 +54,7 @@ The path to the JavaScript runtime executable to use when generating the parser.
 Note that you can also set this with `TREE_SITTER_JS_RUNTIME`. Starting from version 0.26.0, you can
 also pass in `native` to use the native QuickJS runtime that comes bundled with the CLI. This avoids
 the dependency on a JavaScript runtime entirely.
+
+### `--disable-optimization`
+
+Disable optimizations when generating the parser. Currently, this only affects the merging of compatible parse states.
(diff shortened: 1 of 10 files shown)


docs/src/cli/generate.md
```diff
@@ -38,6 +38,10 @@ What generated files to emit. Possible values:
 - `parser` (default): Generate `parser.c` and related files.
 - `lib`: Compile to a library (equivalent of the deprecated `--build` option)
 
+### `-b/--build`
```
clason: I remember there being some doubt about this flag. I think this was motivated by split/multiple parser projects?

Would it not be better to have the CLI check and rebuild all parsers? It's a more complex but robust (and useful) solution, so it might make sense to not prematurely add this API, only to deprecate it later.

docs/src/cli/version.md
```diff
@@ -42,3 +42,11 @@ tree-sitter version
 ### `-p/--grammar-path <PATH>`
 
 The path to the directory containing the grammar.
+
+### `--bump`
```
clason, line 46: I'm not a fan of this feature at all, to be honest -- way too brittle, and it has _never_ worked correctly for me. It also relies on external tooling for some of the  bindings, which needs to be documented.

(Again, parser maintainers get these whether they want to or not -- in particular whether they use or know the language or not.)

docs/src/cli/generate.md
WillLillis: I thought that we deprecated --build when --emit=lib was added, so taking that out meant that we weren't deprecated. But if we still want to deprecate it in favor of the actual build command, I'm fine with that.

clason: Yeah, my point was to _still_ deprecate this, but in favor of something else. And deprecation means that for the release before removing it, you 
1. add a warning to its invocation (unless that is _too_ irritating in legitimate use)
2. remove the documentation, or at least note the deprecation and alternatives.

(There are different schools of thought on how to handle 2.; we should decide on one and apply it consistently.)

clason: But you are right: This deprecation would be an independent decision (and should be an independent discussion), so not blocking this PR (or the release).

docs/src/cli/version.md
clason: (Not a blocker for this PR, of course, which should document all existing flags. But now would be a great time to deprecate this _if_ we want to remove it ;))

PR #5019, opened 2025-11-02
fix(docs): remove multilingual config field

clason:
## Problem

"deploy docs" always pulls in the `latest` release of `mdbook`,
which now is a v0.5.0 prerelease with breaking changes -- including
removing an (apparently unused) `multilingual` config field in the TOML
that is now an error (another breaking change).

## Solution

Delete the line. Add `workflow_dispatch` to the docs workflow
in case follow-up changes are needed; see
https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide

Admonitions are now enabled by default, so we may no longer need to install `mdbook-admonish`?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,7 +26,7 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           jq_expr='.assets[] | select(.name | contains("x86_64-unknown-linux-gnu")) | .browser_download_url'
-          url=$(gh api repos/rust-lang/mdbook/releases/latest --jq "$jq_expr")
+          url=$(gh api repos/rust-lang/mdbook/releases/v0.4.52 --jq "$jq_expr")
           mkdir mdbook
           curl -sSL "$url" | tar -xz -C mdbook
           printf '%s/mdbook\n' "$PWD" >> "$GITHUB_PATH"


WillLillis:
Were you able to get `mdbook serve` to succeed locally? I ran into the following when testing out this fix:

```
 INFO Book building has started
[2025-11-02T20:44:53Z ERROR mdbook_admonish] Fatal error: Unable to parse the input
[2025-11-02T20:44:53Z ERROR mdbook_admonish]   - Unable to parse the input
[2025-11-02T20:44:53Z ERROR mdbook_admonish]   - invalid type: null, expected any valid TOML value at line 1 column 198
 WARN Error writing the RenderContext to the backend, Broken pipe (os error 32)
ERROR The "admonish" preprocessor exited unsuccessfully with exit status: 1 status
```

clason:
To be honest, I didn't even test -- hence the comment. 

Not surprised there are other breakages; let me also pin the version pre-0.5 (we shouldn't pull in random versions anyway).

WillLillis:
I may be misunderstanding, but it looks like `mdbook-admonish` has to be migrated to support `mdbook`'s refactoring into different crates (https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#rust-api). We definitely should just pin to the newest working version.

clason:
As far as I understand the changelog, `mdbook-admonish` is no longer needed. But in any case, my point is that we should migrate when _we_ choose to, not when upstream decides to put out a "prerelease".

(Remind you of anything? ;))

PR #5022, opened 2025-11-02
fix(ci): correct mdbook release url

WillLillis:
Followup to #5019


clason:
😬

clason (approved):

WillLillis:
@clason The trigger path for the doc workflow doesn’t include changes to the workflow file itself so it didn’t re-run after this got merged. I’ll be back at my computer in a few hours so we can “actually” fix it this time but if you want to beforehand feel free.

clason:
And that's exactly why I added the workflow dispatch trigger :) Already ran it, and it passed. Thanks!

PR #5024, opened 2025-11-03
fix(cli)!: deprecate `--build` flag for `generate` command

WillLillis:
This PR is meant to serve as a proposal/place of discussion around deprecating the `--build` flag for the `generate` command. 

Some points for removal:

- The `--build` flag was added in https://github.com/tree-sitter/tree-sitter/pull/2013 *before* the CLI had a separate `build` command. I'm not sure it still makes sense to support this when the functionality has a real place to "live".
- It doesn't offer feature parity with the `build` command, so we're effectively offering two slightly different ways to do the same thing.
- For commands like `parse`, `query`, etc., the CLI tool already "automagically" recompiles parsers when needed. This works fairly well for common use cases, and we offer the `--lib-path` argument for more advanced use cases. Point being, offering `--build` may be misleading to some end users.

I'd like to hear any arguments for keeping it, or potential alternatives the CLI could provide to fill a use case this currently covers.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -114,13 +114,13 @@ struct Generate {
     /// Only generate `grammar.json` and `node-types.json`
     #[arg(long)]
     pub no_parser: bool,
-    /// Compile all defined languages in the current dir
+    /// Deprecated: use the `build` command
     #[arg(long, short = 'b')]
     pub build: bool,
-    /// Compile a parser in debug mode
+    /// Deprecated: use the `build` command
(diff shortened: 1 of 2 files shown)


clason:
I think this
> It doesn't offer feature parity with the build command

is the clincher. We should deprecate this in 0.26 if possible (don't forget to mention this in the documentation!) to be removed in 0.27.

wetneb:
This makes intuitive sense to me. Replacing `--build` by `&& tree-sitter build` should generally not be too much hassle for users.

PR #5032, opened 2025-11-04
fix(cli)!: deprecate `version` command

WillLillis:
Here's another deprecation proposal: the `version`/`publish` subcommand. This command was originally added in https://github.com/tree-sitter/tree-sitter/pull/3786.

Points for removal:

- The functionality provided by this tool is flaky at best. (This issue has been raised several times by @clason).
- Relies on external tooling (cargo, npm, and potentially others in the future). These dependencies are currently undocumented and can lead to problems that are out of our control.
- Introduces unnecessary maintenance for (presumably) a small or zero amount of value to grammar authors.

Once removed, a better approach may be simply providing step-by-step documentation to grammar authors on how to bump their grammar versions instead of attempting to automate it through the CLI.

As an alternative, we could focus developer effort into improving the issues with the command, but I'm not sure how far that will go.

As with the other deprecation proposal, I'd like to hear any thoughts, use cases, concerns etc. regarding its removal.


wetneb:
I find the `version` command useful and rely on it for each grammar release that I publish.
Specifically, I follow the workflow proposed by @amaanq in #3786: use `tree-sitter version`, commit and add a tag, push, and let the CI deal with the packaging and publishing in various places.

What would the alternative step-by-step workflow look like?

Are there more details about the flakiness of this command? Depending on what the issues are, I could be motivated to lend a hand to resolve some of them.

clason:
The issue is about the bindings, which need to be bumped in parallel. (The grammar itself is trivial, you just bump the version in `tree-sitter.json`, which is simple enough to do by hand, and call `tree-sitter generate`, done.)

So the issue is that the command is overkill for the core job (updating the grammar itself) and unreliable for the (optional) bindings since it relies on external, partly unstable, and wholly undocumented external tooling. It simply fails for half of them, requiring me to hunt down and bump the version strings in the different files by hand. Much simpler (both for parser and -- more importantly -- tree-sitter maintenance) to simply grep for the old version string and edit the half dozen or so occurrences by hand.

wetneb:
Thanks for the details.
The issue about `tree-sitter version` depending on undocumented external tools is one that seems quite easy to solve, I would be happy to add this information to the docs if the command is kept.

The process of doing all updates by hand does not seem ideal to me, first because of the effort involved, but also because it involves editing lock files by hand, which should generally be avoided.

It seems that this command currently aborts as soon as any of the subprocesses fail, leaving a partially updated worktree behind. I imagine it could make sense to change this behavior, so that when one such command fails, the command goes on to update the other bindings. At the end it would display an error with the list of bindings files which could not be updated automatically, prompting the user to do those updates manually. This would still produce value over doing all of the updates manually, in my opinion. If the command is kept, I would be happy to propose a PR going in that direction if it's helpful.

clason:
> The issue about tree-sitter version depending on undocumented external tools is one that seems quite easy to solve, I would be happy to add this information to the docs if the command is kept.

The problem is that this information is not static and can bitrot quite quickly (Zig, I'm looking at you).

> The process of doing all updates by hand does not seem ideal to me, first because of the effort involved, but also because it involves editing lock files by hand, which should generally be avoided.

They're not lockfiles, though, but version manifest. Those _can_ be edited by hand.

> It seems that this command currently aborts as soon as any of the subprocesses fail, leaving a partially updated worktree behind.

Yes, that is one of the issues. The question (which is still open) is whether the (real but relatively minor) convenience benefit outweighs the increasing maintenance effort to fix _and keep fixed_ all the rough edges. This usually follows a 90-10 rule, where it's easy to get something that works some of the time for some people, and fixing each edge case that appears costs about the same effort as all the previous work together...

(And real talk: This project has already added so many moving parts that it has exceeded the current maintenance budget by a wide margin. If we don't batten down the hatches, I seriously worry about bitrot and broken windows killing it.)

So for me the reliance on an external unstable toolchain is the no-go here; if there's a reasonably simple, robust, and self-contained(!) implementation, we could give it a try. (For example, the command can simply spit out a diff to check and apply via patch.) But since the manifests are language dependent and may not be able to be parsed without language-specific tools, I am skeptical of the chances here.

ObserverOfTime:
> They're not lockfiles, though, but version manifest. Those can be edited by hand.

Some _are_ lockfiles and _cannot_ be edited by hand.

The zig lockfile is immutable so we could edit it through the command but cargo and _especially_ npm may expect version bumps for other dependencies.

WillLillis:
Updating the command to not exit on the first failure should be a pretty trivial change. Additionally, we could probably also display which bindings failed (if any).

These improvements still don't really address the main outstanding problems, though.

> For example, the command can simply spit out a diff to check and apply via patch.

Amaan and I spent some time trying to do something similar to this in order to clean up some of the `init -u` code paths, but ultimately we ran into too many issues for it to be a robust solution. I'd be open to simplifying the command instead of deprecating, but we'd need to nail down what that looks like.

jpt13653903:
I'm all for simplifying it down to the "core" version only (I can always just regenerate the bindings every time).

But please keep the automated bump feature - I like not having to think about it: I simply say `ts version --bump patch` and it's done.

WillLillis:
Thanks for the discussion everyone! We'll leave the `version` command un-deprecated for now, and start with improving it in #5050.

Issue #5037, opened 2025-11-06
Existing tests fail with emcc 4.0.19 on WSL


vladinator1000:
### Problem

Hey folks, I'm working on a reproduction for another bug and as I was doing the setup steps in the [contributing guide](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/CONTRIBUTING.md) I found that the existing tests fail to run using the latest version of Emscripten.

```
LinkError: WebAssembly.instantiate(): Import #0 module="env" function="__stack_pointer": imported mutable global must be a WebAssembly.Global object
```

Is there any reason for requiring a global Emscripten installation as the first order of preference [when building](https://github.com/tree-sitter/tree-sitter/blob/master/crates/xtask/src/build_wasm.rs#L70)? It's quite a pain to set up the project, very confusing and took a lot of setup time for a newbie contributor like me.

### Steps to reproduce

1. Install the latest emscripten following these instructions: https://emscripten.org/docs/getting_started/downloads.html

2. Follow the contributing setup guide in: https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/CONTRIBUTING.md

Once I installed Emscripten (which was a clusterfuck of an experience by itself), I ran these commands:

Went here
```
cd ./lib/binding_web
```

```
npm i
```

Then back to the project root
```
cd ../../
```

Then ran these from the contributing guide
```sh
cargo xtask fetch-fixtures
```

Optionally, to update the generated parser.c files:

```sh
cargo xtask generate-fixtures
```

Then you can build the Wasm modules:

```sh
cargo xtask generate-fixtures --wasm
```

then went back to the web binding 

```
cd ./lib/binding_web
```

and finally ran the tests
```
npm test
```

### Expected behavior

The project is easy to set up for new contributors and tests pass.

### Tree-sitter version (tree-sitter --version)

main branch

### Operating system/version

WSL

sogaiu:
I may be mistaken but ATM I think the recommended version of emscripten is [4.0.15](https://github.com/tree-sitter/tree-sitter/blob/13d4db8bb484064e5bfe0de22382c0e2bf391fd7/crates/loader/emscripten-version).

vladinator1000:
[...]

PR #5038, opened 2025-11-06
docs: Make the contributing guide easier to follow

vladinator1000:
Hi y'all, I spent quite a bit of time trying to get the project set up locally and I think it would have been easier if the steps in the guide were in this order. I had a lot of trouble setting up Emscripten locally when I realized that it actually doesn't work because of a [version mismatch](https://github.com/tree-sitter/tree-sitter/issues/5037) and I believe it would be easier for new contributors to use Docker/Podman instead.

The main idea in these changes is to order all the necessary steps sequentially, so that newbies like me can get started sooner and not have to deal with the horror of cloning and installing Emscripten.

### DIFF
--- a/lib/binding_web/CONTRIBUTING.md
+++ b/lib/binding_web/CONTRIBUTING.md
@@ -10,34 +10,54 @@ Contributors to Tree-sitter should abide by the [Contributor Covenant][covenant]
 
 To make changes to Web-tree-sitter, you should have:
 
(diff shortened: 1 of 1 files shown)


WillLillis:
Thanks for reporting the setup problem. The Emscripten version incompatibility is an issue, but this rework of the docs doesn’t resolve it. The build still prefers an installed emcc over Docker/Podman, and the new instructions also introduce an invalid path and make optional testing setup mandatory. I’m closing this in favor #5883.

PR #5039, opened 2025-11-06
Add node stability test

vladinator1000:
Hi y'all, the [node ID docs](https://github.com/tree-sitter/tree-sitter/blob/13d4db8bb484064e5bfe0de22382c0e2bf391fd7/lib/binding_web/src/node.ts#L47-L59) say that node IDs are supposed to be stable when reusing trees, but that seems not to be the case. I'm not sure how to fix it yet, but I've added a test that reproduces the issue.

It starts with this JS fixture:
1. 
```js
let name = "John Doe";
```
2. gets the ID of the variable declaration node.
3. Calls `tree.edit` and `parser.parse` using the tree
4. Adds another variable declaration and gets their nodes
```js
let age = 31;
let name = "John Doe";
```

5. Checks that the ID we got in step 2 exists in the nodes

Is this expected behavior?  It doesn't seem to be in line with the incremental parsing goal of the library. If incremental parsing is a core tree-sitter principle, I believe supporting stable IDs can enable some really cool use-cases for users. Incremental renderers using ASTs for example.

Seems like other people have also encountered the issue: https://github.com/tree-sitter/tree-sitter/discussions/2553#discussioncomment-14887885

### DIFF
--- a/lib/binding_web/test/node-id-stability.test.ts
+++ b/lib/binding_web/test/node-id-stability.test.ts
@@ -24,9 +24,7 @@ describe("node id stability", () => {
   });
 
   it("node ID of existing variable_declarator is stable after inserting second declaration", () => {
-    const source1 = `
-      let name = "John Doe";
-    `;
+    const source1 = `let name = "John Doe";\n`;
     tree = parser.parse(source1);
 
     const variableDeclaratorsBefore = tree?.rootNode.descendantsOfType(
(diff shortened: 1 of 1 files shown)


WillLillis:
Thanks for the report, but there's no bug here. The documentation for node ids is:

https://github.com/tree-sitter/tree-sitter/blob/ac17c8d003764bfe1df03055be09570bd2a20425/lib/binding_web/src/node.ts#L48-L59

Note the important caveat:

```
A node not marked as having changes does not guarantee it was reused.
```

Issue #3276, opened 2024-04-10
Compile-time kind and field ids that are generated by binding generation. 


VonTum:
### Problem

So while it's possible to use `Language::id_for_node_kind` and `Language::field_id_for_name` to request ids for kinds and fields, these are runtime functions. This means that it is not possible to use these in compile-time contexts, such as in match arms. It would be cleaner to provide some macros or compile time constants to allow getting these values for these contexts, and in the process slightly improves efficiency in using tree-sitter. 

I was able to [build such proc_macros myself](https://github.com/pc2/sus-compiler/blob/330a4d028cdb196ea5bea92d41994b13464c8f90/sus_proc_macro/src/lib.rs) for my own language 'SUS', and they [work](https://github.com/pc2/sus-compiler/blob/d693030d63a76045fd6cabe47899a6eff90fbc98/src/flattening/mod.rs#L217-L234) [wonderfully](https://github.com/pc2/sus-compiler/blob/d693030d63a76045fd6cabe47899a6eff90fbc98/src/flattening/mod.rs#L663-L686). Perhaps it would be a good idea to incorporate this into mainline tree-sitter? 

### Expected behavior

It should be possible to write code like this:

```rust
use mylang::{kind,kw,field};

fn test_fn(cursor : &TreeCursor) {
  if cursor.field() == field!("sub_expr") {
  let node = cursor.node();
  match node.kind_id() {
    kind!("array_expr") => {},
    kind!("binary_expr") => {},
    kind!("unary_expr") => {},
    // etc
  }
  
}
```

So proc_macros such as these need to be part of a separate crate. During our [discussion](https://github.com/tree-sitter/tree-sitter/discussions/3269) on this issue, we found it difficult to find a good spot to place this new crate. 

Sadly it _has_ to be part of the generated code, since for the proc macros to work they need to have the user's Language as a dependency. Sadly again, we couldn't simply include such proc macros into the generated bindings crate itself, because rust makes a strict distinction between proc_macro and non-proc_macro crates. No other functions are allowed in non-proc_macro crates. 

That leaves the issue of where to put it. Following the directory structure, we'd need to put them in `tree-sitter-myLang/bindings/rust/proc/Cargo.toml`, but this is ugly, and brittle. Not to mention it requires having a dependency that goes up the file tree instead of down. 

We also discussed simply adding it on the root of the myLang folder, but that's cluttering up that folder further.

ObserverOfTime:
[...]

PR #5043, opened 2025-11-07
Include `symbol_ids` field in node-types.json

bglgwyng:
This PR addresses issue #3276.
I’ve tested this change with a small toy grammar, but it still needs further validation and review.
Please let me know if there are additional steps or improvements I should make.

# Notes
I separated [`enum ts_symbol_identifiers` value generation](https://github.com/tree-sitter/tree-sitter/blob/877782a8a41d830f9e7f0f22dfcb67b3c2e50418/crates/generate/src/render.rs#L424-L443) from [render.rs](https://github.com/tree-sitter/tree-sitter/blob/master/crates/generate/src/render.rs). Each numeric id of the symbols is generated in [generate_symbol_ids](https://github.com/bglgwyng/tree-sitter/blob/fc747bce5329833ddb6a59ff306c662450531160/crates/generate/src/introspect_grammar.rs#L89) and used both in the generation of `node-types.json` and `parser.c`. This implementation guarantees that both will use the same value for it.

# Questions
- I added `symbol_ids` field to each node types. In my case, it gives enough information to handle the retrieved AST, like [this](https://github.com/github/semantic/blob/main/semantic-json/src/Language/JSON/AST.hs#L107). Is it generally enough?
- I added `introspect_grammar` but it seems needlessly separated from `prepare_grammar`. What about merging `introspect_grammar` into `prepare_grammar`? Then `prepare_grammar` builds parsing tables also.
- To print proper parsing error message, which indicates which symbol is expected to come, Do we need to add `symbol_id` field in `grammar.json` to look for the corresponding symbol name of its id? This job is not done yet.

### DIFF
--- a/crates/cli/src/tests/node_test.rs
+++ b/crates/cli/src/tests/node_test.rs
@@ -1317,11 +1317,6 @@ fn test_cpp_parser_and_node_types_compatibility() {
 }
 
 #[test]
-fn test_embedded_template_parser_and_node_types_compatibility() {
-    test_parser_and_node_types_compatibility("embedded_template");
-}
-
-#[test]
 fn test_go_parser_and_node_types_compatibility() {
     test_parser_and_node_types_compatibility("go");
 }
--- a/crates/cli/src/tests/node_test.rs
+++ b/crates/cli/src/tests/node_test.rs
@@ -1275,11 +1275,18 @@ fn test_parser_and_node_types_compatibility(grammar_name: &str) {
 
     let mut symbol_ids_by_kind_from_language: HashMap<(String, bool), Vec<u16>> = HashMap::new();
     for i in 0..kind_count as u16 {
-        let kind = language.node_kind_for_id(i).unwrap().to_string();
-        let id = language.node_kind_is_named(i);
+        let kind = language.node_kind_for_id(i).unwrap();
+        let named = language.node_kind_is_named(i);
+
+        // workaround maybe?
+        let kind = if kind.is_empty() {
+            "\0".to_string()
+        } else {
+            kind.to_string()
+        };
 
         symbol_ids_by_kind_from_language
-            .entry((kind, id))
+            .entry((kind, named))
             .or_insert_with(Vec::new)
             .push(i);
     }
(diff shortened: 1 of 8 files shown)


m74278803-cmyk (approved):

VonTum:
Does this allow for compile-time access to node and field IDs in the Rust version? I don't think I see an attached crate with proc-macros. 

While C-style global `const`s would work, Rust enables a much more readable syntax with proc-macros (`kind!("<=")`). What is your plan for this PR?

bglgwyng:
@VonTum I didn't modify the rust codegen part in this PR. I think that those can be separated. Indeed, I'm writing a tree-sitter binding for Haskell, and I need node-types.json only.

WillLillis:
Fair warning that the primary motivation behind #3276 is to provide compile time known constants to the rust bindings (a la https://github.com/tree-sitter/tree-sitter/pull/5043#issuecomment-3508501313) and without a concrete implementation for that feature, this is unlikely to be accepted. @amaanq was the one primarily working on this roughly a year ago, so this PR will likely have to wait until he has time to review it with that in mind.

bglgwyng:
@WillLillis Thank you for your feedback! I think we can have two separate PRs: one for additional metadata generation (this PR) and another for the subsequent code generation. This PR focuses on the former, and I need to confirm that the metadata is all correct. Once the metadata is verified, we can move on to the latter. I've added [new tests]((https://github.com/tree-sitter/tree-sitter/blob/3bfbc00bc0cccf9a6a42a7bb6c3e5a7c9ae6cc7d/crates/cli/src/tests/node_test.rs#L1249) for it, but I'm not sure if I've implemented them properly, so I'm waiting for feedback from the maintainers who are familiar with tree-sitter.

bglgwyng:
I realized that what I added as `symbol_ids` here is indeed `grammar_symbol_ids`.
Grammar symbol id has N:1 relationship to kind in node-types.json.
Does symbol id have also N:1 relationship to kind, so that we need an array of it named `symbol_ids`. I thinks so, however I want to convince my self about it.

WillLillis:
This branch has accumulated conflicts, and doesn't fully address what's planned for #3276. Landing this half baked now would just lead to a rewrite of the implementation, and unnecessary breakage for downstream consumers.

PR #5044, opened 2025-11-10
Fix `ts_parser__do_parse` failure path to return `NULL` (C23- & C11-conformant)

takumi-earth:
## Summary

- `ts_parser__do_parse`’s imbalance branch returned the literal `false` even though the function promises `TSTree *`; `C11`/`GNU11` quietly coerce that to `NULL`, but `C23` raises `-Werror=int-conversion` and stops the build even without extra warning flags.
- Changing that branch to `return NULL;` removes the only observed `C23` blocker while keeping the runtime behavior identical for existing `C11`/`GNU11` consumers (failure still surfaces as a null tree pointer), so the code now compiles cleanly under both standards.

## Background

`tree-sitter v0.25.10` is written for `-std=c11`, yet GCC `15` will default to `-std=c23`. Under `C23` rules, returning `_Bool` from a pointer-returning function is a hard error. The imbalance branch in `ts_parser__do_parse` still returns `false`, so any project that compiles `tree-sitter` with `-std=c23`/`-std=gnu23`cannot build the parser even though the intent was always to `return NULL;` when balancing fails. This patch fixes the standards mismatch without altering runtime semantics.

## Discovery

While building `tree-sitter v0.25.10` as part of [`codex-rs`](https://github.com/openai/codex/tree/main/codex-rs) (`v0.57.0`) on Ubuntu `24.04.3 LTS` with GCC `14.2.0` and `CFLAGS="-std=c23"`/`CFLAGS="-std=gnu23"`, the workspace build failed when the `tree-sitter` crate invoked `cc` to compile `lib/src/parser.c`. GCC reported `error: returning ‘_Bool’ from a function with return type ‘TSTree *’ makes pointer from integer without a cast [-Werror=int-conversion]` at the `balance:` label in `ts_parser__do_parse`. No other diagnostics were produced, identifying this single line as the sole `C23` blocker.

## Reproduction

1. Environment: Ubuntu `24.04.3 LTS`, GCC `14.2.0`, `tree-sitter v0.25.10`.
2. In any project that builds the C library (e.g., `cargo build -p codex-core` in [`codex-rs`](https://github.com/openai/codex/tree/main/codex-rs)), set `CFLAGS="-std=c23"`/`CFLAGS="-std=gnu23"`.
3. Start the build; the `cc` step that compiles `lib/src/parser.c` fails with the `_Bool`-to-pointer error at [`lib/src/parser.c#L2217`](https://github.com/tree-sitter/tree-sitter/blob/v0.25.10/lib/src/parser.c#L2217).
4. Re-run with `-std=c11` to see the line emit only a warning, explaining why existing pipelines still pass under older standards.

## Changes

- `lib/src/parser.c`: in the imbalance bail-out after `ts_parser__balance_subtree(self)` fails, emit `return NULL;` instead of `return false;`. This aligns the literal with the `TSTree *` return contract while preserving the existing “NULL means parse failure” semantics. (The diff is localized to [`lib/src/parser.c#L2217`](https://github.com/tree-sitter/tree-sitter/blob/v0.25.10/lib/src/parser.c#L2217) in `v0.25.10`.)
- `.github/workflows/ci.yml`: add a `c23-build` job on Ubuntu `24.04.2 LTS` that installs `gcc-14`, sets `CFLAGS=-std=c23`, and runs `cargo build --release`, `cargo xtask fetch-fixtures`, `cargo xtask generate-fixtures`, and `cargo xtask test` so future regressions surface automatically during CI.

## Validation

- `cargo build --release`
- `cargo xtask fetch-fixtures`
- `cargo xtask generate-fixtures`
- `cargo xtask test`
- `CFLAGS="-std=c11" cargo build -p tree-sitter --release`
- `CFLAGS="-std=c23" cargo build -p tree-sitter --release`
- `CFLAGS="-std=gnu23" cargo build -p tree-sitter --release`  

All commands complete successfully, and both `CFLAGS="-std=c23"` and `CFLAGS="-std=gnu23"` builds now produce the same binaries as the default `-std=c11` configuration. The new `c23-build-check` CI job runs this sequence on every PR to catch regressions automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,3 +44,28 @@ jobs:
 
   build:
     uses: ./.github/workflows/build.yml
+
+  c23-build-check:
+    runs-on: ubuntu-24.04
+    env:
+      CC: gcc-14
+      CXX: g++-14
+      CFLAGS: -std=c23
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install GCC 14
+        run: sudo apt-get update && sudo apt-get install -y gcc-14 g++-14
+
+      - name: Set up stable Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Build and test under C23
+        run: |
+          cargo build --release
+          cargo xtask fetch-fixtures
+          cargo xtask generate-fixtures
+          cargo xtask test


clason:
First question here is "_why_ do you set C23? We do not advertise support for that, nor do I see much need to do so.

(If the fix itself is backward-compatible, I'm not necessarily against it, but I draw the line at adding to our -- already extensive -- CI load with an unsupported build.)

clason:
That `false` seems like an unintentional inconsistency (@amaanq?), so that fix is welcome.

But please don't add an additional C23 workflow -- if we want to be more strict about validation, we should add more `-Werr` flags in our normal build.

(I am running this once just to see the output.)

clason:
Also, do not make a PR directly to the backport branch. All changes need to go through `master`.

takumi-earth:
Hi @clason, understood, apologies I had read through the contributing guide but it wasn't quite clear where I should submit the PR & if it should be specific to the fix or also include tests/CI tasks. If I submitted just the fix to the master branch would that be ok? Thank you.

clason:
No worries, and yes.

PR #5050, opened 2025-11-12
feat(cli): don't bail after first version update fails

WillLillis:
This addresses some of issues raised with the `version` command in https://github.com/tree-sitter/tree-sitter/pull/5032.

- Attempt all updates rather than exiting at the first failure. Display each failure via a error message
- Document the dependencies on cargo and npm.
- Extend the command to update `build.zig.zon`.

Further discussion is needed regarding how we can strip down/ improve the maintainability of this command rather than deprecate it.


clason:
Don't forget `zig`.

clason (approved):

PR #5052, opened 2025-11-12
fix(docs): don't show mdbook help popup when using query editor

Marcono1234:
Follow-up for #4550
That PR only fixed it for the Code editor, but not the Query editor. Typing a `?` there still opened the mdbook help dialog instead.


WillLillis:
Thanks!

PR #5059, opened 2025-11-14
fix(docs): Install tree-sitter-cli globally using npm

juanramirezc2:
I was following the instructions, and the `npm install tree-sitter-cli` command will install the tool in the current working directory (cwd) and not as a global dependency. I assume we want this to be globally available; if not, feel free to reject this PR :)


sogaiu:
> I assume we want this to be globally available

I don't think so, please see https://github.com/tree-sitter/tree-sitter/pull/4976.

PR #3188, opened 2024-03-17
fix(cli): keep default cc flags in build 

ObserverOfTime:
<details>

<summary>Complete commands from CI</summary>

#### release

```bash
i686-linux-gnu-gcc -O2 -ffunction-sections -fdata-sections -fPIC -m32 -march=i686 -std=c11 -I /project/test/fixtures/grammars/bash/src -Wall -DNDEBUG -Werror=implicit-function-declaration -shared /project/test/fixtures/grammars/bash/src/parser.c /project/test/fixtures/grammars/bash/src/scanner.c -o /project/parser.so
```

#### debug

```bash
i686-linux-gnu-gcc -O0 -ffunction-sections -fdata-sections -fPIC -gdwarf-4 -fno-omit-frame-pointer -m32 -march=i686 -std=c11 -I /project/test/fixtures/grammars/bash/src -Wall -Wextra -Werror=implicit-function-declaration -shared /project/test/fixtures/grammars/bash/src/parser.c /project/test/fixtures/grammars/bash/src/scanner.c -o /project/parser.so
```

#### release

```bash
powerpc64-linux-gnu-gcc -O2 -ffunction-sections -fdata-sections -fPIC -m64 -std=c11 -I /project/test/fixtures/grammars/bash/src -Wall -DNDEBUG -Werror=implicit-function-declaration -shared /project/test/fixtures/grammars/bash/src/parser.c /project/test/fixtures/grammars/bash/src/scanner.c -o /project/parser.so
```

#### debug

```bash
powerpc64-linux-gnu-gcc -O0 -ffunction-sections -fdata-sections -fPIC -gdwarf-4 -fno-omit-frame-pointer -m64 -std=c11 -I /project/test/fixtures/grammars/bash/src -Wall -Wextra -Werror=implicit-function-declaration -shared /project/test/fixtures/grammars/bash/src/parser.c /project/test/fixtures/grammars/bash/src/scanner.c -o /project/parser.so
```

#### release

```bash
arm-linux-gnueabi-gcc -O2 -ffunction-sections -fdata-sections -fPIC -march=armv6 -marm -mfloat-abi=soft -std=c11 -I /project/test/fixtures/grammars/bash/src -Wall -DNDEBUG -Werror=implicit-function-declaration -shared /project/test/fixtures/grammars/bash/src/parser.c /project/test/fixtures/grammars/bash/src/scanner.c -o /project/parser.so
```

#### debug

```bash
[...]

PR #5068, opened 2025-11-17
fix(loader): set correct runtime host for cc while cross-compiling

betonmischer86:
Pass the BUILD_TARGET variable from the build environment as 'host' for the cc crate. Otherwise, when cross-compiled, cc will keep looking for a cross-compiler instead of the native one on the target system. The initial BUILD_TARGET represents a runtime host after all. The change will have no effect outside of the case of cross-compiling tree-sitter's CLI tool.

Fixes #5067.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -557,7 +557,6 @@ impl Config {
 }
 
 const BUILD_TARGET: &str = env!("BUILD_TARGET");
-const BUILD_HOST: &str = env!("BUILD_HOST");
 
 pub struct LanguageConfiguration<'a> {
     pub scope: Option<String>,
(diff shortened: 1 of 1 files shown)


clason:
> The change will have no effect outside of the case of cross-compiling tree-sitter's CLI tool.

Turns out this was a lie ;)

(Errors are lint errors.)

betonmischer86:
> > The change will have no effect outside of the case of cross-compiling tree-sitter's CLI tool.
> 
> Turns out this was a lie ;)
> 
> (Errors are lint errors.)

Since `BUILD_HOST` was left unused, I removed it. The warning should be gone now.

meroamir99881-hue:
مطعم وسناك الداعي
يرحب بكم ابي سمراء شارع روضة الزيتون
70424506
06424506

في الاثنين، 17 نوفمبر 2025 6:04 م Valeriy Kosikhin ***@***.***>
كتب:

clason:
So I did a bit of spelunking, and it turns out that this is exactly how it was originally implemented before [being changed here](https://github.com/tree-sitter/tree-sitter/pull/3188/commits/59336fab67289968fd4a8f8e5332393a4a4eccdb) in an apparently unrelated PR without any explanation that I can see.

PR #5076, opened 2025-11-20
fix(docs): final updates before 0.26.1

WillLillis:
This PR aims to address the docs task in #5013 

- Indicate where xtask looks for wasi-sdk
- Mark native runtime as experimental, describe limitations

- docs were updated to reflect deprecations and flags in #5018 
- [x] What about ABI 13 needs to be described, and where?

### DIFF
--- a/docs/src/6-contributing.md
+++ b/docs/src/6-contributing.md
@@ -83,6 +83,18 @@ cargo xtask generate-fixtures --wasm
 cargo xtask test-wasm
 ```
 
+#### Wasm Stdlib
+
+The tree-sitter Wasm stdlib can be built via xtask:
+
+```sh
+cargo xtask build-wasm-stdlib
+```
+
+This command looks for the [Wasi SDK][wasi_sdk] indicated by the `TREE_SITTER_WASI_SDK_PATH`
+environment variable. If you don't have the binary, it can be downloaded from wasi-sdk's [releases][wasi-sdk-releases]
+page.
+
 ### Debugging
 
 The test script has a number of useful flags. You can list them all by running `cargo xtask test -h`.
@@ -220,4 +232,6 @@ and the tree-sitter module is fetched from [here][js url]. This, along with the 
 [pypi]: https://pypi.org
 [rust]: https://rustup.rs
 [ts repo]: https://github.com/tree-sitter/tree-sitter
+[wasi_sdk]: https://github.com/WebAssembly/wasi-sdk
+[wasi-sdk-releases]: https://github.com/WebAssembly/wasi-sdk/releases
 [web-ts]: https://www.npmjs.com/package/web-tree-sitter
--- a/docs/src/cli/build.md
+++ b/docs/src/cli/build.md
@@ -18,7 +18,9 @@ will attempt to build the parser in the current working directory.
 
 ### `-w/--wasm`
 
-Compile the parser as a Wasm module.
+Compile the parser as a Wasm module. This command looks for the [Wasi SDK][wasi_sdk] indicated by the `TREE_SITTER_WASI_SDK_PATH`
+environment variable. If you don't have the binary, the CLI will attempt to download it for you to `<CACHE_DIR>/tree-sitter/wasi-sdk/`, where
+`<CACHE_DIR>` is resolved according to the [XDG base directory][XDG] or Window's [Known_Folder_Locations][Known_Folder].
 
 ### `-o/--output`
 
@@ -36,3 +38,7 @@ in the external scanner does so using their allocator.
 ### `-0/--debug`
 
 Compile the parser with debug flags enabled. This is useful when debugging issues that require a debugger like `gdb` or `lldb`.
+
+[Known_Folder]: https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid
+[wasi_sdk]: https://github.com/WebAssembly/wasi-sdk
+[XDG]: https://specifications.freedesktop.org/basedir/latest/
--- a/docs/src/using-parsers/7-abi-versions.md
+++ b/docs/src/using-parsers/7-abi-versions.md
@@ -1,8 +1,9 @@
 # ABI versions
 
-Parsers generated with tree-sitter have an associated ABI version. This version establishes hard compatibility boundaries between the generated parser and the tree-sitter library.
+Parsers generated with tree-sitter have an associated ABI version, which establishes hard compatibility boundaries
+between the generated parser and the tree-sitter library.
 
-A given version of the tree-sitter library is only able to load parsers which have certain ABI versions:
+A given version of the tree-sitter library is only able to load parsers generated with supported ABI versions:
 
 | tree-sitter version | Min parser ABI version | Max parser ABI version |
 |---------------------|------------------------|------------------------|
@@ -14,7 +15,8 @@ A given version of the tree-sitter library is only able to load parsers which ha
 | >=0.20.3, <=0.24    | 13                     | 14                     |
 | >=0.25              | 13                     | 15                     |
 
-By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
-```
-tree-sitter generate --abi=<DESIRED-ABI>
-```
+By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version, but an older ABI (supported by the CLI) can be selected by passing the [`--abi` option][abi_option] to the `generate` command.
+
+Note that the ABI version range supported by the CLI can be smaller than for the library: When a new ABI version is released, older versions will be phased out over a deprecation period, which starts with no longer being able to generate parsers with the oldest ABI version.
+
+[abi_option]: ../cli/generate.md#--abi-version
--- a/docs/src/cli/generate.md
+++ b/docs/src/cli/generate.md
@@ -52,8 +52,9 @@ Report conflicts in a JSON format.
 
 The path to the JavaScript runtime executable to use when generating the parser. The default is `node`.
 Note that you can also set this with `TREE_SITTER_JS_RUNTIME`. Starting from version 0.26.0, you can
-also pass in `native` to use the native QuickJS runtime that comes bundled with the CLI. This avoids
-the dependency on a JavaScript runtime entirely.
+also pass in `native` to use the experimental native QuickJS runtime that comes bundled with the CLI.
+This avoids the dependency on a JavaScript runtime entirely. The native QuickJS runtime is compatible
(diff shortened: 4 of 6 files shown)


clason:
> What about ABI 13 needs to be described, and where?

`tree-sitter generate` will no longer create ABI 13 parsers, while the lib still supports them. This is part of a deprecation strategy to be able to eventually (TBD when) drop support for ABI 13 entirely.

docs/src/6-contributing.md
```diff
@@ -82,6 +82,17 @@ Similarly, to test the Wasm binding, you need to compile these parsers to Wasm:
...
+
+The tree-sitter Wasm stdlib can be built via xtask:
+
+```sh
+cargo xtask build-wasm-stdlib
+```
+
+This command looks for the clang wasi-sdk binary indicated by the `TREE_SITTER_WASI_SDK_PATH`
+environment variable. If you don't have the binary, it can be downloaded from wasi-sdk's [releases][wasi-sdk-releases]
+page.
```
clason, line 96: I was referring to `tree-sitter build --wasm` silently downloading and using `wasi-sdk` in favor of emscripten.

docs/src/cli/build.md
```diff
@@ -18,7 +18,8 @@ will attempt to build the parser in the current working directory.
 
 ### `-w/--wasm`
 
-Compile the parser as a Wasm module.
+Compile the parser as a Wasm module. This command looks for the clang wasi-sdk binary indicated by the `TREE_SITTER_WASI_SDK_PATH`
```
clason: I think it's not just the binary, but the full SDK?

docs/src/cli/build.md
```diff
@@ -18,7 +18,8 @@ will attempt to build the parser in the current working directory.
...
+environment variable. If you don't have the binary, the CLI will attempt to download it for you to `<CACHE_DIR>/tree-sitter/wasi-sdk/`.
```
clason: where is `<CACHE_DIR>` set (i.e., what does _tree-sitter_ think this directory is)?

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -14,7 +15,13 @@ A given version of the tree-sitter library is only able to load parsers which ha
...
-By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
+By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can
+specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option
+to the `generate` command:
+
 ```
 tree-sitter generate --abi=<DESIRED-ABI>
 ```
+
+Note that while the CLI doesn't support generation for ABIs prior to 14, existing ABI 13 parsers are still supported by the
```
clason: Hmm, that feels like not the right place for this, since this section is related to the lib, not the CLI.

Maybe we can add this information to the `generate` documentation? Doesn't need to be a table, just noting the "min" and "max/default" versions. (I worry a bit about this bitrotting, but it'll probably caught quickly and be easy to fix.)
Then we can link this here (and remove the parenthetical remark in the previous paragraph)?

clason: That doesn't address the deprecation strategy, though -- which to be fair we have not discussed and decided yet.

Maybe we can just write something like "when a new ABI version is released, older versions can be phased out, starting with no longer being able to generate parsers with the oldest ABI".

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -14,7 +15,13 @@ A given version of the tree-sitter library is only able to load parsers which ha
...
 | >=0.25              | 13                     | 15                     |
 
-By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
+By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can
+specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option
+to the `generate` command:
+
 ```
 tree-sitter generate --abi=<DESIRED-ABI>
 ```
```
clason: I see that the ABI constraints for generating are already mentioned at https://tree-sitter.github.io/tree-sitter/cli/generate.html#--abi-version, so I think it's enough to link to this section from here and remove the parenthetical remark.

docs/src/6-contributing.md
clason: I could test this, but would the `xtask` also download the sdk if it can't find it? If so, that would be good to mention.

WillLillis: `xtask` doesn't download the sdk if it can't find it. Looking at the code a little closer now, the treatment of the wasi binary in `xtask` and `build --wasm` has diverged a bit. It wouldn't be too hard to bring the `xtask` up to par with `build --wasm`, but since that's a developer tool I don't think that should block the release. I can put a PR up to do that and update the contributing docs accordingly.

docs/src/cli/build.md
WillLillis: It only looks for the clang binary:

https://github.com/tree-sitter/tree-sitter/blob/120f74723e694be4dc2f0033e01b24350dd73f19/crates/loader/src/loader.rs#L1376-L1402

I assume the binary is standalone, but maybe it depends on other files from the sdk at runtime? I can try and test this out a bit later.

docs/src/using-parsers/7-abi-versions.md
WillLillis: That makes sense, yeah. I think it's still valuable to give the concrete "can generate 14, library supports 13" info here though.

docs/src/cli/build.md
WillLillis: We use [etcetera](https://github.com/lunacookies/etcetera) to resolve the cache directory here, which currently links to a [404'd page](https://specifications.freedesktop.org/basedir/basedir-spec-latest.html) for their base strategy documentation. :laughing: (It looks like the new page is [here](https://specifications.freedesktop.org/basedir/latest/))

I think this should be 
    - XDG: `~/.cache`
    - Apple: `~/Library/Caches`
    - Windows: `~\AppData\Local\temp` (should be the same as `%temp%`)

Do you think it would be too much to enumerate the options as above? I think it makes sense to link the spec at least.

clason: Link is fine!

WillLillis: It should actually be `~/.cache` on both macos and linux since we're using `base_strategy`. I'll add the appropriate links.

WillLillis:
> That doesn't address the deprecation strategy, though -- which to be fair we have not discussed and decided yet.
> 
> Maybe we can just write something like "when a new ABI version is released, older versions can be phased out, starting with no longer being able to generate parsers with the oldest ABI".

Not sure if my last comment got lost, but I'll re-make it here. I think that makes sense, but I still think it's valuable to have the "can only generate >=14, library supports >=13" information in there. Let me know what you think! :)

clason:
I saw it but haven't formed an opinion yet ;)

My main thought was that having these numbers in one place is easier to maintain than in two places :)

WillLillis:
> I saw it but haven't formed an opinion yet ;)
> 
> My main thought was that having these numbers in one place is easier to maintain than in two places :)

Agreed yeah, although unfortunately I don't think the ABI 13 library support is mentioned anywhere else. Maybe we should just keep that bit?

clason:
It's mentioned in the new ABI page?

WillLillis:
> It's mentioned in the new ABI page?

Ope, oops you're right. I'm good with removing it then.

docs/src/6-contributing.md
clason: Yeah, out of scope for now. I just wanted to ask to make sure, since I don't understand the web build at all.

docs/src/cli/build.md
clason: Nope, needs the full sdk including headers. (Probably not all of it, but enough to make sense to just say "SDK" here.)

Maybe also add a link to https://github.com/WebAssembly/wasi-sdk?

docs/src/using-parsers/7-abi-versions.md
```diff
@@ -14,7 +15,11 @@ A given version of the tree-sitter library is only able to load parsers which ha
 | >=0.20.3, <=0.24    | 13                     | 14                     |
 | >=0.25              | 13                     | 15                     |
 
-By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the `--abi` option to the `generate` command:
-```
-tree-sitter generate --abi=<DESIRED-ABI>
-```
+By default, the tree-sitter CLI will generate parsers using the latest available ABI for that version. Grammar authors can
+specify an older ABI (within the constraints _of the CLI_, which may be stricter than the library!) via the [`--abi` option][abi_option]
```
clason:
```suggestion
specify an older ABI via the [`--abi` option][abi_option]
```
I don't think that callout is needed with the link.

sogaiu: This feels like a decent thing to bring up even if it might be mentioned elsewhere (or will be).

I didn't clearly realize this until recently.

clason: It's already mentioned explicitly (with the exact constraints) in the linked section, as well as in the next paragraph (as part of the deprecation strategy). Since I'm not happy with the tone of that remark (which I added), I'd prefer to remove it.

sogaiu: Ok, but if the tone is the main issue, can't the `_`s and `!` be removed to remedy that?

clason:
Ran across something else: `ts test --wasm` (and friends) requires `tree-sitter` to be built with the `wasm` feature; otherwise it (silently!) does nothing...

docs/src/using-parsers/7-abi-versions.md
WillLillis: I went ahead and removed the callout, I think the necessary info is summarized well in other places Clason listed.

WillLillis:
> Ran across something else: `ts test --wasm` (and friends) requires `tree-sitter` to be built with the `wasm` feature; otherwise it (silently!) does nothing...

Good catch! https://github.com/tree-sitter/tree-sitter/pull/5088

clason:
Could still add a short `(if tree-sitter is build with the `wasm` feature)` to the docs?

WillLillis:
> Still needs the `--wasm` caveat (which I can add later), but here's a pre-approval to unblock.

Forgot to include that with the last push, should have that up sometime Monday!

PR #5077, opened 2025-11-20
docs: fix typo in the page about ABI version

wetneb:
Of course I only catch that once they are already published…


clason (approved):

clason:
The PR didn't get squashed, so it doesn't matter.

PR #5083, opened 2025-11-22
build(deps): bump wasi-sdk to v29

clason:
This may be a bit dangerous, since I don't think the wasi-sdk codepath is exercised in CI? Seems to work fine locally, though (`ts t --wasm`).

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1441,9 +1441,9 @@ impl Loader {
             return Err(LoaderError::WasiSDKPlatform);
         };
 
-        let sdk_filename = format!("wasi-sdk-25.0-{arch_os}.tar.gz");
+        let sdk_filename = format!("wasi-sdk-29.0-{arch_os}.tar.gz");
         let sdk_url = format!(
-            "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/{sdk_filename}",
+            "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-29/{sdk_filename}",
         );
 
         info!("Downloading wasi-sdk from {sdk_url}...");


clason:
`ts_query_ls` SIGSEGVs with wasm parsers from wasi-sdk v29, but turns out that happens already with v25. (Tested with tree-sitter-query: `ts b --wasm && ts_query_ls check queries/`.)

And also with the release artifacts (built with emscripten).

@ribru17

clason:
(Also, not exactly happy about having yet another dependency hardcoded as a string in some source file...)

(But please, no xtask!)

ribru17:
> `ts_query_ls` SIGSEGVs with wasm parsers from wasi-sdk v29, but turns out that happens already with v25. (Tested with tree-sitter-query: `ts b --wasm && ts_query_ls check queries/`.)
> 
> And also with the release artifacts (built with emscripten).
> 
> @ribru17

Is this only on `master`? I'm running `ts b --wasm` with `0.25.10` without issues with the lua parser

(might be dependent on the parser)

clason:
Happens also with `tree-sitter-lua` for me. May be platform-specific? (macos-arm here)

ribru17:
It wouldn't depend on emscripten version, would it? (wasm ecosystem confuses me a lot, i'll be honest)

clason:
No, because it happens without emscripten (wasi-sdk, which is how I caught it ;))

ribru17:
But doesn't `ts b --wasm` use emscripten to compile? I assumed that wasi-sdk is for consuming a wasm blob that's already been produced (by e.g. emscripten)

clason:
> But doesn't ts b --wasm use emscripten to compile? I assumed that wasi-sdk is for consuming a wasm blob that's already been produced (by e.g. emscripten)

Nope, it's exactly for `ts b --wasm` (on `master`/0.26; I forgot whether it's only downloaded and used if `emscripten` is not found).

ribru17:
Ah, ok ignore my input that it was working then. I used the emscripten-created wasm blob

clason:
Me, too -- it happens no matter how the blob is created.

clason:
Hmmm, turns out `ts t --wasm` doesn't really work, either -- it just builds and uses the `dylib`...

So this is starting to look more like a tree-sitter issue. (Not surprised; "optional" features bitrot quickly.)

ObserverOfTime:
I think `ts t --wasm`, like `ts p --wasm`, doesn't actually compile and use a Wasm parser. :)

clason:
It does, if you have compiled tree-sitter with `--features=wasm`.

ObserverOfTime:
It's broken in 0.25 but works in 0.26.

PR #5085, opened 2025-11-23
fix(query): prevent infinite loop with `+` and `?` quantifiers

ribru17:
**Problem:** A query with a `?` quantifier followed by a `+` quantifier would hang at 100% CPU usage while iterating through a tree, regardless of the source content.

**Solution:** Collect all quantifiers in one step, and then add the required repeat/optional step logic *after* we have determined the composite quantifier we need to use for the current step.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -5032,6 +5032,26 @@ fn test_query_quantified_captures() {
                 ("comment.documentation", "// quuz"),
             ],
         },
+        Row {
+            description: "multiple quantifiers should not hang query parsing",
+            language: get_language("c"),
+            code: indoc! {"
+            // foo
+            // bar
+            // baz
+        "},
+            pattern: r"
+                ((comment) ?+ @comment)
+            ",
+            // This should be identical to the `*` quantifier.
+            captures: &[
+                ("comment", "// foo"),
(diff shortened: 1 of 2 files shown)


clason:
Can you give an example of such a pattern (preferably one found in the wild)?

clason:
Running nvim-treesitter's `check-queries.lua` with Neovim built with this branch, I get a new "impossible pattern" warning from the haskell (of course) highlights:
```
(decl
  [
    name: (variable) @function
    names: (binding_list
      (variable) @function)
  ])
```

Looks like this broke grouping in some way? (Other queries seem to parse fine.)

clason:
Never mind, that error already happens on  `master`.

clason:
Bisected (again) to https://github.com/tree-sitter/tree-sitter/pull/4894/commits/87d5c5c1715423c993dfe18d9d8fcead593b7e01

ribru17:
I feel like that commit should be reverted

ribru17:
> Can you give an example of such a pattern (preferably one found in the wild)?

At least in nvim-treesitter, there are none in that wild, but the test case i added includes a very simple faulty pattern

clason:
Nvim-treesitter tests pass, so I'd be in favor of merging this and test-driving this for a bit before release.

PR #5092, opened 2025-11-26
feat(xtask): bring wasi-sdk treatment up to par with the loader

WillLillis:
The loader package's `ensure_wasi_sdk_exists` private method checks for the wasi-sdk, fetching it if it can't be found. This logic was re-implemented in xtask for `build-wasm-stdlib`, but without the fetching functionality. We can have nice things in xtask too! Rather than make this function a public member of `tree-sitter-loader`, we just re-implement and leave a nice comment asking people to keep the two in sync.

Follow up to  https://github.com/tree-sitter/tree-sitter/pull/5076#discussion_r2551000356 (CC @clason)


clason:
The main issue for me is that this also duplicates the pinned version information. Is there some way of factoring this out (if necessary, to a separate file like the emscripten version)?

crates/loader/src/loader.rs
```diff
@@ -1436,9 +1438,9 @@ impl Loader {
             return Err(LoaderError::WasiSDKPlatform);
         };
 
-        let sdk_filename = format!("wasi-sdk-29.0-{arch_os}.tar.gz");
+        let sdk_filename = format!("wasi-sdk-{WASI_SDK_MAJOR_VERSION}.0-{arch_os}.tar.gz");
```
clason: Theoretically, we might want to bump to a minor/patch version at some point, so it'd be better to have the full version (`29.0`) in the external file and extract the major version from that for URL.

Practically, that has never happened (wasm isn't like the other kids...) so not a blocker; we can cross that bridge when we come to it. Just wanted to note this down.

WillLillis: Yeah, I think it makes sense to just account for this while it's still fresh in mind.

PR #5097, opened 2025-11-30
fix(cli): infinite loop in highlight test

### LINKED ISSUE #5645: tree-sitter test hangs instead of failing on a mismatched highlight assertion
### AI Policy

### Problem

// Disclaimer: I used AI to do the detective work for this issue - but these meaty fingers did type this myself including all the typos

I'm a rank amateur tree-sitter consumer who is vibing away with my project [tree-sitter-irules](https://github.com/dekobon/tree-sitter-irules). I think I've hit a real bug and I want to share it. I've logged the issue on my project as [#32](https://github.com/dekobon/tree-sitter-irules/issues/32).

When a highlight test (I think this verifies that a grammar's syntax is highlighting correctly) contains a `^` or `<`- assertion that fails in one particular configuration, the test runner doesn't report a clean "expected X, got Y" failure.

What it does instead, is that it gets stuck appending to a list forever and the memory for that list keeps doubling. Eventually, an allocation request (with an error like "memory allocation of 4294967296 bytes failed") aborts the process. When running without a memory limit Linux just OOM kills it.

This logic is in [crates/cli/src/test_highlight.rs](https://github.com/tree-sitter/tree-sitter/blob/master/crates/cli/src/test_highlight.rs#L147) and I believe the runaway allocation is [here on line 169](https://github.com/tree-sitter/tree-sitter/blob/master/crates/cli/src/test_highlight.rs#L169).

The good news is this bug seems to only affect people writing tests and building tree-sitter grammars and not the users of tree-sitter grammars.

I can provide a PR with a fix if someone tells me this isn't stupid. Or maybe I will do it anyways to help someone else who stumbles on this. Anyways, thank you for your work on this project!

### Steps to reproduce

```sh
git clone --depth=1 https://github.com/tree-sitter/tree-sitter-json
cd tree-sitter-json
tree-sitter --version
# outputs: tree-sitter 0.26.9

tree-sitter generate

# Now, let's make a highlight test with a comment that fails as a assertion:
mkdir -p test/highlight
echo '// x\n// ^x\n' > test/highlight/repro.json

tree-sitter test
# OOMed: [1]    2567855 killed     tree-sitter test

# expected: a normal failure message (expected highlight 'x', actual highlights: 'comment') and a non-zero exit
# actual: the runner consumes all the memory and dies or gets OOMed
```

Cursed test file content:
```
$ cat test/highlight/repro.json
// x
// ^x
```

### Expected behavior

A normal failure message (expected highlight 'x', actual highlights: 'comment') and a non-zero exit. Definitely, not use all the memory and die

tree-sitter 0.26.9

### Operating system/version

Linux 6.8.0-117-generic #117~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC aarch64 aarch64 aarch64 GNU/Linux

jannschu:
Hi,

I noticed an infinite loop bug when testing highlights using the test CLI. I added a test case, `test_assertion_with_non_matching_highlight_at_same_position`, that shows the bug if run on master.

I fixed this by changing these nested loops:

https://github.com/tree-sitter/tree-sitter/blob/df8b62fc509d45e2ebbfa1574a9818beca19eec5/crates/cli/src/test_highlight.rs#L151-L181

The inner while loop could terminate regularly, in which case `i` was not updated.

- Closes #5645

### DIFF
--- a/crates/cli/src/test_highlight.rs
+++ b/crates/cli/src/test_highlight.rs
@@ -13,10 +13,10 @@ use crate::{
 
 #[derive(Debug)]
 pub struct Failure {
-    row: usize,
-    column: usize,
-    expected_highlight: String,
-    actual_highlights: Vec<String>,
+    pub(crate) row: usize,
+    pub(crate) column: usize,
+    pub(crate) expected_highlight: String,
+    pub(crate) actual_highlights: Vec<String>,
 }
 
 impl std::error::Error for Failure {}
@@ -122,6 +122,7 @@ pub fn test_highlights(
 
     if failed { Err(anyhow!("")) } else { Ok(()) }
 }
+
 pub fn iterate_assertions(
     assertions: &[Assertion],
     highlights: &[(Utf8Point, Utf8Point, Highlight)],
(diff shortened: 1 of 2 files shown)


jannschu:
Oh, I also added a `!` before the `expected_highlight` if the assertion is supposed to be negative. This gives a better error message I hope. I can also this split this up into a separate PR if you like.

jannschu:
Further using this, I also noticed that this line's logic is wrong:

https://github.com/tree-sitter/tree-sitter/blob/df8b62fc509d45e2ebbfa1574a9818beca19eec5/crates/cli/src/test_highlight.rs#L162

I will address this in an upcoming PR. Another PR will address the missing checks for the assertion's length condition, which is currently only used for early exits.

Is there anything I can do about the CI failure?

clason:
No, that's just a network flake. (Those tests are irrelevant to the highlight crate.)

WillLillis (approved):

PR #4925, opened 2025-10-04
refactor!: replace anyhow with thiserror

WillLillis:
This removes all usage of the anyhow crate from the project outside of the cli and xtask crates. In particular, the loader and config crates now return errors in the form of structured data via [thiserror](https://crates.io/crates/thiserror).

I'm opening this as a draft for now, as this shouldn't block the 0.26.1 release. This has a large set of user-facing changes (error messages are important!), so regardless of the release this shouldn't be rushed in without a thorough check over.

I also slightly reworked the usage of thiserror in the generate crate to include the relevant path (if available) when an IO error occurs. This is more verbose at call sites, but should be very helpful in both debugging and giving more context to end users.

- Closes #4997

PR #4919, opened 2025-10-02
Add "containing range" APIs to QueryCursor

maxbrunsfeld:
Follow-up to https://github.com/tree-sitter/tree-sitter/pull/2085

### Background

Running a query on a very large syntax tree is generally much faster if you provide a smaller *range* within which the query cursor should search. But the current `QueryCursor::set_byte_range` and `::set_point_range` APIs are now always guaranteed to speed things up by the same amount for all queries. Those existing methods have the semantics that a match should only be returned if some node in the match *intersects* the given range. In some cases, in order to find all of the relevant matches, it's still necessary to walk a large portion of the tree, including nodes arbitrarily distant from the given range.

This is a problem when editing very large files in code editors. For example, in https://github.com/zed-industries/zed/issues/4701, users are observing slowness in Zed's `enclosing_bracket_ranges` and `suggested_indents` queries when editing a large JSON document. These code paths are slow despite looking only for matches that contain the cursor's position.

### Solution

This PR introduces new range-based `QueryCursor` APIs: `set_containing_byte_range` and `set_containing_point_range`, which have a different range-filtering semantics. These allow you to search only for matches where *all* nodes are *fully contained* within the given range. These APIs are independent of the existing range-filtering methods, and can be used in conjunction with them, for more advanced filtering.

For example, if you wanted to search for any matches that *intersect* line 5000, as long as they are fully contained within lines 4500-5500, you could do this:

```rust
let mut query_cursor = QueryCursor::new();
query_cursor.set_point_range(Point::new(5000, 0)..Point::new(5001, 0));
query_cursor.set_containing_point_range(Point::new(4500, 0)..Point::new(5500, 0));
```

The benefit of these new APIs is that they make possible a more aggressive optimization inside the query cursor, where it will *definitively* not descend into any syntax nodes outside of the given range.

####  Todo

* [x] Implementation
* [x] Rust bindings
* [ ] Web bindings
* [ ] API docs
* [ ] Docs

PR #5100, opened 2025-12-02
Add "containing range" APIs to QueryCursor rebased

WillLillis:
This PR is just a rebased copy of the branch from #4919.

Notable changes:
- Drop [847737b](https://github.com/tree-sitter/tree-sitter/pull/4919/commits/847737b6162183f3aa75f8c490d57b220a9e010d)
	- The context added to error messages in this commit was addressed in  #4925
- Add f0f9284a725d76fba27f333f8c89cc691d2b4637
	- Not a change from the original branch, but reporting parsing errors for a malformed range argument is much better than silently failing. I remember this being incredibly confusing to me when I first started using the CLI a few years ago.

#### TODO: 

* [x] Implementation
* [x] Rust bindings
* [x] Web bindings
* [x] API docs (What and where? More complete doc comments in the header?)
* [x] Docs

- Supersedes #4919

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1482,30 +1482,11 @@ impl Query {
         loader.find_all_languages(&loader_config)?;
         let query_path = Path::new(&self.query_path);
 
-        let byte_range = self.byte_range.as_ref().and_then(|range| {
-            let mut parts = range.split(':');
-            let start = parts.next()?.parse().ok()?;
-            let end = parts.next().unwrap().parse().ok()?;
-            Some(start..end)
-        });
-        let point_range = self.row_range.as_ref().and_then(|range| {
-            let mut parts = range.split(':');
-            let start = parts.next()?.parse().ok()?;
-            let end = parts.next().unwrap().parse().ok()?;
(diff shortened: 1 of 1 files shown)


lib/include/tree_sitter/api.h
```diff
@@ -1101,6 +1101,16 @@ bool ts_query_cursor_set_byte_range(TSQueryCursor *self, uint32_t start_byte, ui
  */
 bool ts_query_cursor_set_point_range(TSQueryCursor *self, TSPoint start_point, TSPoint end_point);
 
+/**
+ * Set the range of bytes within which all matches must be fully contained.
```
clason:
```suggestion
 * Set the range of bytes in which matches will be searched for. In contrast to `ts_query_cursor_set_byte_range`, this will restrict the query cursor to only return matches where _all_ nodes are fully contained within the given range. Both functions can be used together, e.g. to search for any matches that intersect line 5000, as long as they are fully contained within lines 4500-5500.
```

and similarly below?

maxbrunsfeld (approved):

maxbrunsfeld:
Thank you @WillLillis !

PR #5114, opened 2025-12-08
Revert "fix(rust): specify workspace dependency of `tree-sitter-language` crate"

WillLillis:
Apparently not. Let's try that again.


clason:
I don't think this is needed.

PR #5115, opened 2025-12-08
0.26.x

WillLillis:
It looks like the last publish workflow failed because the `tree-sitter-language` crate already had version `0.1.5` published (likely directly via crates.io) before other changes went into it in this repo. Bumping the language crate's version to `0.1.6` should allow publishing to succeed, which will then in turn allow the rest of the repo to go through its publishing motions.

Leaving this as a draft for now, to be picked up tomorrow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.13)
 
 project(tree-sitter
-        VERSION "0.26.1"
+        VERSION "0.26.2"
         DESCRIPTION "An incremental parsing system for programming tools"
         HOMEPAGE_URL "https://tree-sitter.github.io/tree-sitter/"
         LANGUAGES C)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1972,7 +1972,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter"
-version = "0.26.1"
+version = "0.26.2"
 dependencies = [
  "bindgen",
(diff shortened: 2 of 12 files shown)


clason:
Shouldn't the Cargo.lock entry be bumped as well?

WillLillis:
> Shouldn't the Cargo.lock entry be bumped as well?

I'm seeing that the Cargo.lock entry for tree-sitter-language was bumped to 0.1.6. Unless I'm missing it in another spot?

clason:
Sorry, I missed that (it was late).

clason:
let's give this a try when you're ready (of course, as `0.26.2`, and the commit message should mention the language bump to `0.1.16` as well so you can grep for it).

clason:
Idea: bump the language crate to `0.1.26` to make the coupling to the minor version explicit? (I.e., it's not an independently meaningful version but just a "down-shift" of the minor to a patch version for dependency resolution purposes.)

(If it's good enough for Apple...)

ObserverOfTime:
And then we can't patch it without breaking the coupling. :^)

clason:
Why would we need to patch it? AFAIUI, the crate is a simple "gear shift" layer, and bumps are needed if and only if the library is bumped.

WillLillis:
Opted to leave the language crate at 0.1.6. Bumping to 0.1.26 could help clear things up for us short term, but would ultimately be more confusing if/when the language crate gets bumped out of the normal release cycle again.

PR #5118, opened 2025-12-10
fix(docs): TREE_SITTER_JS_RUNTIME=native was available from 0.26.0

mavit:


clason:
I don't know what you're talking about, there is no 0.26.0.

mavit:
I don't think you want to bump this number with every release, as you appear to be doing now (commits 8a3dcc6 and 8b81997)?

clason:
No, we don't. But 0.26.2 is the "real" 0.26 release, so this is still correct.

PR #5120, opened 2025-12-10
ci(release): use node 24

ObserverOfTime:
Node 20 installs an outdated version of npm which doesn't support trusted publishing.


clason:
We haven't synced the release branch yet so we can't backport. Once we _finally_ have a working release, that is the next step.

PR #5121, opened 2025-12-10
ci(release): dispatch individual jobs

ObserverOfTime:
Useful when individual jobs fail, but requires all the jobs to depend on `build` instead of having `crates` & `npm` depend on `release`, since I don't know whether the `needs` field supports conditions.

This means that the crates & npm packages may be published even when the GitHub release fails, but this change will allow us to manually dispatch the workflow with just the release enabled in that case.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,20 @@ name: Release
 
 on:
   workflow_dispatch:
+    inputs:
+      release:
+        description: GitHub release
+        type: boolean
+        default: true
+      crates:
+        description: crates packages
+        type: boolean
+        default: true
+      npm:
+        description: npm packages
+        type: boolean
+        default: true
+
   push:
     tags:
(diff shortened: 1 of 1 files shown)


clason:
You can always just rerun failed jobs, no? Why would you want to re-run a successful one?

ObserverOfTime:
Failed jobs sometimes require fixes to the workflow (e.g. #5120) so it can't be rerun directly.

clason:
But won't that then push a non-versioned commit from master?

ObserverOfTime:
If we only use it for commits that fix broken workflows and call it from the release branch that won't be an issue.

ObserverOfTime:
Gonna have to adapt `$GITHUB_REF_NAME` since it will point to the branch through `workflow_dispatch`.

clason:
I don't think that is necessary; just adding a new release is cleaner.

PR #5130, opened 2025-12-14
fix(wasm): add `iswpunct` to stdlib

clason:
Revisit https://github.com/tree-sitter/tree-sitter/pull/3230

This function is admittedly niche but used by a few [parsers in the wild](https://github.com/search?q=iswpunct+path%3A**%2Fsrc%2Fscanner.c&type=code)

### DIFF
--- a/crates/xtask/src/build_wasm.rs
+++ b/crates/xtask/src/build_wasm.rs
@@ -593,6 +593,7 @@ pub fn run_wasm_stdlib() -> Result<()> {
         ])
         .args(&export_flags)
         .arg("crates/language/wasm/src/stdlib.c")
+        .arg("crates/language/wasm/src/wctype.c")
         .output()?;
 
     bail_on_err(


WillLillis (approved):
Shouldn't add any maintenance burden for us and allows a few parsers to target wasm in the wild, LGTM.

ObserverOfTime:
> This function is admittedly niche but used by a few [parsers in the wild](https://github.com/search?q=iswpunct+path%3A**%2Fsrc%2Fscanner.c&type=code)

Actually, if you check that list, it's still just Vim. :^)

lib/src/wasm/wasm-stdlib.h
```diff
@@ -1174,36 +1534,42 @@ unsigned char STDLIB_WASM[] = {
...
+  0x72, 0x65, 0x6e, 0x63, 0x65, 0x2d, 0x74, 0x79, 0x70, 0x65, 0x73, 0x2b,
+  0x0a, 0x6d, 0x75, 0x6c, 0x74, 0x69, 0x76, 0x61, 0x6c, 0x75, 0x65, 0x2b,
+  0x0e, 0x65, 0x78, 0x74, 0x65, 0x6e, 0x64, 0x65, 0x64, 0x2d, 0x63, 0x6f,
+  0x6e, 0x73, 0x74, 0x2b, 0x0f, 0x62, 0x75, 0x6c, 0x6b, 0x2d, 0x6d, 0x65,
+  0x6d, 0x6f, 0x72, 0x79, 0x2d, 0x6f, 0x70, 0x74, 0x2b, 0x16, 0x63, 0x61,
+  0x6c, 0x6c, 0x2d, 0x69, 0x6e, 0x64, 0x69, 0x72, 0x65, 0x63, 0x74, 0x2d,
+  0x6f, 0x76, 0x65, 0x72, 0x6c, 0x6f, 0x6e, 0x67
 };
-unsigned int STDLIB_WASM_LEN = 14463;
+unsigned int STDLIB_WASM_LEN = 18860;
```
clason: That's a roughly 30% increase (in the size of the `stdlib.wasm` in bytes).

(On a related note: It's a bit worrying that the `wasm-opt` output depends on the platform...)

WillLillis: (Documenting what I have so far, still some details to iron out) After some digging, I believe the functions in the wasm stdlib's export list are built from/using wasi-sdk's headers by default, the implementation you added to [‎`crates/language/wasm/include/wctype.h`](https://github.com/tree-sitter/tree-sitter/pull/5130/changes#diff-9ebaf837119fa149b4b2296f3b90f882dccdc2bb8f856f8b829d62e8822dd0d0) is used when targeting wasm32-unknown-unknown a la https://github.com/tree-sitter/tree-sitter/pull/5158. I suspect adding `iswpunct` to the export list pulled in something other than just the implementation for that specific function. The diff for the wasm file suggests this at least:

image

I tried moving your `iswpunct` implementation to a file reachable to the stdlib compilation job (just shoved it inside `stdlib.c for testing purposes, with the corresponding decl in `stdlib.h`) and the wasm blob size reduced down to 14899 bytes, which seems much more reasonable.

clason: ...and wasi-sdk's headers seem to pull this in from the system stdlib (at least I couldn't find any implementation there).

clason: Looking a bit more into the system headers (on macOS), it looks like the implementation of `iswpunct` is shared with `iswalpha` and friends, so this is most curious.

WillLillis:
I moved the implementation of `iswpunct` to `wctype.c` (added in #5158) and tweaked the xtask to include that source file as part of its build (since the header isn't included in `stdlib.c`). I also changed the definition to use the simplified boolean statement (as in https://github.com/tree-sitter-grammars/tree-sitter-vim/commit/5cb66512c12cdff1f2581781ac17647d1a16ff4e) because that resulted in a slightly smaller blob.

It was also necessary to change the return type from `bool` to `int` to match the declaration from the wasi headers.

PR #5131, opened 2025-12-14
chore(templates)!: add generated code to gitignore

ObserverOfTime:
Is it time yet?

### DIFF
--- a/crates/cli/src/templates/gitignore
+++ b/crates/cli/src/templates/gitignore
@@ -36,6 +36,10 @@ zig-out/
 # Example dirs
 /examples/*/
 
+# Generated files
+src/parser.c
+src/tree_sitter/*
+
 # Grammar volatiles
 *.wasm
 *.obj


clason:
No.

ObserverOfTime:
T_T

wetneb:
In my experience, not checking in those files into git works pretty well. The generated files are included in packages published in package repositories (Crates.io, PyPI, NPM) and as an archive in the releases. The main obstacle to that is the downstream reuses which expect a git repository with those files checked in, but that's a factor that's external to tree-sitter itself.

@clason could you elaborate on which changes in tree-sitter you think should be required before making this switch?

clason:
>  The main obstacle to that is the downstream reuses which expect a git repository with those files checked in, but that's a factor that's external to tree-sitter itself.

This. I am not comfortable forcing the hands of people in this way. This should be a deliberate choice of grammar authors, one way or another. Not every downstream stack relies on repository packages, after all.

ObserverOfTime:
We can make it an init flag.

wetneb:
I think it would still be sensible to have tree-sitter nudge grammar authors towards more sensible defaults, meaning not to track generated files in repositories, as proposed in #930. Downstream reusers currently relying on cloning Git repos should be able to switch to downloading release archives which contain generated files.
This is a transition that should be quite doable, since the auto-generated tarballs (by GitHub or other forge platforms) already contain generated files for repos which currently include them. So downstream reusers should already be able to switch to release tarballs without waiting for grammar maintainers to do the switch.

In my opinion, the `.gitignore` file generated by `tree-sitter init` doesn't force anyone's hands, as it can be freely edited. So @ObserverOfTime's suggestion to change it makes sense to me.

clason:
> In my opinion, the .gitignore file generated by tree-sitter init doesn't force anyone's hands, as it can be freely edited.

But this argument cuts both ways. I honestly don't see whom this change helps, and I can see how this adds (even more) maintainer friction. 

> Downstream reusers currently relying on cloning Git repos should be able to switch to downloading release archives which contain generated files.

This assumes that grammar projects actually make trustable versioned releases using the "official" upstream actions. That is far from the case.

wetneb:
So do you disagree with the goal stated in #930 of "Remove generated files from all the grammar repos in the tree-sitter org"? My understanding was that this was a consensual objective, which was only held back by technical limitations in tree-sitter, but I get a sense that you disagree with the overall goal, perhaps?

If you are aware of any improvements to tree-sitter itself that should be done before making this switch, then I think this is worth stating clearly here, so that people can help by tackling those.

clason:
That is a goal _for the tree-sitter org_. I am much more concerned with the impact on the 300+ grammars outside.

Again, this is not about committing vs. not committing the generated `parser.c` -- my point is that _this change_ helps nobody and is needlessly intrusive. (Just as an example: Imagine you want or need to commit the `parser.c`; then there's a drive-by "update bindings" commit. Now you still have the parser, but changes risk no longer being picked up, which means that your grammar and parser are out of sync. That's the worst possible outcome.)

(And personally, I am much more interested in not having the generated _bindings_ committed.)

ObserverOfTime:
The bindings can't be removed from git because they are often manually edited (in particular for split parsers) and are sometimes used through git (especially in Swift and Go).

The source files on the other hand are generated automatically and all they do is cause unresolvable merge conflicts.

sogaiu:
Unfortunately though, there are probably a fair number of projects (that none of us know about comprehensively) that depend on generated files.

ObserverOfTime:
And those files will be in the release tarball.

sogaiu:
Sure, but folks are likely already depending on git to retrieve things, right?

clason:
I don't think we should mandate that; instead we'll add a section to the documentation on "should I commit the generated files? Pros and Cons" with tips for managing each scenario (e.g., keep the json in either case, add `-diff` or `-merge` for the parser if you do commit etc.)

PR #5133, opened 2025-12-14
fix(cli): trailing whitespace after multiline text nodes in CST

skewb1k:
## Problem

The CST printer emits trailing whitespace after multiline text nodes. With 1704c604bf663801876572fe08b746e787cd7fdb and `:cst` corpus tests this causes trailing spaces to appear on `test --update`. These spaces cannot be removed afterward, as the test runner expects an exact character-for-character match for CST tests.

For example, using the https://github.com/tree-sitter/tree-sitter-rust grammar:

1765752365_screenshot

## Solution

Print whitespace only if node is not multiline.

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -887,7 +887,7 @@ fn write_node_text(
                 write!(
                     out,
                     "{}{}{}{}{}{}",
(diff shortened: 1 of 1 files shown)


WillLillis (approved):
Thanks!

PR #5134, opened 2025-12-15
fix(cli): correct discrepancy with cst for `--no-ranges`

WillLillis:
Followup to https://github.com/tree-sitter/tree-sitter/pull/4147. Note that because we currently always include ranges for the `cst` test type, this change isn't breaking.

The improved display of node text wasn't applied when `--no-ranges` was passed.

Before:

image

After:

image

This is more consistent with the display when ranges are included:

image

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -883,35 +883,24 @@ fn write_node_text(
                     0
                 };
             let formatted_line = render_line_feed(line, opts);
-            if !opts.no_ranges {
-                write!(
-                    out,
-                    "{}{}{}{}{}{}",
-                    if multiline { "\n" } else { " " },
-                    if multiline {
-                        render_node_range(opts, cursor, is_named, true, total_width, node_range)
-                    } else {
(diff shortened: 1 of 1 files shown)


crates/cli/src/parse.rs
arkriny: It looks like the only difference between the branches is the range handling. Could this be simplified to this?
```rust
write!(
    out,
    "{}{}{}{}{}{}",
    if multiline { "\n" } else { " " },
    if multiline && !opts.no_ranges {
        render_node_range(opts, cursor, is_named, true, total_width, node_range)
    } else {
        String::new()
    },
    if multiline {
        "  ".repeat(indent_level + 1)
    } else {
        String::new()
    },
    paint(quote_color, &String::from(quote)),
    &paint(color, &render_node_text(&formatted_line)),
    paint(quote_color, &String::from(quote)),
)?;
```

WillLillis: :man_facepalming: yeah, nice catch!

PR #5151, opened 2025-12-22
docs: Add troubleshooting section to tree-sitter-cli README

Mattchine:
### Summary

Added a troubleshooting section with commands for diagnosing Tree-sitter issues in both Neovim and the terminal.

### Background

I had trouble returning to Neovim because Tree-sitter kept spamming errors related to `build-wasm`.  
I initially tried installing newer `tree-sitter-cli` via `cargo`, but the issue persisted.

After further investigation, I discovered that an **older `tree-sitter-cli` installed via `yarn`** was still present in `$PATH` and was being used instead of the newer version. 

Uninstalling the older CLI resolved the issue.

### Notes

The commands included in the troubleshooting section are the same ones used to debug and diagnose this problem.

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -42,3 +42,18 @@ The `tree-sitter` binary itself has no dependencies, but specific commands have 
 
 [the documentation]: https://tree-sitter.github.io/tree-sitter/creating-parsers
 [the releases page]: https://github.com/tree-sitter/tree-sitter/releases/latest
+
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but Neovim issues are completely out of scope here. And the "terminal" section is very low value -- that applies to pretty much every CLI tool, especially once you start mixing and matching package managers (which you shouldn't do).

PR #5157, opened 2025-12-22
Implement BitOr for Range.

ide-outils:
Helper to quickly join ranges. Maybe, implementing a `join` method would be better, but at a first look, it seems correct to me to use BitOr, but it is not really an union or an intersect.

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -109,15 +109,16 @@ pub struct Range {
     pub end_point: Point,
 }
 
-impl std::ops::BitOr for Range {
+impl core::ops::BitOr for Range {
     type Output = Self;
 
     fn bitor(self, rhs: Self) -> Self::Output {
+        use core::cmp::{max, min};
         Self {
-            start_byte: std::cmp::min(self.start_byte, rhs.start_byte),
-            end_byte: std::cmp::max(self.end_byte, rhs.end_byte),
-            start_point: std::cmp::min(self.start_point, rhs.start_point),
-            end_point: std::cmp::max(self.end_point, rhs.end_point),
+            start_byte: min(self.start_byte, rhs.start_byte),
+            end_byte: max(self.end_byte, rhs.end_byte),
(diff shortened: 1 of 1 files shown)


lib/binding_rust/lib.rs
```diff
@@ -109,6 +109,19 @@ pub struct Range {
     pub end_point: Point,
 }
 
+impl std::ops::BitOr for Range {
```
ObserverOfTime:
```suggestion
impl core::ops::BitOr for Range {
```
Same below.

ObserverOfTime:
So what is the use-case for this?

ide-outils:
By example you could do something like this :
```rust
let whole_range = match.captures.iter().map(|c| c.node.range).reduce (|r1,
r2|  r1 | r2).unwrap();

Let range = node_1.range | node_2.range;

...
```
Basically, if you want to have a range that includes multiple items.

So we could even create a method for it :

```rust
fn join(self, ranges: impl Iterator<Item = Range>) -> Range
{ranges.fold(self, Todo!("The logic I put in BitOr"))}
```

Le jeu. 25 déc. 2025 à 13:09, ObserverOfTime ***@***.***> a
écrit :

clason:
Yes, but why does this have to be part of the tree-sitter bindings (instead of whatever project needs it)?

ide-outils:
It's what I have done for my project, wrapping range into a structure to
implement the BitOr, I was just thinking it might be convenient to have it
directly available, so I proposed it. If it is not revalent or commonly
used, you can just close the MR.

Le jeu. 25 déc. 2025 à 18:54, Christian Clason ***@***.***> a
écrit :

WillLillis:
I don't see a compelling enough use case to add this to the rust bindings. The core C library doesn't expose any similar functionality, and this is pretty straightforward to implement for downstream consumers (either by wrapping `Range` in a tuple struct that implements `BitOr`, or by writing a small helper function). Thanks for the thought, though!

PR #5168, opened 2025-12-25
docs(index): Mention 3rd party Elisp bindings

jcs090218:
Mention https://github.com/emacs-tree-sitter/elisp-tree-sitter.


ObserverOfTime:
Those bindings are specific to Emacs. We're not mentioning the Neovim Lua bindings either.

sogaiu:
I wonder if the following ought to be removed for consistency:

```
[ELisp](https://www.gnu.org/software/emacs/manual/html_node/elisp/Parsing-Program-Source.html)
```

Looking back at https://github.com/tree-sitter/tree-sitter/pull/3881, I think there was a suggestion for having a section for editors that use tree-sitter.

May be I missed it, but if that doesn't exist, would it make sense to create that?

clason:
Yes, that would be cleaner (and more useful to the reader).

(Although I still think this sort of list is too ephemeral and more appropriate to the wiki rather than the main docs.)

jcs090218:
> Those bindings are specific to Emacs. We're not mentioning the Neovim Lua bindings either.

What does this have to do with Neovim Lua bindings? I saw Elisp bindings and I think it makes sense to include this good 3rd party Elisp binding as well. 😕

clason:
The point is that you _shouldn't_ have seen "Elisp bindings". The "language bindings" are general purpose and can be used for _any_ standalone project written in that language. By contrast, these "bindings" (like the Neovim Lua bindings) are not general purpose but an _editor_ API that exposes tree-sitter functionality for plugins. That's a different kettle of fish.

jcs090218:
Yes, I understand that, and second to have an editor section. But I don't think that's a good way to just close a PR down. 🤔 It just sounds irrational, but okay. 😅

clason:
We just gave you the rationale. (And these discussions whenever someone wants to add their favorite third-party related project are exactly the reason why I didn't want to have such a list in the first place.)

jcs090218:
Yes, I understand and agree with the idea.

I’m simply not a fan of handling PR in this way, for the reasons I explained in my previous comment. 🤔 It’s a matter of personal preference rather than a technical concern. That said, I accept it, as it’s ultimately up to others to decide how they want to run their community. 👌

PR #5172, opened 2025-12-26
docs(web): add WASM version compatibility section

helloiamlee:
## Summary

Add documentation about WASM version compatibility requirements between `web-tree-sitter` and `tree-sitter-cli`.

## Problem

Users who use pre-built WASM files from third-party packages (like `tree-sitter-wasms`) may experience `Language.load()` failures when upgrading `web-tree-sitter` to 0.26.x. This happens because the WASM files were built with older `tree-sitter-cli` versions (e.g., 0.20.x) that have incompatible ABI.

Currently, there's no documentation explaining this version compatibility requirement, leading to confusion when things break after upgrades.

## Solution

Add a "WASM Version Compatibility" section to `lib/binding_web/README.md` that:

1. Explains the ABI compatibility requirement
2. Provides a version compatibility matrix
3. Warns users about third-party pre-built WASM packages
4. Recommends matching `tree-sitter-cli` version with `web-tree-sitter`

## Changes

- `lib/binding_web/README.md`: Added new "WASM Version Compatibility" section after "Generating .wasm files"

## Related

Closes #5171

Issue #5171, opened 2025-12-26
web-tree-sitter 0.26.x incompatible with WASM files built by tree-sitter-cli 0.20.x


wtdlee:
## Summary

WASM language files built with `tree-sitter-cli` 0.20.x are incompatible with `web-tree-sitter` 0.26.x due to ABI changes. This affects users of third-party WASM packages like `tree-sitter-wasms`.

## Problem

When using `web-tree-sitter@0.26.3` with WASM files from the [`tree-sitter-wasms`](https://github.com/Gregoor/tree-sitter-wasms) package (which builds WASM files using `tree-sitter-cli@0.20.8`), `Language.load()` fails to load the language.

### Reproduction Steps

1. Install dependencies:
```bash
npm install web-tree-sitter@0.26.3 tree-sitter-wasms@0.1.13
```

2. Try to load a language:
```javascript
import Parser from 'web-tree-sitter';

await Parser.init();
const parser = new Parser();

// This fails with WASM files built by tree-sitter-cli 0.20.x
const language = await Parser.Language.load('node_modules/tree-sitter-wasms/out/tree-sitter-ruby.wasm');
parser.setLanguage(language);
```

### Expected Behavior

Either:
1. `Language.load()` works with WASM files built by older `tree-sitter-cli` versions, or
2. Clear documentation about version compatibility requirements

### Actual Behavior

Loading fails silently or throws an error due to ABI mismatch between:
- `web-tree-sitter` 0.26.x (expects tree-sitter 0.26.x ABI)
- WASM files built with `tree-sitter-cli` 0.20.x (provides tree-sitter 0.20.x ABI)

## Workaround

Pin `web-tree-sitter` to `^0.25.10` which remains compatible with WASM files built by older `tree-sitter-cli` versions.

## Suggestion

Consider one of the following:
1. **Documentation**: Add a compatibility matrix showing which `web-tree-sitter` versions work with which `tree-sitter-cli` versions for WASM generation
2. **Backward compatibility**: If feasible, support loading older WASM formats
3. **Error messages**: Provide clear error messages when ABI version mismatch is detected

## Environment

- `web-tree-sitter`: 0.26.3
- `tree-sitter-wasms`: 0.1.13 (uses `tree-sitter-cli@0.20.8`)
- Node.js: 18+
- OS: macOS / Linux

## Related

- [tree-sitter-wasms Issue #34](https://github.com/Gregoor/tree-sitter-wasms/issues/34)
[...]

PR #5173, opened 2025-12-26
fix(web): add early version check in Language.load()

helloiamlee:
## Summary

Add ABI version compatibility check immediately after loading a WASM language file in `Language.load()`.

## Problem

Currently, when loading a WASM file built with an incompatible `tree-sitter-cli` version:

1. `Language.load()` succeeds
2. `parser.setLanguage()` fails with: `Incompatible language version X. Compatibility range Y through Z.`

This is confusing because:
- The error occurs at `setLanguage()`, not at load time
- The error message doesn't indicate the root cause (incompatible WASM build)
- Users don't know how to fix it

This commonly affects users of third-party WASM packages like `tree-sitter-wasms` that may be built with older `tree-sitter-cli` versions.

## Solution

Add version check in `Language.load()` right after loading the WASM module:

```typescript
const version = language.abiVersion;
if (version < MIN_COMPATIBLE_VERSION || LANGUAGE_VERSION < version) {
  throw new Error(
    `Incompatible language version ${version}. ` +
    `Compatibility range ${MIN_COMPATIBLE_VERSION} through ${LANGUAGE_VERSION}. ` +
    `The WASM file may have been built with an incompatible version of tree-sitter-cli. ` +
    `Please regenerate the WASM file using a compatible tree-sitter-cli version.`
  );
}
```

## Benefits

1. **Earlier failure** - Fails at load time instead of `setLanguage()`
2. **Better error message** - Clearly indicates the WASM file might be built with incompatible CLI
3. **Actionable guidance** - Tells users to regenerate the WASM file

## Changes

- `lib/binding_web/src/language.ts`:
  - Import `LANGUAGE_VERSION` and `MIN_COMPATIBLE_VERSION` from `parser.ts`
  - Add version compatibility check after loading WASM
  - Add `@throws` JSDoc annotation

## Related

- #5171 - Issue describing the compatibility problem
- #5172 - Documentation PR for version compatibility

### DIFF
--- a/lib/binding_web/src/language.ts
+++ b/lib/binding_web/src/language.ts
@@ -1,7 +1,7 @@
 import { C, INTERNAL, Internal, assertInternal, SIZE_OF_INT, SIZE_OF_SHORT } from './constants';
 import { LookaheadIterator } from './lookahead_iterator';
 import { unmarshalLanguageMetadata } from './marshal';
-import { TRANSFER_BUFFER } from './parser';
+import { TRANSFER_BUFFER, LANGUAGE_VERSION, MIN_COMPATIBLE_VERSION } from './parser';
 
 const LANGUAGE_FUNCTION_REGEX = /^tree_sitter_\w+$/;
 
@@ -226,6 +226,8 @@ export class Language {
   /**
    * Load a language from a WebAssembly module.
    * The module can be provided as a path to a file or as a buffer.
+   *
+   * @throws {Error} If the WASM file was built with an incompatible version of tree-sitter-cli.
    */
   static async load(input: string | Uint8Array): Promise<Language> {
     let binary: Uint8Array | WebAssembly.Module;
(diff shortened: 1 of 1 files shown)


lib/binding_web/src/language.ts
```diff
@@ -226,6 +226,8 @@ export class Language {
   /**
    * Load a language from a WebAssembly module.
    * The module can be provided as a path to a file or as a buffer.
+   *
+   * @throws {Error} If the WASM file was built with an incompatible version of tree-sitter-cli.
```
ObserverOfTime:
```suggestion
   * @throws {RangeError} If the Wasm file was built with an incompatible version of tree-sitter-cli.
```

lib/binding_web/src/language.ts
```diff
@@ -263,6 +265,19 @@ export class Language {
         throw new Error('Language.load failed: no language function found in Wasm file');
     }
     const languageAddress = mod[functionName]();
-    return new Language(INTERNAL, languageAddress);
+    const language = new Language(INTERNAL, languageAddress);
+
+    // Early version compatibility check
+    const version = language.abiVersion;
+    if (version < MIN_COMPATIBLE_VERSION || LANGUAGE_VERSION < version) {
+      throw new Error(
```
ObserverOfTime:
```suggestion
      throw new RangeError(
```

lib/binding_web/src/language.ts
```diff
@@ -263,6 +265,19 @@ export class Language {
...
+        `Incompatible language version ${version}. ` +
+        `Compatibility range ${MIN_COMPATIBLE_VERSION} through ${LANGUAGE_VERSION}. ` +
+        `The WASM file may have been built with an incompatible version of tree-sitter-cli. ` +
+        `Please regenerate the WASM file using a compatible tree-sitter-cli version.`
```
ObserverOfTime:
```suggestion
        `The Wasm file may have been built with an incompatible version of tree-sitter-cli. `
```

helloiamlee:
Thank you for the review @ObserverOfTime! 

Applied all suggestions

- Changed `Error` to `RangeError`
- Fixed `WASM` → `Wasm` casing
- Removed the regeneration hint for a cleaner message

Let me know if there's anything else!

WillLillis:
Thanks for investigating this. The follow-up analysis in #5171 established that the affected Wasm files fail inside Emscripten’s loader because they use the legacy dylink section format, before a `Language` is constructed or its ABI can be checked. Genuine ABI mismatches are already checked by `Parser.setLanguage`, so duplicating that validation in `Language.load` would not address the reported failure.

PR #5174, opened 2025-12-26
feat(cli): concurrent builds of the same grammar on different locations on disk

stackmystack:
Hello,

The test suite of [`tsdl`](https://github.com/stackmystack/tsdl) runs many builds of the same grammars (same name ofc) in parallel, and the tests were flaky, which required me to put a [`--retries 2`](https://github.com/stackmystack/tsdl/blob/master/justfile#L40) defensively, and it's not always guaranteed to work.

I skimmed through the cli, and I figured there shouldn't be a reason to disallow concurrent builds like this.

So I did the fix, and it works for me now.

The choices I made with the hashing and lock file names are ofc arbitrary :)

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -8,6 +8,7 @@ use std::sync::Mutex;
 use std::{
     collections::HashMap,
     env, fs,
+    hash::{Hash as _, Hasher as _},
     io::{BufRead, BufReader},
     marker::PhantomData,
     mem,
@@ -1025,20 +1026,26 @@ impl Loader {
             return Ok(wasm_store.load_language(&config.name, &wasm_bytes)?);
         }
 
+        // Create a unique lock path based on the output path hash to prevent
+        // interference when multiple processes build the same grammar (by name)
+        // to different output locations
+        let lock_hash = {
+            let mut hasher = std::hash::DefaultHasher::new();
+            output_path.hash(&mut hasher);
+            format!("{:x}", hasher.finish())
+        };
+
         let lock_path = if env::var("CROSS_RUNNER").is_ok() {
             tempfile::tempdir()
-                .unwrap()
+                .expect("create a temp dir")
                 .path()
-                .join("tree-sitter")
-                .join("lock")
-                .join(format!("{}.lock", config.name))
+                .to_path_buf()
         } else {
-            etcetera::choose_base_strategy()?
-                .cache_dir()
-                .join("tree-sitter")
-                .join("lock")
-                .join(format!("{}.lock", config.name))
-        };
+            etcetera::choose_base_strategy()?.cache_dir()
+        }
+        .join("tree-sitter")
+        .join("lock")
+        .join(format!("{}-{lock_hash}.lock", config.name));
 
         if let Ok(lock_file) = fs::OpenOptions::new().write(true).open(&lock_path) {
             recompile = false;


crates/loader/src/loader.rs
```diff
@@ -8,6 +8,7 @@ use std::sync::Mutex;
 use std::{
     collections::HashMap,
     env, fs,
+    hash::{Hash, Hasher},
```
WillLillis:
```suggestion
    hash::{Hash as _, Hasher as _},
```

crates/loader/src/loader.rs
```diff
@@ -1025,19 +1026,28 @@ impl Loader {
             return Ok(wasm_store.load_language(&config.name, &wasm_bytes)?);
         }
 
+        // Create a unique lock path based on the output path hash to prevent interference
+        // when multiple processes build the same grammar to different output locations
+        let lock_hash = {
+            use std::collections::hash_map::DefaultHasher;
+            let mut hasher = DefaultHasher::new();
```
WillLillis:
```suggestion
            let mut hasher = std::hash::DefaultHasher::new();
```

crates/loader/src/loader.rs
```diff
@@ -1089,6 +1099,22 @@ impl Loader {
             }
         }
 
+        // Safety check: ensure the compiled library exists before trying to load it.
+        // This catches race conditions where compilation was skipped but the output file
+        // still doesn't exist.
+        if !output_path.exists() {
```
WillLillis, line 1106: It's not clear to me how this helps with any race conditions. If `output_path` doesn't exist then that should be caught by the `load_language` call below.

crates/loader/src/loader.rs
```diff
@@ -1025,19 +1026,28 @@ impl Loader {
...
+        };
+
         let lock_path = if env::var("CROSS_RUNNER").is_ok() {
             tempfile::tempdir()
                 .unwrap()
                 .path()
                 .join("tree-sitter")
                 .join("lock")
-                .join(format!("{}.lock", config.name))
+                .join(format!("{}-{}.lock", config.name, lock_hash))
```
WillLillis: Nit, but `lock_hash` can be placed directly inside the format string here (as well as below.) This entire `if/else` block could also be cleaned up to this if you'd like to :)

```rust
        let lock_path = if env::var("CROSS_RUNNER").is_ok() {
            tempfile::tempdir().unwrap().path().join("tree-sitter")
        } else {
            etcetera::choose_base_strategy()?.cache_dir().join("tree-sitter")
        }
        .join("lock")
        .join(format!("{}-{lock_hash}.lock", config.name));
```

crates/loader/src/loader.rs
stackmystack: True, it will err in `load_language` but the error is ambiguous and IMHO wrong. I will complain that it could not open the shared object, as if the compilation produced a corrupted lib, where in fact the issue is that the file doesn't exist, and that took a while to figure out ...

We could push this check to `load_language` itself, wdyt ?

crates/loader/src/loader.rs
stackmystack: I eventually settled on this:

```rust
        let lock_path = if env::var("CROSS_RUNNER").is_ok() {
            tempfile::tempdir()
                .expect("create a temp dir")
                .path()
                .to_path_buf()
        } else {
            etcetera::choose_base_strategy()?.cache_dir()
        }
        .join("tree-sitter")
        .join("lock")
        .join(format!("{}-{lock_hash}.lock", config.name));
```

stackmystack:
But on a more general note, I am not quite sure why the locking is necessary: to protect output path or the parser path ? If the former, then this PR addresses it, but if it's the latter, shouldn't we hash `config.parser_path` ?

crates/loader/src/loader.rs
WillLillis: The error message being unclear is fair enough. I think leaving the check where it is now makes more sense than moving it into `load_language`. I would remove/edit the comment above the check though to clarify things, though.

WillLillis:
> But on a more general note, I am not quite sure why the locking is necessary: to protect output path or the parser path ? If the former, then this PR addresses it, but if it's the latter, shouldn't we hash `config.parser_path` ?

My understanding is that the locking is protecting the output path. This PR addresses that issue and (presumably) fixes the problem you were running into with tsdl, so I think it's good as is :)

crates/loader/src/loader.rs
stackmystack: I rewrote that part. and I addressed the other comments. It's ready for re-review. Thanks a lot!

PR #5178, opened 2025-12-27
fix(cli): increase verbosity of `tree-sitter init -u` updates

WillLillis:
Also, use `info` logs rather than `warn`

Opening this to start a discussion regarding what (and how much) information should be displayed for the `tree-sitter init` command. In this PR's current state, I've updated all log statements to use `info` rather than `warn`. As any updates made to a grammar's files through this command was explicitly requested via the `-u` flag, this seems more appropriate. I've also clarified "updating" in several logs where this could be confused with updating the dependencies in a lock file.

After some feedback/discussion here, I hope we can settle on a more uniform stream of logs for the actions taken for this command. A few questions to get started:
- Should the command output a log when a file is generated for the first time?
- Should specific information be displayed regarding *what* in a file is being updated (as I've done for a few logs here), or just that changes are being done?
- Should a log be displayed if a particular language's bindings are disabled, stating that no generation/updates will be done?

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -123,7 +123,7 @@ const BUILD_ZIG_ZON_TEMPLATE: &str = include_str!("./templates/build.zig.zon");
 const ROOT_ZIG_TEMPLATE: &str = include_str!("./templates/root.zig");
 const TEST_ZIG_TEMPLATE: &str = include_str!("./templates/test.zig");
 
-const TREE_SITTER_JSON_SCHEMA: &str =
+pub const TREE_SITTER_JSON_SCHEMA: &str =
     "https://tree-sitter.github.io/tree-sitter/assets/schemas/config.schema.json";
 
 #[derive(Serialize, Deserialize, Clone)]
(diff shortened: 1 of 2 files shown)


clason:
My main concern is when things _don't_ get updated, because that leads to discoverability issues, e.g., new fields in `tree-sitter.json` that aren't added, and people will have no idea that they even exist. (This is not academic: `tree-sitter init` will create all binding fields now, but `init -u` won't add them to an existing file.)

So in my opinion:
* no need to output anything on `init` (the files on disk are "log" enough");
* no real need to output _successful_ changes on `init -u` (since it is reasonable to assume that people who want change tracking work in a git repository and can just do `git status` and `git diff`);
* more concrete info (not warning) on what is changed in existing files is an improvement over the status quo.

WillLillis:
@clason The latest commit fills in the `schema` and any new bindings fields (i.e. if they're missing, they'll get an explicit `false` instead) to tree-sitter.json. I'll have some time to take a pass on your other bullet points hopefully soon!

WillLillis:
> * no real need to output _successful_ changes on `init -u` (since it is reasonable to assume that people who want change tracking work in a git repository and can just do `git status` and `git diff`);
> * more concrete info (not warning) on what is changed in existing files is an improvement over the status quo.

Could you give an example to help me differentiate between these two points here?

clason:
Ah, sorry, they were not really different -- or rather, the third point was entirely separate: "all this being said, your changes are a strict improvement on the status quo so worth merging in any case" ("_if_ you warn, warn better" ;))

PR #5183, opened 2025-12-29
fix(cli): canonicalize build `--output` path

WillLillis:
This fixes a potential issue with the new lock file hashing mechanism, in which two different path literals pointing to the same location would hash to separate lock files, allowing a race condition.

Followup to #5174

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -955,11 +955,21 @@ impl Build {
         } else {
             let output_path = if let Some(ref path) = self.output {
                 let path = Path::new(path);
-                if path.is_absolute() {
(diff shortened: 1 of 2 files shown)


clason (approved):

clason:
Nvim-treesitter uses `ts b -o <lang>.so` (it just sets the non-default filename and moves stuff itself in the next step).

WillLillis:
> Nvim-treesitter uses `ts b -o <lang>.so` (it just sets the non-default filename and moves stuff itself in the next step).

Stupid late night mistake on my part, glad the crater tests caught it!

clason (approved):

Issue #5190, opened 2026-01-01
Export symbols missing on Windows


jcs090218:
### Problem

I've built the shared library (.dll) file on Windows but `dumpbin` gives me nothing:

```
Microsoft (R) COFF/PE Dumper Version 14.44.35222.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file tree-sitter.dll

File Type: DLL

  Summary

       57000 .data
        4000 .pdata
       4C000 .rdata
        B000 .reloc
      119000 .text
        1000 .tls
```

The strange part is that it works on my second machine (also running Windows 11) but not on my first one. I also tried building it on GitHub Actions, and the symbol appears to be missing there as well. The built binary can be found at https://github.com/jcs090218/csharp-tree-sitter/releases

I also tried building it with CMake, and it works after adding `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` on my first machine (this isn’t necessary on the second machine).

### Steps to reproduce

You can see my workflow file: https://github.com/jcs090218/csharp-tree-sitter/blob/main/.github/workflows/build-tree-sitter.yml

For CMake:

```
git clone https://github.com/tree-sitter/tree-sitter
cd tree-sitter
cmake -S . -B build/windows/
```

Then open Visual Studio to build the shared library.

### Expected behavior

The symbols are successfully appeared in the shared library.

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.26.3

### Operating system/version

Windows 11 (both)

PR #5191, opened 2026-01-01
fix(CMakeLists.txt): Export all symbols by default for MSVC

jcs090218:
For #5190.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
 option(TREE_SITTER_FEATURE_WASM "Enable the Wasm feature" OFF)
 option(AMALGAMATED "Build using an amalgamated source" OFF)
 
+# Exports all symbols by default for MSVC.
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+
 if(AMALGAMATED)
   set(TS_SOURCE_FILES "${PROJECT_SOURCE_DIR}/lib/src/lib.c")
 else()


scutuatua-crypto (approved):

jcs090218:
Is there any problem with this PR that's why it hasn't been merged? 🤔 Thanks!

WillLillis:
> Is there any problem with this PR that's why it hasn't been merged? 🤔 Thanks!

Anytime there's a problem in Windows-land it's nontrivial to get things working for us to test/debug locally. In addition, the worry is the symbols missing is due to some underlying problem with the library on Windows for _any_ build process (make, CMake, Zig, etc.), meaning that this will patch over the problem for CMake but leave the others broken and out of parity.

jcs090218:
By looking at the CMake page, https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html.

> This property is implemented only for MS-compatible tools on Windows.

I think it's safe in this scenario? 🤔

jcs090218:
> Anytime there's a problem in Windows-land it's nontrivial to get things working for us to test/debug locally.

Yeah, I agreed. Testing and development on Windows is often painful and hardly enjoyable. 

I've added a build tests for MSVC in csharp-tree-sitter, see https://github.com/jcs090218/csharp-tree-sitter/blob/main/.github/workflows/build-tree-sitter.yml. Maybe this repo can do the same? 🤔

WillLillis:
I (as well as Amaan) spent some more time looking into this and the issue is much broader than a patch to our CMakeLists.txt file. I'll open a PR with a more complete fix for the symbol visibility issues raised in #5190 during this release cycle.

Issue #4960, opened 2025-10-15
"tree-sitter test" crashes on ppc64el since v0.25.7


jamessan:
After updating tree-sitter from 0.25.6 to 0.25.9 in Debian, all of the tree-sitter grammar packages started crashing during their tests -- only on ppc64el.

```
XDG_CACHE_HOME=/build/reproducible-path/tree-sitter-c-0.24.1/debian/.cache tree-sitter test
tree-sitter: ./lib/src/./././././array.h:175: _array__erase: Assertion `index < self->size' failed.
Aborted
```

After bisecting, it was narrowed down to https://github.com/tree-sitter/tree-sitter/commit/b890e8bea0550d158ccd72f826bdee700f5a6e30. Since there are no scenarios in which the build specifies `-DNDEBUG`, all builds have the assert enabled.

Now, I'm not sure why it's only happening on one specific architecture. The backtrace from the crash is

```
#0  0x00003fff8f2dfd3c in ?? () from /lib/powerpc64le-linux-gnu/libc.so.6
#1  0x00003fff8f27663c in raise () from /lib/powerpc64le-linux-gnu/libc.so.6
#2  0x00003fff8f2565f0 in abort () from /lib/powerpc64le-linux-gnu/libc.so.6
#3  0x00003fff8f26a554 in ?? () from /lib/powerpc64le-linux-gnu/libc.so.6
#4  0x00003fff8f26a6e4 in __assert_fail () from /lib/powerpc64le-linux-gnu/libc.so.6
#5  0x000000012b5ab838 in _array__erase.part.0 ()
#6  0x000000012b947c50 in stack.iter ()
#7  0x000000012b94d028 in ts_parser.reduce ()
#8  0x000000012b959c8c in ts_parser_parse ()
#9  0x000000012b95c47c in ts_parser_parse_with_options ()
#10 0x000000012b66b260 in tree_sitter_cli::test::run_tests ()
#11 0x000000012b66abbc in tree_sitter_cli::test::run_tests ()
#12 0x000000012b66abbc in tree_sitter_cli::test::run_tests ()
#13 0x000000012b6680bc in tree_sitter_cli::test::run_tests_at_path ()
#14 0x000000012b5c4b04 in tree_sitter::Test::run ()
#15 0x000000012b5cc504 in tree_sitter::main ()
#16 0x000000012b65ea94 in std::sys::backtrace::__rust_begin_short_backtrace ()
[...]

PR #5194, opened 2026-01-04
Address strict aliasing violations in array.h

nwf:


lib/src/array.h
```diff
@@ -21,98 +21,106 @@ extern "C" {
...
 /// Reserve `new_capacity` elements of space in the array. If `new_capacity` is
 /// less than the array's current capacity, this function has no effect.
 #define array_reserve(self, new_capacity) \
-  _array__reserve((Array *)(self), array_elem_size(self), new_capacity)
+ (self)->contents = _array__reserve(&(self)->meta, (self)->contents, array_elem_size(self), new_capacity)
 
 /// Free any memory allocated for this array. Note that this does not free any
 /// memory allocated for the array's contents.
-#define array_delete(self) _array__delete((Array *)(self))
+#define array_delete(self) _array__delete((self)->contents, (self), sizeof(self))
```
WillLillis: Since a pointer is passed into this macro, `sizeof(self)` will always incorrectly give the `sizeof(void*)`. Instead, why don't we change  `_array__delete` to this?

```c
static inline void _array__delete(void **self_contents, ArrayMeta *self_meta) {
  if (*self_contents) {
    ts_free(*self_contents);
  }   
  *self_contents = NULL;
  self_meta->size = 0;
  self_meta->capacity = 0;
}
```

and the macro to 

```c
#define array_delete(self) _array__delete((void**)&(self)->contents, &(self)->meta)
```

If passing the double pointer to `contents` violates aliasing, would simply passing `sizeof(*self)` work instead?

lib/src/array.h
```diff
@@ -21,98 +21,106 @@ extern "C" {
...
 /// Assign the contents of one array to another, reallocating if necessary.
 #define array_assign(self, other) \
-  _array__assign((Array *)(self), (const Array *)(other), array_elem_size(self))
+  (self)->contents = _array__assign(&(self)->meta, (self)->contents, &(other)->meta, (other)->contents, array_elem_size(self))
 
 /// Swap one array with another
 #define array_swap(self, other) \
-  _array__swap((Array *)(self), (Array *)(other))
+  do { \
+    __typeof__(*other) swap = *other; \
```
WillLillis: Is there a way we can do this without relying on compiler extensions?

WillLillis:
For functions that may alter `contents`, could we (without violating strict aliasing) accept a `void**` parameter instead of returning them as a `void*`?

WillLillis:
The template header (`crates/generate/src/templates/array.h`) will also need to be updated.

lib/src/array.h
```diff
@@ -21,98 +21,106 @@ extern "C" {
...
 
-#define Array(T)       \
-  struct {             \
-    T *contents;       \
-    uint32_t size;     \
-    uint32_t capacity; \
+typedef struct {
+  uint32_t size;
+  uint32_t capacity;
+} ArrayMeta;
```
ObserverOfTime: Note that this struct will be exposed. Name it `_ArrayMeta` or something.

lib/src/array.h
nwf: Sorry, yes, that should have been `sizeof(*(self))`.

nwf:
> For functions that may alter `contents`, could we (without violating strict aliasing) accept a `void**` parameter instead of returning them as a `void*`?

Alas, AIUI, which I will admit is _far from perfectly_, that would violate strict aliasing, since `T*` and `void*` are not compatible types.

lib/src/array.h
nwf: Yes, a bit more verbosely.

WillLillis:
These changes appear to cause build failures for any parsers that depend on the existing `Array` type (essentially, any parser with an external scanner). Unless there's a way to provide a smooth transition that doesn't disrupt the grammar ecosystem, I think it makes more sense to add the `-fno-strict-aliasing` flag you mentioned previously [here](https://github.com/tree-sitter/tree-sitter/issues/4960#issuecomment-3706937047).

nwf:
Direct access to Array members is... worrying.  (The Array type sort of intrinsically encourages the strict aliasing abuse on display here.)  That said, it's possible to do these changes without ArrayMeta so that those fields remain accessible.  I don't have time to do that right now, though.

nwf:
`-fno-strict-aliasing` is a bad solution to this problem, as it must be present when compiling any file that uses `array.h`.

WillLillis:
> Direct access to Array members is... worrying. (The Array type sort of intrinsically encourages the strict aliasing abuse on display here.) That said, it's possible to do these changes without ArrayMeta so that those fields remain accessible. I don't have time to do that right now, though.

That's fair. If you have an approach in mind feel free to outline it here and I can try to spend some time implementing it (giving you the proper credit for all the legwork of course). Otherwise I'll take a fresh look at the problem when I have the time. 🙂

nwf:
Thanks for being understanding. :)

The next thing I'd try is to drop the 1st commit from this PR -- the one that adds `_ArrayMeta` -- and pass the `size` and `capacity` fields by value or by reference, as needed, to the `array.h` internal functions.  That's probably pretty "easy" but it's a time-consuming mechanical transform.

nwf:
Closing in favor of #5242.

martinez78960-glitch (commented):
Nd go

PR #5199, opened 2026-01-04
feat(wasm): add some common used defs to wasm stdlib

trim21:


clason:
Why?

trim21:
this stdlib will be used when users write a scanner.c themselves and targeting wasm32

clason:
Can you point to a single scanner where those are used?

trim21:
> Can you point to a single scanner where those are used?

Yeah, I just fix a scanner that doesn't compile on wasm32, which use UINT8_MAX...

https://github.com/polarmutex/tree-sitter-beancount/pull/127/files#diff-bbcf6c785b007d9655b443abc204cc3cc32aa03cea190cb865fe44252dc6f1df

clason:
The Rust bindings are auto-generated and not meant to target wasm. To build a WASM-compliant parser, use `tree-sitter build --wasm`.

clason:
(This isn't saying it's impossible to add wasm support to the bindings, but this is not the way.)

trim21:
> (This isn't saying it's impossible to add wasm support to the bindings, but this is not the way.)



> The Rust bindings are auto-generated and not meant to target wasm. To build a WASM-compliant parser, use `tree-sitter build --wasm`.

not really, we already support it https://github.com/tree-sitter/tree-sitter/pull/4820 , and it's currently in our testing matrix.

we just have a very small stdlib that users need to be careful when they write their own `scanner.c`

I don't think we need to support large set of stdlib, for example, there is no need to support printf, but add these static def to stdint.h won't do any harm

clason:
My point is that this needs to be done carefully -- you also need to provide the _implementation_ for any stdlib function you want to add. (The stdlib is minimal by design; we can't simply add any function somebody thinks they may want -- we'd rather invest in providing a better _purpose-built_ infrastructure for external scanners.)

clason:
(This is a general discussion to be had and not specific to _this_ PR, mind you.)

trim21:
I'm ok to not add `strlen`, I just find that it's already in the stdio.h, I simply move it. so I won't argue for it if someone think it's a bad idea to add it.

I mainly want to add missing defs on stdint.h, which are simple `#define`

clason:
@maxbrunsfeld this change looks innocuous so I am leaning towards merging it, but I think it would be helpful to have some sort of "design goal" for the wasm stdlib (since wasm parsers are the future -- and it would be very good if we could start supporting multiple runtimes for which this could serve as a specification).

maxbrunsfeld:
I'm ok with all of these changes. Overall, I just want to provide a minimal subset of libc that enables writing external scanners, so that those scanners compile and behave the same when they are compiled to native code, and to wasm. I don't know exactly what the subset should be. I do want the code size to be reasonable. Constants like don't have any cost AFAICT.

clason:
I agree; this was a more general question (including for guidance on, e.g., the iswpunct addition).

clason:
@trim21 you (and I) forgot to actually add this to the stdlib (`wasm-stdlib.h`).

clason:
(we really need some documentation about this... @WillLillis )

trim21:
> @trim21 you (and I) forgot to actually add this to the stdlib (`wasm-stdlib.h`).

What's inside this file? contents of stdlib files?

clason:
Yes. That's the part that actually ends up in the wasm blob for the parser.

trim21:
If it's generated from other sources we may need a ci to check it or update it

clason:
Something like that, yes.

PR #5201, opened 2026-01-04
fix(rust,wasm32): implement c malloc with rust alloc for wasm32

trim21:
in short words: `crates/language/wasm/src/stdlib.c` is buggy, you will get `indirect call type mismatch` `uninitialized element` `out of bounds memory access` `unreachable` randommly

<details><summary>snapshot</summary>

image

image

</details>

I don't think it's worth fixing.

This is a POC that we can redirect rust's malloc, dealloc and realloc as stdlib for c code, and it works fine (in my local testing).

The rust code does what we do in the original c code. In `malloc`, alloc extra N bytes memory and store the memory size at the beginning, then return the rest memory to caller.

Later in `free` or `realloc`, calculate the read total size from header then use it.

---

maybe off-topic: if we need stdlib c function that hard to implement in c, we can do the same think and implement with rust like what this PR do.


crates/language/src/stdlib.rs
```diff
@@ -0,0 +1,132 @@
+#![cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+
+use std::alloc::Layout;
+use std::mem::{align_of, size_of};
+use std::ptr;
+
+// C `malloc` must return a pointer suitably aligned for any object type.
+// We use a conservative alignment that is at least pointer-sized and typically 16.
```
trim21: this is misleading comment and it's not true, I'll remove it later

WillLillis:
This is a large change, and the first report of its kind for the issue you're describing. 

Before we consider this, could you provide a reproduction with 
- repo link
- exact build steps
- exact running steps

trim21:
> This is a large change, and the first report of its kind for the issue you're describing.
> 
> Before we consider this, could you provide a reproduction with
> 
> * repo link
> * exact build steps
> * exact running steps

I make a re-produce branch here: https://github.com/trim21/beancount-format/tree/broken-malloc

you can re-produce this bug by following the steps in readme

trim21:
also cc amaanq maybe?

WillLillis:
> also cc amaanq maybe?

I've added a request for his review, however all work on tree-sitter is done on a volunteer basis so there's no guarantee he'll have a chance to take a look anytime soon.

crates/language/src/language.rs
```diff
@@ -1,4 +1,3 @@
-#![no_std]
```
ObserverOfTime, line 1: Ideally the language crate shouldn't rely on `std`.

trim21: Yeah, but the malloc need it, and tree sitter parser need a malloc everywhere.

ObserverOfTime: Then we should fix the implementation in C.

trim21: yeah ideally someone fix the c malloc.

Another option is that we make `no_std` when it's not wasm32, and std when it's not wasm32? for example, `#![cfg_attr(not(all(target_arch = "wasm32", target_os = "unknown")), no_std)]` I'm not sure how many users need `no_std` in `wasm32-unknown-unknown`

trim21:
I won't spend time to fix the bug of current c malloc. if using rust allocator is not a option please close this PR.

maxbrunsfeld:
That's fine. I get why you'd suggest switching mallocs btw; totally valid.

I just believe that there are still some advantages to having a special-purpose malloc for these allocations that only live for the lifetime of one `parse` call, in order to achieve faster malloc/free in the common case, and especially smaller code size.

trim21:
There may be some misunderstanding, this use rust's alloc, but not necessary to be a general purpose malloc, users can replace global_allocator in rust the code in this PR will use it.

maxbrunsfeld:
I don't think there's any misunderstanding. I want to provide, with Tree-sitter, a special-purpose malloc. This PR removes that.

PR #5206, opened 2026-01-05
make it build with zig 0.16

dchinmay2:


WillLillis:
This is a very welcome PR, but will have to wait until 0.16.0 is released. Because of the high amount of pre 1.0 churn, we only target tagged releases for Zig.

clason:
There's no guessing how many breaking changes between here and 0.16, so no point in keeping this open.

clason:
(Also, don't forget the bindings. We do _not_ want to mix and match.)

PR #5212, opened 2026-01-07
Fix python free-threading compatibility

kevmo314:
Under free-threaded python we cannot use the stable ABI. I think this check was missed, since it seems like the other cases are caught? In either case, this results in the downstream languages producing incorrectly tagged wheels when building with a free-threaded python interpreter.

### DIFF
--- a/crates/cli/src/templates/setup.py
+++ b/crates/cli/src/templates/setup.py
@@ -32,7 +32,7 @@ class BuildExt(build_ext):
 class BdistWheel(bdist_wheel):
     def get_tag(self):
         python, abi, platform = super().get_tag()
-        if python.startswith("cp"):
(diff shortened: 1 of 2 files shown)


talebsabreen-a11y (dismissed):

crates/cli/src/templates/setup.py
```diff
@@ -32,7 +32,7 @@ def build_extension(self, ext: Extension):
 class BdistWheel(bdist_wheel):
     def get_tag(self):
         python, abi, platform = super().get_tag()
-        if python.startswith("cp"):
+        if python.startswith("cp") and not get_config_var("Py_GIL_DISABLED"):
```
talebsabreen-a11y: - [ ]

talebsabreen-a11y: - [ ]

PR #5220, opened 2026-01-10
fix: Naming; Lifetimes; `deny(rust_2018_idioms)`

DanikVitek:
- One has to think about lifetimes if a type has one:
  - `<&'a Node<'tree>>::language` now returns `Language<'tree>` instead of `Language<'a>`, as it should;
- Renamed `struct TreeCursor<'cursor>` into `struct TreeCursor<'tree>`, to be consistant with the usages and reduse confusion;
- Remove explicit "outlives" requirements from `QueryMatches`, `QueryCaptures`, and their impl blocks, because they're inferred
> [!NOTE]
> Should `'query` be renamed into `'cursor` in `QueryMatches`, `QueryCaptures` as it is named in `QueryMatch`?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,7 +79,7 @@ unused_self = "allow"
 used_underscore_items = "allow"
 
 [workspace.lints.rust]
-mismatched_lifetime_syntaxes = "allow"
+rust_2018_idioms = "deny"
 
 [profile.optimize]
 inherits = "release"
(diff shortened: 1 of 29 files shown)


clason:
Sorry. You're welcome to use any tool you want to write the code (as long as you understand it), but LLM comment spam is not welcome here.

DanikVitek:
> Sorry. You're welcome to use any tool you want to write the code (as long as you understand it), but LLM comment spam is not welcome here.

I understand. It got autoenabled when I got my student subscription and I need to figure out where to disable it. The changes themselves were written manually by me

clason:
Yeah; this was a general comment. It boggles my mind that it's impossible for repos to opt-out of "AI review" entirely.

(New PR welcome.)

DanikVitek:
> (New PR welcome.)

So I can PR tge same changes? Should I maybe open an issue first?

clason:
New PR is fine, but no guarantees of course. A good PR description starts with a description of the problem you're trying to solve, followed by a (high-level) description of the solution. It should make clear a) why this solves the problem and b) that this is an appropriate approach. (Of course, we might disagree that this is an actual problem that needs to be fixed.)

Issue #2160, opened 2023-03-28
WASM Support for Vercel Edge Functions


dqbd:
When attempting to use Web Tree-Sitter on Vercel Edge Functions, we need to be able to provide the following:

1. An already instantiated WebAssembly.Module, both for the `Parser.Language` and for `Parser.init`
2. The instantiated WebAssembly.Module might not come from the same V8 context, we cannot depend on `instanceof`
3. Avoid using `require("fs")` and `require("path")` directly, as ESBuild statically analyses the bundle and will throw an error, even if `ENVIRONMENT_IS_NODE=false`

An example repository, where the changes above are made, can be seen here: https://github.com/dqbd/tree-sitter-edge. 

Currently, the [script](https://github.com/dqbd/tree-sitter-edge/blob/main/scripts/edge-prepare.ts) operates on the output JS file directly, but I am eager to contribute these changes back to the upstream `tree-sitter/tree-sitter` repository.

ahlinc:
A PR is very welcome!

ObserverOfTime:
If this *must* be implemented with patching then it should be a separate package (e.g. `web-tree-sitter-vercel`).
If instead it can be implemented with an optional module parameter and some conditions, then a PR is welcome.

amaanq:
I'm inclined to mark this as a "won't fix" for now since there doesn't seem to be anything actionable we can do, since we don't want to rely on a hacky script to remove references to fs or path, or modify the output of the emscripten JS file at all.

PR #5222, opened 2026-01-11
wasm: add `Language.loadSync` for synchronous loading

WilcoKruijer:
Hi all,

I ran into the same problem as #2160. On Cloudflare Workers (and Vercel Edge) WASM binaries can be directly imported using `import wasmBlob from "language.wasm";`, producing a `WebAssembly.Module` type. Before these changes there was no way to load this object as language.

This PR:
- Adds `Language.loadSync(wasmModule: WebAssembly.Module): Language`
- Updates the generated `loadWebAssemblyModule` typing so `loadAsync: true` returns a `Promise`, `loadAsync: false` returns synchronously, and a widened `boolean` flag falls back to the union of both
- Includes test and documentation

This is also useful for [source phase imports](https://nodejs.org/api/esm.html#wasm-source-phase-imports), which produce a `WebAssembly.Module` directly.

Note that it was already possible to load `web-tree-sitter` itself as a Module, by doing something like (not sync, sadly):

```ts
    await Parser.init({
      instantiateWasm: async (
        imports: WebAssembly.Imports,
        successCallback: (instance: WebAssembly.Instance) => void,
      ) => {
        const instance = await WebAssembly.instantiate(treeSitterWasm, imports);
        successCallback(instance);
        return instance.exports;
      },
    });
```


amaanq:
Thanks for this! I pushed a few tweaks on top, namely I extracted the shared symbol-resolution into a private helper so `load` and `loadSync` aren't duplicating some code, and collapsed the `loadWebAssemblyModule` overloads into a single generic conditional that handles both literal and widened `loadAsync` flags.

amaanq (approved):

PR #5223, opened 2026-01-11
fix: Clarify lifetime parameters; Rust 2018 idioms

DanikVitek:
My first reason to make a PR was just to fix naming of the lifetime parameters on `TreeCursor`, as the previous name `'cursor` didn't indicate the actual reference target (the `Tree`), and was not consistent with most usages of the type.

Then I noticed that the codebase chose to hide the elided lifetime parameters in most places. Out of curiosity/being pedantic, and also having some unpleasant experience with missed lifetime parameters in APIs I decided to enable the `rust_2018_idioms` lint. And I did in fact find a couple of places where the lifetime parameter value was inferred as overly restrictive:
- `<&'a Node<'tree>>::language` had been returning `LanguageRef<'a>` instead of `LanguageRef<'tree>`, as it should have.
- `<&'a Loader>::language_configuration_*` had been returning `LanguageConfiguration<'a>` instead of `LanguageConfiguration<'static>`

I believe, that even though the elided lifetime parameters may incur a seamingly useless syntax noise, they make you think about them and one has to think about them, especially in the scenarios of assigning lifetimes to raw pointers, as is the case here.

Other changes include:
- As I first intended, renamed `struct TreeCursor<'cursor>` into `struct TreeCursor<'tree>`, to be consistant with the usages and remove confusion;
- In accordance to another sub-lint of `rust_2018_idioms`, removed explicit "outlives" requirements from `QueryMatches`, `QueryCaptures`, and their impl blocks, because they're inferred.

> [!NOTE]
> Should `'query` be renamed into `'cursor` in `QueryMatches` & `QueryCaptures`, as is in `QueryMatch`?

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -953,7 +953,7 @@ fn render_node_range(
 
 fn cst_render_node(
     opts: &ParseFileOptions,
-    cursor: &mut TreeCursor,
+    cursor: &TreeCursor,
     source_code: &[u8],
     out: &mut impl Write,
     total_width: usize,
--- a/crates/highlight/src/highlight.rs
+++ b/crates/highlight/src/highlight.rs
@@ -189,7 +189,7 @@ struct HighlightIterLayer<'a> {
     depth: usize,
 }
 
-pub struct _QueryCaptures<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]>> {
+pub struct _QueryCaptures<'query, 'tree, T: TextProvider<I>, I: AsRef<[u8]>> {
     ptr: *mut ffi::TSQueryCursor,
     query: &'query Query,
     text_provider: T,
@@ -225,7 +225,7 @@ impl<'tree> _QueryMatch<'_, 'tree> {
     }
 }
 
-impl<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]>> Iterator
+impl<'query, 'tree, T: TextProvider<I>, I: AsRef<[u8]>> Iterator
     for _QueryCaptures<'query, 'tree, T, I>
 {
     type Item = (QueryMatch<'query, 'tree>, usize);
@@ -594,6 +594,7 @@ impl<'a> HighlightIterLayer<'a> {
                     }
                 }
 
+                // SAFETY:
                 // The `captures` iterator borrows the `Tree` and the `QueryCursor`, which
                 // prevents them from being moved. But both of these values are really just
                 // pointers, so it's actually ok to move them.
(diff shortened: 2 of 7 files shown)


clason:
I think it would be better if these unrelated changes would be done in separate commits.

DanikVitek:
@clason, Split the commit

DanikVitek:
I understand. I'll revert the explicit elided lifetimes. Though exactly this change allowed me to notice the misinferred lifetimes on `LanguageRef` and `LanguageConfiguration`, which I assume were missed simply by accident and wouldn't be found otherwise for a long time

crates/highlight/src/highlight.rs
```diff
@@ -240,7 +240,10 @@ impl<'query, 'tree: 'query, T: TextProvider<I>, I: AsRef<[u8]>> Iterator
                     m.as_mut_ptr(),
                     core::ptr::addr_of_mut!(capture_index),
                 ) {
-                    let result = std::mem::transmute::<_QueryMatch, QueryMatch>(_QueryMatch::new(
+                    let result = std::mem::transmute::<
+                        _QueryMatch<'query, 'tree>,
+                        QueryMatch<'query, 'tree>,
+                    >(_QueryMatch::new(
```
WillLillis:
```suggestion
                    let result = std::mem::transmute::<_QueryMatch, QueryMatch>(_QueryMatch::new(
```

crates/cli/src/parse.rs
```diff
@@ -953,7 +953,7 @@ fn render_node_range(
 
 fn cst_render_node(
     opts: &ParseFileOptions,
-    cursor: &mut TreeCursor,
+    cursor: &TreeCursor<'_>,
```
WillLillis:
```suggestion
    cursor: &TreeCursor,
```

DanikVitek:
@WillLillis And also, what about this one?
> Should `'query` be renamed into `'cursor` in `QueryMatches` & `QueryCaptures`, as is in `QueryMatch`?

WillLillis:
> @WillLillis And also, what about this one?
> 
> > Should `'query` be renamed into `'cursor` in `QueryMatches` & `QueryCaptures`, as is in `QueryMatch`?

I think those names are correct as-is. The `cursor` lifetime on `QueryMatch` ties directly to a reference to a `QueryCursor`, while `'query` on `QueryMatches` and `QueryCaptures` ties directly to a reference to the `Query`. These objects are created via `QueryCursor::matches` and `QueryCursor::captures`, which provide bounds that `'cursor` (`&self`) lives as long as `'query` .

WillLillis:
@DanikVitek Just noticed your TODO in the commit message. Since that's been addressed, if you'd like to amend the message to remove it feel free, but no worries if you don't get to it. I'm planning on merging this sometime tomorrow night at the latest.

PR #5224, opened 2026-01-11
Update build.zig for 0.16

mdm5995:
Updates the build.zig to use new 0.16 bindings/functions/io


clason:
Thank you, but 0.16 is not released yet. We don't want to chase constant breaking changes on the dev branch.

PR #5225, opened 2026-01-12
feat: free memory automatically

theanarkh:
Manual memory management can lead to memory leak easily. I think it's better to manage memory through the `FinalizationRegistry` automatically.

test case.
```js
import { Parser, Language } from './web-tree-sitter.js';
import path from 'path';

async function main() {
  await Parser.init();
  const wasm =  path.join(process.cwd(), `../../target/release/tree-sitter-javascript.wasm`);
  const javascript = await Language.load(wasm);
  const parser = new Parser();
  parser.setLanguage(javascript);
  parser.parse('abc + cde');
  setInterval(() => {
    gc();
  }, 100);
}

main();
```
1. Add `console.log('finalizationRegistry', value);` in callback of `FinalizationRegistry`(`finalization_registry.ts`).
2. Run `npm run build && node --expose-gc demo.js`, the output is as follow.
```
finalizationRegistry { type: 1, argv: [ 474240 ] }
finalizationRegistry { type: 2, argv: [ 460032, 462512 ] }
```

### DIFF
--- a/lib/binding_web/src/finalization_registry.ts
+++ b/lib/binding_web/src/finalization_registry.ts
@@ -1,0 +1,8 @@
+export function newFinalizer<T>(handler: (value: T) => void): FinalizationRegistry<T> | undefined {
+  try {
+    return new FinalizationRegistry(handler);
+  } catch(e) {
+    console.error('Unsupported FinalizationRegistry:', e);
+    return;
+  }
+}
--- a/lib/binding_web/src/lookahead_iterator.ts
+++ b/lib/binding_web/src/lookahead_iterator.ts
@@ -1,5 +1,10 @@
 import { C, Internal, assertInternal } from './constants';
 import { Language } from './language';
+import { newFinalizer } from './finalization_registry';
+
+const finalizer = newFinalizer((address: number) => {
+  C._ts_lookahead_iterator_delete(address);
+});
 
 export class LookaheadIterator implements Iterable<string> {
   /** @internal */
@@ -13,6 +18,7 @@ export class LookaheadIterator implements Iterable<string> {
     assertInternal(internal);
     this[0] = address;
     this.language = language;
+    finalizer?.register(this, address, this);
   }
 
   /** Get the current symbol of the lookahead iterator. */
@@ -27,6 +33,7 @@ export class LookaheadIterator implements Iterable<string> {
 
   /** Delete the lookahead iterator, freeing its resources. */
   delete(): void {
+    finalizer?.unregister(this);
     C._ts_lookahead_iterator_delete(this[0]);
     this[0] = 0;
   }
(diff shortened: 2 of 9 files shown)


maxbrunsfeld:
Does this work in all browsers?

theanarkh:
I think most browsers support `FinalizationRegistry`, we do nothing if unsupported.
image

lib/binding_web/src/disposable.ts
```diff
@@ -0,0 +1,19 @@
+export class Disposable<T> {
```
maxbrunsfeld: Curious why you called this `Disposable`. Is there some precedent for that somewhere? Elsewhere, I've seen the word `Disposable` refer to a *single* cleanup task (as opposed to a registry of cleanup tasks).

theanarkh: Thanks. You are right, `dispose` is used to release resources. How about `Cleaner`?

maxbrunsfeld: Can we just have one FinalizationRegistry for the whole library? Then `registerFinalizer` and `unregisterFinalizer` could just be free functions, we wouldn't really need to have a wrapper clas..

theanarkh: Yes, I have updated the code. Please help review again. Thanks.

maxbrunsfeld: I'm sorry, I wasn't thinking through the fact that the `FinalizationRegistry` takes a handler as a constructor argument. I actually think my suggestion was bad, and the code structure that you had before was better; I liked how the deletion function appeared in the same module that it belonged to (like `ts_tree_delete` was in `tree.ts`).

I actually think we should go back to something close to what you had originally, except instead of a wrapper class, we could just have a helper function that builds a `FinalizationRegistry`, or returns `null` if the feature is not supported. Then, at the call sites where we'd be calling `register` and `unregister` (in constructors and destructors), we'd just use a null guard (`finalizers?.register(this)`).

maxbrunsfeld: Thanks for following up; I think this is a very big improvement to the library; I did not realize that JS had finalizers now.

theanarkh: Done ! This approach is clearer than before😄.

theanarkh:
@maxbrunsfeld  Hi ! Can you please help review again. Thanks !

theanarkh:
Are there any dependencies between these objects (`LookaheadIterator`, `TreeCursor`, `Query`, `Tree`, `Parser`) ? for example. Can we call  `Tree.delete` before `TreeCursor.delete` ? Or is it safe to call `free` directly in `FinalizationRegistry` handler ?

PR #5228, opened 2026-01-13
docs: various updates

WillLillis:
- Partially addresses https://github.com/tree-sitter/tree-sitter/pull/5199#issuecomment-3714376444
  -  CC @clason Different versions of wasi-sdk will produce slightly different wasm blobs, which means we can't do a dumb generate and compare in the CI run. I think we _should_ be able to check and make sure `lib/src/wasm/wasm-stdlib.h` is changed if anything inside `crates/language/wasm/`is changed. That should catch the majority of cases where we simply forgot to run the xtask command. I'll try to have separate PR up for that in a bit
- In general the markdown docs follow the rule that no new words start after column 120, instead breaking onto a newline. This is simply for ease of editing (i.e. without having to worry about `:set nowrap`). The last commit in this PR just enforces that for all of the existing files.


crates/cli/README.md
```diff
@@ -27,16 +28,22 @@ You can also download a pre-built binary for your platform from [the releases pa
...
 ### Dependencies
 
-The `tree-sitter` binary itself has no dependencies, but specific commands have dependencies that must be present at runtime:
+The `tree-sitter` binary itself has no _required_ dependencies, but specific commands have dependencies that must be present
+at runtime:
 
 * To generate a parser from a grammar, you must have [`node`](https://nodejs.org) on your PATH.
 * To run and test parsers, you must have a C and C++ compiler on your system.
 
+In order to use the native js runtime for the `generate` command, clang is required at build-time for the CLI. If clang
```
clason: That is actually not quite the issue:

https://github.com/quickjs-ng/quickjs/issues/817#issuecomment-3720847517

So it's a Debian-only quirk.

docs/src/6-contributing.md
```diff
@@ -93,7 +93,7 @@ cargo xtask build-wasm-stdlib
 
 This command looks for the [Wasi SDK][wasi_sdk] indicated by the `TREE_SITTER_WASI_SDK_PATH`
 environment variable. If you don't have the binary, it can be downloaded from wasi-sdk's [releases][wasi-sdk-releases]
-page.
+page. Note that any changes to `crates/language/wasm/**` requires rebuilding the tree-sitter Wasm stdlib.
```
clason: via `cargo xtask build-wasm-stdlib`.

crates/cli/README.md
WillLillis: Oh interesting. I ran into a similar problem when trying to build the CLI from source on a Windows machine and assumed it was the same issue. For the sake of our docs, which requirement do you think we should list?

clason: Good question. It's about some headers but I am not sure if we can give a general guideline.

WillLillis: Yeah, I don't think there's any good general info we can give here. We may just have to deal with the issues case by case as they come in. It seems like the problem is google-able enough with the issue you linked, and at least one other place (https://github.com/mimblewimble/grin-wallet/issues/219).

PR #5229, opened 2026-01-13
feat(ci): ensure `wasm-stdlib.h` is regenerated when wasm stdlib source files are modified

WillLillis:
Follow up to https://github.com/tree-sitter/tree-sitter/pull/5199#issuecomment-3714376444

If any files inside `crates/language/wasm/` are changed and `lib/src/wasm/wasm-stdlib.h` is *not* modified, the check will fail. This is a fairly rough check, but should catch the common case of PR authors updating wasm stdlib source files and forgetting (or not knowing) to run `cargo xtask build-wasm-stdlib`.

Tested in my fork here: https://github.com/WillLillis/tree-sitter/pull/8 (ignore [f62cf2b](https://github.com/WillLillis/tree-sitter/pull/8/commits/f62cf2b5826bd19151ae89d65a75f598b7f7b75d) and [d5e4bcd](https://github.com/WillLillis/tree-sitter/pull/8/commits/d5e4bcd6fd9a36841f0a7d0079900d209636eec8), those changes have been squashed into the parent commit)

### DIFF
--- a/.github/workflows/wasm_stdlib.yml
+++ b/.github/workflows/wasm_stdlib.yml
@@ -1,0 +1,19 @@
+name: Check Wasm Stdlib build
+
+on:
+  workflow_call:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Check directory changes
+        uses: actions/github-script@v7
+        with:
+          script: |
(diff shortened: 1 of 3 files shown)


.github/workflows/wasm_stdlib.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
clason: v6

clason (approved):

Issue #5205, opened 2026-01-05
malloc for rust wasm32 target is buggy


trim21:
### Problem

`crates/language/wasm/src/stdlib.c` is buggy, you will get `indirect call type mismatch` `uninitialized element` `out of bounds memory access` `unreachable` randomly, I'm not sure why.

<details><summary>snapshot</summary>

<img width="2556" height="591" alt="image" src="https://github.com/user-attachments/assets/27c68cff-e8fa-465c-bd60-7493119a9235" />

<img width="2457" height="579" alt="image" src="https://github.com/user-attachments/assets/5c1afc8a-7067-4563-9102-2af7816388d2" />

</details>

### Steps to reproduce

```
git clone --branch broken-malloc --depth -1 https://github.com/trim21/beancount-format

# build wasm32 file
export RUSTFLAGS="-C link-args=--max-memory=4294967296 -C link-arg=--initial-heap=104857600"
cargo build -p dprint-plugin-beancount --target wasm32-unknown-unknown --features wasm --release
cp ./target/wasm32-unknown-unknown/release/dprint_plugin_beancount.wasm ./plugin.wasm


# install dprint
pnpm install

# dprint fill load wasm32 plugin and execute our code
pnpm exec dprint fmt
```

### Expected behavior

it should exit without error

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.26.3

### Operating system/version

windows

### workaround

If someone hit this bug and don't want to fix the c malloc, you can try the code in this PR https://github.com/tree-sitter/tree-sitter/pull/5201

maxbrunsfeld:
I haven't debugged in detail, but the Beancount grammar works fine in Zed, which relies on this malloc implementation.

<img width="1879" height="1558" alt="Image" src="https://github.com/user-attachments/assets/ce029dea-a0da-4075-87c4-25052fde7da0" />

trim21:
> I haven't debugged in detail, but the Beancount grammar works fine in Zed, which relies on this malloc implementation.

Does that parser in zed also using the malloc provided by `crates/language/wasm/src/stdlib.c` when building wasm32?

maxbrunsfeld:
I can't tell what you mean. Zed uses the native Tree-sitter rust crate, and all grammars, such as Beancount, are distributed as wasm32 files. The stdlib file is compiled to wasm32 and included as part of the rust crate, for use with Wasm grammars.

trim21:
[...]

PR #5234, opened 2026-01-14
 fix(wasm): realloc should handle smaller new size

trim21:
fix https://github.com/tree-sitter/tree-sitter/issues/5205

I know I said I won't spend time on it, but I tried LLM on it and it's amazing, it find this very easily and I think this is right. also tested locally in dprint it works as expected.

In realloc we didn't check the src size and dest size and truncate the memory in memcpy.


crates/language/wasm/src/stdlib.c
```diff
@@ -129,7 +129,10 @@ void *realloc(void *ptr, size_t new_size) {
   }
 
   void *result = malloc(new_size);
-  memcpy(result, &region->data, region->size);
+  if (result == NULL) return NULL;
+
+  size_t copy_size = region->size < new_size ? region->size : new_size;
+  memcpy(result, ptr, copy_size);
```
WillLillis: What's the motivation behind changing the `src` argument here from `&region->data` to `ptr`? Is that part of the fix or just an incidental change?

trim21: They should be eq here.   When I wrote this I was thinking from what realloc should do, not think from our implement details.

trim21:
After I create this pr I realize we still need to free old ptr, I'll it it later

WillLillis:
Changes look good to me, but I'd like to spend some time testing them out. In the meantime, if you could try to come up with a test case so we could have some coverage here that would help a great deal. This can be especially difficult for wasm but it's worth a shot. You may be able to mimic the allocation pattern that causes the failure in a new fixture grammar's external scanner. There are some examples you can work off of in `crates/cli/src/tests/wasm_language_tests.rs` (in particular, loading a fixture grammar in `test_load_fixture_language_wasm`).

crates/language/wasm/src/stdlib.c
WillLillis: They _should_ be equal, but all of our wasm memory APIs internally deal with `Region`s instead of raw pointers from the user. Unless you have a strong reason to change it, let's respect Chesterton's Fence here and keep the old argument.

trim21:
> Changes look good to me, but I'd like to spend some time testing them out. In the meantime, if you could try to come up with a test case so we could have some coverage here that would help a great deal. This can be especially difficult for wasm but it's worth a shot. You may be able to mimic the allocation pattern that causes the failure in a new fixture grammar's external scanner. There are some examples you can work off of in `crates/cli/src/tests/wasm_language_tests.rs` (in particular, loading a fixture grammar in `test_load_fixture_language_wasm`).

Sorry, I can't reproduce this in testing

WillLillis:
> > Changes look good to me, but I'd like to spend some time testing them out. In the meantime, if you could try to come up with a test case so we could have some coverage here that would help a great deal. This can be especially difficult for wasm but it's worth a shot. You may be able to mimic the allocation pattern that causes the failure in a new fixture grammar's external scanner. There are some examples you can work off of in `crates/cli/src/tests/wasm_language_tests.rs` (in particular, loading a fixture grammar in `test_load_fixture_language_wasm`).
> 
> Sorry, I can't reproduce this in testing

I'll spend some time trying to come up with one when I have a chance then.

WillLillis:
CC @maxbrunsfeld if you have a minute. Are the `free` calls added to `realloc` here ok? I'm assuming they were left out because the `free` implementation pre- 69c42450c32a528fc716afbce6c0f52e77776ddf didn't re-use free'd blocks in most cases, so it made sense to "leak" then. Given that's been updated, is there any reason to not `free` when appropriate in `realloc` now?

crates/language/wasm/src/stdlib.c
```diff
@@ -14,6 +14,7 @@ extern void tree_sitter_debug_message(const char *, size_t);
 
 #define PAGESIZE 0x10000
 #define MAX_HEAP_SIZE (4 * 1024 * 1024)
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
```
ObserverOfTime: Prefer inline functions over macros for min / max to avoid potential recomputation of values.

trim21:
> CC @maxbrunsfeld if you have a minute. Are the `free` calls added to `realloc` here ok? I'm assuming they were left out because the `free` implementation pre- [69c4245](https://github.com/tree-sitter/tree-sitter/commit/69c42450c32a528fc716afbce6c0f52e77776ddf) didn't re-use free'd blocks in most cases, so it made sense to "leak" then. Given that's been updated, is there any reason to not `free` when appropriate in `realloc` now?

Since https://github.com/tree-sitter/tree-sitter/issues/5205#issuecomment-3714533767 I never be able to connect the dots how wasm32 grammer file works, do you create a new wasm instance for each parsing so `free` in `realloc` is not necessary?

PR #5241, opened 2026-01-16
wasm: give binaryen the wasi-sdk treatment

### LINKED ISSUE #5235: Wasm stdlib size depends on external tools
### Problem

The size of the compiled wasm-stdlib depends on whether `wasm-opt` (from https://github.com/WebAssembly/binaryen) is available on path, since `clang` calls that by default for `-O3` _if it is on `$PATH`_.

So different people will get different stdlib blobs/headers.

### Steps to reproduce

```sh
cargo xtask build-wasm-stdlib
ll -B target/stdlib.wasm
```

### Expected behavior

1. Document this.
2. Run `wasm-opt` explicitly (with aggressive flags) as part of `build-wasm-stdlib`.
3. Add a check to the `build-wasm` xtask for `wasm-opt` and download it to `.cache/tree-sitter` if not there.

tree-sitter 0.26.3

### Operating system/version

macOS 26.3

WillLillis:
Open to discussion on what flags we should pass in addition to/instead of `-O3`.

Before merging this I'd also like to take a second look at our options for sharing this code between the loader and xtask. Keeping the two copies in sync will be error prone and annoying.

- Closes #5235


docs/src/6-contributing.md
```diff
@@ -93,7 +93,9 @@ cargo xtask build-wasm-stdlib
 
 This command looks for the [Wasi SDK][wasi_sdk] indicated by the `TREE_SITTER_WASI_SDK_PATH`
 environment variable. If you don't have the binary, it can be downloaded from wasi-sdk's [releases][wasi-sdk-releases]
-page. Note that any changes to `crates/language/wasm/**` requires rebuilding the tree-sitter Wasm stdlib via
+page. Similarly, this command also looks for [Binaryen][binaryen] indicated by the `TREE_SITTER_BINARYEN_PATH`
```
clason:
```suggestion
page. Similarly, this command also looks for [the `wasm-opt` tool from Binaryen][binaryen] indicated by the `TREE_SITTER_BINARYEN_PATH`
```

maxbrunsfeld:
Thanks for doing this. For the optimization flags, I think by far the main concern is binary size (as opposed to execution speed). Is there a `-Os` flag.

WillLillis:
> Thanks for doing this. For the optimization flags, I think by far the main concern is binary size (as opposed to execution speed). Is there a `-Os` flag.

Agreed. xtask and the loader are actually already passing `-Os` currently (I either made a typo or forgot to edit the PR description). There's also `-Oz`, which "super" focuses on code size:

```
   -Os                                          execute default optimization
                                                passes, focusing on code size

   -Oz                                          execute default optimization
                                                passes, super-focusing on code
                                                size
```

Testing locally, passing `-Oz` instead of `-Os` only nets us another 8 bytes. This could provide more significant savings in the future as the stdlib changes, though. Do you think it's worth passing?

maxbrunsfeld:
I guess let's just stick with `-Os`

clason:
I noticed a regression: `tree-sitter build --wasm`  _always_  redownloads `wasi-sdk` even when it's already cached.

PR #5242, opened 2026-01-17
fix(lib): address strict aliasing violations with `Array` type

### LINKED ISSUE #4960: "tree-sitter test" crashes on ppc64el since v0.25.7

WillLillis:
Altering the `Array` type itself isn't feasible, as this causes unacceptable breakage with existing parsers that depend on it. Instead, pass in individual `Array` fields for to various `_array__*` functions.

Any time the `contents` of an array may be modified (`free`d, `realloc`d, etc), return the potentially new address out by value. This prevents any strict aliasing violations as we're no longer writing to a type-casted pointer.

Note that the `array.h` included in a grammar's `src/tree-sitter/` directory is updated upon running `tree-sitter generate`.

This is a rework of #5194, thanks @nwf for taking an initial pass and discussing solutions!

- Closes #4960
- Supersedes #5194

### DIFF
--- a/crates/generate/src/templates/array.h
+++ b/crates/generate/src/templates/array.h
@@ -52,67 +52,91 @@ extern "C" {
 
 /// Reserve `new_capacity` elements of space in the array. If `new_capacity` is
 /// less than the array's current capacity, this function has no effect.
-#define array_reserve(self, new_capacity) \
-  _array__reserve((Array *)(self), array_elem_size(self), new_capacity)
+#define array_reserve(self, new_capacity)        \
+  ((self)->contents = _array__reserve(           \
+    (void *)(self)->contents, &(self)->capacity, \
+    array_elem_size(self), new_capacity)         \
+  )
 
 /// Free any memory allocated for this array. Note that this does not free any
 /// memory allocated for the array's contents.
-#define array_delete(self) _array__delete((Array *)(self))
+#define array_delete(self) _array__delete((self), (void *)(self)->contents, sizeof(*self))
 
 /// Push a new `element` onto the end of the array.
-#define array_push(self, element)                            \
-  (_array__grow((Array *)(self), 1, array_elem_size(self)), \
-   (self)->contents[(self)->size++] = (element))
+#define array_push(self, element)                                 \
+  do {                                                            \
+    (self)->contents = _array__grow(                              \
+      (void *)(self)->contents, (self)->size, &(self)->capacity,  \
+      1, array_elem_size(self)                                    \
+    );                                                            \
+   (self)->contents[(self)->size++] = (element);                  \
+  } while(0)
 
 /// Increase the array's size by `count` elements.
 /// New elements are zero-initialized.
-#define array_grow_by(self, count) \
-  do { \
-    if ((count) == 0) break; \
-    _array__grow((Array *)(self), count, array_elem_size(self)); \
+#define array_grow_by(self, count)                                               \
+  do {                                                                           \
+    if ((count) == 0) break;                                                     \
+    (self)->contents = _array__grow(                                             \
+      (self)->contents, (self)->size, &(self)->capacity,                         \
+      count, array_elem_size(self)                                               \
+    );                                                                           \
     memset((self)->contents + (self)->size, 0, (count) * array_elem_size(self)); \
-    (self)->size += (count); \
+    (self)->size += (count);                                                     \
   } while (0)
 
 /// Append all elements from one array to the end of another.
-#define array_push_all(self, other)                                       \
+#define array_push_all(self, other) \
   array_extend((self), (other)->size, (other)->contents)
 
 /// Append `count` elements to the end of the array, reading their values from the
 /// `contents` pointer.
-#define array_extend(self, count, contents)                    \
-  _array__splice(                                               \
-    (Array *)(self), array_elem_size(self), (self)->size, \
-    0, count,  contents                                        \
+#define array_extend(self, count, other_contents)                 \
+  (self)->contents = _array__splice(                              \
+    (void*)(self)->contents, &(self)->size, &(self)->capacity,    \
+    array_elem_size(self), (self)->size, 0, count, other_contents \
   )
 
 /// Remove `old_count` elements from the array starting at the given `index`. At
 /// the same index, insert `new_count` new elements, reading their values from the
 /// `new_contents` pointer.
-#define array_splice(self, _index, old_count, new_count, new_contents)  \
-  _array__splice(                                                       \
-    (Array *)(self), array_elem_size(self), _index,                \
-    old_count, new_count, new_contents                                 \
+#define array_splice(self, _index, old_count, new_count, new_contents) \
+  (self)->contents = _array__splice(                                   \
+    (void *)(self)->contents, &(self)->size, &(self)->capacity,        \
+    array_elem_size(self), _index, old_count, new_count, new_contents  \
   )
 
 /// Insert one `element` into the array at the given `index`.
-#define array_insert(self, _index, element) \
-  _array__splice((Array *)(self), array_elem_size(self), _index, 0, 1, &(element))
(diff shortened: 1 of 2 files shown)


WillLillis:
I believe the CI failures are now due to this line in tree-sitter-php's scanner:

https://github.com/tree-sitter/tree-sitter-php/blob/master/common/scanner.h#L468

(Here's the snippet since GH won't render it inline)

```c
array_delete(&array_pop(&scanner->heredocs).word);
```

Taking the address of a non-lvalue is undefined behavior, and this seems to be it. I've opened a PR to fix the issue here (https://github.com/tree-sitter/tree-sitter-php/pull/293), so once that's merged and a new release is published, we can update the tag in `test/fixtures/fixtures.json`, and *then* (barring any other issues with other corpus grammars) this PR can proceed.

clason:
Once the php scanner is fixed, I'd like to run a crater test with Neovim built against this PR. (I'm sure that's not the only parser with... questionable behavior out there...)

WillLillis:
I've added [29281b1](https://github.com/tree-sitter/tree-sitter/pull/5242/commits/29281b1f9320b739ea1aa47a4eb31b226de88be2), which allows `fixtures.json` to specify an override branch for a test fixture grammar to be checked out to. I've created the [upstream_test_fixture](https://github.com/tree-sitter/tree-sitter-php/tree/upstream_test_fixture) branch in the tree-sitter-php repo, and have pointed the `fixtures.json` file to use that branch instead of the `v0.24.2` tag. The changes in this branch include fixing the undefined behavior described in this [comment](https://github.com/tree-sitter/tree-sitter/pull/5242#issuecomment-3764959700), as well as regenerating the parser in order to update the repo's `array.h` copies.

CC @clason I think this is a better solution than disabling the PHP tests, especially given how hard to predict/reason about some of these crashes are.

nwf:
I am not entirely sure that this will *close* https://github.com/tree-sitter/tree-sitter/issues/4960, though it is surely a huge improvement, since, as I noted over in https://github.com/tree-sitter/tree-sitter/issues/4960#issuecomment-3707523955, there is also some strict-aliasing violation going on in `tree_cursor.c`.  I wouldn't be surprised if that isn't biting anyone _yet_, but, strictly speaking...

WillLillis:
> I am not entirely sure that this will _close_ #4960, though it is surely a huge improvement, since, as I noted over in [#4960 (comment)](https://github.com/tree-sitter/tree-sitter/issues/4960#issuecomment-3707523955), there is also some strict-aliasing violation going on in `tree_cursor.c`. I wouldn't be surprised if that isn't biting anyone _yet_, but, strictly speaking...

I was planning on extracting the violations in `tree_cursor.c` to a separate issue and keep #4960 scoped to the crashes observed on ppc64el (I still need to fight with docker in order to test these changes on there again.). I'm planning on taking a look at `tree_cursor.c` after we get this resolved. :slightly_smiling_face:

nwf:
Sounds like a plan!  Thank you!

WillLillis:
I've gone through _most_ of the grammars with external scanners in nvim-treesitter to check for further fallout from this change. Besides tree-sitter-php, The following repos crashed when regenerated and tested:
- [tree-sitter-xml](https://github.com/tree-sitter-grammars/tree-sitter-xml): https://github.com/tree-sitter-grammars/tree-sitter-xml/pull/36
- [tree-sitter-tlaplus](https://github.com/tlaplus-community/tree-sitter-tlaplus): https://github.com/tlaplus-community/tree-sitter-tlaplus/pull/142

clason:
also
* tree-sitter-vue: https://github.com/tree-sitter-grammars/tree-sitter-vue/pull/11

All three PRs are now merged, so this is good to go?

nwf:
Belatedly (sorry!), can say that `cargo xtask test` looks to have only unrelated failures on my powerpc64le Debian box.  (Not quite sure why "tests::corpus_test::test_feature_corpus_files" is failing with "called `Result::unwrap()` on an `Err` value: ScannerSymbols(ScannerSymbolError { missing: ["tree_sitter_external_unicode_column_alignment_external_scanner_create", ...", but it is, _but_ it doesn't seem relevant).

WillLillis:
> Belatedly (sorry!), can say that `cargo xtask test` looks to have only unrelated failures on my powerpc64le Debian box. (Not quite sure why "tests::corpus_test::test_feature_corpus_files" is failing with "called `Result::unwrap()` on an `Err` value: ScannerSymbols(ScannerSymbolError { missing: ["tree_sitter_external_unicode_column_alignment_external_scanner_create", ...", but it is, _but_ it doesn't seem relevant).

That shouldn't be relevant to these changes, but I'm not sure why you'd get that error. Have you `cargo xtask fetch-fixtures` and `cargo xtask generate-fixtures`?

WillLillis:
Following up on this, the fixture corpus tests are now passing for me in a ppcle64 ubuntu docker container. I have ran into the same missing symbol issue raised by @nwf above. The problem seems to be here:

https://github.com/tree-sitter/tree-sitter/blob/4f8983b89f0e891085f758930cf7238f1275ea4d/crates/loader/src/loader.rs#L1298-L1302

For some background, on non-Windows platforms we use the system installed `nm` to verify symbols in the generated parser shared library. There is a special case for powerpc64 to use the `D` section of nm's output, and `T` for all other architectures. It looks like with the a more recent version of binutils, ld now places the functions in the correct section, or something in nm changed to indicate the correct section. Potentially related discussion:

https://lists.gnu.org/archive/html/bug-binutils/2021-11/msg00024.html

PR #5158, opened 2025-12-22
Fix tree-sitter-ruby crate failing to build to Wasm

kossnocorp:
This PR makes tree-sitter-ruby crate compile to Wasm.

See [reproduction repo](https://github.com/kossnocorp/tree-sitter-ruby-wasm-repro) for detailed steps and more details about the described issues. The repo branch `fix-env` contains fully resolved problem using both tree-sitter and tree-sitter-ruby forks.

---

This is 1/2 PRs required to address the problem:

- tree-sitter tree-sitter/tree-sitter#5158
- tree-sitter-ruby tree-sitter/tree-sitter-ruby#283

---

### Missing `#include <stdlib.h>`

The problem starts when trying to build simple Rust code using tree-sitter-ruby to Wasm:

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn parse_ruby(source: &str) -> String {
    let mut parser = tree_sitter::Parser::new();
    parser
        .set_language(&tree_sitter_ruby::LANGUAGE.into())
        .unwrap();
    let tree = parser.parse(source, None).unwrap();
    tree.root_node().to_sexp()
}
```

To demonstrate the problem, I used `wasm-pack build --target nodejs`:

```rs
use wasm_bindgen_test::*;

#[wasm_bindgen_test]
fn test_parse() {
    let result = tree_sitter_ruby_wasm_repro::parse_ruby("def foo; end");
    assert!(result.contains("method"));
}
```

It fails with missing `#include <stdlib.h>` error:

```
[INFO]: 🎯  Checking for the Wasm target...
   Compiling serde_core v1.0.228
   Compiling equivalent v1.0.2
   Compiling serde_json v1.0.145
   Compiling hashbrown v0.16.1
   Compiling ryu v1.0.21
   Compiling itoa v1.0.16
   Compiling memchr v2.7.6
   Compiling tree-sitter-ruby v0.23.1
   Compiling wasm-bindgen-macro v0.2.106
   Compiling serde v1.0.228
   Compiling indexmap v2.12.1
   Compiling wasm-bindgen v0.2.106
   Compiling js-sys v0.3.83
warning: tree-sitter-ruby@0.23.1: In file included from src/parser.c:1:
warning: tree-sitter-ruby@0.23.1: src/tree_sitter/parser.h:10:10: fatal error: 'stdlib.h' file not found
warning: tree-sitter-ruby@0.23.1: #include <stdlib.h>
[...]

PR #5255, opened 2026-01-24
wasm: correct unicode support in stdlib

WillLillis:
This PR changes our minimal wasm stdlib such that all of the functions we export are our own implementations, instead of those from the wasi headers. 

There are a few benefits to this, namely the >7x size reduction for the  stdlib header (~14k -> ~2k), as well as matched behavior between parsers built for wasm (`ts b --wasm`) and rust projects built targeting `wasm32-unknown-unknown`). 

The downside is mainly the lack of full unicode support for `towupper` and `towlower`. A lot (most, it seems) of the bloat from wasi's implementations comes from unicode tables, which we don't cover in our own stdlib impl. Instead, we only handle ASCII characters, which is the approach already taken for the existing stdlib implementation. We could also just use the wasi implementations for these functions, bringing the header size back up to `7315` bytes. 

@maxbrunsfeld Do you think the tradeoff here is worth it? 

~~Edit: Not sure why the sanitizer checks are failing, will look into that later tonight.~~ Cache issue


WillLillis:
Looking into this further, I think we may not be sacrificing any functionality with this change. If I'm reading the documentation correctly for the various `wint_t` functions in `wctype.c` (for example, [`iswupper`](https://en.cppreference.com/w/c/string/wide/iswupper.html)), the unicode paths are only hit if `setlocale` was previously called (i.e. `setlocale(LC_ALL, "en_US.utf8")`). We don't export `setlocale` for wasm builds, so it actually seems very reasonable to treat these code paths as unreachable and just handle the ASCII cases as I've done in this PR.

These re-implementations will need to be tested further before I think this is ok to merge, but I believe the previous unicode concerns I raised are actually a non-issue.

A quick [GH search](https://github.com/search?q=path%3A**src%2Fscanner.c+setlocale&type=code) shows that no parsers use `setlocale` at all, besides [tree-sitter-haskell](https://github.com/tree-sitter/tree-sitter-haskell) and [tree-sitter-synquid](https://github.com/lemueldls/tree-sitter-synquid). For both of these parsers, the calls to `setlocale` are gated behind `#ifdef TREE_SITTER_DEBUG`, so they're not actually used outside of development.

maxbrunsfeld:
Oh, I actually thought those functions would use UTF-8 character categories by default. I think it would be better to keep the unicode support compiled in, and look into setting the locale somehow. I think those functions are used in some scanners where UTF8 support is intended.

WillLillis:
> Oh, I actually thought those functions would use UTF-8 character categories by default. I think it would be better to keep the unicode support compiled in, and look into setting the locale somehow. I think those functions are used in some scanners where UTF8 support is intended.

Gotcha. In that case, the recent additions to `wctype.c` should probably be moved to  `wctype.h` as `static inline`s so that the wasi implementations are used. This will grow the stdlib size by about 4k to ~18,984. Using our own implementations for the functions in `string.c` brings the size back down to `15559`. As long as those minimal impls are sufficient, that seems like a reasonable tradeoff. 

Since we don't export `setlocale`, the default `C` locale should always be in use by wasm parsers. Given this, maybe we can use a pared down version of wasi's [casemap table](https://github.com/WebAssembly/wasi-libc/blob/3f1ee99f7756015dcf5d7f33792cfa92d755e529/libc-top-half/musl/src/ctype/casemap.h#L3) and its surrounding functionality to get some more size savings since we don't need to support other locales. Just out of curiosity, I tried adding `setlocale` to the export list (`stdlibsymbols.txt`) and run into some odd duplicate symbol errors:

```
Failed to compile the Tree-sitter Wasm stdlib:
wasm-ld: error: duplicate symbol: malloc
>>> defined in /tmp/stdlib-3eba72.o
>>> defined in /home/lillis/.cache/tree-sitter/wasi-sdk/bin/../share/wasi-sysroot/lib/wasm32-wasi/libc.a(dlmalloc.o)

wasm-ld: error: duplicate symbol: free
>>> defined in /tmp/stdlib-3eba72.o
>>> defined in /home/lillis/.cache/tree-sitter/wasi-sdk/bin/../share/wasi-sysroot/lib/wasm32-wasi/libc.a(dlmalloc.o)

wasm-ld: error: duplicate symbol: calloc
>>> defined in /tmp/stdlib-3eba72.o
>>> defined in /home/lillis/.cache/tree-sitter/wasi-sdk/bin/../share/wasi-sysroot/lib/wasm32-wasi/libc.a(dlmalloc.o)

wasm-ld: error: duplicate symbol: realloc
>>> defined in /tmp/stdlib-3eba72.o
>>> defined in /home/lillis/.cache/tree-sitter/wasi-sdk/bin/../share/wasi-sysroot/lib/wasm32-wasi/libc.a(dlmalloc.o)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I also tested a rust `wasm32-unknown-unknown` build via the repro from https://github.com/tree-sitter/tree-sitter/pull/5158 and that use case is still supported with these changes.

WillLillis:
Quick update here. I worked out how to get `setlocale` working in wasm for external scanners, and have a branch up here: https://github.com/WillLillis/tree-sitter/tree/wasm_setlocale.

Locales seem to be working, at least by testing the example from the `iswupper` docs [here](https://en.cppreference.com/w/c/string/wide/iswupper.html). It does bloat the stdlib header size up to `26486` bytes. In my own [testing](https://github.com/WillLillis/tree-sitter-haskell/tree/wasm_setlocale_testing), the default locale in wasm is `C`, which I believes means that no parsers are currently able to support the UTF8 locale in wasm (unless it's possible to set the locale via other means). If we _do_ want to add support for other locales by adding `setlocale` in the wasm stdlib for external scanners, it makes sense to stick to the wasi implementations for all the exported `wint_t` functions. Otherwise, if we _don't_ want to add support for other locales, then I think it makes sense to depend on our own ASCII-only implementations and take the size savings.

Presently, there aren't any grammars (on github, at least) that set an alternative locale for non-debugging purposes.

clason:
I think there's two separate issues here:
* (custom) locale support -- I think that is out of scope and actually a bad idea;
* support for utf8 character classes -- I think that is something that would be expected nowadays and should just be unconditionally enabled.

In any case, the most important bit is to have parity between native and wasm parsers. So if we don't support utf8 for native parsers now, we don't have to do so for wasm parsers. (But I absolutely think we should add it for both; the question is whether it's worth taking the "size win" now if we re-add things in the near future.)

WillLillis:
> I think there's two separate issues here:
> 
>     * (custom) locale support -- I think that is out of scope and actually a bad idea;
> 
>     * support for utf8 character classes -- I think that is something that would be expected nowadays and should just be unconditionally enabled.
> 

:+1: 

> In any case, the most important bit is to have parity between native and wasm parsers. So if we don't support utf8 for native parsers now, we don't have to do so for wasm parsers. (But I absolutely think we should add it for both; the question is whether it's worth taking the "size win" now if we re-add things in the near future.)

I believe utf8 is supported out of the box for native parsers, it's just that none of them use it. This is assuming the default locale is "C" across the board (it seems to be using `tree-sitter build`, at least). @amaanq and I discussed providing our own implementation for `setlocale`, supporting _just_ utf8 for wasm, which should hopefully offset a majority of the size penalty seen from using wasi-sdk's.

clason:
Can't we just hardcode utf8 support (if we provide our own stdlib)? Why is messing with locale necessary?

WillLillis:
> Can't we just hardcode utf8 support (if we provide our own stdlib)? Why is messing with locale necessary?

My understand is that the default locale is "C", which does not include utf8 support. So hardcoding it in the wasm stdlib would cause parsers to behave differently in wasm vs. native.

maxbrunsfeld:
Yeah, it would be great if we could just get `iwalpha` and friends to just assume utf8. Not sure how to make that happen though.

clason:
That would be an argument for just rolling our own (utf8 version) regardless of platform and target?

WillLillis:
Superseded by #5851

PR #5261, opened 2026-01-26
fix(wasm): correct several bugs in `realloc`

WillLillis:
- free memory if 0 size is passed in
- Don't `memcpy` contents if new pointer is `NULL`
- Copy old region's contents only up to size of new region
- free old region

- also exit early from `calloc` if call to `malloc` fails

Supersedes #5234 with more _minimal_ changes and test coverage for the bug.

CC @Trim21 I wasn't able to get `beancount-format` to build locally to give this change a try on your original repro from #5205. If you get a chance to try it out let me know!

- Closes #5205

PR #5264, opened 2026-01-27
fix(wasm): correct several bugs in realloc

WillLillis:
Backport of #5261 failed. Here's a manual one.


trim21:
is there a ETA on next 0.26 release?

clason:
Soon(tm). We need to make sure these fixes actually fix things (so testing and confirming helps move this along).

PR #5267, opened 2026-01-27
fix(generate): error if a supertype is defined as a terminal

### LINKED ISSUE #5265: Misuse of `supertypes` leads to spurious errors.
### Problem

https://github.com/user-attachments/assets/b062ce36-7466-42e5-9da2-aa1d02a9f5d8

If a rule is added to the `supertypes` array but that rule isn't really a supertype as described in [Tree-sitter docs](https://tree-sitter.github.io/tree-sitter/creating-parsers/3-writing-the-grammar.html#using-supertypes), it may be the reason for unexpected behaviour described below.

An error message is also displayed naming another rule as an being an invalid supertype. This is misleading, making it difficult to debug.

### Steps to reproduce

1. Use the below Grammar to reproduce the issue.
2. Run Tree-sitter cli command `tree-sitter generate`.

The misleading error is printed. You can further experiment by altering the grammar and again running the `generate` command to observe the unexpected behaviour as described in the notes below.

```javascript
module.exports = grammar({
  name: 'foo',

  // This seems to be the root of the problem as $.A is not really a supertype
  supertypes: $ => [$.A],

  extras: $ => [$.lineComment, $.lineComment2],

  rules: {
    source_file: $ => choice(
      $.B,
      $.A,
    ),

    lineComment: $ => /;;.*/,

    lineComment2: $ => /##.*/,

    A: $ => /xyz/,

    // It seems if the rule C appears before it is used, then problem occurs
    C: $ => "abc",

    B: $ => seq('abc', $.C),
  }
});
```

#### Notes to better understand the issue

1. I guess the actual error was to add rule `A` to `supertypes`. Rule `A` is not a valid supertype as per the [Tre-sitter docs](https://tree-sitter.github.io/tree-sitter/creating-parsers/3-writing-the-grammar.html#using-supertypes)

However the below error message names rule `B` as not being a valid supertype:

> Grammar error: Supertype symbols must always have a single visible child, but `B` can have multiple

2. Another observation: If the string in rule `C` is changed from `"abc"` to `"xyz"`, it leads to a panic:

>     thread 'main' (176176) panicked at crates/generate/src/node_types.rs:350:18:
>     index out of bounds: the len is 2 but the index is 2
>     note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

3. Another observation: Re-ordering of rules (move rule B before rule C) gets rid of the error.  Sometimes re-ordering of rules names yet another rule to be the invalid supertype. 

4. I guess the actual fix is to remove rule `A` from `supertypes` because it does not belong there in the first place. After doing so, everything works fine but it was difficult to debug.  (The original Grammar where this error first occurred has many more rules making it very difficult to figure out the problem)

### Expected behavior

The error shoud not indicate a problem with rule `B` as a being a invalid supertype because that rule isn't in the `supertypes` and the actual problem seems to be with rule `A`.

tree-sitter library **0.26.3** (compiled from source)
tree-sitter cli 0.26.3 (installed via Node.js)

### Operating system/version

Debian Trixie (13.3)

WillLillis:
The assumption that supertypes are nonterminals (and thus *not* moved into the lexical grammar following token extraction) is very baked into the generate code. When this assumption is broken, a common crash (or worse, silent corruption) happens when the terminal supertype's `index` is used to access the grammar's `variables`. There are several places where this error could be caught, but this was the earliest reasonable place. If a supertype symbol has been marked as a terminal, it must have been moved into the lexical grammar, so we grab its name that way and report the error to the user.

Open to suggestions for improving the error message.

- Closes #5265

### DIFF
--- a/crates/generate/src/prepare_grammar/extract_tokens.rs
+++ b/crates/generate/src/prepare_grammar/extract_tokens.rs
@@ -21,6 +21,8 @@ unless they are used only as the grammar's start rule.
 "
     )]
     EmptyString(String),
+    #[error("Terminal rule '{0}' cannot be used as a supertype")]
+    SupertypeTerminal(String),
     #[error("Rule '{0}' cannot be used as both an external token and a non-terminal rule")]
     ExternalTokenNonTerminal(String),
     #[error("Non-symbol rules cannot be used as external tokens")]
@@ -128,11 +130,18 @@ pub(super) fn extract_tokens(
         })
         .collect();
 
-    let supertype_symbols = grammar
+    let supertype_symbols: Vec<Symbol> = grammar
         .supertype_symbols
         .into_iter()
         .map(|symbol| symbol_replacer.replace_symbol(symbol))
         .collect();
+    for supertype_symbol in &supertype_symbols {
+        if supertype_symbol.is_terminal() {
+            Err(ExtractTokensError::SupertypeTerminal(
+                lexical_variables[supertype_symbol.index].name.clone(),
+            ))?;
+        }
+    }
 
     let variables_to_inline = grammar
         .variables_to_inline
--- a/test/fixtures/test_grammars/terminal_supertype/expected_error.txt
+++ b/test/fixtures/test_grammars/terminal_supertype/expected_error.txt
@@ -1,0 +1,1 @@
+Terminal rule 'A' cannot be used as a supertype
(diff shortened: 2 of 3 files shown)


kayomarz:
I compiled the cli after cherry-picking your fix in version 0.26.3. It's working fine.
Thank you.

PR #5268, opened 2026-01-28
fix(loader): account for nm/ld fix on newer powerpc linux toolchains

WillLillis:
Relying on a user's system's installation of `nm` has proven to be bug
prone and flaky. Instead, we can enforce that these symbols are defined
by requiring the linker to resolve them. This is already the default on
macos, but linux has looser requirements which defers the error to when
the library is opened. This check was not run on Windows (because
there's no `nm`), but the msvc linker is similiarly strict to macos's
w.r.t. resolving symbols at build time rather than runtime, so there's
no issue here.

The first commit here provides a patch for the issue raised in https://github.com/tree-sitter/tree-sitter/pull/5242#issuecomment-3795456090. We had an workaround for buggy `ld`/`nm` on ppc64, where functions were reported to be in the data section rather than the text section. The text section is now reported correctly for Ubuntu latest, but we also need to maintain compatibility with legacy systems.

The main change here is that If a grammar with an external scanner is built via `tree-sitter b` and one of the required functions is missing, the user will receive a linker error about a missing symbol. For example, removing `tree_sitter_vim_external_scanner_create` from tree-sitter-vim's `src/scanner.c` will result in:

```
[lillis@cachyos-x8664] ~/projects/grammars/tree-sitter-vim (master) ⚡
❯ tree-sitter b                                                                                                                                             ⏎
Error: Failed to compile parser

Caused by:
    Parser compilation failed.
    Stdout:
    Stderr: /home/lillis/projects/grammars/tree-sitter-vim/src/scanner.c: In function 'try_lex_heredoc_marker':
    /home/lillis/projects/grammars/tree-sitter-vim/src/scanner.c:158:3: warning: 'strncpy' output may be truncated copying between 1 and 31 bytes from a string of length 254 [-Wstringop-truncation]
      158 |   strncpy(scanner->heredoc_marker, marker, marker_len);
          |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    /usr/bin/ld: /tmp/ccfQLauO.o:(.data.rel.ro.language.0+0xb8): undefined reference to `tree_sitter_vim_external_scanner_create'
    collect2: error: ld returned 1 exit status
```

I've opted to leave the `nm` check in for non tree-sitter non-static symbols in, as 
- this only produces a warning rather than a hard error preventing creation of the parser library
- Leaving such a nonstatic function in can be non-trivial to catch (i.e. the parser will build and test without issues)

One open question, do we want to provide an "escape hatch" to skip the `nm` check? We could use the `NM` environment variable and not perform the check if it's equal to `skip` or something similar.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1,6 +1,8 @@
 #![cfg_attr(not(any(test, doctest)), doc = include_str!("../README.md"))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
+#[cfg(unix)]
+use std::fmt::Write as _;
 #[cfg(any(feature = "tree-sitter-highlight", feature = "tree-sitter-tags"))]
 use std::ops::Range;
 #[cfg(feature = "tree-sitter-highlight")]
@@ -110,8 +112,6 @@ pub enum LoaderError {
     NoLanguage,
     #[error(transparent)]
     Query(LoaderQueryError),
-    #[error(transparent)]
-    ScannerSymbols(ScannerSymbolError),
     #[error("Failed to load language for scope '{0}':\n{1}")]
     ScopeLoad(String, Box<Self>),
     #[error(transparent)]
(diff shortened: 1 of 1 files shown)


clason:
> One open question, do we want to provide an "escape hatch" to skip the nm check? We could use the NM environment variable and not perform the check if it's equal to skip or something similar.

I think long-term (possibly for 0.27 already) we can just drop the check?

WillLillis:
> > One open question, do we want to provide an "escape hatch" to skip the nm check? We could use the NM environment variable and not perform the check if it's equal to skip or something similar.
> 
> I think long-term (possibly for 0.27 already) we can just drop the check?

I'm not sure how often this issue arises when working on a grammar's external scanner. It would be very nice to simplify things and remove the check, but I am a bit worried that it may introduce some issues downstream.

PR #5271, opened 2026-01-29
Fix aliased nodes being missing from node_types.json

### LINKED ISSUE #1654: Aliasing to a Grammar Symbol without a Production Produces an AST without a Corresponding Node in node-types.json

nike4613:
Fixes #1654

### DIFF
--- a/crates/generate/src/node_types.rs
+++ b/crates/generate/src/node_types.rs
@@ -650,6 +650,71 @@ pub fn generate_node_types_json(
         }
     }
 
+    // Handle aliases that don't correspond to any existing named type.
+    // These are aliases that are only used as aliases and never as direct references.
+    for (symbol, aliases) in &aliases_by_symbol {
+        for alias_opt in aliases {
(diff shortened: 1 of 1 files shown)


clason:
Is this still an issue after #5389? (Please check the latest nightly build.)

If it is, this needs to be rebased on those changes (which change the logic, so probably a different approach is needed).

WillLillis:
Stale, had some issues with the implementation, and superseded by #5874

PR #5274, opened 2026-01-30
fix: wasm32 stdlib realloc on heap end

trim21:
Like I said before, https://github.com/tree-sitter/tree-sitter/pull/5261#issuecomment-3798091474 , there is bug when `region_end == next` in realloc.

The previous code call `malloc(new_size)`. but `malloc()` first searches the `free_list`, so it may return a different block and leave next at the old region. but since we didn't do any memcpy, it will definitely break something.

The easy fix is to remove this "optimize" and just call malloc and memcpy, 
there another fix that we can grow the heap in realloc, but it's complex and hard to maintain IMO.

https://github.com/tree-sitter/tree-sitter/pull/5261 plus this, you will finally get what I had in https://github.com/tree-sitter/tree-sitter/pull/5234 ...

@WillLillis

Also a proof it still has bug:

image


WillLillis:
Thanks for identifying the bug! This definitely needs a test case though, and I'd prefer to not remove this optimization. I'll take care of it

PR #5276, opened 2026-01-31
wasm: fix `realloc` optimization

WillLillis:
There is an optimization in the wasm stdlib `realloc` implementation where, if we are reallocating the most recently allocated region, we can simply resize that region and return the same pointer. This skips a potentially expensive `memcpy`.

This optimization in `realloc` was tightly coupled with the internal implementation of `malloc`, however. The `malloc` implementation was altered in #4818, which silently broke a built in assumption by pulling regions off of the free list. When this happens, we get whatever data was left in the freed region, returning bad data to the user since no `memcpy` was performed between the old and new regions.

To fix this, we just need to resize the region in place. This is what the old implementation intended to do implicitly through its call to `malloc`.

This PR also adds a matching test case. Because we now have multiple wasm realloc-related test cases, I renamed the older one to be more specific in the first commit.

CC @trim21 I've added you as a Co-author to the second commit here since you were able to narrow down the issue, thanks! If you have a chance to try this out with your project that would be greatly appreciated. Otherwise I should have time to get to it sometime this weekend.

- Supersedes #5274


trim21:
I think this fixed the issue.

this is my idea, not sure if it's easy to implement, can we has a dedicated fuzzy testing for wasm32 allocator directly? it can replace the previous tests trying to re-produce the bugs

WillLillis:
> I think this fixed the issue.

👍 

> this is my idea, not sure if it's easy to implement, can we has a dedicated fuzzy testing for wasm32 allocator directly? it can replace the previous tests trying to re-produce the bugs

Fuzzing tests for our wasm allocators is definitely something we could explore, but it's a nontrivial problem and not a huge priority at the moment. Such fuzz testing would definitely _not_ replace the existing tests. These tests are added so we can reliably verify that certain behaviors work. If a test fails, we know exactly what's broken. Fuzz testing can be useful for uncovering _new_ bugs, but it isn't a replacement for existing coverage.

PR #5285, opened 2026-02-01
feat(cli): add `--verbose` flag to `build` command

WillLillis:
This displays the working directory of the command (if present), compiler used, all of its arguments, and any environment variables set.

tree-sitter-vim with various build commands:

image

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -188,6 +188,9 @@ struct Build {
     /// Compile a parser in debug mode
     #[arg(long, short = '0')]
     pub debug: bool,
+    /// Display verbose build information
+    #[arg(short, long)]
+    pub verbose: bool,
 }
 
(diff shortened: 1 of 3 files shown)


clason:
I like it! But I think we should not backport this. It's better for CLI flags to be clearly tied to a "major" version (so "this is available in 0.27 but not 0.26").

(Once it's in, we are more free to change _what_ is printed.)

WillLillis:
> I like it! But I think we should not backport this. It's better for CLI flags to be clearly tied to a "major" version (so "this is available in 0.27 but not 0.26").
> 
> (Once it's in, we are more free to change _what_ is printed.)

Works for me. I forgot to add stdout/stderr on my first pass, so that's included now, i.e.:

image

This should be good to go now.

clason (approved):

PR #5292, opened 2026-02-02
ci: add a dry-run workflow to test rust releases

WillLillis:
This is intended to be a manual safety check run locally before opening a PR to bump version numbers for a release.

Ideally this would check for changes to a crate if a version is already published so we could run this as part of the release ci job, but given that:

- Releases can (and have been) created directly via crates.io without a corresponding tag in the repo
- Tags don't line up with release versions for every crate (tree-sitter-language specifically, which has to stay at 0.1.x)
- This would require getting information via the github API, which would mean a lot of slow, awkward testing in CI runners
- This level of automation will almost surely come with edge cases and leave us back where we started

I've opted to keep the check a manual one and leave the decision to bump or not to the maintainer making the release.

Also includes a few related docs improvements.

### DIFF
--- a/docs/src/6-contributing.md
+++ b/docs/src/6-contributing.md
@@ -136,8 +136,13 @@ several packages that are published to package registries for different language
 
 * Rust crates on [crates.io][crates]:
   * [`tree-sitter`][lib crate] — A Rust binding to the core library
-  * [`tree-sitter-highlight`][highlight crate] — The syntax-highlighting library
+  * [`tree-sitter-config`][config crate] — User configuration of the command-line tool
   * [`tree-sitter-cli`][cli crate] — The command-line tool
+  * [`tree-sitter-generate`][generate crate] — The parser-generation library
+  * [`tree-sitter-highlight`][highlight crate] — The syntax-highlighting library
+  * [`tree-sitter-language`][language crate] — A language type for grammars to interact with the core library
+  * [`tree-sitter-loader`][loader crate] — The parser building and loading library
+  * [`tree-sitter-tags`][tags crate] — The syntax-tagging library
 
 * JavaScript modules on [npmjs.com][npmjs]:
   * [`web-tree-sitter`][web-ts] — A Wasm-based JavaScript binding to the core library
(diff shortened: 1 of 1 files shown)


clason:
Running this as part of the release workflow would be the real value here, so I think it would make sense to invest more time into seeing if we can extract the [part of the publish action that does this check](https://github.com/katyo/publish-crates/blob/main/src/package.ts).

Actually, the release workflow is already too late -- this needs to be a regular workflow that is run on the "bump" commit (or any commit that updates the language crate).

clason:
See also https://github.com/katyo/publish-crates/pull/663

docs/src/6-contributing.md
```diff
@@ -152,6 +157,20 @@ published as [`tree-sitter`][py package] on [PyPI.org][pypi].
 * [`tree-sitter/go-tree-sitter`][go ts] — Go bindings to the core library,
 published as [`tree_sitter`][go package] on [pkg.go.dev][go.dev].
 
+### Publishing a Release
+
+Tree-sitter's develoment follows two main release schedules. Primary development of the next major release is conducted
+on the `master` branch. Non-breaking fixes are backported to the `release-<previous-major-version>` branch. In order to
+publish a new release for either, you should:
```
clason: 💯 for documenting this, but we should distinguish here between patch and minor releases (since the workflow is slightly different).

WillLillis:
> Actually, the release workflow is already too late -- this needs to be a regular workflow that is run on the "bump" commit (or any commit that updates the language crate).

Couldn't we run it as part of the release workflow _before_ the steps creating the releases (e.g. alongside the tests)? I'm not sure how we'd set up triggers for the workflow to only run on "bump" commits, and if we run it on _every_ PR then it may get ignored.

clason:
If it's part of the release workflow, it runs _after_ we bumped and tagged -- so if we do need to bump the language crate, we can only do it after the bump, so we'd have to fix, bump, and tag again, same as now.

As to how: we can gate this either behind the commit message (regex) or a label we apply to release commits.

WillLillis:
> Running this as part of the release workflow would be the real value here, so I think it would make sense to invest more time into seeing if we can extract the [part of the publish action that does this check](https://github.com/katyo/publish-crates/blob/main/src/package.ts?rgh-link-date=2026-02-02T09%3A32%3A50Z).

After spending a few too many hours on the xtask implementation, I think I agree with this approach. We may even be able to run the workflow with `dry-run=true` and not have to extract anything (needs to be tested).

Solely so I can document this _somewhere_ the xtask approach consisted mainly of the following:

- Grab crate versions from the local checkout
- See if those versions exist on crates.io
- If they do, get their sha256 checksum
- Create package of corresponding crate locally
- Compare checksums (a small amount of shenanigans are required to get the repo's current commit sha out of `.cargo_vcs_info.json`)

The issue is that the `cargo package` command has a bug in its `tar` dependency, which seems to sometimes append 4 `\0` bytes to the end of the `.crate` file, sometimes not. This changes the checksum, and everything kind of falls apart. I also tried another approach where we pulled down the published `.crate` file, extracted it, and then manually hashed that against a locally created `.crate` file, but the same `tar` issue shows up and needs workarounds, and there's still the issue with `.cargo_vcs_info.json`. Too much complexity and way too many places where this could break.

docs/src/6-contributing.md
WillLillis: I've pushed a first pass on instructions differentiating minor and patch releases, let me know what should be improved!

WillLillis:
I've added a new workflow, which is just the normal rust publishing job `katyo/publish-crates@v2` with `dry-run` set to `true`. (TBD if we need to supply our token to use github's API to check for changes to crates). This should be tested on this PR, on its own backport PR (once this is merged), and again on a separate backport PR to the release branch (https://github.com/tree-sitter/tree-sitter/pull/5295 should be backported, but I'm waiting to so we can test both cases). I also need to add the option to trigger this workflow manually, but that is going to be a tomorrow problem.

.github/workflows/crate_versions.yml
```diff
@@ -0,0 +1,21 @@
+name: Crate Versions Check
+
+on:
+  pull_request_target:
+    types: [labeled]
```
clason, line 5:
```suggestion
    types: [labeled]
  workflow_dispatch:
```

clason:
I pushed a commit with a proposed release workflow listing individual steps. (It's a bit of a pity that we don't have any changelog or similar that could be edited for the minor release PR; we might want to put a nonce somewhere, e.g. in the README, for that purpose.)

WillLillis:
Going to merge this now as-is. This is an opt-in change and the docs are also a nice win to get in. If there's any followup issues we run into, further testing can be done on my fork before fixing in one PR.

PR #5295, opened 2026-02-02
fix(lib): correct ts_parser_parse docstring

bertie-wheen:
The ts_parser_parse docstring had become inconsistent, stating the existence of four possible failure types but then listing only two.

The other two (that made four) were removed as part of 821cf79, but this part of the edit was missed.


WillLillis:
Thanks for catching this! Could you run `cargo xtask generate-bindings` so the corresponding comment in the `lib/binding_rust/bindings.rs` is also updated? If you don't have cargo installed, you could also `git apply` the diff from the failing [Check Bindgen Output](https://github.com/tree-sitter/tree-sitter/actions/runs/21608390908/job/62279955937?pr=5295) failing workflow.

bertie-wheen:
Sorry - yep. Have now done so!

dundargoc (approved):

PR #5296, opened 2026-02-03
fix(test): acquire lock before downloading wasm tools

WillLillis:
Because Rust tests run in parallel, two tests compiling a parser to wasm can both try to download wasi-sdk and wasm-opt to a common location, causing corrupted files and/or other failures. To prevent this, we can guard access to these tools. Because testing is the only way two different threads will try to access them at the same time (from the same executable at least, I don't think we should worry about multiple tree-sitter-cli processes downloading these tools concurrently), this lock only needs to be present during testing.

This issue has caused several CI failures over the last few weeks, for example: https://github.com/tree-sitter/tree-sitter/actions/runs/21580322569/job/62176136461?pr=5293

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1632,7 +1632,13 @@ impl Loader {
         })?;
 
         info!("Downloading {tool_name} from {url}...");
-        let temp_tar_path = cache_dir.join(filename);
+        let temp_tar_dir = tempfile::tempdir_in(&cache_dir).map_err(|e| {
+            LoaderError::IO(IoError {
+                error: e,
+                path: Some(cache_dir.to_string_lossy().to_string()),
+            })
+        })?;
+        let temp_tar_path = temp_tar_dir.path().join(filename);
 
         let status = Command::new("curl")
             .arg("-f")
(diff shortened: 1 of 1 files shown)


clason:
We could also simply pre-download (and cache?) these as a setup step in our workflows? (I'm wary of shoving too much "defensive" logic into the CLI; it's doing too much already...)

(Unless this is also a common issue for interactive use.)

WillLillis:
> We could also simply pre-download (and cache?) these as a setup step in our workflows? (I'm wary of shoving too much "defensive" logic into the CLI; it's doing too much already...)
> 
> (Unless this is also a common issue for interactive use.)

My main thought was distro packagers. If you delete your own cached tools, this failure reproduces locally very reliably with `cargo test --features=wasm`. Anyone cloning the repo and testing it with cargo (as you'd expect with a rust project) is likely to hit this. 

Also, it turns out that `cfg(test)` is _not_passed across crate boundaries (https://github.com/rust-lang/cargo/issues/8379), which explains the CI failures before my last push. Given that we're interacting with the filesystem, network, and or shelling out to external programs, I don't think acquiring an uncontested mutex is a significant cost to unconditionally guard this code path.

I've also swapped the tar path to use a `temp_dir`, which is cleaned up even in case of an error. In case there's an error during download or extraction, the corrupted tarball will be removed so the next invocation can start from a clean slate.

PR #5306, opened 2026-02-07
fix(cli): allow for both debug logs and graphs

### LINKED ISSUE #5304: -d and -D are incompatible
### Problem

IF you run `tree-sitter test -d -D` you will only get the results of -D. You will not also get the debug info dumped to the terminal.

### Steps to reproduce

```
tree-sitter test -D -d
```

### Expected behavior

I expect the log to be generated (per -D) and also have the log info dumped to screen (per -d). But instead, TS behaves as if only the -D flag had been used.

0.26.3

### Operating system/version

macOS 26.2

WillLillis:
- Closes #5304


WillLillis:
- To be backported after #5292 is merged

PR #5309, opened 2026-02-08
lib: clean up strict aliasing fixes in array.h

amaanq:
### Problem

In #5242, strict aliasing violations were fixed though there was this funky dance with a `Swap` struct that was unnecessary. This PR tidies it up a bit more.


WillLillis (approved):

amaanq:
TLA+ is the only grammar that fails to compile with this change, which was a bug on their side. I've opened https://github.com/tlaplus-community/tree-sitter-tlaplus/pull/143 to fix this.

PR #5313, opened 2026-02-09
fix(rust)!: return `u32` from `Node::child_count`

### LINKED ISSUE #5312: Node::child_count returns a usize but Node::child takes a u32
### Problem

If I want to loop over the children of a node using a range[1] then my code looks something like this, with a required type conversion:

```
    for i in 0..node.child_count() {
        let child = node.child(i as u32);
    }
```

I'm pretty sure `child` should take a usize, or `child_count` should return a u32. I don't have an opinion on which is correct. I'll note that this inconsistency arose during https://github.com/tree-sitter/tree-sitter/issues/4515 which might mean that work simply wasn't comprehensive enough.

[1] Yeah, I should probably be using a cursor, but this is where I'm at at the moment. :-)

### Steps to reproduce

N/A

### Expected behavior

N/A

0.26.5

### Operating system/version

Debian Linux 13

WillLillis:
This matches the type returned from the underlying C library, and also composes more cleanly with other functions like `Node::child`.

One _could_ make the argument for `usize` here because it's the natural "indexing" type in Rust, but (in my opinion) it makes more sense to more closely match what's going on in the core C library.

- Followup to #4664
- Closes #5312

### DIFF
--- a/crates/cli/src/fuzz/corpus_test.rs
+++ b/crates/cli/src/fuzz/corpus_test.rs
@@ -23,7 +23,7 @@ pub fn check_consistent_sizes(tree: &Tree, input: &[u8]) {
         let mut some_child_has_changes = false;
         let mut actual_named_child_count = 0;
         for i in 0..node.child_count() {
-            let child = node.child(i as u32).unwrap();
+            let child = node.child(i).unwrap();
             assert!(child.start_byte() >= last_child_end_byte);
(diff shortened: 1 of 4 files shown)


clason:
I know you're on a roll, but can we _please_ not merge PRs with such a broad CI failure (especially for what is supposed to be a stable backport branch)? At least rerun them once or twice first.

PR #5315, opened 2026-02-09
feat: allow `-` in grammar names

jfly:
A number of grammars in the wild have this character in their name:

- https://github.com/sogaiu/tree-sitter-janet-simple/pull/7
- https://github.com/tree-sitter/tree-sitter-c-sharp/pull/408
- https://github.com/tree-sitter/tree-sitter-embedded-template/pull/45
- https://github.com/tree-sitter/tree-sitter-ql-dbscheme/pull/7

I read through https://github.com/tree-sitter/tree-sitter/pull/3700 and https://github.com/tree-sitter/tree-sitter/issues/3637, and it doesn't look like there was much discussion about this name regex, so hopefully this is an acceptable change?

### DIFF
--- a/docs/src/assets/schemas/config.schema.json
+++ b/docs/src/assets/schemas/config.schema.json
@@ -13,7 +13,7 @@
           "name": {
             "type": "string",
             "description": "The name of the grammar.",
-            "pattern": "^[a-z0-9_]+$"
+            "pattern": "^[a-z0-9_-]+$"
           },
           "camelcase": {
             "type": "string",


WillLillis:
Thanks!

sogaiu:
Please have a look at [this discussion](https://github.com/tree-sitter/tree-sitter/discussions/5321).

clason:
Should we revert this then (and document the requirement)?

WillLillis:
> Should we revert this then (and document the requirement)?

That seems  reasonable to me. The motivation behind merging this was to align with existing grammars that have already used `-` instead of `_`.  If this is  causing real issues upstream, it makes more sense to enforce this more strictly (and try to fixup what grammars we can).

sogaiu:
FWIW, I suspect that in my case the entry that ended up in `tree-sitter.json` was generated, but I'm not quite certain.

I tend to be hyper-sensitive to the `-` / `_` issue because I've had to cope with it in a number of places, but may be I messed up in this case :sweat_smile:

PR #5322, opened 2026-02-10
ci: speed it up

amaanq:
### Problem

Our CI has some inefficients and can be sped up.

### Solution

I've made a bunch of small changes in our CI to reduce our CI times by over 50% ideally.

- We don't need to run wasm tests on macos-x86 or linux-arm64, they are slow platforms
- Wasmtime provides pre-built c api libraries for us, which saves a lot of time
- setup-emsdk can be cached, unfortunately this isn't doable with the action so I got rid of it in favor of a few extra lines of yaml
- MSYS2 doesn't need to run updates, the runners already have pretty up to date packages to begin with
- We don't need to run the `cargo build --release` step, binary uploads are only used during releases so we gate this behind that
- cargo clippy already does what cargo check does, so we can drop that invocation in the makefile
- We can cache our npm packages for the web bindings

This makes our CI blazingly fast ⚡!

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1756787288,
-        "narHash": "sha256-rw/PHa1cqiePdBxhF66V7R+WAP8WekQ0mCDG4CFqT8Y=",
+        "lastModified": 1770562336,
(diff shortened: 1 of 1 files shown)


WillLillis (approved):

clason:
Don't forget that we can use native runners instead of emulation for some platforms now (linux-arm64 being one, so it will be blazingly fast by itself).

clason:
And if CI is really so fast now, we should consider running them unconditionally so we can mark them as required (and thus can use automerge).

PR #5330, opened 2026-02-11
feat(cli) `tree-sitter test --show-diff-symbols`

buckynbrocko:
Added `tree-sitter test` option `--show-diff-symbols` so both color and symbols can use simultaneously.

image

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -343,6 +343,9 @@ struct Test {
     /// Force showing fields in test diffs
     #[arg(long)]
     pub show_fields: bool,
+    /// Force showing '+' and '-' in test diffs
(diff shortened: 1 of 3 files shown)


WillLillis:
@buckynbrocko any plans on moving this out of the draft stage? I haven't looked over the code changes too closely yet but the end result looks nice :slightly_smiling_face:

WillLillis:
- Rebased the changes on top of master
- Used diff _marker_ terminology over symbol (too  easily confused with grammar symbols imo)
- Added an entry for the new flag in the docs
- Cleaned up some now unnecessary special casing of trailing newlines in `TestDiff`'s `Display` impl.

WillLillis (approved):

PR #5340, opened 2026-02-17
perf(generate): optimize TokenSet

WillLillis:
## The bottom line

A 28-35% reduction in wall time for `tree-sitter generate`.

image

image

image

For certain other large/complicated grammars (don't want to name and shame here), there is a 50% reduction in wall time.

## The how

Examining a flamegraph for `tree-sitter generate` reveals that a lot of time is spent unnecessarily in a few places: 
image

First, note the `smallbitvec` `reallocate` and `resize` calls, highlighted in purple. `smallbitvec` can be useful when you have a small amount of data that is likely to be kept inline, however this is not the case for tree-sitter. Instead, nearly every instances of this object begins inline, undergoes an expensive transfer (and allocation!) to the heap, as well as several reallocations after the fact. Furthermore, the bit flags in our use of `smallbitvec` are iterated over _one by one_. You can see the cost of this when looking at `TokenSet` operations (highlighted in purple) within the graph:

image

Computers are incredibly good at performing bitwise operations on word-sized pieces of data, which is something we can take advantage of! Here's what our flamegraph looks like now (`TokenSet` again highlighted in purple):

image

#### [perf(generate): replace SmallBitVec with word-level BitVec for TokenSet](https://github.com/tree-sitter/tree-sitter/commit/0b3bf7985dc629435012225523c1612218dd7e7f)

Because of the aforementioned issues with `smallbitvec` for tree-sitter's use case, I replaced it with a custom `BitVec` implementation based on `Vec<u64>`. This implementation is optimized to take advantage of word-level bitwise operations, which you can see in particular with its `insert_all` member function as well as `SetBitsIter::next`. This commit is responsible for most of the performance gains, yielding about a 20-25% reduction from `master`

#### [perf(generate): pre-alloc some BitVec sizes](https://github.com/tree-sitter/tree-sitter/commit/1e56b0a48c368c166bc10d50f595bac5a2aa2df5)

I originally tried pre-allocating the inner buffers for _all_ of `TokenSet`'s `BitVec` members, but this turned out to be a bad idea. Many `TokenSet`s are sparse and incredibly short lived. Because of the arena allocator I added below, this resulted in a 20% increase in peak rss with little to no wall time improvements. Instead, we only reserve where we can verify that the `TokenSet`s are long lived and are fully populated.

#### [perf(generate): back BitVec with a custom arena allocator.](https://github.com/tree-sitter/tree-sitter/commit/ff42b8d107ff3830cf39cea3e5d4d9139cd956df)

Those short lived `BitVec` instances I mentioned in the above commit's explanation still have a cost. When we're creating _many_ of these objects in a few hot loops, their continuous allocation and dropping become significant. This is the perfect use case for an arena allocator, which makes allocations incredibly cheap (you're just incrementing a pointer!). This yielded about a 10% reduction of wall time on top of the `BitVec` changes. The cost here is about a 10% increase in peak rss for most grammars, which should be an acceptable tradeoff. 

#### [perf(generate): pre-collect terminal indices in CoincidentTokenIndex](https://github.com/tree-sitter/tree-sitter/commit/71fbfd3079fb012d5730b7c23059fe9f2dc19ee8)

This is much less important optimization. Because we only care to iterate over items that are terminals, we can pre-compute that list once per state instead of skipping over the ones we don't care about. While it has a much smaller (immeasurable for some grammars) impact, I verified this does yield either a net zero or net reduction in wall time.

#### [perf(generate): iterate set bits only in merge_token_set](https://github.com/tree-sitter/tree-sitter/commit/027d65498dff1ebd9272fe64422787a32e54d252)

This is also a much less significant optimization. We simply reduce the number of objects we're iterating over inside `merge_token_set` instead of skipping items we aren't interested in. I verified this does yield either a net zero or net reduction in wall time. If I have the time, I'd like to take a look at the generated assembly for `check_token_conflicts` here. If the `for` loops are replaced with the iterator `.all` method, some grammars see a 1-2% boost, while others see a similar slow down. It's not clear to me (now, at least) why this is happening.

#### [perf(generate): don't continually reserve space in render buffer](https://github.com/tree-sitter/tree-sitter/commit/1f9e3379d4efd5b1ea5d92f75221b71d1da065ce)

This fixes a past mistake of mine. Allocating space up front is great! Doing so in a loop (as we are in `add_whitespace!`) can harm performance. Since the buffer will always have the necessary capacity, it will never use the default doubling allocation strategy, potentially leading to needless small allocations when less (larger) allocations would have been sufficient.

#### [perf(generate): add basic free list to BitVec's arena allocator](https://github.com/tree-sitter/tree-sitter/pull/5340/commits/a97e9c34716d0f07364ce6b3a72f5d9965113d74)

The 8-13% increase in peak rss due to the arena allocator was a bit concerning, so we can mitigate this by adding a simple free list to the arena. Anytime a block is freed, it's added to a free list keyed by block size. When allocating, the free list is checked before a new chunk is created. This comes at a ~1.5-3.5% performance cost, but brings peak rss down to within <1% of master.

NOTE: Marking this PR as a draft for now, as I'd like to have some time to fully read over these changes with a fresher mind. 

- [x] ~~Change new usages of `std` over to `core` where appropriate~~ Don't have to worry about no-std builds for the `generate` crate
- [x] ~~Should this be backported?~~ No, risky enough of a change that we should dog food this on master
- [x] Test on all nvim-treesitter grammars and ensure generated parsers are identical to those generated by master -> https://github.com/tree-sitter/tree-sitter/pull/5340#issuecomment-3938721668
- [x] Cleanup and review

### DIFF
--- a/crates/generate/src/rules.rs
+++ b/crates/generate/src/rules.rs
@@ -346,6 +346,7 @@ impl TokenSet {
         SetBitsIter::new(self.terminal_bits.as_slice()).map(Symbol::terminal)
     }
 
+    #[inline]
     pub fn contains(&self, symbol: &Symbol) -> bool {
         match symbol.kind {
             SymbolType::NonTerminal => panic!("Cannot store non-terminals in a TokenSet"),
@@ -356,6 +357,7 @@ impl TokenSet {
         }
     }
 
+    #[inline]
     pub fn contains_terminal(&self, index: usize) -> bool {
         self.terminal_bits.get(index).unwrap_or(false)
     }
--- a/crates/generate/src/build_tables/coincident_tokens.rs
+++ b/crates/generate/src/build_tables/coincident_tokens.rs
@@ -20,16 +20,23 @@ impl<'a> CoincidentTokenIndex<'a> {
             grammar: lexical_grammar,
             entries: vec![Vec::new(); n * n],
         };
+        // Pre-collect terminal indices up front rather than continuously recomputing within the
+        // loop below.
+        let mut terminal_indices = Vec::new();
         for (i, state) in table.states.iter().enumerate() {
-            for symbol in state.terminal_entries.keys() {
-                if symbol.is_terminal() {
-                    for other_symbol in state.terminal_entries.keys() {
-                        if other_symbol.is_terminal() {
-                            let index = result.index(symbol.index, other_symbol.index);
-                            if result.entries[index].last().copied() != Some(i) {
-                                result.entries[index].push(i);
-                            }
-                        }
+            terminal_indices.clear();
+            terminal_indices.extend(
+                state
+                    .terminal_entries
+                    .keys()
+                    .filter(|s| s.is_terminal())
+                    .map(|s| s.index),
+            );
+            for (j, &a) in terminal_indices.iter().enumerate() {
+                for &b in &terminal_indices[j..] {
+                    let index = result.index(a, b);
+                    if result.entries[index].last().copied() != Some(i) {
+                        result.entries[index].push(i);
                     }
                 }
             }
--- a/crates/generate/src/build_tables/item.rs
+++ b/crates/generate/src/build_tables/item.rs
@@ -160,6 +160,7 @@ impl<'a> ParseItem<'a> {
 }
 
 impl<'a> ParseItemSet<'a> {
+    #[inline]
     pub fn insert(&mut self, item: ParseItem<'a>) -> &mut ParseItemSetEntry<'a> {
         match self.entries.binary_search_by(|e| e.item.cmp(&item)) {
             Err(i) => {
@@ -357,6 +358,7 @@ impl Hash for ParseItem<'_> {
 }
 
 impl PartialEq for ParseItem<'_> {
+    #[inline]
     fn eq(&self, other: &Self) -> bool {
         if self.variable_index != other.variable_index
             || self.step_index != other.step_index
@@ -394,6 +396,7 @@ impl PartialEq for ParseItem<'_> {
 }
 
 impl Ord for ParseItem<'_> {
+    #[inline]
     fn cmp(&self, other: &Self) -> Ordering {
         self.step_index
             .cmp(&other.step_index)
--- a/crates/generate/src/build_tables/minimize_parse_table.rs
+++ b/crates/generate/src/build_tables/minimize_parse_table.rs
@@ -340,6 +340,7 @@ impl Minimizer<'_> {
         false
     }
 
+    #[inline]
     fn token_conflicts(
         &self,
         left_id: ParseStateId,
@@ -396,9 +397,6 @@ impl Minimizer<'_> {
             if self
                 .token_conflict_map
                 .does_conflict(new_token.index, token.index)
-                || self
-                    .token_conflict_map
-                    .does_match_same_string(new_token.index, token.index)
             {
                 debug!(
                     "split states {} {} - token {} conflicts with {}",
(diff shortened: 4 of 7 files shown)


clason:
One big (arguably bigger) issue with generating and building tree-sitter parsers is memory usage. (This is what limits the number of parallel jobs in the nvim-treesitter tests.) Is this affected by these changes?

WillLillis:
> One big (arguably bigger) issue with generating and building tree-sitter parsers is memory usage. (This is what limits the number of parallel jobs in the nvim-treesitter tests.) Is this affected by these changes?

There is roughly an 8-15% increase in peak memory usage depending on the grammar with these changes (due to the arena allocator). I'm looking into ways to potentially reduce that, but I'm not sure how much (if any) progress I'll be able to make.

WillLillis:
Surprisingly, if we add a simple free list (keyed by block size), we stay within ~1.5% for most grammars (~3.5% for ocaml) in terms of wall time reductions, and peak rss stays within 1% of master.  New benchmark results:

image
image
image

WillLillis:
This is ready for review now. I've gone down a bit of a performance rabbit hole, and there's lots more to come. However these other changes are unrelated to `TokenSet`, make some non-trivial changes that will require careful review, and shouldn't hold up these wins.

clason:
Rebased on latest master to remove the `unsafe` from the generated parser; this should make it easier to verify that these changes don't affect the output.

clason:
Here's the result of running `install-parsers.lua --generate --max-jobs=4`:

Master:
```
Executed in  332.62 secs    fish           external
   usr time  690.76 secs    0.38 millis  690.76 secs
   sys time   45.27 secs    1.60 millis   45.27 secs
```

This PR:
```
________________________________________________________
Executed in  227.03 secs    fish           external
   usr time  546.93 secs    0.33 millis  546.93 secs
   sys time   44.03 secs    1.52 millis   44.03 secs
```

All binaries are identical (according to diff).

crates/generate/src/rules.rs
```diff
@@ -73,14 +77,408 @@ pub enum Rule {
...
+            return ptr;
+        }
+        if self.offset + n_words > ARENA_CHUNK_WORDS {
+            let size = ARENA_CHUNK_WORDS.max(n_words);
+            self.chunks.push(vec![0u64; size]);
+            self.offset = 0;
+        }
+        let chunk = self.chunks.last_mut().unwrap();
+        // SAFETY: `self.offset + n_words <= ARENA_CHUNK_WORDS.max(n_words) == chunk.len()`,
+        // so `self.offset` is an in-bounds offset into `chunk`.
```
amaanq:
```suggestion
        // SAFETY: Either a new chunk was just created with len = ARENA_CHUNK_WORDS.max(n_words),
        // in which case offset = 0 and n_words <= len; or an existing chunk is reused, in which
        // case offset + n_words <= ARENA_CHUNK_WORDS <= chunk.len().
```

maxbrunsfeld:
Nice work!

crates/generate/src/rules.rs
```diff
@@ -73,14 +77,416 @@ pub enum Rule {
...
+            arena_free(self.data, self.capacity as usize, self.words_in_use());
+        }
+    }
+}
+
+impl Default for BitVec {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
maxbrunsfeld: This is minor, but could you put `BitVec` in its own module, separate from the more application-specific logic in `rules.rs`?

WillLillis:
CI is failing due to network flakes. I'll kick off a fresh set of runs tomorrow, and hopefully GH is in a better state then. Otherwise, this should be ready to merge.

clason:
I'll rerun the tests until they pass (having _one_ Windows test pass would be good). Actual blocker for merging is the unresolved change anyway.

PR #5342, opened 2026-02-17
use provided by RedoxOS byte orders implementations

greyshaman:
Hi!
It adds basic Redox OS support to portable/endian.h by including __redox__ in the platform checks. Redox OS uses Relibc, which provides <endian.h> declarations for htobeXX/beXXtoh functions. This avoids the #error "platform not supported" .

The change is minimal and non-breaking for other platforms.
It was created as part of porting Helix editor to Redox OS.

Thanks.

### DIFF
--- a/lib/src/portable/endian.h
+++ b/lib/src/portable/endian.h
@@ -20,14 +20,14 @@
     defined(__GNU__) || \
     defined(__HAIKU__) || \
     defined(__illumos__) || \
+    defined(__redox__) || \
     defined(__NetBSD__) || \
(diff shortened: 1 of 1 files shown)


clason:
And commit message should be `fix(lib): add RedoxOS support to portable/endian.h`

greyshaman:
> WillLillis

Thanks for the quick review! Done — moved `__redox__` right below `__illumos__` for better organization.  
New commit pushed.

Let me know if anything else needs tweaking!

clason (approved):

PR #5343, opened 2026-02-18
fix(lib): cast NULL in ts_subtree_children macro

lucasew:
Explicitly casting NULL to (Subtree *) in the ternary expression ensures
type consistency. This resolves ambiguous type inference issues encountered
by strict static analysis tools and C-to-Go transpilers (like ccgo).

Signed-off-by: lucasew <lucas59356@gmail.com>

### DIFF
--- a/lib/src/subtree.h
+++ b/lib/src/subtree.h
@@ -252,7 +252,7 @@ static inline size_t ts_subtree_alloc_size(uint32_t child_count) {
 // Get a subtree's children, which are allocated immediately before the
 // tree's own heap data.
 #define ts_subtree_children(self) \
-  ((self).data.is_inline ? NULL : (Subtree *)((self).ptr) - (self).ptr->child_count)
+  ((self).data.is_inline ? (Subtree *)NULL : (Subtree *)((self).ptr) - (self).ptr->child_count)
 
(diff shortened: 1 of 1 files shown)


lucasew:
This looks like the hashbang situation from NixOS users. I am working on using ccgo from modernc to compile this into Go code along with the grammars. I got success in that with pretty printing the AST in a fully static Golang binary with about 4MB. That compiler found some issues dealing with this edge case.

lib/src/subtree.h
```diff
@@ -252,7 +252,7 @@ static inline size_t ts_subtree_alloc_size(uint32_t child_count) {
 // Get a subtree's children, which are allocated immediately before the
 // tree's own heap data.
 #define ts_subtree_children(self) \
-  ((self).data.is_inline ? NULL : (Subtree *)((self).ptr) - (self).ptr->child_count)
+  ((self).data.is_inline ? (Subtree *)NULL : (Subtree *)((self).ptr) - (self).ptr->child_count)
```
ObserverOfTime: Better to cast the entire expression instead of repeating the cast.

lucasew: Done

amaanq:
CI is failing across the board @lucasew

amaanq (approved):

PR #5360, opened 2026-02-22
feat(wasm): update wasi-sdk and binaryen

clason:
- **feat(wasm): bump wasi-sdk to v30**
- **feat(wasm): bump binaryen to v126**


clason:
Note: tree-sitter will _not_ update an already downloaded toolchain; you'll have to manually remove the directories in `.cache/tree-sitter` to trigger downloading the new version. (Right now, there's no real need to upgrade since we're not relying on critical bugfixes or newer flags -- although v126 does come with two new ones -- but at some point we might wish to look into automating that on version bumps.)

amaanq:
Yeah, that's a good point.

amaanq (approved):

PR #5363, opened 2026-02-22
loader: invalidate cached wasm toolchains on version mismatch

amaanq:
### Problem

Previously, `get_existing_tool` checked only for the existence of binaries in the cache directory without verifying their version, meaning bumping the version files had no effect until users manually deleted the cache directories. 

### Solution

This commit writes a `.version` marker file after downloading and checks it on subsequent runs, removing stale caches automatically when the expected version changes.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1498,6 +1498,7 @@ impl Loader {
         if let Some(path) = Self::get_existing_tool(
             "clang",
             "wasi-sdk",
+            WASI_SDK_VERSION,
             &possible_executables,
             "TREE_SITTER_WASI_SDK_PATH",
         )? {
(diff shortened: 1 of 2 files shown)


crates/loader/src/loader.rs
```diff
@@ -1593,7 +1598,23 @@ impl Loader {
             })
         })?;
 
-        let tool_dir = cache_dir.join(toolchain).join("bin");
+        let toolchain_dir = cache_dir.join(toolchain);
+        let version_file = toolchain_dir.join(".version");
+
+        // If a cached toolchain exists but the version doesn't match, remove it
+        if toolchain_dir.exists() {
+            let cached_version = fs::read_to_string(&version_file).unwrap_or_default();
```
WillLillis:
```suggestion
            let cached_version = fs::read_to_string(&version_file).unwrap_or_else(|_| "unknown".to_string());
```

I think this is a little nicer to display to the user, but minor/feel free to ignore.

WillLillis (approved):

PR #5380, opened 2026-02-26
ci(release): publish zip archives

### LINKED ISSUE #5378: Release: don't gzip artifacts on Windows
> It's true that a winget package cannot be added while release binaries are gzipped. 

 _Originally posted by @ObserverOfTime in [#5377](https://github.com/tree-sitter/tree-sitter/issues/5377#issuecomment-3965194628)_

clason:
Problem: Publishing the built tree-sitter-cli artifacts as gzipped
binaries prevents installing them with `cargo binstall` or packaging
them for Windows Package Manager (`winget`).

Solution: Create a zip archive for each binary, named `tree-sitter-cli`
to match now-common downstream usage. (Most package managers have
separate `tree-sitter-cli` and `tree-sitter` (library) packages.)

Closes #5378

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,7 @@ jobs:
           for platform in $(cd artifacts; ls | sed 's/^tree-sitter\.//'); do
             exe=$(ls artifacts/tree-sitter.$platform/tree-sitter*)
             gzip --stdout --name $exe > target/tree-sitter-$platform.gz
+            zip -j9 target/tree-sitter-cli-$platform.zip $exe
           done
           rm -rf artifacts
           ls -l target/
@@ -54,12 +55,14 @@ jobs:
         with:
           subject-path: |
             target/tree-sitter-*.gz
+            target/tree-sitter-cli-*.zip
             target/web-tree-sitter.tar.gz
 
       - name: Create release
         run: |-
           gh release create $GITHUB_REF_NAME \
             target/tree-sitter-*.gz \
+            target/tree-sitter-cli-*.zip \
             target/web-tree-sitter.tar.gz
         env:
           GH_TOKEN: ${{ github.token }}


clason:
The problem here is that `zlib` does not support zip archives (it can handle the contained deflated stream, but not the zip archive format itself). I had hoped that single-file zip archive would still be fine, but no such luck. So the install script needs extra work (or dependencies, which I won't touch).

Simple alternative would be to publish files both as gzip and zipped archive and leave the npm install script untouched (except for the `-cli` rename)?

clason:
What _is_ that file even used for? And is this really (still) the proper way for "installing" a binary in this context? (I don't think `npm install` is a reasonable "official" installation method in general; it's much simpler to say "download from Github releases" -- `curl` exists on every platform.)

clason:
@amaanq

ObserverOfTime:
The way I see it we have the following options:

1. Release all of the binaries uncompressed.
1. Add a npm dependency (e.g., `fflate`) for unzipping.
1. Publish each binary as a platform-specific npm package and get rid of the install script.

clason:
1. wouldn't solve the winget issue.

`3`. would be a reasonable approach (more reasonable than a JS wrapper for "just download the binary from GH smh"), but out of scope for this PR.

4. just publish everything as gz and zip, let downstream sort out which one they need, and move on with my life

ObserverOfTime:
winget supports uncompressed `.exe` releases just fine.

clason:
Then let's just drop the compression. Github's network reliability is pretty much trashed anyway.

clason:
New plan: Let's just _add_ the zip archives to the published artifacts; this way we can backport this without potentially breaking external tooling.

For 0.27, we then can -- and should -- decide on a new (less npm-centric) installation story.

PR #5384, opened 2026-02-27
fix: skip missing Makefile in version command

### LINKED ISSUE #5383: cli: `tree-sitter version` does not respect `tree-sitter.json`
### Problem

`tree-sitter version` tries to update non-existent files for disabled bindings (e.g., `Makefile`), erroring out bef

### Steps to reproduce

```sh
git clone --depth=1 https://github.com/tree-sitter-grammars/tree-sitter-vim
cd tree-sitter-vim
tree-sitter version 0.8.1
```

```
Bumping version 0.8.0 to 0.8.1
Error: Failed to update one or more files:

Failed to update /Users/clason/build/parsers/tree-sitter-vim/Makefile:
No such file or directory (os error 2)
```

### Expected behavior

No error, update all files for bindings that are _not_ disabled.

tree-sitter 0.26.6

### Operating system/version

macOS 26

MFS-code:
Fixes #5383 `tree-sitter version` fails when the project has no `Makefile` (e.g., grammars created before the current `tree-sitter init` flow, or projects with C bindings disabled).
- `update_makefile` was the only `update_*` method in `version.rs` that did not check for file existence before attempting to read it. All other methods (`update_cargo_toml`, `update_package_json`, `update_cmakelists_txt`, `update_pyproject_toml`, `update_zig_zon`) already guard with `if !path.exists() { return Ok(()); }`.

This adds the same check, consistent with the rest of the file.
Tested against [tree-sitter-vim](https://github.com/tree-sitter-grammars/tree-sitter-vim) (the reproduction case from the issue).

### DIFF
--- a/crates/cli/src/version.rs
+++ b/crates/cli/src/version.rs
@@ -292,6 +292,9 @@ impl Version {
         } else {
             self.current_dir.join("Makefile")
         };
+        if !makefile_path.exists() {
+            return Ok(());
+        }
 
         Self::update_file_with(&makefile_path, |content| {
             content


clason:
Arguably, that `Makefile` shouldn't even be part of the bindings (it's more useful for project-local maintenance than as a cross-project build script), but this is a good (backportable) fix in the meantime.

PR #5385, opened 2026-02-27
fix(wasm): pass target triple to clang

maribu:
While the downloader WASI-SDK defaults to building WASM binaries, users might want to use the system's clang instead to safe some disk storage.

By adding the target triple an `export TREE_SITTER_WASI_SDK_PATH=/usr` allows using the system's `clang`. (Tested using 0.26.6 release.)

Passing the default target triple explicitly to WASI-SDK's `clang` should have no impact.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1388,6 +1388,7 @@ impl Loader {
 
         let mut compile_command = Command::new(&clang_exe);
         compile_command.current_dir(src_path).args([
+            "--target=wasm32-unknown-wasi",
             "-o",
             output_path,
             "-fPIC",


maribu:
> (Tested using 0.26.6 release.)

I tested this together with https://github.com/tree-sitter/tree-sitter/pull/5387 and it I was able to run `tree-sitter build --wasm` with it using the system's clang.

So tested with both the release and `master` now.

clason:
> By adding the target triple an export TREE_SITTER_WASI_SDK_PATH=/usr allows using the system's clang. (Tested using 0.26.6 release.)

I just tested that on macOS and it failed with
```
Error: wasi-sdk clang command failed: In file included from parser.c:3:
./tree_sitter/parser.h:10:10: fatal error: 'stdlib.h' file not found
   10 | #include <stdlib.h>
      |          ^~~~~~~~~~
1 error generated.
In file included from scanner.c:1:
./tree_sitter/parser.h:10:10: fatal error: 'stdlib.h' file not found
   10 | #include <stdlib.h>
      |          ^~~~~~~~~~
1 error generated.
```

So it's not quite that simple.

maribu:
That would indicate that WASI-SDK is not installed. On Alpine Linux, the C lib for WASM is packaged as `wasi-libc` which is installed as a dependency of the `wasi-sdk` package.

If you run `~/.cache/tree-sitter/wasi-sdk/bin/clang -print-target-triple` it should print `wasm32-unknown-wasi`.

I just did so by mounting the WASI-SDK downloaded by tree-sitter in a container:

```
$ ./wasi-sdk/bin/clang -print-target-triple
wasm32-unknown-wasi
```

That would mean that adding `--target=wasm32-unknown-wasi` is indeed a no-op for when the downloaded toolchain is used. But it does users to use the system's clang, if the WASI-SDK correctly installed.

clason:
I think when you are writing "system clang", you mean something very different... (You do know that this is the standard compiler (not just) on macOS, right? And is therefore available even without a WASI-SDK installation?)

maribu:
clang is by definition always a cross compiler, as the native target is the low level virtual machine. Unlike e.g. GCC that can only build for one target, clang can emit code for multiple targets.

E.g. if you run `clang -print-targets` it will list all targets it can emit code for. You'll still need C headers and libraries to link against, though. I haven't used Mac OS, but WASM is among the targets enabled by default these days. E.g. my system's clang lists:

```
$ clang -print-targets

  Registered Targets:
    aarch64     - AArch64 (little endian)
    aarch64_32  - AArch64 (little endian ILP32)
    aarch64_be  - AArch64 (big endian)
    amdgcn      - AMD GCN GPUs
    arm         - ARM
    arm64       - ARM64 (little endian)
    arm64_32    - ARM64 (little endian ILP32)
    armeb       - ARM (big endian)
    avr         - Atmel AVR Microcontroller
    bpf         - BPF (host endian)
    bpfeb       - BPF (big endian)
    bpfel       - BPF (little endian)
    hexagon     - Hexagon
    lanai       - Lanai
    loongarch32 - 32-bit LoongArch
    loongarch64 - 64-bit LoongArch
    mips        - MIPS (32-bit big endian)
    mips64      - MIPS (64-bit big endian)
    mips64el    - MIPS (64-bit little endian)
    mipsel      - MIPS (32-bit little endian)
    msp430      - MSP430 [experimental]
    nvptx       - NVIDIA PTX 32-bit
    nvptx64     - NVIDIA PTX 64-bit
    ppc32       - PowerPC 32
    ppc32le     - PowerPC 32 LE
    ppc64       - PowerPC 64
    ppc64le     - PowerPC 64 LE
    r600        - AMD GPUs HD2XXX-HD6XXX
    riscv32     - 32-bit RISC-V
    riscv64     - 64-bit RISC-V
    sparc       - Sparc
    sparcel     - Sparc LE
    sparcv9     - Sparc V9
    spirv       - SPIR-V Logical
    spirv32     - SPIR-V 32-bit
    spirv64     - SPIR-V 64-bit
    systemz     - SystemZ
    thumb       - Thumb
    thumbeb     - Thumb (big endian)
    ve          - VE
    wasm32      - WebAssembly 32-bit
    wasm64      - WebAssembly 64-bit
    x86         - 32-bit X86: Pentium-Pro and above
    x86-64      - 64-bit X86: EM64T and AMD64
    xcore       - XCore
```

clason:
So let's go back to the beginning and try to get on the same page for what is actually the problem and your desired behavior.

Do I understand correctly that the situation is the following:
1. you already have a (reasonably recent) version of WASI-SDK installed, say in directory `foo/`
2. you `cd` into a parser project and do `TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm`

Then what happens?
3a. `tree-sitter` still downloads the SDK and uses that (correctly) _or_
3b. it doesn't but you get an error because you don't _really_ have the (actual) WASI-SDK distribution installed but a Frankenbuild of headers and "system `clang`?

Is this PR enough to address _either_ issue?

(Here's mine:
```
clang -print-targets

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_32 - AArch64 (little endian ILP32)
    aarch64_be - AArch64 (big endian)
    arm        - ARM
    arm64      - ARM64 (little endian)
    arm64_32   - ARM64 (little endian ILP32)
    armeb      - ARM (big endian)
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
```
which means relying on existence of a `clang` binary is not enough; you need the full LLVM clang. By contrast, WASI-SDK bundles a stripped LLVM that _only_ has `wasm32` and `wasm64` targets. Hence I think we should keep treating the SDK as an atomic distribution -- especially since we may want to rely on a specific version.)

maribu:
OK, here is what happens when I run a fresh build of `tree-sitter` from `master`:

```
$ ~/Repos/software/tree-sitter/target/debug/tree-sitter build --wasm
Error: wasi-sdk clang command failed: terminate called after throwing an instance of 'std::system_error'
  what():  No error information
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.	Program arguments: /home/maribu/.cache/tree-sitter/wasi-sdk/bin/clang -o /home/maribu/Repos/software/tree-sitter-proverif/tree-sitter-proverif.wasm -fPIC -shared --no-wasm-opt -Os -Wl,--export=tree_sitter_proverif -Wl,--allow-undefined -Wl,--no-entry -nostdlib -fno-exceptions -fvisibility=hidden -I . parser.c
#0 0x00007f108b86ff11 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/maribu/.cache/tree-sitter/wasi-sdk/bin/../lib/libLLVM.so.21.1-wasi-sdk+0x86ff11)
```

It fails, becauses the downloaded WASI-SDK is dynamically linked and I do not have matching dynamic libraries installed. Specifically, even a trivial call like this fails:

```
$ ~/.cache/tree-sitter/wasi-sdk/bin/clang --version           
terminate called after throwing an instance of 'std::system_error'
  what():  No error information
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.	Program arguments: /home/maribu/.cache/tree-sitter/wasi-sdk/bin/clang --version
#0 0x00007f21b466ff11 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/maribu/.cache/tree-sitter/wasi-sdk/bin/../lib/libLLVM.so.21.1-wasi-sdk+0x86ff11)
[1]    15641 abort      ~/.cache/tree-sitter/wasi-sdk/bin/clang --version
```

Running with `TREE_SITTER_WASI_SDK_PATH=/usr` on a build from `master` results in:

```
$ TREE_SITTER_WASI_SDK_PATH=/usr ~/Repos/software/tree-sitter/target/debug/tree-sitter build --wasm
Error: wasi-sdk clang command failed: clang: warning: argument unused during compilation: '--no-wasm-opt' [-Wunused-command-line-argument]
/usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: unrecognized option '--allow-undefined'
/usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: use the --help option for usage information
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is because on my system, the default target triple of clang is:

```
$ clang -print-target-triple
x86_64-alpine-linux-musl
```

So calling `clang` without specifying a target and flags that only work for the WASM target and the WASM linker fail. Even if they wouldn't fail, the output won't be WASM but an x86_64 ELF file.

When I now use a build of https://github.com/tree-sitter/tree-sitter/pull/5387 that I compiled with `cargo build --features 
prefer-system-toolchain` I get

```
$ ~/Repos/software/tree-sitter/target/debug/tree-sitter build --wasm
$ echo $?   
0
```

So I am able to build the WASM with the `clang` installed via the package manager (provided I also install `wasm-sdk` and `lld`), but I'm unable to get `tree-sitter build --wasm` working with the WASI-SDK toolchain that get's auto-downloaded.

clason:
> When I now use a build of https://github.com/tree-sitter/tree-sitter/pull/5387 that I compiled with cargo build --features  prefer-system-toolchain I get

We're not talking about #5387 here; that is not on the table. I am asking about the change _here_.

(And let me be clear: It should be possible in your specific niche setup to be able to compile parsers to WASM, but it need not be out of the box.)

maribu:
OK, going back to the change here: The WASI-SDK's `clang` has an implicit `--target=wasm32-unknown-wasi` set. Similar as to how my `clang` has an implicit `--target=x86_64-alpine-linux-musl`.

The change here makes the implicit target explicit. It is a no-op for when a `clang` is used that already has an implicit `--target=wasm32-unknown-wasi` set. So merging this PR would not be a semantic change and should have no side-effect for the current workflow.

It is however a requirement to pass the target explicitly if one were to use a `clang` that does not has implicitly `--target=wasm32-unknown-wasi` set.

This change alone sadly is not sufficient to allow using the system's `clang` with `master` anymore. In the `0.26.6` release an `export TREE_SITTER_WASI_SDK_PATH=/usr` did cause `tree-sitter` to use `clang` from `/usr/bin/clang`, but with `master` it will download the WASI-SDK anyway.

maribu:
> but it need not be out of the box.

That is a regression from 0.25.10 btw, as before docker / podman was used to fetch the toolchain.

clason:
Yes, because it needs to check against the expected version and update the cached version if it is older. But that is a separate (minor) issue; the problem I am concerned with is whether you can build a wasm parser at all.

So, again, does `TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm` build without error for you **with only the change in this PR**?

> That is a regression from 0.25.10 btw, as before docker / podman was used to fetch the toolchain.

Which is why this was introduced as a breaking change in a minor (zerover) release. The benefits (massively) outweigh the regression.

maribu:
> So, again, does TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm fork for you with only the change in this PR?

As I said: Applying on this change on top of `0.26.6` allows using `TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm`. With `master`, `TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm` would still download WASI-SDK no matter what.

clason:
I don't care about the download (right now), will it _work_?

> With master, TREE_SITTER_WASI_SDK_PATH=/usr tree-sitter build --wasm would still download WASI-SDK no matter what.

(It doesn't for me, fwiw.)

maribu:
| Version  | `export TREE_SITTER_WASI_SDK_PATH=/usr` | This PR Applied | Works | Why                                                       |
| :------- | :-------------------------------------- | :-------------- | :---- | :-------------------------------------------------------- |
| 0.26.6   | No                                      | No              | No    | WASI-SDK binaries used, not compatible with my system     |
| 0.26.6   | Yes                                     | No              | No    | system's clang used, but x86_64 as target                 |
| 0.26.6   | No                                      | Yes             | No    | WASI-SDK binaries used, not compatible with my system     |
| 0.26.6   | Yes                                     | Yes             | Yes   | system's clang used with correct target triple            |
| `master` | No                                      | No              | No    | WASI-SDK binaries used, not compatible with my system     |
| `master` | Yes                                     | No              | No    | WASI-SDK binaries used, not compatible with my system     |
| `master` | No                                      | Yes             | No    | WASI-SDK binaries used, not compatible with my system     |
| `master` | Yes                                     | Yes             | No    | WASI-SDK binaries used, not compatible with my system     |

clason:
Ok, then the first step is to find out why it doesn't use the clang binary from the explicit path you provide. (It should, going by the code, and it does for me.)

Stick to the master branch, please, and don't worry about this PR. We are only interested in debugging why the provided WASI_SDK_PATH is ignored on your machine.

maribu:
Nevermind, running

```
$ TREE_SITTER_WASI_SDK_PATH=/usr TREE_SITTER_BINARYEN_PATH=/usr tree-sitter build --wasm
```

allows using the system toolchain.

With master:

```
$ TREE_SITTER_WASI_SDK_PATH=/usr TREE_SITTER_BINARYEN_PATH=/usr ~/Repos/software/tree-sitter/target/debug/tree-sitter build --wasm
Error: wasi-sdk clang command failed: clang: warning: argument unused during compilation: '--no-wasm-opt' [-Wunused-command-line-argument]
/usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: unrecognized option '--allow-undefined'
/usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: use the --help option for usage information
clang: error: linker command failed with exit code 1 (use -v to see invocation)
$ echo $?
1
$ ls *.wasm 
zsh: no matches found: *.wasm
```

With this PR:

```
$ TREE_SITTER_WASI_SDK_PATH=/usr TREE_SITTER_BINARYEN_PATH=/usr ~/Repos/software/tree-sitter/target/debug/tree-sitter build --wasm
$ echo $?  
0
$ ls *.wasm
tree-sitter-proverif.wasm
```

clason:
Ok, thank you. That makes more sense. I'll merge and backport this PR then. 

A followup PR could add a TREESITTER_CLANG_PATH (defaulting to the sdk path / bin) for people who have a downloaded sdk but need a local clang. 

@amaanq ?

maribu:
> A followup PR could add a TREESITTER_CLANG_PATH (defaulting to the sdk path / bin) for people who have a downloaded sdk but need a local clang.

That would be awesome.

Specifically for distributions packaging tree-sitter, it would still be nice if there would be build flag to prefer the system's toolchain. Many distributions have the policy to debundle different components to avoid duplicates. (See e.g. [this lwn.net article with regard to Debian](https://lwn.net/Articles/842319/) or the [Fedora's Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling); I'm aware that for Arch Linux, Ubuntu, Gentoo, and Alpine Linux also try debundle software and using system packages to provide dependencies rather than vendoring them in. Such policies are also one of the most common reason for downstream patches.) If tree-sitter were to provide a way for distros that would allow packaging it in accordance with such policies, it would reduce the a lot of downstream patching.

clason:
That is misunderstanding how tree-sitter is designed to work, so out of the question. You should be able to override defaults if you need to and know what you are doing, but I would rather archive this project than having to deal with every tree-sitter CLI behaving differently.

maribu:
> but I would rather archive this project than having to deal with every tree-sitter CLI behaving differently.

An open source project (such as the MIT licensed tree-sitter) allows everyone to modify the source code, build binaries from the modified source code, and share the modified source code and binaries with very limited conditions (such as releasing the changes under a compatible license and providing recipients of source and binaries the license terms).

Contributors to open source projects modify the source code and test their changes before opening a PR, and therefor are running a tree-sitter CLI that is behaving differently from upstream. This is generally considered an intentional part of how open source works. I'm a bit puzzled that you appear to be threatening to archive the repo over modifying software, as this part of the standard open source development workflow and explicitly allowed by the MIT license that tree-sitter chose.

If the concern here is about people obtaining a tree-sitter CLI binary behaving differently from upstream without having to modify the source code themselves, rather than active contributors to the project: Linux users regardless of what distribution they choose will be running dozens of software components that have been patched downstream (and therefore behave not 100% identical to upstream). Here a few examples of high profile packages on high profile distros:

- [distro: NixOS, package: bash](https://github.com/NixOS/nixpkgs/tree/3cd1d797d606d4b46e9a98d84f55126d362e3853/pkgs/shells/bash): 5 downstream patches
- [distro: Debian, package: firefox](https://tracker.debian.org/pkg/firefox): Debian is even maintaining a downstream [git fork](https://salsa.debian.org/mozilla-team/firefox/-/commits/release/master) of the [upstream repo](https://github.com/mozilla-firefox/firefox). E.g. the [current HEAD commit](https://salsa.debian.org/mozilla-team/firefox/-/commit/973328337c3a56741b8e194244cd0eb1686fa04a) in Debian is [not part of the upstream repo](https://github.com/mozilla-firefox/firefox/commit/973328337c3a56741b8e194244cd0eb1686fa04a)
- [distro: Gentoo, package: binutils](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-devel/binutils/files): 8 downstream patches
- [distro: Arch Linux, package: clang](https://gitlab.archlinux.org/archlinux/packaging/packages/clang): 3 downstream patches
- [distro: Fedora, package: GCC](https://src.fedoraproject.org/rpms/gcc/tree/rawhide): 14 downstream patches

And to show that I haven't hand picked a few exceptions and misrepresenting that downstream patches are an integral part of how distros maintain their packages:

```
$ git clone --depth=1 'https://anongit.gentoo.org/git/repo/gentoo.git'
[...]
$ find gentoo -name '*.patch' | wc -l
13604
```

```
$ git clone --depth=1 'https://github.com/NixOS/nixpkgs'
[...]
$ find nixpkgs -name '*.patch' | wc -l 
4920
```

```
$ git clone --depth=1 'https://gitlab.alpinelinux.org/alpine/aports.git'
[...]
$ find aports -name '*.patch' | wc -l      
6510
```

(Most other distros do not have all their patches in a single repo, so getting total number of patches there is not as trivial as for the above three)

The fact that there will be modified software (and, therefore differently behaving versions) really is an integral part of the open source deal.

The question I would discuss here is whether the tree-sitter project is willing to make package maintainer's life a bit easier by allowing them to set a build time flag, or whether they have to apply downstream patches. Your opinion has been pretty clearly formulated, but maybe other project members have a different stance?

PS: Please keep in mind that package maintainers apply the downstream patches to deliver their users the best possible experience. Making life of upstream developers hard is definitely **NOT** a goal of package maintainers. In fact, a well packaged software component (e.g. where `tree-sitter build --wams` works out of the box) will result in less issues being reported upstream. Also note that distros have their own bug trackers that reduces the load of issue reports reaching upstream, or that package maintainers are often actively upstreaming bug fixes that may profit all users.

sogaiu:
> I'm a bit puzzled that you appear to be threatening to archive the repo over modifying software, 

I very much doubt the comment was meant as a threat.  I took it as an expression of not finding the proposed idea to be desirable from the perspective of folks looking after things here.

This project has traditionally lacked resources to cope even with its own issues adequately.  clason and others have worked hard to keep things working as well as they do.

maribu:
@sogaiu Thanks for the clarification.

For the PR at hand: I think this is a minimal, technically correct change with no risk of increased maintenance, as it just explicitly passes the exact same target that the currently used `clang` is using due to baked-in defaults anyway.

clason:
As @sogaiu said, this was hyperbole to stress the fact that having this be an official configuration option would be a _support_ nightmare. (I know distributions patch all manner of things, and that is _already_ a nightmare, but then we can wash our hands of it and when things go wrong -- as they inevitably do -- say "complain to the distro, this is not how it's meant to be".)

And we stopped discussing this patch a long time ago; I already agreed that it was a reasonable fix and promised to merge it.

clason (approved):

PR #5387, opened 2026-02-27
feat(loader): allow preferring system toolchain

maribu:
This allows users to prefer a system toolchain over the automatically downloaded ones. This can be useful to safe storage, for distros packaging tree-sitter (as they can handle dependencies in their package manager and install WASI-SDK as dependency to tree-sitter-cli), or for systems on which the downloaded WASI-SDK is not compatible.

Depends on and includes: https://github.com/tree-sitter/tree-sitter/pull/5385

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -33,6 +33,7 @@ name    = "benchmark"
 default = [ "qjs-rt" ]
 qjs-rt  = [ "tree-sitter-generate/qjs-rt" ]
 wasm    = [ "tree-sitter/wasm", "tree-sitter-loader/wasm" ]
+prefer-system-toolchain = [ "tree-sitter-loader/prefer-system-toolchain" ]
 
 [dependencies]
 ansi_colours.workspace          = true


clason:
I don't think that should be a build-time option. And can you explain with what sort of system clang you have tested? It probably not only requires a binary with wasm support but also header files?

maribu:
The problem I am facing is that `tree-sitter build --wasm` started to download WASI-SDK binaries into `~/.cache/tree-sitter/wasi-sdk` and was using `~/.cache/tree-sitter/wasi-sdk/bin/clang` to build the WASM code. On my system, those binaries crash on startup, as they are incompatible with my system.

Instead, I have to the `clang` installed on my system via the package manager.

Making this behavior a build time option allows to provide a `tree-sitter-cli` package in Alpine on which `tree-sitter build --wasm` works out of the box, rather than crashing.

clason:
> The problem I am facing is that tree-sitter build --wasm started to download WASI-SDK binaries into ~/.cache/tree-sitter/wasi-sdk and was using ~/.cache/tree-sitter/wasi-sdk/bin/clang to build the WASM code.

Yes, that is the intended behavior (if `emscripten` is not installed).

> On my system, those binaries crash on startup, as they are incompatible with my system.

So the first question is "why does it download incompatible binaries?"

> Making this behavior a build time option allows to provide a tree-sitter-cli package in Alpine on which tree-sitter build --wasm works out of the box, rather than crashing.

Making this a runtime option will make _every_ build work without potentially crashing. For example, exposing environment variables that let users point to a different (system) toolchain is a much better option. (But note that these are version-locked right now.)

So for the sake of documentation, we need to know what is expected at these custom locations.

maribu:
> Yes, that is the intended behavior (if emscripten is not installed).

But I do have have WASI-SDK, clang and binaryen installed. If the intended behavior is to first try the installed toolchain and fallback to a downloaded if that toolchain is not available, then there is a regression.

> So the first question is "why does it download incompatible binaries?"

The downloaded toolchain is dynamically linked. Those will require compatible shared libraries to be installed on the system, which on my machine is not the case. A statically linked toolchain could be reasonably expected to run on any Linux machine of matching architecture without having to account for availability of shared libraries in compatible versions.

clason:
> If the intended behavior is to first try the installed toolchain and fallback to a downloaded if that toolchain is not available, then there is a regression.

No, that is not the intended behavior. It always downloads unless `emscripten`(!) is available. Testing for system installation of wasi-sdk is not (yet) implemented (but obviously desirable, although complicated because of versioning).

clason:
Let's keep the discussion focused on one PR. The build-time option is a no-go anyway.

PR #5389, opened 2026-02-28
generate: refactor `node_types` JSON generation logic

### LINKED ISSUE #5218: Inlined supertype is listed in `node-types.json` despite not actually existing
### Problem

When a type is listed both in `supertypes` and `inline` in a grammar, the generated `node-types.json` refers to that type despite it not actually existing in the generated C code (probably due to `inline`).

Note: This is not just a contrived corner case, this situation actually exists in [tree-sitter-java](https://github.com/tree-sitter/tree-sitter-java/blob/v0.23.5/grammar.js) for the `_simple_type` rule.

### Steps to reproduce

`grammar.js`:
```javascript
// @ts-check

module.exports = grammar({
  name: "debug",

  supertypes: $ => [$.t_super],
  inline: $ => [$.t_super],

  rules: {
    source_file: $ => $.t_super,
    t_a: $ => choice("a", "A"),
    t_b: $ => choice("b", "B"),
    t_super: $ => choice($.t_a, $.t_b),
  }
});
```

1. Run `./tree-sitter generate`
2. Inspect the generated `src/parser.c` file
  ℹ️ As expected: It does not contain `t_super`
3. Inspect the generated `src/node-types.json`
  ❌ Bug: It refers to the `t_super` type

### Expected behavior

Either:
- since the inlined type does not actually exist as node type, it should not appear in `node-types.json`
- OR, if it does not make sense to specify a type both in `supertypes` and in `inline`, the CLI should emit a warning and in future versions maybe an error

tree-sitter 0.26.3

### Operating system/version

Ubuntu 24.04 (on Windows WSL 2)

amaanq:
- Closes #5218

### Problem

The `node_types` json code is a bit of a mess, and had some buggy behavior, where a rule that was both a `supertype` and `inline` would produce incorrect JSON output.

### Solution

This PR refactors the code a bit by refactoring the larger functions in `node_tpyes.rs` that involve generating the JSON, and fixes the aforementioned bug. Additionally, I reworked the container types holding `String`s to instead borrow over a `&str` instead, allowing us to make them `Copy`able and save wasteful allocations on cloning.

### DIFF
--- a/crates/generate/src/node_types.rs
+++ b/crates/generate/src/node_types.rs
@@ -4,6 +4,8 @@ use rustc_hash::FxHashMap;
 #[cfg(feature = "load")]
 use rustc_hash::FxHashSet;
 
+#[cfg(feature = "load")]
+use serde::ser::{SerializeMap, SerializeSeq, SerializeStruct};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


clason:
As CI shows, this is a breaking change and should be marked as such; it breaks 29 parsers in the wild (including official fixtures):

Upstream:
* Python
* Bash
* Go
* Java
* Ocaml
* Ruby
* Rust
* Scala

Community:
* Apex
* Bitbake
* Cairo
* Fsharp
* Gdscript
* Ispc
* Julia
* Kcl
* Nqc
* Prisma
* Qmljs
* Snakemake
* Sourcepawn
* Starlark
* Sway
* Systemtap
* Templ
* V
* Zsh

I don't think fixing all those will be feasible... Does it _need_ to be an error? To fix the linked issue, 
can't you just silently skip these nodes when generating `node-types.json`? (And add -- yet another -- warning.)

dianecloud (approved):

SidoJain:
Tested a minimal alternative that doesn't touch `intern_symbols.rs`, so it shouldn't break any of the 29 parsers - just skips the symbol in `build_supertype_entries` and warns instead. Confirmed it fixes the original repro without needing the hard error. Happy to open this as a PR if useful.

crates/generate/src/node_types.rs
```diff
@@ -2553,38 +2553,40 @@ mod tests {
...
-    fn get_node_types(grammar: InputGrammar) -> SuperTypeCycleResult<Vec<NodeInfoJSON>> {
+    fn get_node_types(grammar: InputGrammar) -> SuperTypeCycleResult<Vec<NodeInfoJSON<'static>>> {
         let PreparedGrammar {
             syntax_grammar,
             lexical_grammar,
             default_aliases,
             str_pool,
             ..
         } = prepare_grammar(grammar, &mut Vec::new()).unwrap();
+        // Leak the pool so the JSON, which borrows from it, can be returned.
```
WillLillis: This works, but will make some noise if we try to run these tests under miri or other kinds of sanitizers. Could we instead ~~return the string pool back out alongside the `NodeInfoJSON` in a tuple? Another approach I tried while working on this involved~~ make all of the struct fields hold `StrId`s instead of borrowed `&str`s, which gets rid of the lifetimes, but does make the diff a little noisier in the tests.

WillLillis (approved):

clason:
Should we do the same for aliased nodes?

WillLillis:
> Should we do the same for aliased nodes?

Should be taken care of in #5874

PR #5395, opened 2026-03-01
refactor(query): remove `alternative_is_immediate`

ribru17:
This is implied by `is_pass_through`.


lib/src/query.c
```diff
@@ -4181,13 +4167,13 @@ static inline bool ts_query_cursor__advance(
                   copy->pattern_index,
                   copy->step_index,
                   next_step->alternative_index,
-                  next_step->alternative_is_immediate,
+                  next_step->is_pass_through,
```
WillLillis: Should we also update `immediate:` to `pass_through:` in the format string?

ribru17: good point :+1:

Issue #1788, opened 2022-06-29
Build WASM — undefined exported symbol


wkordalski:
Hi!

I'm trying to build Tree-sitter WASM library with debug symbols. So I do:
```
~/tree-sitter > ./script/build-wasm --debug
emcc: error: undefined exported symbol: "__ZNKSt3__220__vector_base_commonILb1EE20__throw_length_errorEv" [-Wundefined] [-Werror]
```

...and I get an error. Building without `--debug` also results with this error. Building `tags/v0.20.6` results with this error similarly to `master` branch.

Could you give me some hints?

----

OS: ArchLinux (Linux 5.18.3-arch1-1 x86_64 GNU/Linux)

```
~/tree-sitter > emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.13-git (531257621816c200bc7c3be53129494afd029aec)
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

ahelwer:
Might have to do with how the WASM imports are hard coded to a handful of entities in the STL and vectors are not among them; try searching past issues about this.

wkordalski:
Using Docker (i.e. older Emscripten) fixes this issue.

Right now I get
```
LinkError: WebAssembly.instantiate(): Import #4 module="wasi_snapshot_preview1" function="fd_write" error: function import requires a callable
```

However this is not so much important now, as I now can build correctly grammars (using Docker and correct version of Emscripten) and use official Tree-sitter library WASM builds.

xtalax:
I'm getting the same error trying to make web-tree-sitter, I run
```
$ yarn add -D tree-sitter-julia 
```
and the same error is thrown.
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.34-git (57b21b8fdcbe3ebb523178b79465254668eab408)
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

sogaiu:
[...]

PR #5401, opened 2026-03-02
 fix(wasm): export missing C stdlib symbols for language grammar side modules

Kenzo-Wada:
## Summary

Certain language grammar WASM side modules crash when loaded via `Language.load()` in web-tree-sitter because `tree-sitter.wasm` does not export C stdlib symbols their external scanners depend on (`isalpha`, `__assert_fail`).

- **0.25.x**: `TypeError: resolved is not a function`
- **0.20.x**: `TypeError: Cannot read properties of undefined (reading 'apply')`

9 out of ~36 tested grammars are affected (bash, cpp, html, php, python, ruby, tlaplus, vue, yaml).

## Changes

- Add `isalpha` to `stdlib-symbols.txt` (used by both Emscripten and native paths)
- Add `__assert_fail` to Emscripten `EXPORTED_FUNCTIONS` only — the native path already provides it as a host builtin in `wasm_store.c`

## Future work

A more robust approach would be to port the native `wasm-stdlib.h` blob injection to the Emscripten path, guaranteeing symbol coverage without manual tracking.

## Related

#1906, #949, #1788


clason:
Coding agent spam is unacceptable. If you require LLM review, do this on your own fork before opening the PR here. (This includes the PR description!)

Kenzo-Wada:
@clason   
Sorry about that. I had GitHub's Copilot PR review enabled in my settings and didn't realize it would trigger automatically. I also used AI to help translate the PR description since I'm not a native English speaker. I apologize.

Is the fix itself something that could be accepted? If so, I'll open a new clean PR without any AI involvement.

clason:
I'd suggest using a dedicated translation tool to help with PR description and comments; LLMs change the style and meaning so it's impossible to tell whether we're reading your words or its interpolation. (And your English seems fine; it doesn't need to be perfect in any case.)

I'm not sure about the fix either: the stdlib has been deliberately pared down for size, and most of the languages you mention are first-party and hence would have been kept compatible with any change here.

So the first question is whether you're sure you're using the latest version of all(!) relevant components (tree-sitter 0.26.6, tree-sitter-bash 0.25.1 -- better yet, the tip of the respective `master` branch). A minimal reproducible example would go a long way.

clason:
(In particular, `isalpha` is a no-go since we already have `iswalpha`. The other issue with the emscripten path I cannot speak about.)

Issue #880, opened 2021-01-13
Specify descendant or ancestor in query


robbym:
Is there a way to pattern match a descendant of arbitrary depth?

I'm parsing C code and I'm trying to get all identifiers of static declarations.
Something like the following works for simple cases:

Given:
```c
static int test;
static int test = 1;
static int test();
static int* test;
```
This works:
```
(declaration
    (storage_class_specifier "static")
    declarator: [
        (identifier)@name
        (_ declarator: (identifier)@name)
    ]
)
```

However, with more complex declarations:
```c
static int* test();
static int** test;
static int*** test;
```
Require me to keep adding alternates:
```
(declaration
    (storage_class_specifier "static")
    declarator: [
        (identifier)@name
        (_ declarator: (identifier)@name)
        (_ declarator: (_ declarator: (identifier)@name))
        (_ declarator: (_ declarator: (_ declarator: (identifier)@name))
        ...
    ]
)
```

Is there any way to specify the identifier whose ancestor (of arbitrary depth) is a static declaration (or declaration in general)?

robbym:
Maybe something like introducing a subquery syntax?

```
(declaration (storage_class_specifier "static") => (_ declarator: (identifier)@name)
```

This basically reads as: For each declaration with a static storage_class_specifier, match any "sub-declarator" that is an identifier.

robbym:
Or maybe something like:
```
(declaration
    (storage_class_specifier "static")
    { (_ declarator: (identifier)@name) }
)
```

Where anything in a set of braces is just an arbitrary query, with the only exception that the root node being traversed by said subquery must match the parent pattern.

MunifTanjim:
> I think all mentioned cases with catastrophic recursion actually have an issue in a grammars' composition and instead of attempts to handle unlimited recursions by queries it's better to reorganize grammars and flatten recursions by replacing a recursive rule by a rule with a repetition of the part that produces recursion.

Most of the parsers for popular languages in under https://github.com/tree-sitter org is made that way. That makes it hard to argue that it's not the recommended way of doing it.

tmillr:
> > I think all mentioned cases with catastrophic recursion actually have an issue in a grammars' composition and instead of attempts to handle unlimited recursions by queries it's better to reorganize grammars and flatten recursions by replacing a recursive rule by a rule with a repetition of the part that produces recursion.
> 
[...]

PR #5396, opened 2026-03-01
fix(query): don't add copies for quantifier steps outside alternations

ribru17:
Fixes an issue discussed in the Matrix channel. Related to #5382.

- Fixes #5400

PR #5403, opened 2026-03-02
query: add recursive back-reference (^) syntax for alternations (for #880)

arinar:
Introduce "(^)"-expressions inside "[...]"-alternations to express recursive descent to arbitrary tree depth. Allows matching deeply nested elements without specifying each possible nesting alternative.

I need a feature like this to do complex pattern detection (on unpreprocessed C-code in a code-checker-tool I'm writing for work). Without it, I'd have to write many nested queries for each of my cases and it almost defeats the purpose of having a powerful query-language.

example recursive query:
```query
    (function_definition
        body: (compound_statement [
            (if_statement consequence: (^))
            (if_statement consequence: (compound_statement (^)))
            (return_statement) @ret
        ])
    ) @func
```

-> matches function definitions and their "return"s if they are in the "consequence:"-lineage of an arbitrarily nested if-structure.

example syntax data:
```c
    int shallow(void) {           // <- [0].func
        return 1;                 // <- [0].ret
    }
    int one_deep(int x) {         // <- [1]func, [2].func
        if (x) { return 1; }      // <- [2].ret
        return 0;                 // <- [1].ret
    }
    int two_deep(int x, int y) {  // <- [3].func, [4].func, [5].func
        if (x) {
            if (y) { return 2; }  // <- [5].ret
            return 1;             // <- [4].ret
        }
        return 0;                 // <- [3].ret
    }
```

Each (^) jumps back to the enclosing [...], retrying all alternatives one level deeper. (^^) refers to the next outer alternation (De Bruijn indices), enabling multi-state recursive patterns.

The implementation reuses the existing dead-end step mechanism but with backward-pointing alternative_index. No new fields on QueryStep or QueryState — only a small linked list during parsing to track enclosing [...] positions.

Parsing:
* new (^) branch in the "("-dispatch chain
* validation that recursive branches consume input and a base case exists.

Analysis:
* backward dead-ends treated as successful terminations.

Execution:
* backward jump adjusts start_depth to keep compile-time step depths aligned with runtime tree position.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -6528,3 +6528,469 @@ fn test_query_recursive_ref_nested_alternations() {
         );
     });
 }
+
+// -----------------------------------------------------------------------
+// Recursive back-reference: capture ordering along the recursion path
+//
+// All tests use C pointer/array declarations for compact nesting:
+//   int ***c;          - 3 pointer levels
+//   int c2[10][20][30]; - 3 array levels
+//   int **g[10];       - mixed: 2 ptr + 1 arr (3 total)
+//   int *h[10][20];    - mixed: 1 ptr + 2 arr (3 total)
+// -----------------------------------------------------------------------
+
+// Test 1: Capture only root type and leaf identifier - no intermediate
+// recursion captures. Each declaration produces a single match with
+// scalar (non-repeated) captures regardless of nesting depth.
+#[test]
+fn test_query_recursive_ref_capture_root_and_leaf_only() {
+    allocations::record(|| {
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            "(declaration type: (_) @type declarator: [(pointer_declarator declarator: (^)) (identifier) @name])",
+        ).unwrap();
+
+        assert_query_matches(
+            &language,
+            &query,
+            "int *a; int **b; int ***c;",
+            &[
+                (0, vec![("type", "int"), ("name", "a")]),
+                (0, vec![("type", "int"), ("name", "b")]),
+                (0, vec![("type", "int"), ("name", "c")]),
+            ],
+        );
+    });
+}
+
+// Test 2: Capture root, leaf, AND each recursion step.
+// The @ptr capture appears once per pointer_declarator traversed -
+// including the outermost one - in DFS (= source) order.
+#[test]
+fn test_query_recursive_ref_capture_all_steps() {
+    allocations::record(|| {
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            "(declaration type: (_) @type declarator: [(pointer_declarator declarator: (^)) @ptr (identifier) @name])",
+        ).unwrap();
+
+        assert_query_matches(
+            &language,
+            &query,
+            "int *a; int **b; int ***c;",
+            &[
+                // depth 1: single ptr captured, then base case
+                (0, vec![("type", "int"), ("ptr", "*a"), ("name", "a")]),
+                // depth 2: two ptrs in source order (outer first)
+                (0, vec![("type", "int"), ("ptr", "**b"), ("ptr", "*b"), ("name", "b")]),
+                // depth 3: three ptrs in source order
+                (0, vec![("type", "int"), ("ptr", "***c"), ("ptr", "**c"), ("ptr", "*c"), ("name", "c")]),
+            ],
+        );
+    });
+}
+
+// Test 3: Three recursive alternatives (pointer_declarator,
+// array_declarator, parenthesized_declarator) captured under the SAME
+// name @step. parenthesized_declarator is transparent (no capture),
+// so paren levels are invisible in the capture list.
+#[test]
+fn test_query_recursive_ref_mixed_alternatives_same_name() {
+    allocations::record(|| {
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            concat!(
+                "(declaration type: (_) @type declarator: [",
+                "  (pointer_declarator declarator: (^)) @step",
+                "  (array_declarator declarator: (^)) @step",
+                "  (parenthesized_declarator (^))",
+                "  (identifier) @name",
+                "])",
+            ),
+        ).unwrap();
+
+        // int ***c;          ~> ptr(ptr(ptr(id)))
+        // int d[10][20][30]; ~> arr(arr(arr(id)))
+        // int **g[10];       ~> ptr(ptr(arr(id)))
+        // int *h[10][20];    ~> ptr(arr(arr(id)))
+        // int (*e)[10];      ~> arr(paren(ptr(id))) - paren level skipped
+        assert_query_matches(
+            &language,
+            &query,
+            "int ***c; int d[10][20][30]; int **g[10]; int *h[10][20]; int (*e)[10];",
+            &[
+                (0, vec![("type", "int"), ("step", "***c"), ("step", "**c"), ("step", "*c"), ("name", "c")]),
+                (0, vec![("type", "int"), ("step", "d[10][20][30]"), ("step", "d[10][20]"), ("step", "d[10]"), ("name", "d")]),
+                (0, vec![("type", "int"), ("step", "**g[10]"), ("step", "*g[10]"), ("step", "g[10]"), ("name", "g")]),
+                (0, vec![("type", "int"), ("step", "*h[10][20]"), ("step", "h[10][20]"), ("step", "h[10]"), ("name", "h")]),
+                (0, vec![("type", "int"), ("step", "(*e)[10]"), ("step", "*e"), ("name", "e")]),
+            ],
+        );
+    });
+}
+
+// Test 4: Recursive alternatives captured under SEPARATE names
+// @ptr and @arr. parenthesized_declarator passes through without
+// capture. Each list only contains matches from its own branch.
+#[test]
+fn test_query_recursive_ref_mixed_alternatives_separate_names() {
+    allocations::record(|| {
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            concat!(
+                "(declaration type: (_) @type declarator: [",
+                "  (pointer_declarator declarator: (^)) @ptr",
+                "  (array_declarator declarator: (^)) @arr",
+                "  (parenthesized_declarator (^))",
+                "  (identifier) @name",
+                "])",
+            ),
+        ).unwrap();
+
+        // int ***c;              ~> ptr, ptr, ptr (3 @ptr)
+        // int d[10][20][30];     ~> arr, arr, arr (3 @arr)
+        // int **g[10];           ~> ptr, ptr, arr (2 @ptr + 1 @arr)
+        // int *h[10][20];        ~> ptr, arr, arr (1 @ptr + 2 @arr)
+        // int (*e)[10];          ~> arr, paren, ptr (1 @arr + 1 @ptr)
+        // int (*(*f)[10])[20];   ~> arr, paren, ptr, arr, paren, ptr (2 @arr + 2 @ptr)
+        assert_query_matches(
+            &language,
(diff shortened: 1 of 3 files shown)


arinar:
So. I've updated this, now basing on master with the "[A+ B]"-stuff (#5396) merged. And related to this, I've also snuck in some extra-handling for ?- and *-quantifiers on branch alternants:
- "[A* B]" -> now matches like "[A+ B]?" instead of like "[A+ B]"
- "[A? B]" -> now matches like "[A B]?" instead of like "[A B]"

Regarding use of "AI"-assistants: Do you prefer more elaboration for what and how I've made use of Claude Code for this, or is the "Co-authored"-by-line in commits enough?

maxbrunsfeld:
I think it’s very very cool that you got this to work with such a minimal addition to the engine. I worry that it’s not very intuitive that the feature is tied to alternations specifically. A related feature I’ve thought about adding is some kind of arbitrary “pattern_a, occurring somewhere within node_b, within pattern_c” syntax, where you can match on containment relationships between without having to explicitly match on the nodes at every depth. I’m curious if that would solve the same problem for you.

arinar:
> where you can match on containment relationships between without having to explicitly match on the nodes at every depth. I’m curious if that would solve the same problem for you.

Yes this would also have helped with my usecase, it's exactly where I was originally coming from actually. I basically want the query to do all the work when looking for complex nonlocal patterns/constellations.

> some kind of arbitrary “pattern_a, occurring somewhere within node_b, within pattern_c” syntax

... initially looked for syntax-inspiration in glob, xpath, jq - in these the "arbitrary-nesting"-syntax is derived from _doubling up_ wildcards or hierarchy-characters ("**" in glob, "//" in xpath, ".." in in jq). So I tried doubling the parentheses `((_))`, i tried `(.. (_) )`, `(__ (_) )` and also `(* (_) )` but neither gave an easy way to integrate into the existing query parser. Then I saw the comment in #880 mentioning `(^)` for recursion and it worked well also with the step-machine so i went with that.

maxbrunsfeld:
I think we should use the `..` or `...` operators to mean arbitrary descendant.

For example, to match any return statement within a function:

```clj
(function_definition
  body: (compound_statement
     ... (return_statement) @ret
  )) @func
```

or to match any identifier within a declarator:

```clj
(declaration
  declarator: ...(identifier) @name) @decl
```

Over the years since releasing the query API, a need for this type of feature has come up several times. My worry with the alternation and `^` syntax is that I think it's a bit harder to understand.

@arinar Are you interested in exploring how this could be implemented? I wonder if we could use a similar approach to what you did with this PR. Personally, I think it would be ok if there were some restrictions on how the `...` syntax worked, at least for the initial version.

arinar:
> @arinar Are you interested in exploring how this could be implemented?

Yes, I will try. `...` would just mean: a step follows that must behave like the initial query step (set initial depth, scan everything, on match: split the query state).

christoph-logicstar:
We also find this feature very interesting.
Our use cases involve matching nodes along the way, and thus would not be possible with arbitrary descent (`...`).
#### 1. Allowing to match nodes as if the grammar was flat.
This would be particularly helpful for the C++ grammar, e.g. to match the parts of a qualified names:
```tsq
[
    (identifier) @part
    (qualified_identifier
        scope: (namespace_identifier) @part
        name: (^)
    )
]
```
It would also be useful in other grammars, e.g. when matching the parts of Java imports:
```tsq
(import_declaration
    [
        (identifier) @part
        (scoped_identifier
            scope: (^)
            name: (identifier) @part
        )
    ]
)
```

#### 2. Querying C/C++ functions
When querying for the name of a declared function, matching all identifiers in its subtree is too coarse. The following query also matches parameter names:
> ```tsq
> (declaration
>   declarator: ...(identifier) @name) @decl
> ```
Additionally, querying for the return type is currently infeasible. Using the `^` operator we could match its components:
```tsq
(declaration
    (type_qualifier)? @type_part
    type: (_) @type_part
    declarator: [
        (pointer_declarator
            "*" @type_part
            (type_qualifier)? @type_part
            declarator: (^)
        )
        (reference_declarator
            "&" @type_part
            (type_qualifier)? @type_part
            declarator: (^)
        )
        (function_declarator
            declarator: (identifier) @name
        )
    ]
)
```
This would extract `type_part` == `["const", "char", "*", "const", "&"]` for the following example:
```cpp
const char* const& get_c_str_ref();
```


What do you think? Are these use cases already too advanced for tree-sitter?

maxbrunsfeld:
Ok, that’s a valid reason to do the recursive matching thing instead, @christoph-logicstar . Thanks for chiming in with your use case.

@arinar is there a way we can mae it more explicit which sub-expression we must jump back to? Like some kind of labeled subexpression, rather than it just being based on alternations?

arinar:
> Like some kind of labeled subexpression, 

I can only think of something like the following, using ^ (or other special characters) to make essentially "labels and gotos" within the query step-machine.

```query
entry^ . [
    (identifier) @part
    (qualified_identifier
        scope: (namespace_identifier) @part
        name: ^entry
    )
]
```

The label part `entry^ .` would put the current step index into a name lookup structure and the goto-part `^entry` would retrieve that index and unconditionally link to that step like the recursion  thing does in the current PR. 

That said, I think such syntax doesn't clearly communicate that it's only meant for recursion. Users could be encouraged to try other patterns: forwards jumps, inward jumps, optional jumps, arbitrary jumps (spaghetti queries). I thought very long about the implications of having to meet those potential expectations or guarding against such uses with additional checks, error messages and sanitizing... I can't see it being only a small addition to the parser or to the step-machine.

> rather than it just being based on alternations?

I want to point out:

1) an alternation is _always required_ to terminate the recursion (reach an inner base-case).

2) both nesting and siblingness distribute over alternation
  - `(A [(B) (C)])` == `[(A (B)) (A (C))]`
  - `((A) [(B) (C)])` == `[((A) (B)) ((A) (C))]`

This suggests that query authors should always be able to structure their query in a way that recurses to a common entry point and have that entry point be the required alternation. If we also would allow unary alternations for just marking any node as recursion target, then with (^^), (^^^), ... repetitions can even be kept to a minimum, i think:

- `(A (B [(A (B (^))) (C)]))` simplifies to `[(A (B [(^^) (C)]))]`

Also just thought of this: The standard pattern for an "i don't care about types"-descent could get a special syntax

- standard: `(A [(_ (^)) (B)])`
- special: `(A [_^ (B)])`
- maybe even: `(A [.. (B)])`

maxbrunsfeld:
> I want to point out ...

Ok, I am totally sold that a recursion syntax is better than an arbitrary descendant syntax. Appreciate both of you arguing for that.

I am *still* not totally sold on automatically recurring to the innermost alternation being the *only* way to recur. What if the `^` operator behaved as follows:

* When appearing by itself, it jumps to the innermost alternation (I can see how this would *often* be what you want)
* When followed by an *identifier*, it recurs back to the innermost node that was *captured* as that name (just using the normal capture syntax).

---

> Users could be encouraged to try other patterns: forwards jumps, inward jumps, optional jumps, arbitrary jumps (spaghetti queries). I thought very long about the implications of having to meet those potential expectations or guarding against such uses with additional checks, error messages and sanitizing... I can't see it being only a small addition to the parser or to the step-machine.

To me, this doesn't seem like a blocker.

arinar:
@maxbrunsfeld Would a syntax not based on @capture also work for you? The problem with these is they only declare the name after the closing `)` of the node to which the recursion would want to jump. The target's step index can't be looked up during encounter of jumps, because it's not captured yet.

The sequence/group-syntax could be extended to allow marking recursion targets explicitly by name. Where i had `[(A (B [(^^) (C)]))]` before, it would be `(^label (A (B [(^label) (C)])))`.

A `(^name`-parse would be label or jump:

- when it is followed by something: capture a the current step index as a named recursion target (on the stack allocated linked list)
- when it's followed by `)`: walk the list and find the first step index with that name and make the special dead-end link.

The label being inside the parentheses reinforces the notion that it is scoped and can only be reached from jumps within that parentheses. It discourages spaghetti.

The current alternation-counting syntax could still be supported (either on the same list or with another list).

good compromise?

The arbitrary descent would look like
```query
(^label
    [
        (tail)
        (_ (^label))
    ]
)
```

@christoph-logicstar's example
```query
(^entry
    [
        (identifier)
        (qualified_identifier
            scope: (namespace_identifier)
            name: (^entry)
        )
    ] @part
)
```

maxbrunsfeld:
> The problem with these is they only declare the name after the closing ) of the node to which the recursion would want to jump. The target's step index can't be looked up during encounter of jumps, because it's not captured yet.

~I still think it might be cleaner to use captures, because it has the least conceptual overhead. Captures already serve a very similar purpose - to assign a name to a particular subtree of a pattern. PCRE regular expressions have a similar feature where you can use recursion with a named (or numbered) capture group.~

~It's slightly more work when parsing, because we need to do a follow-up step after parsing each top-level pattern, to resolve the capture name to a step index, but that doesn't seem prohibitive to me.~

Sorry, I thought about this more and I don't think using captures is a good idea.

maxbrunsfeld:
Alternatively, I wonder if it'd be more useful and clear if we introduced a concept of *named patterns*, where a top-level pattern can be declared with a certain name, which would cause it not be invoked by default, unless referenced in some other pattern. The recursion feature could be built on top of that.

Something like:

```clj
(#def
  name_chain [
    (identifier) @part
    (qualified_identifier
        scope: (namespace_identifier) @part
        name: (^name_chain))
  ])

(call_expression
  function: (^name_chain)) @call
```

arinar:
Yeah, like a macro, reusable in any tails inside other queries.

I do like `(#define name ...)` better than `name: ...` or `name = ...`, because of the s-expr style.

For recursive chains I'm still leaning towards an inline `(^name (... (^name) ))` again because of the scoping.


I'm currently trying to make the recursion more correct regarding field-constraints `(root f1: [ (_ f2: (^)) (tail)])`, since this is probably currently broken, because constraints are taken from the alternation step chain as is. And to resolve this there would have to be additional copies of alternation-chain-entries for each different field constraint used on recursion steps - and this becomes very complicated.

I want to explore a change to query execution, where (positive) field-constraints are first offloaded into a separate _field-gating_ step that goes before the actual step. The gate would not consume input, it would only prime the matcher to check for the field constraint. Then the entry into an alternation step-chain would be reusable with different field-constraints. Also the field-constraint would no longer have to be stamped onto all alternatives in postprocessing.

And then after this experiment, i will tackle "jump-to-name" recursion.

maxbrunsfeld:
That sounds great; appreciate all of the exploration you're doing.

clason:
Looks like that is leaking debug information and suffers from _massive_ performance increases. (This is something that can't be vibe-coded...)

clason:
@arinar I rebased your PR on top of master to make sure the workflows cover your lib changes here; please don't force-push over that.

arinar:
@clason Sorry, I didn't mean to bump this thread by pushing into my fork today. I'm not sure where the branch even lives right now - thought it's in my fork, guess not. I'll try to rebase on master at regular points myself.

clason:
It's still your branch; you have to `git pull -f` to sync your local checkout with the changes on the GitHub side. 

And normally it's not necessary to constantly rebase the branch, but in this case there were directly relevant changes on master that should be included in any CI runs.

quentinLeDilavrec:
@arinar Just in case, considering the initial example and your use case not being highlighting,
I think that https://docs.rs/tree-sitter-graph/ would be a good alternative.

clason:
friendly ping @arinar Are you still interested in getting this to the finish line (which seems to require significant effort, but for significant value)?

WillLillis:
This PR is a great data point for a future implementation, but in its current state the query engine is far to prone to regressions (both correctness and performance) for a change this large. Soon (in the next release cycle or two), we're planning on reworking a large part of the query engine to help address this and make such a feature feasible.

Thanks so much for the PR and the discussion!

PR #5418, opened 2026-03-09
perf(generate): More optimizations

WillLillis:
This is the followup I  mentioned in https://github.com/tree-sitter/tree-sitter/pull/5340#issuecomment-3937947448.

These optimizations don't really have a central theme like #5340 did, instead I found lots of smaller changes to make by examining `perf` traces for various grammars.

### Performance Gains

Gains vary widely depending on grammar (~14% for tree-sitter-c, ~35% for tree-sitter-cpp). Comparisons are made against the current state of master, which already benefits from #5340.

image

image

image

image

image

While working on this, I also noticed the [`optimize`]() profile in the repo's root Cargo.toml file. We should ensure this is used  when creating  our release artifacts, as it nets an additional ~2% speed boost compared to the default `release` profile. The only cost is increased build time.

#### Disclaimer:

I utilized an LLM  to aid in examining the traces from perf,  as well as to help devise the  code changes to address the hot spots found as a result. These changes were made over the course of roughly 20 commits, which I've squashed down  to the 5 here, attempting to group them logically. Most of this work was done  a few weeks ago, so both  I and @amaanq will need to give  this an incredibly thorough review before it's merged.

### DIFF
--- a/crates/generate/src/bitvec.rs
+++ b/crates/generate/src/bitvec.rs
@@ -234,6 +234,7 @@ impl BitVec {
     }
 
     /// Word-level OR: self |= other. Returns true if any new bits were set.
+    #[inline]
     pub fn insert_all(&mut self, other: &Self) -> bool {
         let other_words = other.words_in_use();
         if other_words == 0 {
--- a/crates/generate/src/build_tables/build_lex_table.rs
+++ b/crates/generate/src/build_tables/build_lex_table.rs
@@ -286,20 +286,27 @@ fn check_token_conflicts(
     token_conflict_map: &TokenConflictMap,
     coincident_token_index: &CoincidentTokenIndex,
 ) -> bool {
-    let symbol = Symbol::terminal(i);
-    for existing_token in set_without_terminal.terminals() {
-        if token_conflict_map.does_conflict(i, existing_token.index)
-            || token_conflict_map.does_match_prefix(i, existing_token.index)
-        {
-            return true;
-        }
-        if !coincident_token_index.contains(symbol, existing_token)
-            && (token_conflict_map.does_overlap(existing_token.index, i)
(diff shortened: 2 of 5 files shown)


WillLillis:
This should be ready for review now.

clason:
> While working on this, I also noticed the [optimize](https://github.com/tree-sitter/tree-sitter/pull/5418) profile in the repo's root Cargo.toml file. We should ensure this is used when creating our release artifacts, as it nets an additional ~2% speed boost compared to the default release profile. The only cost is increased build time.

I've always been using that for my own builds ;) (It does strip symbols, though, which makes it harder to debug issues. As the release artifacts are what we (should) point users to for reproducing issues, that's not ideal.)

Speaking of: We might wish to set up a nightly/tip release workflow for that same purpose.

clason:
Noting this down here for reference: We should (also) optimize for memory, not (just) CPU time, as that is  what usually leads to catastrophic performance, especially when parallelizing (as in our CI).

For example, https://github.com/alaviss/tree-sitter-nim requires 8 Gb when generating (which already exhausts GH runners) -- add another parser to this and you're swapping constantly. (This PR seems to help a bit already, though: now it peaks at slightly over 7 Gb and the generate time drops from 31s to 17s.)

clason:
|platform | master | PR |
|---------|--------|----|
|Ubuntu | 13:17 | 10:26|
|macOS | 15:30| 12:00|
|Windows| 18:39| 16:55|

So this translates into a real 25% wall time improvement 🚀

WillLillis:
Little bit of final cleanup, no functional changes. This PR has been tested extensively over the last couple weeks and should be as good to go as possible. If there's any regressions we've missed, they'll be caught via testing/usage, not reading the diff. Any stylistic fixups can be delivered in a followup PR. I'm going to go ahead and merge this as-is in order to unblock related work.

PR #5419, opened 2026-03-10
revert allowing dashes in parser name

WillLillis:
This reverts https://github.com/tree-sitter/tree-sitter/pull/5315

The original intention behind merging that fix was to match parsers that were erroneously (either via a previous version of tree-sitter-cli or manual intervention) using dashes in their parser name. This was the more straightforward path compared to opening PRs for all affected repos and hoping they're still actively maintained. However, this change is leading to issues downstream, so it makes the most sense to revert this change and explicitly document this requirement.

### DIFF
--- a/docs/src/creating-parsers/1-getting-started.md
+++ b/docs/src/creating-parsers/1-getting-started.md
@@ -41,6 +41,10 @@ cd tree-sitter-${LOWER_PARSER_NAME}
 The `LOWER_` prefix here means the "lowercase" name of the language.
 ```
 
+```admonish warning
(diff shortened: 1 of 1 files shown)


WillLillis:
Unsure why CI isn't running, maybe a  GH outage?

clason:
Works on #5422, so maybe it's your fork? Can you try syncing it with upstream and rebasing the branch?

clason:
CI isn't running because it's told in `ci.yml` not to run...

I suggest removing the `path_ignore` entirely -- it's no longer necessary with the faster CI and clearly is incompatible with required checks (which are much higher value).

clason (approved):

PR #5439, opened 2026-03-17
fix(loader): unconditionally import `Mutex`

WillLillis:
Also, remove  the `LazyLock` for `WASM_TOOL_LOCK`. This isn't necessary as `Mutex::new` is valid in const contexts


amaanq (approved):

WillLillis:
Actually looks like this doesn't need to be backported.

PR #5446, opened 2026-03-21
feat: add deprecation attribute to private _array__* functions

### LINKED ISSUE #5443: Use `deprecation` attribute for private array functions
### Problem

```c
/// This is not what you're looking for, see `array_erase`.
static inline void _array__erase
```
doesn't guaranties end user will actually see it.

### Expected behavior

Use 
```c
__attribute__((deprecated("This is not what you're looking for, see `array_erase`.")))
static inline void _array__erase
```
to, at least, show deprecation warning during compilation.

Also, looks like, it is possible to define compiler specific macro, similar to 

```c
#if defined(__GNUC__) || defined(__clang__)
    #define DEPRECATED(msg) __attribute__((deprecated(msg)))
#elif defined(_MSC_VER)
    #define DEPRECATED(msg) __declspec(deprecated(msg))
#else
    #define DEPRECATED(msg)
    #pragma message("WARNING: Deprecated feature mechanism not supported on this compiler.")
#endif
```

eyupcanakman:
Mark all private `_array__*` helper functions with a cross-compiler deprecation attribute. Direct calls to these functions now produce a compiler warning pointing to the correct public macro.

Public macros suppress the warning internally, so existing code using the macros is unaffected. Works on GCC, Clang, and MSVC.

Note: some macros (`array_reserve`, `array_extend`, `array_splice`, `array_insert`, `array_erase`, `array_assign`) changed from expression-form to `do { ... } while(0)` statement-form to accommodate the pragma wrapping. All call sites in tree-sitter use these as statements, but external scanners using them as sub-expressions would need to update.

Closes #5443

### DIFF
--- a/crates/generate/src/templates/array.h
+++ b/crates/generate/src/templates/array.h
@@ -21,6 +21,29 @@ extern "C" {
 #pragma GCC diagnostic ignored "-Wunused-variable"
 #endif
 
+/// Cross-compiler deprecation attribute for private array functions.
+/// Direct calls to `_array__*` functions will produce compiler warnings;
+/// use the corresponding public `array_*` macros instead.
+#if defined(__GNUC__) || defined(__clang__)
+#define TS_DEPRECATED_(msg) __attribute__((deprecated(msg)))
+#define TS_SUPPRESS_DEPRECATED_START_ \
+  _Pragma("GCC diagnostic push") \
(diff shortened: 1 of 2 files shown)


WillLillis:
Thanks, but we're not going to mark these  functions as deprecated. See https://github.com/tree-sitter/tree-sitter/issues/5443#issuecomment-4102768288.

PR #5451, opened 2026-03-23
perf(parser): use `LookaheadIterator` in error recovery's all-symbol scan

WillLillis:
In `ts_parser__do_all_potential_reductions`, when `lookahead_symbol` is 0 (error recovery), the code scanned every symbol from 1 to `token_count` calling ts_language_table_entry for each with most returning empty. Replace with `LookaheadIterator` which efficiently visits only symbols with valid actions, yielding 7-20x fewer lookups for typical grammars.

I tested this by comparing the `benchmark` xtask's average parsing speeds, as well as parsing javascript code as c (and vise versa) to force error recovery:

image

master vs. this branch for error recovery in `cargo xtask benchmark` (in bytes/ms):

bash:          6968 -> 8255  (+18%)
c:                 1831 -> 2460  (+34%)
cpp:            1855 -> 2347  (+27%)
go:              2422 -> 3165  (+31%)
java:            3087 -> 4198  (+36%)
javascript:  2644 -> 3517  (+33%)
php_only:    1737 -> 2649  (+52%)
python:       1734 -> 2477  (+43%)
ruby:           2402 -> 3246  (+35%)
rust:            1653 -> 2254  (+36%)
tsx:              3359 -> 4353  (+30%)
typescript:  2964 -> 3864  (+30%)

This isn't ready to merge yet, as I'd like to work out how to perform some measurements to see what effect this will have on downstream (Neovim).

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1098,6 +1098,35 @@ static void ts_parser__accept(
   ts_stack_halt(self->stack, version);
 }
 
+static bool ts_parser__process_candidate_recovery_actions(
+  TSParser *self,
+  const TSParseAction *actions,
+  uint32_t action_count
+) {
+  bool has_shift_action = false;
+  for (uint32_t i = 0; i < action_count; i++) {
+    TSParseAction action = actions[i];
+    switch (action.type) {
+      case TSParseActionTypeShift:
+      case TSParseActionTypeRecover:
+        if (!action.shift.extra && !action.shift.repetition) has_shift_action = true;
+        break;
+      case TSParseActionTypeReduce:
+        if (action.reduce.child_count > 0)
+          ts_reduce_action_set_add(&self->reduce_actions, (ReduceAction) {
+            .symbol = action.reduce.symbol,
+            .count = action.reduce.child_count,
+            .dynamic_precedence = action.reduce.dynamic_precedence,
+            .production_id = action.reduce.production_id,
+          });
+        break;
+      default:
+        break;
+    }
+  }
+  return has_shift_action;
+}
+
 static bool ts_parser__do_all_potential_reductions(
   TSParser *self,
   StackVersion starting_version,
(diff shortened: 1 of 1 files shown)


lib/src/parser.c
```diff
@@ -1124,37 +1153,16 @@ static bool ts_parser__do_all_potential_reductions(
...
-            break;
-          default:
-            break;
-        }
+      ts_language_table_entry(self->language, state, lookahead_symbol, &entry);
+      has_shift_action = ts_parser__collect_reductions(self, entry.actions, entry.action_count);
+    } else {
+      LookaheadIterator iter = ts_language_lookaheads(self->language, state);
+      while (ts_lookahead_iterator__next(&iter)) {
+        if (iter.symbol == 0 || iter.symbol >= self->language->token_count) continue;
```
amaanq: Better to use `ts_builtin_sym_end` here


```suggestion
        if (iter.symbol == ts_builtin_sym_end || iter.symbol >= self->language->token_count) continue;
```

amaanq:
```suggestion
        // only terminal tokens are valid lookaheads for reduction decisions
        if (iter.symbol == 0 || iter.symbol >= self->language->token_count) continue;
```

lib/src/parser.c
```diff
@@ -1098,6 +1098,35 @@ static void ts_parser__accept(
   ts_stack_halt(self->stack, version);
 }
 
+static bool ts_parser__collect_reductions(
```
amaanq: The name is okay but I'm not a huge fan of it, I sorta like `ts_parser__process_actions` but maybe something else is better..

lib/src/parser.c
```diff
@@ -1098,6 +1098,35 @@ static void ts_parser__accept(
   ts_stack_halt(self->stack, version);
 }
 
+static bool ts_parser__process_actions(
```
maxbrunsfeld: Since this method is pretty specific in what it does and when it should be used, what do you think about a more specific name like `ts_parser__process_candidate_recovery_actions`?

maxbrunsfeld:
Wow, great find. Big perf win while cleaning up the code and re-using an existing code path!

lib/src/parser.c
WillLillis: Amaan and I both weren't super happy about the name here, that sounds much better!

WillLillis:
I tested this PR's impact by replicating `cargo xtask benchmark` in lua to run in neovim for most of the fixture grammars. The neovim results match the same general trend from the xtask:

bash: 15931 -> 17297 (+9%)
c: 1864 -> 2398 (+25%)
cpp: 2129 -> 2499 (+17%)
go: 2903 -> 3651 (+26%)
java: 3408 -> 4710 (+36%)
javascript: 2964 -> 3913 (+32%)
php_only: 1964 -> 2871 (+46%)
python: 1695 -> 2231 (+32%)
ruby: 2449 -> 3141 (+28%)
rust: 1985 -> 2598 (+31%)
tsx: 4130 -> 5051 (+22%)
typescript: 3360 -> 4058 (+21%)

There's a few differences to the xtask results due to which example files got copied over, but the general trend still holds.

CC @clason So positive impact for neovim in error recovery, but I don't think it's worth downstreaming the benchmarking code. It's a bit hacked together and (at least in the current state) would require committing a fair amount of corpus files.

clason:
Thanks! Yeah, Neovim benchmarks shouldn't duplicate upstream but target specific and representative editor workloads.

clason:
Ok, I think I found a regression: The nvim-treesitter indent test for `php` fail now. (Bisected to this commit.)

PR #5453, opened 2026-03-23
Add SLSA generic generator workflow

Tawhidsdk:
This workflow generates SLSA provenance files for the project, satisfying level 3 requirements. It includes steps to build artifacts and generate provenance subjects.


clason:
This is clearly agentic spam.

PR #5456, opened 2026-03-25
Prevent a buffer over-read when parsing 4-byte UTF-16 characters.

danielpunkass:
When an input stream includes a 4-byte character (such as 🦋, for example), and the character's leading surrogate lands as the last character of a chunk, the unicode parser will erroneously read past the end of the chunk's buffer, creating a likely corrupt code point value.

The length parameter to ts_decode_utf16_** is in bytes, but it was being passed as-is to the U16_NEXT_** macros, which expect length to be in code units (2-bytes each). length parameter to ts_decode_utf16_** is in bytes and was being passed as-is to the U16_NEXT_** macros, which expect length to be in code units (2-bytes each).

You can see how the ts_decode_utf16 functions multiply the resulting code unit count by 2 in order to convert BACK to bytes, but the corresponding conversion from bytes TO code units was not occurring.

In practice this probably didn't show up too often, because it requires the 4-byte character to land at the end of a chunk, and because the "length" in this context is strictly treated as a limit for whether or not the U16_NEXT macros should consider the subsequent code unit. For every unit in the chunk that is not flush against the end of the buffer, the length value could be anything but 2 and not limit the macro's ability to fetch the trailing surrogate.


WillLillis:
@danielpunkass Thanks for catching this issue! Your first commit looks good. It took me some time but I worked out a test case so we can have some coverage for this. I've pushed that up to your branch. I'd lean against dropping the second commit, however. I think it should be the responsibility of the caller to pass in valid text, and catching that here would paper over bugs elsewhere. I'll leave the final decision up to @amaanq / @maxbrunsfeld, however. :slightly_smiling_face:.

danielpunkass:
Thanks @WillLillis for adding the tests! I don't have a strong feeling about the second, so happy to let the more experienced and authoritative folks decide.

amaanq (approved):

amaanq:
thanks!

PR #5457, opened 2026-03-25
Update the Dockerfile so it works with current source base

danielpunkass:
Update the Dockerfile so it works with current source base so that the resulting image is suitable for building and running tests against the current source base. The previous Dockerfile specified an image that didn't have the right versions of rust or nodejs for all the dependencies.


clason:
Hmm, I think that is already outdated. I would rather remove the Dockerfile altogether.

PR #5463, opened 2026-03-28
Fix/supertype unwrap panic 0.26.7

### LINKED ISSUE #5462: `tree-sitter generate` panics when supertype wraps a hidden external token
## Problem

`tree-sitter generate` panics when a supertype rule's only child is a hidden external token:

```
thread 'main' panicked at crates/generate/src/node_types.rs:677:71:
called `Option::unwrap()` on a `None` value
```

## Reproducer

`grammar.js`:
```js
module.exports = grammar({
  name: "repro",

  externals: $ => [$._hidden_ext],

  supertypes: $ => [$._type_a, $._type_b],

  rules: {
    source_file: $ => repeat(choice($._type_a, $._type_b)),

    // Supertype A: normal choice of named subtypes (works)
    _type_a: $ => choice($.v1, $.v2),
    v1: $ => "x",
    v2: $ => "y",

    // Supertype B: wraps a hidden external token (panics)
    _type_b: $ => $._hidden_ext,
  },
});
```

`src/scanner.c`:
```c
#include <stdlib.h>
enum TokenType { HIDDEN_EXT };
void *tree_sitter_repro_external_scanner_create(void) { return NULL; }
void tree_sitter_repro_external_scanner_destroy(void *p) { (void)p; }
unsigned tree_sitter_repro_external_scanner_serialize(void *p, char *b) { (void)p; (void)b; return 0; }
void tree_sitter_repro_external_scanner_deserialize(void *p, const char *b, unsigned n) { (void)p; (void)b; (void)n; }
bool tree_sitter_repro_external_scanner_scan(void *p, TSLexer *l, const bool *v) { (void)p; (void)l; (void)v; return false; }
```

Steps:
```bash
tree-sitter generate  # panics
```

## Root Cause

`get_variable_info()` filters hidden children from supertype `children.types` (line ~366: `retain(child_type_is_visible)`). When the only child is a hidden external token, this leaves zero subtypes.

`generate_node_types_json()` then pushes `(supertype, [])` into `subtype_map` (line 587). The topological sort at line 660 has no edges for this supertype, so its kind never enters `sorted_kinds`. The `sort_by` at line 682 panics when looking up the missing kind.

### Expected behavior

Expect to not crash.

tree-sitter 0.26.7

### Operating system/version

macOS 26.4

FranklinChen:
## Fix: prevent panic when supertype wraps a hidden external token

Fixes #5462

Yes, I used an LLM, but I also manually looked at the changes and ran these commands myself with my own fingers to double check everything this time.

### Problem

`tree-sitter generate` panics when a supertype rule's only child is a hidden external token:

```
thread 'main' panicked at crates/generate/src/node_types.rs:677:71:
called `Option::unwrap()` on a `None` value
```

### Root Cause

`get_variable_info()` filters hidden children from supertype `children.types` (line ~366: `retain(child_type_is_visible)`). When the only child is a hidden external token, this leaves zero subtypes.

`generate_node_types_json()` then pushes `(supertype, [])` into `subtype_map` (line 587). The topological sort at line 660 adds no edges for this supertype, so its kind never enters `sorted_kinds`. The `sort_by` at line 682 panics when `position()` returns `None`.

### Fix

Skip pushing to `subtype_map` when subtypes is empty:

```rust
if !subtypes.is_empty() {
    subtype_map.push((supertype, subtypes.clone()));
}
```

A supertype with zero visible subtypes has no dependency ordering to compute, so excluding it from the sort is correct.

### Test

`test_node_types_supertype_with_only_hidden_child`: two supertypes, one with normal named subtypes, one wrapping a hidden external. Panics without the fix, passes with it. All 60 existing tests pass.

Without the fix:

```
cargo test -p tree-sitter-generate --features qjs-rt test_node_types_supertype_with_only_hidden_child
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running unittests src/generate.rs (target/debug/deps/tree_sitter_generate-7ae5e72afb297ba1)

running 1 test
test node_types::tests::test_node_types_supertype_with_only_hidden_child ... FAILED

failures:

---- node_types::tests::test_node_types_supertype_with_only_hidden_child stdout ----

thread 'node_types::tests::test_node_types_supertype_with_only_hidden_child' (25867235) panicked at crates/generate/src/node_types.rs:677:71:
called `Option::unwrap()` on a `None` value


failures:
    node_types::tests::test_node_types_supertype_with_only_hidden_child

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 59 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p tree-sitter-generate --lib`
```

### DIFF
--- a/crates/generate/src/node_types.rs
+++ b/crates/generate/src/node_types.rs
@@ -584,7 +584,14 @@ pub fn generate_node_types_json(
                 kind: node_type_json.kind.clone(),
                 named: true,
             };
-            subtype_map.push((supertype, subtypes.clone()));
+            // Only add to subtype_map if there are visible subtypes.
+            // A supertype may have zero subtypes if its children are all
+            // hidden (e.g., wrapping a hidden external token). In that case,
+            // the supertype won't participate in the topological sort and
+            // must be excluded from the sort at line ~682.
+            if !subtypes.is_empty() {
+                subtype_map.push((supertype, subtypes.clone()));
+            }
             node_type_json.subtypes = Some(subtypes);
         } else if !syntax_grammar.variables_to_inline.contains(&symbol) {
             // If a rule is aliased under multiple names, then its information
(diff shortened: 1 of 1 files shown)


WillLillis:
@FranklinChen It looks like you created your fix off of the release-0.26 branch and are now targeting master. You should perform your fix off of the master branch instead. We can backport any fix to the release branch afterward.

PR #5471, opened 2026-03-30
perf(cli): replace regex with manual parsing in test file parser PLUS loader lock file rewrite

WillLillis:
This started as a simple one to one rewrite, just removing the regexes, and quickly devolved into a rewrite of the test parsing logic. In addition to the memory enhancements, the general flow should be much clearer now. A few data points for `tree-sitter test`:

- JS:   walltime -1.3%, peak rss -10.2%
- C:    walltime -5.7%, peak rss -4.5%
- Rust: walltime -3.8%, peak rss -2.2%

While the test case format (and its associated error reporting) isn't a pain point I've heard much from users, this should make it significantly easier to improve error reporting and/or add features to the test format in the future. One thing I noticed during this rewrite is that many format issues lead to test cases being silently dropped. I'm happy to start adding warnings for these cases, but wanted to keep the diff smaller for this first set of changes.

Edit: The second commit in this PR includes a rewrite of the loader crate's build locking logic. We've seen enough spurious errors in CI to fix this. Open to feedback here, especially the poll-style check in the loser's case.

### DIFF
--- a/crates/loader/src/loader.rs
+++ b/crates/loader/src/loader.rs
@@ -1142,14 +1142,7 @@ impl Loader {
             format!("{:x}", hasher.finish())
         };
 
-        let mut lock_path = if env::var("CROSS_RUNNER").is_ok() {
-            tempfile::tempdir()
-                .expect("create a temp dir")
-                .path()
-                .to_path_buf()
-        } else {
-            etcetera::choose_base_strategy()?.cache_dir()
-        };
+        let mut lock_path = etcetera::choose_base_strategy()?.cache_dir();
         lock_path.push(format!(
             "tree-sitter{slash}lock{slash}{name}-{lock_hash}.lock",
             slash = std::path::MAIN_SEPARATOR,


WillLillis:
CI failures are an unrelated race condition in the loader. I was working on fixing that a few days ago but haven't fully gotten it yet. I'll spend some more time on that now.

crates/loader/src/loader.rs
```diff
@@ -711,6 +713,66 @@ impl<'a> CompileConfig<'a> {
...
+fn temp_path(path: &Path) -> PathBuf {
+    let filename = path.file_name().unwrap().to_string_lossy();
+    path.with_file_name(format!(
+        ".{filename}.{}.{:?}",
+        std::process::id(),
+        std::thread::current().id()
+    ))
+}
+
+/// RAII lock file guard. The lock file is created atomically via `create_new`
```
amaanq: this is a little confusing because it could be mistaken for `LockFile::create` as a typo?

```suggestion
/// RAII lock file guard. The lock file is created atomically via
/// [`create_new`](`fs::OpenOptions::create_new`) and removed on drop.
```

crates/loader/src/loader.rs
```diff
@@ -711,6 +713,66 @@ impl<'a> CompileConfig<'a> {
...
+            Ok(_) => Ok(Some(Self {
+                path: path.to_path_buf(),
+            })),
+            Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => Ok(None),
+            Err(e) => Err(LoaderError::IO(IoError::new(e, Some(path)))),
+        }
+    }
+
+    /// Wait for an existing lock file to be removed by whoever created it.
+    /// If the lock file persists beyond `timeout`, return `LoaderError::LockFile`
```
amaanq:
```suggestion
    /// If the lock file persists beyond `timeout`, return [`LoaderError::LockFile`]
```

crates/loader/src/loader.rs
```diff
@@ -711,6 +713,66 @@ impl<'a> CompileConfig<'a> {
...
+
+/// RAII lock file guard. The lock file is created atomically via `create_new`
+/// and removed on drop.
+struct LockFile {
+    path: PathBuf,
+}
+
+impl LockFile {
+    /// Attempt to atomically create the lock file. Returns `Ok(Some)` if we won the
+    /// race, `Ok(None)` if we lost, or the underlying IO error otherwise.
```
amaanq:
```suggestion
    /// Attempt to atomically create the lock file.
    ///
    /// Returns `Ok(Some)` if we won the race, `Ok(None)` if we lost,
    /// or the underlying IO error otherwise.
```

crates/loader/src/loader.rs
amaanq:
```suggestion
    /// If the lock file persists beyond `timeout`, return [`LoaderError::LockFileTimeout`]
```

crates/cli/src/test.rs
```diff
@@ -32,36 +29,56 @@ use crate::{
...
-static DIVIDER_REGEX: LazyLock<ByteRegex> = LazyLock::new(|| {
-    ByteRegexBuilder::new(r"^(?P<hyphens>(?:-+){3,})(?P<suffix>[^-\r\n][^\r\n]*)?\r?\n")
-        .multi_line(true)
-        .build()
-        .unwrap()
-});
+/// Check if a line consists of 3+ repetitions of `ch` followed by an optional suffix.
+/// Returns `Some((delim_len, suffix))` if so, where `suffix` is the part after the
+/// repeated characters (empty string if no suffix). Returns `None` otherwise.
+fn parse_delimiter_line(line: &str, c: char) -> Option<(usize, &str)> {
```
amaanq, line 36:
```suggestion
///
/// Returns `Some((delim_len, suffix))` if so, where `suffix` is the part after the
/// repeated characters (empty string if no suffix). Returns `None` otherwise.
fn parse_delimiter_line(line: &str, c: char) -> Option<(usize, &str)> {
```

crates/cli/src/test.rs
```diff
@@ -1268,230 +1302,334 @@ pub fn parse_tests(path: &Path) -> io::Result<TestEntry> {
...
+}
 
-                let (output, has_fields) = if prev_attributes.cst {
-                    (output.trim().to_string(), false)
-                } else {
-                    // Remove all comments
-                    let output = COMMENT_REGEX.replace_all(output, "").to_string();
+fn parse_test_content(name: String, content: &str, file_path: Option<PathBuf>) -> TestEntry {
+    let mut children = Vec::new();
+    let lines: Vec<&str> = content.split_inclusive('\n').collect();
```
amaanq: #teamturbofish

```suggestion
    let lines = content.split_inclusive('\n').collect::<Vec<_>>();
```

crates/cli/src/test.rs
```diff
@@ -1268,230 +1302,334 @@ pub fn parse_tests(path: &Path) -> io::Result<TestEntry> {
...
+    }
+
+    let (divider_delim_len, divider_line) = best_divider?;
+
+    // Input: lines before the divider (as bytes), with trailing newline stripped.
+    let mut input: Vec<u8> = body_lines[..divider_line]
+        .iter()
+        .flat_map(|l| l.as_bytes())
+        .copied()
+        .collect();
```
amaanq:
```suggestion
    let mut input = body_lines[..divider_line]
        .iter()
        .flat_map(|l| l.as_bytes())
        .copied()
        .collect::<Vec<_>>();
```

crates/cli/src/test.rs
```diff
@@ -1268,230 +1302,334 @@ pub fn parse_tests(path: &Path) -> io::Result<TestEntry> {
...
+    // Remove trailing newline.
+    if input.last() == Some(&b'\n') {
+        input.pop();
+    }
+    if input.last() == Some(&b'\r') {
+        input.pop();
+    }
+
+    // Output: lines after the divider.
+    let output_str: String = body_lines[divider_line + 1..].iter().copied().collect();
```
amaanq:
```suggestion
    let output_str = body_lines[divider_line + 1..].iter().copied().collect::<String>();
```

crates/loader/src/loader.rs
```diff
@@ -711,6 +713,66 @@ impl<'a> CompileConfig<'a> {
 
 unsafe impl Sync for Loader {}
 
+/// Generate a temporary file path for atomic writes. The temp file is a hidden
+/// dotfile alongside the target, tagged with the current process and thread id
+/// for debuggability (e.g. `.javascript.so.12345.ThreadId(3)`).
+fn temp_path(path: &Path) -> PathBuf {
+    let filename = path.file_name().unwrap().to_string_lossy();
```
amaanq:
```suggestion
    let filename = path.file_name().expect("output_path must have a filename).to_string_lossy();
```

crates/loader/src/loader.rs
```diff
@@ -1091,85 +1134,81 @@ impl Loader {
...
+                            Self::check_external_scanner(&output_path);
+                        }
+                    }
+                    // _lock dropped here, removing the lock file.
+                }
+                None => {
+                    // Another thread/process is compiling (or a previous run
+                    // crashed and left a stale lock). Wait for it to finish.
+                    LockFile::wait_for_removal(&lock_path, Duration::from_secs(30))?;
+                }
```
amaanq:
```suggestion
                // Another thread/process is compiling (or a previous run
                // crashed and left a stale lock). Wait for it to finish.
                None => LockFile::wait_for_removal(&lock_path, Duration::from_secs(30))?,
```

PR #5475, opened 2026-03-31
Fix wasm loading of languages w/ multiple reserved word sets

maxbrunsfeld:
Previously, languages with multiple reserved word sets were not handled properly when loaded from WASM. This would cause crashes during lexing.

### DIFF
--- a/crates/cli/src/tests/wasm_language_test.rs
+++ b/crates/cli/src/tests/wasm_language_test.rs
@@ -315,6 +315,55 @@ fn test_load_wasm_errors() {
 }
 
 #[test]
+fn test_load_wasm_language_with_reserved_words() {
+    // This test exercises a grammar with multiple reserved word sets loaded via WASM.
+    allocations::record(|| {
+        let store = WasmStore::new(&ENGINE).unwrap();
+        let language = get_test_fixture_language_wasm("reserved_words");
+
+        let mut parser = Parser::new();
+        parser.set_wasm_store(store).unwrap();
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
@WillLillis @amaanq Thanks for the review and backport. Are you cool with doing a patch release on the 0.26 line, so I can get this crash fix into Zed? I can do it, just wanted to check in since I typically cause chaos when I do releases these days 😓 .

clason:
The workflow is [documented now](https://github.com/tree-sitter/tree-sitter/blob/master/docs/src/6-contributing.md#release-workflow) (for no particular reason 😇 ) but I can do it if you prefer. Is it urgent or can it wait till the end of the week? (There's one or two backportable PRs that would be nice to include, but nothing that needs to hold up a release.)

maxbrunsfeld:
That would be wonderful @clason , and thanks for the docs pointer. Sooner would be better, since this crash is reachable for end users depending on what grammars they install.

clason:
https://github.com/tree-sitter/tree-sitter/pull/5479

WillLillis:
@maxbrunsfeld I don't have any issue with a patch release. If it's not super urgent like @clason  said, I can take a look at those other PRs later tonight after work to get them in. If you need it sooner, that's fine too :)

maxbrunsfeld:
Tonight would be totally fine

clason:
Too late, it's already on crates.io!

maxbrunsfeld:
Just got the emails. Thank you!

PR #5484, opened 2026-04-01
Add Solaris support to the portable endian header

psumbera:
Solaris does not provide <endian.h> or <sys/endian.h>, but it does expose byte-order definitions and conversion helpers via <sys/isa_defs.h> and <sys/byteorder.h>.

Add a __sun branch so the portable header defines __BYTE_ORDER and the htobe*/le*toh conversions on Solaris.

### DIFF
--- a/lib/src/portable/endian.h
+++ b/lib/src/portable/endian.h
@@ -41,6 +41,25 @@
 
 # include <sys/endian.h>
 
+#elif defined(__sun)
+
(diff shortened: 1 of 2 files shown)


lib/src/portable/endian.h
m74278803-cmyk: psumbera:solaris

amaanq (approved):

PR #5491, opened 2026-04-03
Separate `init` crate

### LINKED ISSUE #5490: Package `tree_sitter_cli::init` as a crate
### Problem

Currently the `tree-sitter-cli` crate exposes the module `init`. 
In my use-case, I depend only on this module to generate a Tree-sitter directory programmatically. But then, I am forced to depend on all the external dependencies of `tree-sitter-cli`. 

Since several cli/lib parts of Tree-sitter are already split in several sub-crates, would it be considered to also package `init` in a similar way, eg., `tree-sitter-init`? 
I suppose other modules of the crate could benefit the same process but I have no specific use of them at the moment so I cannot say.

### Expected behavior

`tree-sitter-init` is a crate.

fosskers:
This PR extracts `tree_sitter_cli::init` into its own crate in order to free it of the many dependencies imposed by the rest of that crate. This helps us downstream as we only need `init` and a few things from `tree_sitter_generate`, not the entire `tree_sitter_cli` crate itself.

Another option would be to relocate such logic directly into `tree_sitter_generate`.

Closes #5490


maxbrunsfeld:
Could you explain your use case a bit more? Is there any existing crate we could put this into that your app depends on already? We can do a new crate if absolutely necessary, but it adds a moving part to the maintenance going forward, so it’d be nice to avoid.

clason:
We were actually talking about going the other way: Splitting up the monolithic `tree-sitter` binary into a bundle of two or three independent binaries; e.g., the playground (and with it all the web bindings) could be separate, very few people use the `highlight` or `tags` function these days, etc. The goal would be to have a slim "grammar compiler" with the core functionality for grammar developers and consumers and leave the rest as standalone binaries or modules.

We also need to rethink how we handle bindings, some of which straddle the `init` and `generate` steps. (We actually moved a large part of these from `generate` to `init` to reduce the churn during grammar development and for downstream consumers; simply moving them back would be a regression, I think.)

(Long-term, it would be nice to extend `tree-sitter` with format, lint, clippy, and even language server capabilities for grammars, and shoving these into the main binary would simply be too much.)

Lelio-Brun:
On behalf of @fosskers, @maxbrunsfeld our use case is a parser generator: we take a grammar file in our custom format and generate automatically a TS parser from it. Instead of generating custom directories, we use `tree_sitter_cli::init`.
Now the problem is that this makes our codebase depending on all the dependencies of `tree-sitter-cli`.

Since we also use `tree-sitter-generate`, moving `init` into it would solve our problem indeed. But if it is a regression as @clason says, then going with separate crates seems like the only way.

clason:
@Lelio-Brun I don't mean this dismissively, but this seems like an XY problem: If all you want is to translate a stand-alone `grammar.mydsl` to `parser.c` (plus necessary headers), `tree-sitter generate` should be all you need. So what part exactly of `tree-sitter init` (which does _a lot_) do you actually _need_? Maybe we can remove that requirement, or move _that_ back into `generate` (if it makes sense).

Lelio-Brun:
We use `tree_sitter_cli::init::generate_grammar_files`. In theory yes, we could manually build those files, and the important part of the generation is in `tree_sitter_generate`. But it is convenient to use `init` to ensure generation of compliant directories with automatically generated bindings.

vxnkjyffmq-code (approved):

alvarezlupe16442025-droid (approved):

fosskers:
Spam approvals?

clason:
Yes, spam approvals. (Unfortunately, Github doesn't offer any tools to deal with those.)

Anyway, as mentioned above, this is not a direction we want to go into, sorry.

PR #5492, opened 2026-04-03
feat(generate): don't require `tree-sitter.json` for ABI 15

clason:
Problem: Many parser projects (about 100 out of ~330 tracked by
nvim-treesitter) still do not contain a `tree-sitter.json` config file,
so `tree-sitter generate` refuses to generate ABI 15 parsers due to the
lack of version information. This prevents these parsers from profiting
from other improvements in ABI 15+ and will eventually make them
unusable when ABI 14 support is dropped.

Solution: Simply use a default version `0.0.0` if no `tree-sitter.json`
is found.

This will allow the `nvim-treesitter-generate` workflow to give much wider
coverage of the ABI 15 code path.

(Don't know why I didn't think of that initially 🤦 )

### DIFF
--- a/crates/generate/src/render.rs
+++ b/crates/generate/src/render.rs
@@ -6,7 +6,6 @@ use std::{
 };
 
 use crate::LANGUAGE_VERSION;
-use indoc::indoc;
 use serde::Serialize;
 use thiserror::Error;
 
@@ -109,6 +108,7 @@ struct LargeCharacterSetInfo {
     is_used: bool,
 }
 
+#[derive(Clone, Copy, Default)]
 struct Metadata {
     major: u8,
     minor: u8,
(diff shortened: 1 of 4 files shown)


crates/generate/src/render.rs
```diff
@@ -1658,15 +1658,7 @@ impl Generator {
...
-            let Some(metadata) = &self.metadata else {
-                panic!(
-                    indoc! {"
-                        Metadata is required to generate ABI version {}.
-                        This means that your grammar doesn't have a tree-sitter.json config file with an appropriate version field in the metadata table.
-                    "},
-                    self.abi_version
-                );
-            };
+            let metadata = self.metadata.unwrap_or_default();
```
clason, line 1661: This should also make sure future metadata is handled gracefully.

maxbrunsfeld:
Should we fall back on looking at package.json, or a Cargo.toml, for a version number?

clason:
No, I think making tree-sitter.json the single source of truth was the right call. If the project is abandoned, none of the manifest files are useful. The point of this is to unblock consumers who know what they're doing and want to use the grammar despite it being abandoned.

WillLillis:
Need to cargo build (or similar) to update Cargo.lock

PR #5496, opened 2026-04-03
ci(nvim-ts): minor improvements

clason:
* remove obsolete msvc-dev-cmd action (runners come with build tools)
* build tree-sitter CLI with --profile optimize (longer build time due
  to LTO, but generate workflow should profit)
* increase generate parallelism to 3

### DIFF
--- a/.github/workflows/nvim_ts.yml
+++ b/.github/workflows/nvim_ts.yml
@@ -37,14 +37,13 @@ jobs:
           ref: main
 
       - if: runner.os != 'Windows'
-        run: echo ${{ github.workspace }}/target/release >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


clason:
Bit of a wash -- the LTO increases build time by 2x (30-60s), slightly hurting the build steps while the generate steps mostly profit from the increased parallelism (by ~15%). Higher parallelism is blocked by memory pressure (`N_I_m`, I'm looking at you!).

Probably won't hurt merging this; I'll keep an eye out in the hope that performance improvements unlock further options here.

(I am keenly aware of the cost of these workflows, but they've been very useful in catching unexpected consequences of changes. Ideally, I would even like to expand them so we can also test with Neovim built against the lib from the PR, but I should start pruning before that...)

PR #5508, opened 2026-04-07
NULL check for length before dereferencing it.

Adarshupase:
Caller can pass NULL if they don't want to use length. And if they accidentally do check if it is NULL without having to dereference a NULL pointer.


amaanq:
The caller is never supposed to pass `NULL` to this. I'm okay with a minor adjustment to the api header to mention this, but this isn't necessary.

Issue #4801, opened 2025-09-07
`grammar.json` schema extremely laggy


RedCMD:
### Problem

Trying to open a TreeSitter `grammar.json` file in VSCode freezes all JSON language features without error
https://github.com/RedCMD/Lag-Syntax-Highlighter/blob/main/_test/grammar.json

dup:
* https://github.com/microsoft/vscode/issues/256401

### Steps to reproduce

Steps to Reproduce:
1. open VSCode
1. open a large `grammar.json` tree-sitter file
4. notice once the schema loads, all JSON language features stop working
  Hover, format, Intellisense, diagnostics etc
```json
{
	"$schema": "https://tree-sitter.github.io/tree-sitter/assets/schemas/grammar.schema.json",
	"name": "laggy-schema",
	"rules": {
		"json": {
			"type": "REPEAT",
			"content": {
				"type": "REPEAT",
				"content": {
					"type": "REPEAT",
					"content": {
						"type": "REPEAT",
						"content": {
							"type": "REPEAT",
							"content": {
								"type": "REPEAT",
								"content": {
									"type": "REPEAT",
									"content": {
										"type": "REPEAT",
										"content": {
											"type": "REPEAT",
											"content": {
												"type": "REPEAT",
												"content": {
													"type": "REPEAT",
													"content": {
														"type": "REPEAT",
														"content": {
															"type": "REPEAT",
															"content": {
																"type": "REPEAT",
																"content": {
																	"type": "REPEAT",
																	"content": {
																		"type": "REPEAT",
																		"content": {
																			"type": "BLANK"
																		}
																	}
																}
															}
														}
													}
												}
											}
										}
									}
								}
							}
						}
					}
				}
			}
		}
	}
}
```

### Expected behavior

no lag

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.25.9

### Operating system/version

Windows 11

clason:
How is this a tree-sitter problem? VS Code uses TextMate and LSP semantic tokens, not tree-sitter.

ObserverOfTime:
The issue is about the grammar schema, not tree-sitter itself.
I can't reproduce the lag in Neovim but hover doesn't respond and formatting times out.

Not sure if there's anything we can do about this though.
The `grammar.json` file isn't meant to be edited by hand anyway.

RedCMD:
> How is this a tree-sitter problem

@clason cause the schema is hosted in this repo 🙃
[...]

PR #5509, opened 2026-04-07
Improve efficency of grammar schema

jdesrosiers:
It came to my attention that the grammar JSON Schema is extremely inefficient. See, https://github.com/microsoft/vscode-json-languageservice/issues/284. Because of the `oneOf` and the recursive nature of the schema, every level of nesting results in exponentially longer evaluation. Vscode ended up working around the problem by implementing an optimization, but that shouldn't be expected universally. A more robust solution is to change the `oneOf` to `if`/`then`s so the schema is more intentional and can work efficiently in any validator.

### DIFF
--- a/docs/src/assets/schemas/grammar.schema.json
+++ b/docs/src/assets/schemas/grammar.schema.json
@@ -318,20 +318,124 @@
     },
 
     "rule": {
-      "oneOf": [
-        { "$ref": "#/definitions/alias-rule" },
-        { "$ref": "#/definitions/blank-rule" },
-        { "$ref": "#/definitions/string-rule" },
-        { "$ref": "#/definitions/pattern-rule" },
-        { "$ref": "#/definitions/symbol-rule" },
-        { "$ref": "#/definitions/seq-rule" },
-        { "$ref": "#/definitions/choice-rule" },
-        { "$ref": "#/definitions/repeat1-rule" },
-        { "$ref": "#/definitions/repeat-rule" },
-        { "$ref": "#/definitions/reserved-rule" },
-        { "$ref": "#/definitions/token-rule" },
-        { "$ref": "#/definitions/field-rule" },
-        { "$ref": "#/definitions/prec-rule" }
+      "allOf": [
+        {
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but see https://github.com/tree-sitter/tree-sitter/issues/4801.

jdesrosiers:
If you don't mind, I think there's some misunderstanding about the consequences here.

> 1. "weird schema layout": weird != wrong

It actually is harmful. The other thread points out,

> prob leads to catastrophic backtracking (as it tries every combination, then backtracks, just like bad regexes)

That's exactly what's happening. Anyone using this schema to validate a grammar is going to effectively [ReDoS](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) themselves.

> 2. "bad idea to validate a grammar.json file using the schema": as this is a purely internal representation, and we have our own tooling for working with this file, this is not something we consider a valid use case

If that's the case, I'll submit a PR to remove the entry from schemastore.org citing that it's unsafe in its current form.

> 4. "freezes all other JSON Language Features during the validation time": that is an editor problem and will affect any sufficiently complex schema.

It's not an editor problem. Any validator that tries to use this schema is going to have the same problem, not just just editors. The editor is doing the best it can in a situation that's equivalent to a ReDoS attack. The problem is not that the schema is complex. It's not. Far larger and more complex schemas are handled with no problem at all. The edit that I have proposed fixes the catastrophic backtracking problem and the schema will be handled with no problem in any context.

ObserverOfTime:
Personally I think this change makes sense.

WillLillis:
This is stale, and will need further rework due to future changes to grammar generation. We'll spend some time looking into this issue, but it's considered low priority, especially given that [microsoft has already fixed the problem on their end](https://github.com/microsoft/vscode-json-languageservice/pull/320).

jdesrosiers:
Ping me if I can help.

Also, keep in mind that although vscode has implemented a work around, other validators and editors that use schemastore.org haven't.

PR #5522, opened 2026-04-14
build(zig): update to zig 0.16

dchinmay2:
Updates to other repositories are pending this one:
deps

Updating zig-tree-sitter will need the commit hash from tree-sitter after this PR is merged, then I can update the parser template's build.zig and build.zig.zon after zig-tree-sitter is updated, then I can update tree-sitter-c using the hash from updated zig-tree-sitter and finally update the build.zig.zon in zig-tree-sitter to use the updated tree-sitter-c so that all tests build

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -4,51 +4,56 @@ pub fn build(b: *std.Build) !void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
+    var threaded: std.Io.Threaded = .init(b.allocator, .{});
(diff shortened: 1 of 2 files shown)


bfredl (approved):
I tested to use it as a zig 0.16.0 dependency in neovim, works fine:)

clason:
Please update the `zig.zon` manifest as well (or delete it if it shouldn't be committed); minimum Zig version is now 0.16 after these changes.

dchinmay2:
done

clason (approved):

PR #5524, opened 2026-04-15
build(swift): use single-file amalgamation for SPM target (fixes #5523)

### LINKED ISSUE #5523: Xcode 26 / Swift 6.3: 'import TreeSitter' yields empty module (TSLanguage not in scope)
## Summary

On Xcode 26.4 (Swift 6.3, swiftlang-6.3.0.123.5), Swift consumers of the `TreeSitter` SPM product can `import TreeSitter` without error, but cannot reference any C declarations from `lib/include/tree_sitter/api.h` — `TSLanguage`, `TSInputEncoding`, `TSInput`, etc. all resolve to `cannot find type ... in scope`. Affects every release from 0.25.0 through 0.26.8 (latest as of writing); 0.20.x is fine.

The TreeSitter Clang module *is* loaded — Clang's own lookup tables contain the symbols (see "Evidence" below). The bug is in how Xcode 26's SwiftPM/build system surfaces the publicHeadersPath modulemap to Swift consumers, not in the C sources themselves. Notably, [`swift-tree-sitter`](https://github.com/tree-sitter/swift-tree-sitter)'s CI [pins `DEVELOPER_DIR: /Applications/Xcode_16.4.app`](https://github.com/tree-sitter/swift-tree-sitter/blob/main/.github/workflows/ci.yml), so the upstream Swift binding test matrix does not catch this regression.

## Environment

- Xcode 26.4 (Build 17E192)
- Swift 6.3 (`swift-frontend -version` → `swiftlang-6.3.0.123.5 clang-2100.0.123.102`)
- `swift-tools-version: 5.8` (matches `tree-sitter`'s own Package.swift)
- macOS host, both `iphoneos` and `iphonesimulator` destinations affected
- Reproduces with `xcodebuild` and Xcode UI; **does not reproduce with `swift build` CLI** (different SwiftPM code path — see "Why amalgamation works" below)

## Reproduction

```swift
// Package.swift
// swift-tools-version: 5.9
import PackageDescription

let package = Package(
    name: "Repro",
    products: [.library(name: "Repro", targets: ["Repro"])],
    dependencies: [
        .package(url: "https://github.com/tree-sitter/tree-sitter", exact: "0.26.8"),
    ],
    targets: [
        .target(name: "Repro", dependencies: [
            .product(name: "TreeSitter", package: "tree-sitter"),
        ])
    ]
)
```

```swift
// Sources/Repro/Repro.swift
import TreeSitter
public func foo() -> UnsafePointer<TSLanguage>? { nil }
```

Build inside an Xcode project (or via `xcodebuild -resolvePackageDependencies` + `xcodebuild build`) on Xcode 26.4. `import TreeSitter` succeeds, every reference to `TSLanguage` (or any other api.h type) fails with `Cannot find type 'TSLanguage' in scope`.

## Evidence the C declarations *are* in the module

Running the same compile invocation Swift uses, but as bare Clang:

```sh
$ cat > /tmp/test.c <<EOF
#include "TreeSitter.h"   // or "tree_sitter/api.h"
TSLanguage *g_lang;
EOF
$ clang -fmodules \
    -fmodule-map-file=$(SourcePackages)/checkouts/tree-sitter/lib/include/module.modulemap \
    -I $(SourcePackages)/checkouts/tree-sitter/lib/include \
    -c /tmp/test.c -o /tmp/test.o
# succeeds, produces .o
```

Asking Swift's frontend to dump the ClangImporter lookup table:

```sh
$ swift -Xfrontend -dump-clang-lookup-tables -Xcc -fmodule-map-file=... test.swift
...
<<TreeSitter lookup table>>
  TSLanguage:
    TU: TSLanguage
  TSLanguageMetadata:
    TU: TSLanguageMetadata, TSLanguageMetadata
    TSLanguageMetadata: major_version
    ...
```

`TSLanguage` appears in the lookup table (TU = top-level translation unit). Yet Swift typecheck still emits `Cannot find type 'TSLanguage' in scope` — `import TreeSitter` resolves the module name but exposes nothing.

## Why 0.20.x works on Xcode 26 but 0.25+ does not

The relevant `Package.swift` diff:

| | 0.20.9 | 0.25/0.26 |
|---|---|---|
| `sources` | `["src/lib.c"]` (single amalgamation file that `#include`s every other .c) | `["src"]` (multi-file) |
| `publicHeadersPath` | default (= `"include"`) | `"include"` (explicit) |
| `cSettings.headerSearchPath` | none | `"src"` |

Both end up generating the same modulemap (`module TreeSitter { umbrella "lib/include" export * }`). But Xcode 26's build system handles single-source-file C targets along a different path that produces a Swift-visible module, while the multi-file path produces a module Swift sees as empty.

I cannot point at a specific SwiftPM commit — this looks like an interaction between (a) the new explicit-modules pipeline (`-fno-implicit-module-maps`, `-fno-implicit-modules`) Xcode 26 enables, (b) the umbrella-directory modulemap layout, and (c) the multi-file target's dependency-graph computation.

## Workarounds I tried (mostly didn't work)

These all failed in surprising ways and may help diagnose the underlying SwiftPM bug:

1. **Hand-written `lib/include/module.modulemap`** with explicit `umbrella header "TreeSitter.h"` — SwiftPM detects it (deletes its auto-generated copy) but **never injects `-fmodule-map-file=` into the Swift compile invocation** for that target. Other SPM dependencies in the same project all get `-fmodule-map-file=...GeneratedModuleMaps-iphoneos/X.modulemap`; only `TreeSitter` is missing from that list.
2. **Flat `lib/include/api.h`** (move out of `tree_sitter/` subdir) — SwiftPM does generate a modulemap this time but `-fno-modulemap-allow-subdirectory-search` (passed by Xcode 26 by default) breaks something downstream; same `Cannot find type` error.
3. **`OTHER_SWIFT_FLAGS = -Xcc -fmodule-map-file=$(BUILD_DIR)/../../SourcePackages/.../module.modulemap`** in the consumer project — flag does reach the swift-frontend invocation, but the module is still seen as empty.

The only thing that worked: revert `Package.swift` to 0.20-style amalgamation (`sources: ["src/lib.c"]`, drop `publicHeadersPath` and `headerSearchPath`). Patch: https://github.com/patr1ckzhu/tree-sitter/commit/9b61f80798f4e1d4e66803c8605c1a571468870b

## Proposed direction (maintainer's call)

I'm happy to send a PR but I don't want to guess at the ergonomics tradeoff. Some options:

1. **Revert to amalgamation in Package.swift only.** Same physical code, smallest diff, restores Xcode 26 compatibility today. Costs the granular per-file incremental build of the multi-file layout.
2. **Add a hand-written `lib/include/module.modulemap`.** Cleanest if Apple fixes the SwiftPM regression in a future Xcode — degrades gracefully. Risk: as documented above, didn't work for me on Xcode 26.4 (SwiftPM ignored it). Worth re-testing in case I missed something.
3. **Add a flat umbrella header (`lib/include/TreeSitter.h` that `#include`s `tree_sitter/api.h`)** alongside the subdir. Requires removing/excluding `tree_sitter/` from publicHeadersPath, which has knock-on effects.
4. **Wait for Apple.** Not great, since 0.25 has been broken on the latest Xcode for several months.

Happy to draft option 1 (or another) once you indicate which way you'd like to go.

## Downstream impact

This blocks Swift apps that consume `tree-sitter` directly via SPM from upgrading past 0.20.9 on Xcode 26, even when they want 0.25+'s parse-speed improvements and cancellation API. The amalgamation workaround is fine as a private fork pin, but a proper fix upstream would let downstream consumers track regular releases again.

Thanks for tree-sitter — much appreciated.

patr1ckzhu:
Fixes #5523.

Reverts the `TreeSitter` SPM target to the 0.20-era amalgamation build style (`sources: ["src/lib.c"]`) so Swift consumers can import the module on Xcode 26.

## What changed

`Package.swift`, 3 lines:

- `sources: ["src"]` → `sources: ["src/lib.c"]`
- Drop `"src/lib.c"` from `exclude` (it's compiled now)
- Drop the explicit `publicHeadersPath: "include"` (same as the default)

Everything else is untouched. `lib/src/lib.c` already exists and `#include`s every other `.c` file in `lib/src/`, so the compiled library is byte-for-byte the same as before from the linker's perspective.

## Why

On Xcode 26 (Swift 6.3), the multi-file layout (`sources: ["src"]` + `publicHeadersPath: "include"`) produces a generated modulemap (`umbrella "include"`) that `clang @import TreeSitter` can resolve, but `swift import TreeSitter` sees as an empty module — every reference to `TSLanguage` / `TSInputEncoding` / etc fails with `Cannot find type ... in scope`. Full investigation and dumped evidence in #5523.

The 0.20-era single-source-file layout goes through a different SwiftPM code path and works on Xcode 26. Since `lib/src/lib.c` is already maintained as a working amalgamation include, reusing it is the smallest possible fix that restores Swift SPM support without touching any non-core build surface.

## Non-impact

- **C / Rust / WASM builds**: untouched. These do not read `Package.swift`.
- **`swift build` CLI**: still works (both layouts work there; only `xcodebuild` / Xcode UI on Xcode 26 is broken).
- **Older Xcode (16/17)**: continues to work. 0.20.x used this same layout for years.
- **Grammars / bindings repos**: no ABI change, no header change, no include-path change visible to consumers.

## Tested

- `xcodebuild -scheme <consumer>` on Xcode 26.4 (Build 17E192), Swift 6.3 — build succeeds, consumer can reference `TSLanguage` etc.
- `swift build` CLI — unchanged, still green.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,10 +19,8 @@ let package = Package(
                         "src/unicode/README.md",
                         "src/unicode/LICENSE",
                         "src/wasm/stdlib-symbols.txt",
-                        "src/lib.c",
                 ],
-                sources: ["src"],
-                publicHeadersPath: "include",
+                sources: ["src/lib.c"],
                 cSettings: [
                         .headerSearchPath("src"),
                         .define("_POSIX_C_SOURCE", to: "200112L"),


clason:
@mattmassicotte

Also, would it be feasible to remove this file here and maintain it in the `swift-tree-sitter` bindings? How useful is this here without the bindings?

mattmassicotte:
@clason I'm honestly not sure how many clients there are of the C library directly. However, removing this file would break them. The Swift bindings moved to using this specifically to be compatible with a client that uses both it and the C bindings directly. But I've considered changing that because I cannot imagine that's common.

I don't know - pretty tough call.

clason:
Yeah. I'm asking honestly here since I have no idea about the language ecosystem. (This is one of the many challenges of maintaining this project: people add something for their favorite language and run away giggling -- and we're left months later trying to understand why it was added in the first place...)
So I'm hoping to take this chance to understand the rationale for having this file here instead of in the bindings.

But it's a single file that's already here so no harm in fixing it -- if it is indeed a fix, which I can't tell, which is why I'm asking for a sharper pair of eyes ;)

clason:
Original change in https://github.com/tree-sitter/tree-sitter/pull/4068, which in hindsight looks like LLM slop (the PR description doesn't really touch that change, which seems shoehorned into an unrelated fix).

mattmassicotte:
I did notice some breakage a little while ago. It would take me some time to determine if this change makes sense or not though. The runtime and parsers are very complex and maintaining compatibility over the years has been non-trivial.

I will look at this in a little bit.

mattmassicotte:
Apologies for taking so long here, but I've now had a look. Here's what I have found.

I don't think the Package.swift file is actually broken. It is true that there are some definitions from the headers that are not visible to Swift. But figuring out why would require some work, and I can at least confirm that this change does not fix #5523. Possibly it did at one point, and I'm sorry if my delay here caused further issues. But I have to say that I do not understand the change.

SwiftTreeSitter is also broken on main, but that's because it depends on the `ts_language_version` symbol which looks looks like it might have been replaced by a new mechanism. However, the module import seems just fine.

In short, I do not understand what it would take to improve the definition visibility via this Package.swift, but I don't think this is it. I also think it would be more worthwhile understanding why you'd need the ability to do that at all, given that I think there is API for everything anyways?

ObserverOfTime:
Language bindings are supposed to be high-level so consumers should not be using the low-level C API directly in other languages anyway. I think it makes sense to remove `Package.swift` in v0.27.

Besides the bundled Rust & Wasm bindings, the only other bindings with a build file here are the Zig ones. This is  because Zig can also be used to build the library itself and is used by downstream consumers such as Neovim.

clason:
> Language bindings are supposed to be high-level so consumers should not be using the low-level C API directly in other languages anyway. I think it makes sense to remove Package.swift in v0.27.

Let's go with that then (PR welcome).

PR #5529, opened 2026-04-17
fix(loader): allow filenames with dots

vmx:
Currently all dots/periods in a filename are considered an extension. This way you can match filenames like `foo.blade.php` with `blade.php` and not only `php`.

Though there are cases where filenames have dots and you still want to match on a suffix of the extension (or simply the last one).

This commit introduces best effort matching. It starts with the all dots being considered an extension and then gradually tries to match with less and less extensions.

Example: A filename `foo.bar.baz.blade.php` would try to find a match of these extensions in this order:

 - `bar.baz.blade.php`
 - `baz.blade.php`
 - `blade.php`
 - `php`

It stops on the first match, so if a grammar for `blade.php` is registered, it stops there and won't try `php`.

### DIFF
--- a/crates/cli/src/tests/detect_language.rs
+++ b/crates/cli/src/tests/detect_language.rs
@@ -128,6 +128,52 @@ fn detect_language_by_double_barrel_file_extension() {
 }
 
 #[test]
+fn detect_language_with_dots_in_filename() {
+    let blade_dir = tree_sitter_dir(
+        r#"{
+  "grammars": [
+    {
+      "name": "blade",
+      "path": ".",
+      "scope": "source.blade",
+      "file-types": [
+        "blade.php"
+      ]
+    },
+    {
+      "name": "php",
+      "path": ".",
+      "scope": "source.php",
+      "file-types": [
+        "php"
+      ]
+    }
+  ],
+  "metadata": {
+    "version": "0.0.1"
+  }
+}
+"#,
+        "blade",
+    );
+
+    let mut loader = Loader::with_parser_lib_path(scratch_dir().to_path_buf());
+    let config = loader
+        .find_language_configurations_at_path(blade_dir.path(), false)
+        .unwrap();
+
+    // this is just to validate that we can read the tree-sitter.json correctly
+    assert_eq!(config[0].scope.as_ref().unwrap(), "source.blade");
+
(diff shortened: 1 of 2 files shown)


vmx:
Does anyone have an instant idea why it might fail on Windows? If not, then I'll add some debug information and will try to get it running on some GitHub CI (as I don't have a Windows machine). That might take a while.

crates/loader/src/loader.rs
```diff
@@ -958,8 +958,10 @@ impl Loader {
                     path = PathBuf::from(path.file_stem()?.to_os_string());
                 }
                 extensions.reverse();
-                self.language_configuration_ids_by_file_type
-                    .get(&extensions.join("."))
+                // Try all extension suffixes from coarser to finer and stop at the first match.
```
WillLillis: This comment could be clarified a bit, something like 

```rust
// Try longest extension suffixs first (e.g. "foo.bar.baz"->"bar.baz"->"baz"), stopping at the first match
```

crates/cli/src/tests/detect_language.rs
```diff
@@ -127,6 +127,52 @@ fn detect_language_by_double_barrel_file_extension() {
...
+        "php"
+      ]
+    }
+  ],
+  "metadata": {
+    "version": "0.0.1"
+  }
+}
+"#,
+        "blade",
```
WillLillis: Tests are failing because the grammar name in this test matches another. Rust tests all run in the same process, so even after one of the "blade" tests finish the blade language's dll is still loaded. Windows doesn't allow you to overwrite dlls on disk while they're mapped, so we get the failure. The fix here is to just name the test language something different, e.g. `blade_dots`.

WillLillis (changes requested):
Changes look good overall, just a fix to get tests passing and the comment reword.

vmx:
Thanks @WillLillis for the help! Your wording of the comment is way clearer.

crates/cli/src/tests/detect_language.rs
WillLillis: @vmx You also need  to update the second argument to `tree_sitter_dir`

vmx: Thanks for the hint. I was confused that the tests pass on other platforms.

WillLillis (approved):

PR #5532, opened 2026-04-19
update parser template to zig 0.16

dchinmay2:


dchinmay2:
i asked an llm to write a script to test this, seems to work:
```bash
#!/usr/bin/env bash
set -e

ZIG=${ZIG:-~/Downloads/zig-x86_64-linux-0.16.0/zig}
TMPDIR=$(mktemp -d)
trap 'rm -rf "$TMPDIR"' EXIT

cargo build -p tree-sitter-cli

cat > "$TMPDIR/tree-sitter.json" <<'EOF'
{
  "grammars": [{ "name": "foo", "scope": "source.foo" }],
  "metadata": { "version": "0.0.1" },
  "bindings": { "zig": true }
}
EOF

./target/debug/tree-sitter init -p "$TMPDIR"
(cd "$TMPDIR" && "$OLDPWD/target/debug/tree-sitter" generate grammar.js)

cd "$TMPDIR"
$ZIG build
$ZIG build test -Dtest
```

crates/cli/src/templates/build.zig.zon
```diff
@@ -4,8 +4,8 @@
     .version = "PARSER_VERSION",
     .dependencies = .{
         .tree_sitter = .{
-            .url = "git+https://github.com/tree-sitter/zig-tree-sitter#b4b72c903e69998fc88e27e154a5e3cc9166551b",
-            .hash = "tree_sitter-0.25.0-8heIf51vAQConvVIgvm-9mVIbqh7yabZYqPXfOpS3YoG",
+            .url = "git+https://github.com/tree-sitter/zig-tree-sitter#b22ced81dc72bf6f0238196830e0a8e597e1ad2a",
+            .hash = "tree_sitter-0.25.0-8heIf3CaAQD8ekvpvZRI_dlVdrbEX6GcsJ4WJZ758ceV",
```
clason: Where is that pointing to, exactly?

dchinmay2: just after my zig-tree-sitter PR was merged

clason: Then the version number is wrong

dchinmay2: what is it supposed to be? 0.25.0 is the latest tag

clason: of zig-treesitter? probably should bump that then first @WillLillis

WillLillis: > of zig-treesitter? probably should bump that then first @WillLillis

I bumped zig-tree-sitter to 0.26 and pushed a new tag. We should be good to use that here now.

dchinmay2: updated

WillLillis (approved):

PR #5533, opened 2026-04-20
chore: enable install via `cargo binstall`

### LINKED ISSUE #4351: cargo binstall support
### Problem

It would be cool if `tree-sitter-cli` could be installed through [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall). Since the release binaries are already published, this could be [supported](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md) by either changing their names or adding a `package.metadata.binstall` section to the cargo manifest.

### Expected behavior

_No response_

wetneb:
Closes #4351.

To test this:
- check out this PR
- temporarily change the `version` field in the root `Cargo.toml` to a released version (such as `0.26.8`)
- run `cargo binstall tree-sitter-cli --manifest-path crates/cli/Cargo.toml --verbose --version 0.26.8` and see that it offers to install tree-sitter from a downloaded archive instead of compiling it from scratch

Of course the actual command will become simpler once this change is released: it should become simply `cargo binstall tree-sitter-cli`.

See [the docs of `cargo binstall`](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md) for details about this configuration format. I first tried making a more compact configuration via statements such as `[package.metadata.binstall.overrides.'cfg(target_arch = "x86_64")']` to translate architectures independently from platforms but this format doesn't seem to get picked up on my side. The more verbose format I ended up with is arguably easier to maintain as it matches the CI configuration more explicitly.

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -12,21 +12,15 @@ The Tree-sitter CLI allows you to develop, test, and use Tree-sitter grammars fr
 
 ### Installation
 
-You can install the `tree-sitter-cli` with `cargo`:
+You can install the `tree-sitter-cli` with `cargo-binstall`:
 
-```sh
-cargo install --locked tree-sitter-cli
-```
-
-or if you have `cargo-binstall`, you can avoid compiling the CLI from source:
 ```sh
 cargo binstall tree-sitter-cli
 ```
 
-or with `npm`:
-
+or you can build it from source:
 ```sh
-npm install tree-sitter-cli
+cargo install --locked tree-sitter-cli
 ```
 
 You can also download a pre-built binary for your platform from [the releases page].
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -18,6 +18,11 @@ You can install the `tree-sitter-cli` with `cargo`:
 cargo install --locked tree-sitter-cli
 ```
 
+or if you have `cargo-binstall`, you can avoid compiling the CLI from source:
+```sh
+cargo binstall tree-sitter-cli
+```
+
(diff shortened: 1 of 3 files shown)


clason:
Can you update the relevant documentation (README etc.) to suggest this over `npm` and `cargo install`?

crates/cli/README.md
```diff
@@ -18,6 +18,11 @@ You can install the `tree-sitter-cli` with `cargo`:
 cargo install --locked tree-sitter-cli
 ```
 
+or if you have `cargo-binstall`, you can avoid compiling the CLI from source:
+```sh
+cargo binstall tree-sitter-cli
+```
+
```
clason, line 20: I would promote this as the _first_ method; the other two should be exceptions. ("To build and install from source, call ...".)

In fact, I would suggest removing `npm` completely here; it's just too messy.

(Instead, we could give distro-specific CLI invocations for the major platforms, but that's out of scope for this PR.)

wetneb: I don't know how widespread `cargo-binstall` is, so my worry was that the first command we give isn't actionable for a lot of people… Where is this README generally read? Probably on https://crates.io/crates/tree-sitter-cli, no? There, there are pretty good chances that readers have `cargo` installed, I would say, so `cargo install --locked tree-sitter-cli` should be a reasonably safe bet. But maybe having `cargo-binstall` is a way of spreading the `binstall` gospel and saving some CPU cycles across the user base, okay :)

clason (approved):

PR #5537, opened 2026-04-21
Add highlight generation for LaTeX

Neptunium931:


clason:
No thank you; this is a legacy module that we do not want to actively develop further.

PR #5541, opened 2026-04-23
fix(generate): pass default optimization level in `generate_parser_for_grammar`

WillLillis:
Passing `empty` here causes some grammars (i.e. tree-sitter-cpp) to fail to generate.

### DIFF
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -316,7 +316,7 @@ pub fn generate_parser_for_grammar(
         LANGUAGE_VERSION,
         semantic_version,
         None,
-        OptLevel::empty(),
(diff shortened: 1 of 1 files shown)


clason:
Interesting -- why didn't we see this in CI then?

WillLillis:
> Interesting -- why didn't we see this in CI then?

This is only used to generate the small corpus grammars in `test/fixtures/test_grammars/`, so nothing big enough to cause issues there. The actual fixtures (cpp, php, etc.) are generated via the xtask which builds the CLI and invokes it as a separate process. Generating this way passes `default` optimization  flags unless specified otherwise (via `--disable-optimization`).

clason (approved):

PR #5546, opened 2026-04-25
fix(web): map Node-only specifiers to false via browser field

### LINKED ISSUE #5545: web-tree-sitter: bundlers fail to resolve `fs/promises` and `module` when targeting the browser
### Problem

Bundling `web-tree-sitter@0.26.x` for the browser fails because two `await import()` calls in environment-guarded branches are statically discoverable by bundlers:

- `await import("fs/promises")` at `web-tree-sitter.js:1487` (inside `Language.load` — guarded by `globalThis.process?.versions.node`)
- `await import("module")` at `web-tree-sitter.js:1527` (inside the emscripten module init — guarded by `ENVIRONMENT_IS_NODE`)

Even though those branches never execute in a browser, the bundler tries to resolve the specifiers at build time and errors out.

### Minimal repro

`package.json`:

```json
{
  "name": "wts-bundle-repro",
  "private": true,
  "type": "module",
  "scripts": {
    "repro": "esbuild --bundle --platform=browser --format=esm --outfile=out.js entry.js"
  },
  "dependencies": {
    "esbuild": "^0.27.0",
    "web-tree-sitter": "^0.26.6"
  }
}
```

`entry.js`:

```js
import 'web-tree-sitter';
```

Run `npm install && npm run repro`:

```
✘ [ERROR] Could not resolve "fs/promises"

    node_modules/web-tree-sitter/web-tree-sitter.js:1487:31:
      1487 │       const fs2 = await import("fs/promises");
           ╵                                ~~~~~~~~~~~~~

  The package "fs/promises" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "--platform=node" to do that, which will remove this error.

✘ [ERROR] Could not resolve "module"

    node_modules/web-tree-sitter/web-tree-sitter.js:1527:43:
      1527 │     const { createRequire } = await import("module");
           ╵                                            ~~~~~~~~

  The package "module" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "--platform=node" to do that, which will remove this error.

2 errors
```

Tested with `web-tree-sitter@0.26.8` and `esbuild@0.27.7`. Other bundlers (vite, webpack) likely fail similarly — see historical reports #466, #681, #1121.

### Workaround

Mark them external:

```
esbuild --bundle --platform=browser --format=esm \
  --external:fs/promises --external:module --outfile=out.js entry.js
```

This succeeds (bundle ~150KB) and is safe at runtime: the imports stay as bare specifiers in the output but are guarded by Node-only checks that never fire in a browser.

### Suggested fixes (any of)

1. Add a `"browser"` condition in `package.json#exports` pointing at a Node-stripped entry.
2. Add a top-level `"browser"` field mapping `fs/promises` and `module` to `false`.
3. Move the Node-only loader into a separate file imported through a specifier that bundlers can't statically resolve (e.g. `await import(/* @vite-ignore */ nodeOnlyPath)` with `nodeOnlyPath` a runtime-built string).

### Environment

- `web-tree-sitter` 0.26.8
- `esbuild` 0.27.7
- macOS 24.6.0 (also reproduces on Linux)

reteps:
## Summary

- Adds a top-level `"browser"` field in `lib/binding_web/package.json` mapping `fs/promises` and `module` to `false`.
- Fixes #5545

The `"browser"` field tells bundlers to substitute empty stubs for these specifiers when targeting the browser.

### Other Node references audited

- `require("fs")`, `require("path")`, `require("url")` call a local `var require = createRequire(import.meta.url)` that's hoisted inside `Module2()`. esbuild and webpack both correctly resolve these as local-variable calls and leave them untouched.

## Test plan

Claude reproduced the issue's exact esbuild repro against the published `web-tree-sitter@0.26.8` artifact, then re-ran with the patched `package.json`, and it succeeded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -43,6 +43,10 @@
     },
     "./debug/web-tree-sitter.wasm": "./debug/web-tree-sitter.wasm"
   },
+  "browser": {
+    "fs/promises": false,
+    "module": false
+  },
   "types": "web-tree-sitter.d.ts",
   "keywords": [
     "incremental",
--- a/lib/binding_web/package.json
+++ b/lib/binding_web/package.json
@@ -43,6 +43,11 @@
     },
     "./debug/web-tree-sitter.wasm": "./debug/web-tree-sitter.wasm"
   },
+  "browser": {
+    "fs": false,
+    "fs/promises": false,
+    "module": false
+  },
   "types": "web-tree-sitter.d.ts",
   "keywords": [
     "incremental",


WillLillis:
Added fs to the browser mappings because for the CommonJS entry point.

WillLillis (approved):

PR #5548, opened 2026-04-25
fix(query): overflow in capture pool ids + some perf

### LINKED ISSUE #5011: Segfault in `tree-sitter highlight` on long Zig file
### Problem

It seems like tree-sitter highlighting crashes when trying to highlight [a large Zig file](https://github.com/ziglang/zig/commit/b206b0626ae8e4fd1c50097d2f376890a48d4518#diff-dc16e976d4e6b1e81b3fa6610adff75ddd07b9017f6fa8c40a2e341a7c12a9f1) both on GitHub and in my editor. I don't think this is specific to Zig based on the below results but the highlighting does seem a bit slow and I couldn't reproduce this with other languages, so it could be that the Zig highlighting queries are the issue.

I tried to reduce this file as much as I could in the reproduction to about 1200 lines but it seems like getting rid of any more gets the segfault to go away. I tried debugging in LLDB and I'm not too familiar with the code but it seems like `right` in `ts_query_cursor__compare_captures` is `NULL` perhaps because of some boundary case with `NONE`/`UINT32_MAX`?

```sh
leave node. depth:19, type:,
enter node. depth:19, type:}, field:(null), row:1214 state_count:28, finished_state_count:65529
  start state. pattern:68, step:404
  discard state. pattern:41, step:154
  capture node. type:}, pattern:68, capture_id:41, capture_count:2
  advance state. pattern:68, step:405
  keep state. pattern: 3, start_depth: 1, step_index: 9, capture_count: 4294967295
  keep state. pattern: 4, start_depth: 1, step_index: 12, capture_count: 4294967295
  keep state. pattern: 7, start_depth: 1, step_index: 23, capture_count: 4294967295
Process 1444832 stopped
* thread #2, name = 'tests::highligh', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x10)
    frame #0: 0x0000555555fc79d2 tree_sitter_cli-b38d4d9217d5d765`ts_query_cursor__compare_captures(self=0x00007ffff0028620, left_state=0x00007ffff05a1f50, right_state=0x00007ffff05a1f60, left_contains_right=0x00007ffff7bfc680, right_contains_left=0x00007ffff7bfc688) at query.c:3379:41
   3376	     if (j < right_captures->size) {
   3377	       TSQueryCapture *left = array_get(left_captures, i);
   3378	       TSQueryCapture *right = array_get(right_captures, j);
-> 3379	       if (left->node.id == right->node.id && left->index == right->index) {
   3380	         i++;
   3381	         j++;
   3382	       } else {
(lldb) bt
* thread #2, name = 'tests::highligh', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x10)
  * frame #0: 0x0000555555fc79d2 tree_sitter_cli-b38d4d9217d5d765`ts_query_cursor__compare_captures(self=0x00007ffff0028620, left_state=0x00007ffff05a1f50, right_state=0x00007ffff05a1f60, left_contains_right=0x00007ffff7bfc680, right_contains_left=0x00007ffff7bfc688) at query.c:3379:41
    frame #1: 0x0000555555fca696 tree_sitter_cli-b38d4d9217d5d765`ts_query_cursor__advance(self=0x00007ffff0028620, stop_on_definite_step=true) at query.c:4111:13
    frame #2: 0x0000555555fcb42b tree_sitter_cli-b38d4d9217d5d765`ts_query_cursor_next_capture(self=0x00007ffff0028620, match=0x00007ffff7bfc910, capture_index=0x00007ffff7bfc90c) at query.c:4371:8
    frame #3: 0x0000555555f3c3cc tree_sitter_cli-b38d4d9217d5d765`_$LT$tree_sitter_highlight.._QueryCaptures$LT$T$C$I$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$::next::h904a06a53b1ce6d7(self=0x00007ffff002ae30) at highlight.rs:238:20
    frame #4: 0x0000555555f318c7 tree_sitter_cli-b38d4d9217d5d765`core::iter::adapters::peekable::Peekable$LT$I$GT$::peek::_$u7b$$u7b$closure$u7d$$u7d$::ha0f166f2c787d5d2 at peekable.rs:218:48
    frame #5: 0x0000555555f3a2f4 tree_sitter_cli-b38d4d9217d5d765`core::option::Option$LT$T$GT$::get_or_insert_with::he592c958601f9fcb(self=0x00007ffff002aec0, f={closure_env#0}<tree_sitter_highlight::_QueryCaptures<&[u8], &[u8]>> @ 0x00007ffff7bfcad0) at option.rs:1753:26
    frame #6: 0x0000555555f31831 tree_sitter_cli-b38d4d9217d5d765`core::iter::adapters::peekable::Peekable$LT$I$GT$::peek::he1dcd5e3f3b74d3d(self=0x00007ffff002ae30) at peekable.rs:218:21
    frame #7: 0x00005555556d383d tree_sitter_cli-b38d4d9217d5d765`_$LT$tree_sitter_highlight..HighlightIter$LT$F$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$::next::hfe2c2ded4953c2ec(self=0x00007ffff7bfdf60) at highlight.rs:1018:80
    frame #8: 0x000055555580087e tree_sitter_cli-b38d4d9217d5d765`tree_sitter_cli::tests::highlight_test::to_token_vector::hcdbce8225080df3d(src=(data_ptr = "\nconst Select = struct {\n    const TempSpec = struct {\n        fn create(spec: TempSpec, s: *const Select) InnerError!struct { Temp, bool } {\n            const cg = s.cg;\n            const pt = cg.pt;\n            return switch (spec.kind) {\n                .imm => |imm| .{ try cg.tempInit(spec.type, .{ .immediate = @bitCast(@as(i64, imm)) }), true },\n                .cc => |cc| .{ try cg.tempInit(spec.type, .{ .eflags = cc }), true },\n                .ptest_cc => |cc_spec| {\n                    const ccs: [2]Condition = .{ .z, .c };\n                    const cc = ccs[cc_spec.all ^ @intFromBool(cc_spec.ref.valueOf(s).register_mask.info.inverted)];\n                    return .{ try cg.tempInit(spec.type, .{ .eflags = switch (cc_spec.not) {\n                        false => cc,\n                        true => cc.negate(),\n                    } }), true };\n                },\n                .ref => |ref| .{ ref.tempOf(s), false },\n                .reg => |reg| .{ try cg.tempInit(spec.type, .{ .register = reg }), t"..., length = 90135), language_config=0x0000555556750328) at highlight_test.rs:1994:18
    frame #9: 0x000055555580290e tree_sitter_cli-b38d4d9217d5d765`tree_sitter_cli::tests::highlight_test::test_highlighting_zig::h4fd1dbc397843169 at highlight_test.rs:1340:5
    frame #10: 0x0000555555732657 tree_sitter_cli-b38d4d9217d5d765`tree_sitter_cli::tests::highlight_test::test_highlighting_zig::_$u7b$$u7b$closure$u7d$$u7d$::hf4b1d888115b0ec5((null)=0x00007ffff7bfe436) at highlight_test.rs:119:27
    frame #11: 0x00005555556a1ec6 tree_sitter_cli-b38d4d9217d5d765`core::ops::function::FnOnce::call_once::hf6cdc9c987015f33((null)={closure_env#0} @ 0x00007ffff7bfe436, (null)=<unavailable>) at function.rs:250:5
    frame #12: 0x0000555555967d0b tree_sitter_cli-b38d4d9217d5d765`test::__rust_begin_short_backtrace::h38ffe622e17a164d + 11
    frame #13: 0x0000555555984c22 tree_sitter_cli-b38d4d9217d5d765`test::types::RunnableTest::run::h84bd14d098b23d4e + 50
    frame #14: 0x000055555596727d tree_sitter_cli-b38d4d9217d5d765`test::run_test::_$u7b$$u7b$closure$u7d$$u7d$::he62792e50b2840a6 + 1917
    frame #15: 0x000055555595deb4 tree_sitter_cli-b38d4d9217d5d765`std::sys::backtrace::__rust_begin_short_backtrace::h0badc65700f2a8a4 + 148
    frame #16: 0x000055555598ab28 tree_sitter_cli-b38d4d9217d5d765`core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h8302f868ed379776 + 168
    frame #17: 0x00005555562f46a7 tree_sitter_cli-b38d4d9217d5d765`std::sys::pal::unix::thread::Thread::new::thread_start::h7afb2fc4e6f2f483 + 55
    frame #18: 0x00007ffff7c9a97a libc.so.6`start_thread + 682
    frame #19: 0x00007ffff7d22d2c libc.so.6`__clone3 + 44
(lldb) p right_captures->size 
(const uint32_t) 4294967295
(lldb) p right_captures->contents
(TSQueryCapture *const) NULL
```

I noticed that either changing `max_capture_list_count` back to `NONE` or essentially just changing all `uint16`s to `uint32`s seems to fix the crash but I'm not sure if these are the correct fixes.

### Steps to reproduce

```sh
git clone --depth 1 https://github.com/tree-sitter-grammars/tree-sitter-zig
cd tree-sitter-zig/
tree-sitter generate
wget https://gist.githubusercontent.com/rmehri01/73e9b6422603e18bbf47743925481088/raw/9d20dd5c39ac2430e461668ed022b240e3d7f6de/foo.zig
tree-sitter highlight foo.zig 
```

Output:

```sh
fish: Job 1, 'tree-sitter highlight foo.zig' terminated by signal SIGSEGV (Address boundary error)
```

### Expected behavior

It should not crash.

tree-sitter 0.26.0 (605e58006351ff2d133e69dfc617268ab9887a8d)

### Operating system/version

Linux 6.17.4 x86_64

WillLillis:
#5011 describes a crash when highlighting [a large zig file](https://github.com/ziglang/zig/blob/b206b0626ae8e4fd1c50097d2f376890a48d4518/src/codegen/x86_64/CodeGen.zig). Amaan and I looked at this issue briefly when it was first opened and suspected it was the highlight crate's unsafe usage of various query structures (grep for `std::mem::transmute` and you'll see it). This is necessary to allow for the highlight crate's peeking iterator behavior, which the updated streaming iterator provided by tree-sitter's rust bindings does not allow. The worry there was that fixing this bug would require a significant or complete rewrite of the highlight crate to use the new streaming iterators.

However, after digging into  this issue more the problem is actually an integer overflow in `query.c`. The capture pool uses 16 bit ids, but it really should allow for 32.  That change alone is enough to prevent the crash. If you re-run the repro with this fix though, you may notice that it takes forever. This is because `ts_query_cursor_next_capture` does a linear scan over _all_ finished query states to find the next capture. This is the perfect use case  for a min heap instead!

I've done some spot checking and this doesn't appear to introduce any regressions with small files, and it  greatly helps when working with captures within a large file. For example, on my machine, highlighting the repro from #5011:

### On master:

Segfaults after ~22 seconds. 

### With the first commit from this PR:

Finishes successfully in ~377 seconds. Preventing the segfault exposes this performance cliff.

### With both commits from this PR:

Finishes successfully in ~5 seconds.

## Downstream Considerations

I'm not super sure how/where to test the affects of this boost on downstream. I assume most consumers have limits in place to  prevent such a large number of finished states from accumulating, but if anyone has any ideas (CC @clason @ribru17 ) I'd love to get some more numbers here.

- Closes #5011

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -334,6 +334,7 @@ struct TSQueryCursor {
 static const TSQueryError PARENT_DONE = -1;
 static const uint16_t PATTERN_DONE_MARKER = UINT16_MAX;
 static const uint16_t NONE = UINT16_MAX;
+static const uint32_t CAPTURE_LIST_NONE = UINT32_MAX;
 static const TSSymbol WILDCARD_SYMBOL = 0;
 static const unsigned OP_COUNT_PER_QUERY_CALLBACK_CHECK = 100;
 
@@ -429,7 +430,7 @@ static CaptureListPool capture_list_pool_new(void) {
 }
 
 static void capture_list_pool_reset(CaptureListPool *self) {
-  for (uint16_t i = 0; i < (uint16_t)self->list.size; i++) {
+  for (uint32_t i = 0; i < self->list.size; i++) {
     // This invalid size means that the list is not in use.
     array_get(&self->list, i)->size = UINT32_MAX;
   }
@@ -437,18 +438,18 @@ static void capture_list_pool_reset(CaptureListPool *self) {
 }
 
 static void capture_list_pool_delete(CaptureListPool *self) {
-  for (uint16_t i = 0; i < (uint16_t)self->list.size; i++) {
+  for (uint32_t i = 0; i < self->list.size; i++) {
     array_delete(array_get(&self->list, i));
   }
   array_delete(&self->list);
 }
 
-static const CaptureList *capture_list_pool_get(const CaptureListPool *self, uint16_t id) {
+static const CaptureList *capture_list_pool_get(const CaptureListPool *self, uint32_t id) {
   if (id >= self->list.size) return &self->empty_list;
   return array_get(&self->list, id);
 }
 
-static CaptureList *capture_list_pool_get_mut(CaptureListPool *self, uint16_t id) {
+static CaptureList *capture_list_pool_get_mut(CaptureListPool *self, uint32_t id) {
   ts_assert(id < self->list.size);
   return array_get(&self->list, id);
 }
@@ -459,10 +460,10 @@ static bool capture_list_pool_is_empty(const CaptureListPool *self) {
   return self->free_capture_list_count == 0 && self->list.size >= self->max_capture_list_count;
 }
 
-static uint16_t capture_list_pool_acquire(CaptureListPool *self) {
+static uint32_t capture_list_pool_acquire(CaptureListPool *self) {
   // First see if any already allocated capture list is currently unused.
   if (self->free_capture_list_count > 0) {
-    for (uint16_t i = 0; i < (uint16_t)self->list.size; i++) {
+    for (uint32_t i = 0; i < self->list.size; i++) {
       if (array_get(&self->list, i)->size == UINT32_MAX) {
         array_clear(array_get(&self->list, i));
         self->free_capture_list_count--;
@@ -475,7 +476,7 @@ static uint16_t capture_list_pool_acquire(CaptureListPool *self) {
   // doesn't put us over the requested maximum.
   uint32_t i = self->list.size;
   if (i >= self->max_capture_list_count) {
-    return NONE;
+    return CAPTURE_LIST_NONE;
   }
   CaptureList list;
   array_init(&list);
@@ -483,7 +484,7 @@ static uint16_t capture_list_pool_acquire(CaptureListPool *self) {
   return i;
 }
 
-static void capture_list_pool_release(CaptureListPool *self, uint16_t id) {
+static void capture_list_pool_release(CaptureListPool *self, uint32_t id) {
   if (id >= self->list.size) return;
   array_get(&self->list, id)->size = UINT32_MAX;
   self->free_capture_list_count++;
@@ -3551,7 +3552,7 @@ static void ts_query_cursor__add_state(
   );
   array_insert(&self->states, index, ((QueryState) {
     .id = UINT32_MAX,
-    .capture_list_id = NONE,
+    .capture_list_id = CAPTURE_LIST_NONE,
     .step_index = pattern->step_index,
     .pattern_index = pattern->pattern_index,
     .start_depth = start_depth,
@@ -3571,13 +3572,13 @@ static CaptureList *ts_query_cursor__prepare_to_capture(
   QueryState *state,
   unsigned state_index_to_preserve
 ) {
-  if (state->capture_list_id == NONE) {
+  if (state->capture_list_id == CAPTURE_LIST_NONE) {
     state->capture_list_id = capture_list_pool_acquire(&self->capture_list_pool);
 
     // If there are no capture lists left in the pool, then terminate whichever
     // state has captured the earliest node in the document, and steal its
     // capture list.
-    if (state->capture_list_id == NONE) {
+    if (state->capture_list_id == CAPTURE_LIST_NONE) {
       self->did_exceed_match_limit = true;
       uint32_t state_index, byte_offset, pattern_index;
       if (
@@ -3596,7 +3597,7 @@ static CaptureList *ts_query_cursor__prepare_to_capture(
         );
         QueryState *other_state = array_get(&self->states, state_index);
         state->capture_list_id = other_state->capture_list_id;
-        other_state->capture_list_id = NONE;
+        other_state->capture_list_id = CAPTURE_LIST_NONE;
         other_state->dead = true;
         CaptureList *list = capture_list_pool_get_mut(
           &self->capture_list_pool,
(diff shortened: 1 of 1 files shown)


amaanq (approved):

clason:
I think nvim-treesitter's `indents` function iterates over all captures in a file (since you may want to reindent the whole buffer); Nvim has a built-in (configurable) match limit though -- whose need this PR may obviate?

clason:
But even for functions that restrict the cursor to a section, that happens line-wise so things like minified JSON would still be affected (modulo match limit, which can be disabled).

clason:
Some history: https://github.com/tree-sitter/tree-sitter/pull/1127

PR #5556, opened 2026-04-27
Documentation update to index.md

Thorium:
I've been testing the 0.3.0 version of the F# tree-sitter and that works very well.

### AI Policy

- [x] No AI used.


clason:
That list tracks language bindings (i.e., to use the tree-sitter _library_ from F#) not parsers! Those can be found in the Wiki, where this parser is already listed.

(I know it's confusing -- bindings are called $language-tree-sitter, while parsers are called tree-sitter-$language.)

Thorium:
Oh, right, thanks. Then F# is not needed there, because the "C#" is actually an incorrect label, it's a .NET component that is already usable from F#.

PR #5557, opened 2026-04-27
perf(generate): deduplicate identical small parse table entries

Thorium:
## Summary
- deduplicate identical `ts_small_parse_table` rows during parser generation by reusing the first emitted entry
- build the deduplication key from `Symbol` identity so grammars with nonterminal entries in the small parse table are handled correctly

### AI Policy

Used GitHub Copilot Chat / OpenCode to inspect the larger unmerged CSR branch, isolate the generator-only small-parse-table deduplication change, and help prepare this standalone PR. The code change itself is a cherry-pick of the existing branch work with no additional functional edits.

### DIFF
--- a/crates/generate/src/render.rs
+++ b/crates/generate/src/render.rs
@@ -1426,7 +1426,8 @@ impl Generator {
                     symbols.sort_unstable();
                     canonical_data.push(symbols.len() as u16);
                     for symbol in symbols.iter() {
-                        canonical_data.push(self.symbol_order[symbol] as u16);
+                        canonical_data.push(symbol.kind as u16);
+                        canonical_data.push(symbol.index as u16);
                     }
                 }
 
--- a/crates/generate/src/render.rs
+++ b/crates/generate/src/render.rs
@@ -1371,7 +1371,7 @@ impl Generator {
                     .saturating_sub(self.large_state_count),
             );
             // Deduplication: map canonical data to its first table index
-            let mut seen_data: HashMap<Vec<u16>, usize> = HashMap::new();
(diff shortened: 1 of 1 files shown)


clason:
_Please_ test your changes locally before pushing. I have to manually approve CI after every push, which is simply not sustainable.

Also, do _not_ trust LLMs. The AI policy is there for a reason. It's fine for (local) exploration, but you **must** clean up the PR manually _before_ submitting. If you cannot do that, just open an issue explaining what changes you think should be made (and linking to the branch on your fork) and leave the PR to a maintainer.

(In this case, I had to make a number of changes when rebasing your PR on other improvements to the same code which need to be included here as well. You need to make sure that you do not regress on those -- again, don't trust LLMs to do that for you.)

WillLillis:
Which grammars did you test this change against? After spot checking a few grammars, this PR is either net zero or marginally worse:

tree-sitter-cpp:

image

tree-sitter-rust:

image

tree-sitter-vim:

image

nvim-treesitter install:

This PR: 

image


Master:

image

Thorium:
What this PR deduplicates:
- identical rows in the ts_small_parse_table
- if two small states serialize to the same grouped symbol/action data, they now share one table entry via ts_small_parse_table_map

The de-duplication aims to save memory and compiled wasm size (not nearly as much as CSR would have) while keeping the execution speed around the same.

clason:
@Thorium we know what the PR does (we can read); the question is "why?" and "does it _work_?" (and "is it worth it?") These questions should be answered in your PR description (and commit message) and backed up with hard data.

AI-generated PRs without human context are a significant maintainer burden (and violate our AI policy).

clason:
Superseded by #5578

PR #5558, opened 2026-04-27
fix(test): make lookahead iterator test order-independent

Thorium:
## Summary
- stop assuming a specific iteration order in `test_lookahead_iterator`
- sort the collected symbol names before comparison so the test only checks the documented set of valid symbols

### AI Policy

Used GitHub Copilot Chat / OpenCode to inspect the larger unmerged CSR branch, identify this test-only fix as a standalone change, and help prepare this isolated PR. The final patch here is limited to the lookahead iterator test.


clason:
Superseded by #5578

PR #5565, opened 2026-05-03
build(deps): bump wasmtime-c-api to v36.0.8

clason:
Another LTS security update; plus `cargo update`.

(I think we should keep sticking to LTS, even though incremental updates are easier for working through breaking changes. For reference, [LTS releases are yearly in August](https://docs.wasmtime.dev/stability-release.html).)


lib/Cargo.toml
```diff
@@ -47,11 +47,10 @@ streaming-iterator             = "0.1.9"
 tree-sitter-language.workspace = true
 
 [dependencies.wasmtime-c-api]
-default-features = false
+version = "36.0.8"
 features         = [ "cranelift", "gc-drc" ]
 optional         = true
 package          = "wasmtime-c-api-impl"
-version          = "36.0.7"
```
clason, line 54: formatting the TOML files like that broke a lot of tooling, so I'm manually reverting to the unformatted version wherever necessary.

WillLillis: What tooling did it break?

WillLillis (approved):

lib/Cargo.toml
clason: The bump xtask, and the backport action.

PR #5569, opened 2026-05-04
Validate Wasm language memory reads

ZX41R:
Validate offsets from `LanguageInWasmMemory` before copying language data out of Wasm memory.

The language descriptor is provided by the module being loaded. Before this change, offsets such as `parse_table`, `symbol_names`, `lex_modes`, and the alias/supertype tables were used directly as indexes into the store's memory buffer. A malformed module could point one of those fields outside the current linear memory and make the host process read through an invalid pointer while loading the language.

This adds a small checked-memory wrapper for Wasm language loading and routes descriptor reads, table copies, string reads, and the alias-map scan through it. Invalid descriptor addresses now fail loading with `TSWasmErrorKindInstantiate`.

Local checks:

- Wasm-enabled CMake build with ASan/UBSan
- normal CMake build with ASan/UBSan
- malformed local Wasm module with `parse_table = 0x7ffffff0` is rejected instead of crashing
- in-bounds control module still loads
- `git diff --check`

AI disclosure: I used assistant support during local investigation and repro drafting; I reviewed the code and verified the patch.

### DIFF
--- a/lib/src/wasm_store.c
+++ b/lib/src/wasm_store.c
@@ -365,18 +365,60 @@ typedef struct {
   wasm_functype_t *type;
 } FunctionDefinition;
 
-static void *copy(const void *data, size_t size) {
+typedef struct {
+  const uint8_t *data;
+  size_t size;
+} WasmMemory;
+
+static bool wasm_memory__contains(const WasmMemory *memory, int32_t address, size_t size) {
+  if (address < 0) return false;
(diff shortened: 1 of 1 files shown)


maxbrunsfeld (approved):

maxbrunsfeld:
Thank you! Great improvement.

PR #5488, opened 2026-04-02
feat(generate): compress parse tables with CSR format

Thorium:
## Summary

Reduce the size of generated `parser.c` files by applying **CSR (Compressed Sparse Row) compression** to the parse table, which is the dominant source of file size in large grammars.

### Source size reduction

| Grammar | Before | After | Reduction |
|---------|--------|-------|-----------|
| JSON | 28 KB | 22 KB | -24% |
| C | 3.9 MB | 1.5 MB | -61% |
| JavaScript | 2.8 MB | 1.2 MB | -59% |
| C# | 29 MB | 8.5 MB | -71% |

### Binary size reduction (MSVC /O2, same compiler for both)

| Grammar | ABI 15 DLL | ABI 16 DLL | Reduction |
|---------|-----------|-----------|-----------|
| JSON | 106 KB | 106 KB | 0% (min DLL overhead) |
| C | 735 KB | 627 KB | -15% |
| JavaScript | 505 KB | 468 KB | -7% |
| C# | 5.4 MB | 3.2 MB | -41% |

All existing tests pass for every grammar tested.

Addresses [#1799](https://github.com/tree-sitter/tree-sitter/issues/1799), [#464](https://github.com/tree-sitter/tree-sitter/issues/464), [#3496](https://github.com/tree-sitter/tree-sitter/issues/3496), [#1860](https://github.com/tree-sitter/tree-sitter/issues/1860), [#5337](https://github.com/tree-sitter/tree-sitter/issues/5337).

## Motivation

Large grammars like C# (29 MB) and F# (58 MB with a custom fork, 32 MB upstream) produce enormous `parser.c` files. Analysis of C# showed that the dense parse table (`ts_parse_table[4050][525]`) accounts for **64%** of the file, with the small parse table adding another **31%**. Yet only **24.3%** of the dense table entries are non-zero — the rest are zeros consuming space for nothing.

Tree-sitter's current approach splits states into "large" (dense 2D array) and "small" (sparse list), but doesn't apply any compression to the dense table.

### What was tried first (minor optimizations A1-A4)

Before implementing full compression, four smaller optimizations were explored:

- **A1 (cost-based threshold)**: Dynamically chose which states go to the dense vs sparse table. **Reverted** — caused regressions on small grammars with no benefit for large ones.
- **A2 (lex state dedup)**: Deduplicated identical lex state bodies in the switch statement. Minor benefit for JavaScript, zero effect on C#.
[...]

PR #5578, opened 2026-05-06
perf(generate): CSR revamped

WillLillis:
This PR serves as an alternative to #5488, with an improved CSR/hybrid parse table format heuristic, as well as tighter integration with the `generate` command. In short:

## The problem

The libraries produced by tree-sitter grammars' `parser.c` files can be quite large.

## The Solution

CSR format can provide a smaller compiled library size, but only for some parsers.

In long...

### The heuristic:

I developed a better metric to determine when to use CSR by checking against >300 grammars from nvim-treesitter. This metric is intentionally conservative and introduces no regressions.  In order for a grammar to use CSR format, it must fulfill 3 conditions:

1. `LARGE_STATE_COUNT * SYMBOL_COUNT > STATE_COUNT * 40`

This ensures the dense table is large enough relative to total state count that savings outweigh the small-state grouping penalty.

image

2. dense table density < 45%:

This ensures CSR actually saves space. Above ~50% density, CSR's per-entry column indices cost more than the zeros they eliminate. 45% adds margin below the theoretical crossover.

image

3. `LARGE_STATE_COUNT * SYMBOL_COUNT > 50,000`

Avoids applying a format change to tiny grammars where fixed overhead dominates.

image

Overall:

image

### A4 size wins

The original A4 commit from #5488 yielded some size wins to parsers sticking with the hybrid format, but slowed down parser generation in some cases. I optimized this commit slightly. Functionally it's the same, but with either net zero or a slight reduction in generation time: 

image
image
image

### Runtime performance of compressed parsers

I ran `cargo xtask benchmark`, comparing the performance of the fixture grammars at hand for this branch compared to master. Overall, we see roughly neutral performance (within noise) to substantial gains (especially in error recovery). Most importantly, no major regressions!

``` 

  ┌───────────────────┬────────┬───────────────┬───────────┬──────────┬──────────────┬──────────┬─────────┐
  │      Grammar      │ Format │ Normal master │ Normal PR │ Δ normal │ Error master │ Error PR │ Δ error │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ bash              │ hybrid │ 9,455         │ 9,924     │ +5.0%    │ 8,178        │ 8,411    │ +2.8%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ c                 │ CSR    │ 10,369        │ 10,477    │ +1.0%    │ 2,583        │ 3,075    │ +19.0%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ cpp               │ CSR    │ 6,869         │ 6,881     │ +0.2%    │ 2,400        │ 2,659    │ +10.8%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ embedded-template │ hybrid │ —             │ —         │ —        │ 140,139      │ 140,589  │ +0.3%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ go                │ hybrid │ 18,499        │ 18,925    │ +2.3%    │ 3,329        │ 3,220    │ -3.3%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ html              │ hybrid │ 549           │ 541       │ -1.5%    │ 36,905       │ 36,678   │ -0.6%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ java              │ CSR    │ —             │ —         │ —        │ 4,416        │ 4,946    │ +12.0%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ javascript        │ CSR    │ 18,682        │ 19,275    │ +3.2%    │ 3,762        │ 4,364    │ +16.0%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ jsdoc             │ hybrid │ —             │ —         │ —        │ 34,987       │ 35,462   │ +1.4%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ json              │ hybrid │ —             │ —         │ —        │ 28,719       │ 28,330   │ -1.4%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ php               │ CSR    │ —             │ —         │ —        │ 135,658      │ 135,035  │ -0.5%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ php_only          │ CSR    │ —             │ —         │ —        │ 2,726        │ 3,513    │ +28.9%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ python            │ hybrid │ 14,073        │ 13,906    │ -1.2%    │ 2,622        │ 2,532    │ -3.4%   │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ ruby              │ CSR    │ —             │ —         │ —        │ 3,374        │ 3,842    │ +13.9%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ rust              │ CSR    │ 15,882        │ 15,617    │ -1.7%    │ 2,365        │ 3,088    │ +30.6%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ tsx               │ CSR    │ —             │ —         │ —        │ 4,617        │ 5,165    │ +11.9%  │
  ├───────────────────┼────────┼───────────────┼───────────┼──────────┼──────────────┼──────────┼─────────┤
  │ typescript        │ CSR    │ 12,388        │ 12,515    │ +1.0%    │ 25,928       │ 26,210   │ +1.1%   │
  └───────────────────┴────────┴───────────────┴───────────┴──────────┴──────────────┴──────────┴─────────┘
```

### CLI Integration

The above heuristic works well, but it is empirically driven and there still may be cases where it chooses wrong. I also added a new flag to the generate command, `--table-fmt`, which allows the user to control the parse table output format. Not supplying an argument (or giving `auto`) leaves it up to the heuristic, `hybrid` uses the "legacy" (existing) table format, and `csr` uses CSR  compression. 

### Bonus

Remove the unused `section_index` field from `LookaheadIterator`. (https://github.com/tree-sitter/tree-sitter/pull/5488#discussion_r3040180214)

### AI Policy

I used claude code to help derive the metric used to decide when to use CSR vs. hybrid parse table format, analyze the runtime impact on parser performance, and generate the pretty pictures for this PR.

- Supersedes #5488 
- Supersedes #5558 
- Supersedes #5557


clason:
I think I would recommend bumping ABI version unconditionally; this makes things easier to reason about (no need to worry about combination with `--abi` flag) and opens the door for further ABI changes in 0.27 (e.g., expanding metadata). In fact, once we add any other ABI update in the future, we can no longer use ABI to check for table format, so we best use an explicit discriminator in the API (metadata field?) from the start.

**EDIT:** Tests agree with me ;)

lib/src/language.h
```diff
@@ -12,6 +12,7 @@ extern "C" {
 
 #define LANGUAGE_VERSION_WITH_RESERVED_WORDS 15
 #define LANGUAGE_VERSION_WITH_PRIMARY_STATES 14
+#define LANGUAGE_VERSION_WITH_COMPRESSED_TABLES 16
```
clason: Nit:
```suggestion
#define LANGUAGE_VERSION_WITH_COMPRESSED_TABLES 16
#define LANGUAGE_VERSION_WITH_RESERVED_WORDS 15
#define LANGUAGE_VERSION_WITH_PRIMARY_STATES 14
```

clason:
Excellent work, thank you for tackling this!

clason (approved):

clason:
Out of interest: is it possible to compare memory usage, at least for the "monster" parsers?

maxbrunsfeld:
Is it possible to eliminate the heuristic entirely, and compute the exact size of the CSR representation of a parse state compared to the size of a dense parse state, and use CSR for only the states where it results in fewer bytes?

In other words, partition the parse table into *three* distinct representations (small, CSR, and dense) using the best representation available for each state

WillLillis:
> Is it possible to eliminate the heuristic entirely, and compute the exact size of the CSR representation of a parse state compared to the size of a dense parse state, and use CSR for only the states where it results in fewer bytes?
> 
> In other words, partition the parse table into _three_ distinct representations (small, CSR, and dense) using the best representation available for each state

Still working on this, should have some numbers either later tonight or this weekend.

WillLillis:
@maxbrunsfeld  I tried to implement your suggestion and have it up here: https://github.com/WillLillis/tree-sitter/tree/plan-b-impl. It yields ~10% additional space savings, but with some tradeoffs.

Doing some more focused parser runtime perf testing reveals some issues with that approach as well as with this PR's current implementation. I was able to mitigate this slowdown partially by introducing a bias parameter, but this would have to be exposed to users given how unpredictable the results are. Unless you have some improvements/optimization ideas for the `plan-b-impl` branch, I think this is a no-go. I'm also a bit worried about the impact this current PR has on parser performance. Some grammars benefit, while some regress significantly. 

image
 
Another data point, generation time increases by ~5-10% with `plan-b-impl`. Some of this can probably be optimized away, but definitely not all. Not as important as size or runtime perf, but still worth noting.

WillLillis:
Unless we can find significant runtime stability improvements, I think this may be a no-go. Marking as a draft for now.

maxbrunsfeld:
@WillLillis I agree, we shouldn't do this if it regresses parsing performance for some grammars by > 10%. It is surprising  to me that this is happening. I would have thought that the cost of the binary search during table lookup would be pretty minor.

I was not able to reproduce the runtime performance regression that you observed. When I run `cargo xtask benchmark` on master and on your `plan-b-impl` branch, the benchmarks show a pretty significant improvement on `plan-b-impl`. The improvement is largest for the syntax error cases, which are, I think, some of the most valuable cases to improve, since they represent worst-case performance.

So overall, I would say that this branch still seems very promising.

maxbrunsfeld:
One change I would like to see is combining `parse_table_columns` and `parse_table_values` into one single array. I think that for a given state, the symbols and actions are always access together, so unlike in linear algebra code, there isn't a reason to split them into separate contiguous arrays. This should make it possible to format the C code such that there is *some* source-level visibility into the state structure. Something like:

```c
static const uint16_t ts_compressed_parse_states[] = {
  [0] =
    anon_sym_LPAREN, 24,
    anon_sym_LBRACK, 5,
  [4] =
    sym_integer_literal, 200,
    sym_char_literal, 300,
  // ...
};
```

With this change, we won't see as much reduction in C source file size, but that has never really been a priority for Tree-sitter. Thus far, we've maintained an invariant that it's at least possible to walk through the state machine by reading the C code. I'd like to stick with that philosophy for now.

maxbrunsfeld:
@WillLillis I pushed a commit on top of your `plan-b` branch to this new branch: https://github.com/tree-sitter/tree-sitter/compare/compressed-parse-table

WillLillis:
> @WillLillis I pushed a commit on top of your `plan-b` branch to this new branch: [`compressed-parse-table` (compare)](https://github.com/tree-sitter/tree-sitter/compare/compressed-parse-table)

Thanks! I'll be able to work with this! I'm looking into why I saw so many performance regresions earlier. Most of them seem to actually be within noise, while rust and kotlin appear to be significantly slower.

maxbrunsfeld:
Yeah, I did see some actual performance regressions in some of the Rust benchmarks, oddly. But I see much larger performance gains during error recovery, which I think outweighs the regression in happy path performance. It would be nice to avoid any performance regression. Maybe some kind of bias for dense states vs compressed, similar to what you did w/ compressed vs small?

WillLillis:
> Yeah, I did see some actual performance regressions in some of the Rust benchmarks, oddly. But I see much larger performance gains during error recovery, which I think outweighs the regression in happy path performance. It would be nice to avoid any performance regression. Maybe some kind of bias for dense states vs compressed, similar to what you did w/ compressed vs small?

Yep, I'm working on tuning another bias parameter now. The error recovery gains are very encouraging though! The performance impact is variable enough across grammars that I think it makes sense to expose these to users (with reasonable defaults) so people can tune if needed? These could be persisted in `tree-sitter.json` as optional fields (we already load this file to find the parser version) during the generate process.

WillLillis:
Quick update. I ran a sweep for 12 grammars with the new bias parameter.  There seems to be a pretty consistent size vs. perf tradeoff here.

image

I'm going to make some additional updates to the picker to see if some improvements can be made, as well as to perform a full sweep over more bias combinations to get a more complete picture. I'll post more updates here and push up some updates once I have a better set of defaults.

xxxsenzodyne (approved):

WillLillis:
I spent a significant amount of time trying to find a combination of tuning parameters that satisfied all (most) grammars but wasn't able to. The solution I did settle on gave a default regression for runtime perf of around 1-5%, with marginal at best size savings. This new format _does_ unlock configurations that can yield both binary size and runtime performance wins (or, at the least a win for one and more or less neutral for the other), but this required running a lengthy parameter space search on a quiet machine on a per-grammar basis.

I believe this requires too high a level of buy in from grammar authors, and opens the door to surprising cross platform behavior when tuning is only done on one platform. I'd like to revisit this when other ABI-breaking changes are in the pipeline, but for now I don't think the realistic benefits we'll get are worth such a large amount of breakage.

PR #5598, opened 2026-05-17
docs(rust): expand set_allocator safety documentation

### LINKED ISSUE #5078: Better safety documentation for `tree_sitter::set_allocator` in Rust SDK
### Problem

From the current docs:

```
Sets the memory allocation functions that the core library should use.
                                                                       
# Safety
                                                                       
This function uses FFI and mutates a static global.
```

It is unclear to me what other requirements my allocator handles must fulfill to be memory safe:
- is it actually file to leave `new_free = None`, if I set the other function handles, won't that cause libc free to invoked on allocations made by me?
- what alignment for pointers is expected, etc.
- will calloc default to malloc + memset if I don't define it

I'm not very familiar with the depths of tree sitter so would really appreciate some more guidance here. Considering the default allocation pattern is libc I imagine setting a custom arena allocator or such is like very important for performance? Or does tree sitter internally already store the tree in some kind of arena and Node structs are just handles to that?

Thanks!

### Expected behavior

_No response_

wyf027:
## Summary

Fixes #5078.

Expands the Rust `set_allocator` docs to match the C API behavior in `alloc.c` / `api.h`:

- `None` per parameter restores that function's libc default (including `free`)
- `calloc` is not emulated from `malloc`
- alignment / realloc-free safety expectations
- when it is safe to call after prior allocations
- brief note on arenas vs `Node` handles

## Test plan

- [x] `cargo check -p tree-sitter`
- [x] Documentation only

### DIFF
--- a/lib/binding_rust/lib.rs
+++ b/lib/binding_rust/lib.rs
@@ -3925,9 +3925,35 @@ static mut FREE_FN: unsafe extern "C" fn(ptr: *mut c_void) = free;
 
 /// Sets the memory allocation functions that the core library should use.
 ///
+/// By default, Tree-sitter uses the standard libc allocation functions and aborts
+/// the process when an allocation fails. Pass custom function pointers to override
+/// them at runtime.
+///
+/// Pass `None` for any parameter to keep Tree-sitter's default implementation of
+/// that function. For example, `new_free: None` does **not** route frees through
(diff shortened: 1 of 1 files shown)


wyf027:
Superseded by #5599 (combined CLI/template PR for easier review). Closing this one.

wyf027:
Superseded by #5599

Issue #4688, opened 2025-08-12
A quantifier in a child node implicitly anchors the child node


TheLeoP:
### Problem

Using either a `+` or `*` quantifier in a child node will implicitly anchor it's position within the capture.

### Steps to reproduce

In the playground https://tree-sitter.github.io/tree-sitter/7-playground.html

1. Select `C#`
2. Use the following source code

```cs
public class Foo
{
    //foo
    public static void Roo()
    {
        Console.WriteLine("Hello, World!");
        var bar = "Hello, World!";
        Console.WriteLine(bar);
    }

    //foo
    public static void Woo()
    {
        Console.WriteLine("Hello, World!");
        var bar = "Hello, World!";
        Console.WriteLine(bar);
    }
}
```

3. Enable queries and use the following query

```scheme
(class_declaration
 (declaration_list
 (comment) @output.comment
  .
  [
    (method_declaration)
    (constructor_declaration)
  ] @output.method))
````

Both of the methods will be highlighted. If the query is changed to have a `*` quantifier after the `comment` capture like 

```scheme
(class_declaration
 (declaration_list
 (comment)* @output.comment
  .
  [
    (method_declaration)
    (constructor_declaration)
  ] @output.method))
````

Only the first method will be highlighted (just like if there were a `.` before the `comment` capture)

### Expected behavior

Using a quantifier on a child node should not implicitly anchor the location of the node inside of the parent

### Tree-sitter version (tree-sitter --version)

0.25.8

### Operating system/version

arch

WillLillis:
Fixed by #5603

PR #5603, opened 2026-05-17
Fix issues with queries matching a parent node with anchored children (siblings)

lucasmdjl:
### AI Policy

I used claude.ai web interface for help with debugging and reasoning through debug logs, evaluating my approach and suggesting other directions to explore, and double-checking my code.

### Description
This PR fixes three issues for queries that match parent node with anchored children (siblings).

The first issue was that the check for fallible steps was only considering nodes with children, which resulted in no state split being made for anchored siblings (node1) . (node2). This prevented the query from matching beyond the first occurrence. The fix was to also consider the case where the next step is at the same depth and must be matched immediately after.

The second issue was that with this type of queries where in addition the first anchored sibling had a quantifier, the deduplication logic for states was being overly aggressive, causing the state split on the first capture (with the parent) to be dropped in favor of the loopback state. This resulted in not being able to match beyond the first occurrence. The fix was to prevent the deduplication logic to drop a state that is not seeking for an immediate match for one that is seeking for an immediate match, because the one that is not seeking for an immediate match still could match later nodes.

The third issue was that with this type of queries where in addition the first anchored sibling had a quantifier and wasn't captured, the check for fallible steps was skipping splitting the state, because the next node was a passthrough node (and not an `is_immediate` one like in the first issue). This again prevented the query from matching beyond the first occurrence. The fix was to skip the next steps in the check for fallibility while they are passthrough steps.

The third fix would close issue #4688 .

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -1594,6 +1594,38 @@ fn test_query_matches_with_leading_zero_or_more_repeated_leaf_nodes() {
 }
 
 #[test]
+fn test_matches_with_anchor_sibling_inside_parent() {
+    allocations::record(|| {
+        let language = get_language("rust");
+
+        let query = Query::new(
+            &language,
+            "
+            (source_file
+                (line_comment)
+                .
+                (function_item
+                    name: (identifier) @name)
+            )",
+        )
+        .unwrap();
+
+        assert_query_matches(
+            &language,
+            &query,
+            "
+            // A
+            fn a() {}
+
+            // B
+            fn b() {}
+            ",
+            &[(0, vec![("name", "a")]), (0, vec![("name", "b")])],
+        );
+    });
+}
+
+#[test]
 fn test_query_matches_with_trailing_optional_nodes() {
     allocations::record(|| {
         let language = get_language("javascript");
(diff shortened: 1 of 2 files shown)


clason:
@ribru17

clason:
@lucasmdjl Mixing together multiple fixes like that in a single PR (without cleanly separated commits) makes review much harder (hence the long silence) and prevents "easy" fixes from being merged (and backported) quickly. If you can clean up the commits here in this regard, that would be helpful.

lucasmdjl:
@clason I've cleaned up formatting throughout the commits so that there's no final commit fixing all formatting. Now each fix is structured as a commit adding a (failing) test for the expected behavior and a commit with the fix making the test pass. If it's more helpful to have each fix as a single commit with both the test and the fix, I can do that too, though I think this removes the signal that the test was failing before the fix.

clason:
Yes, please squash test and fix; commits should be atomic (this is important for bisecting).

lucasmdjl:
Alright, done. Now each fix is in its own commit.

clason:
For the sake of completeness (but this can come after review, before the final merge):

1. The commit message should be `fix(query)` or `fix(lib)` (we should standardize that! @WillLillis )
2. The commit message should include the relevant details from the PR description (for when GH goes bust...) A "Problem: ... Solution: ..." format has proved very helpful.

WillLillis (approved):

WillLillis:
Fix looks great, thank you! Happy to merge after the git cleanup that Clason mentioned is done.

lucasmdjl:
Great! I've cleaned up the commit messages adding "fix(query)" to the title and the "Problem:... Solution:..." structure Clarson mentioned.

PR #5605, opened 2026-05-18
fix(wasm): load supertype tables for ABI 15 grammars

maxbrunsfeld:
### Problem

Query construction can crash when grammars that use supertypes, are compiled with ABI version 15, and are loaded via Wasm.

### Background

Supertypes were introduced in ABI version 15, the same ABI version that introduced reserved words. Methods that access a language's supertypes check that the grammar's ABI version is greater than **or equal to** 15.

But when loading a language via Wasm, we were neglecting to load supertype data unless the language's abi version was strictly greater than 15.

Switching the check to `>=` matches the convention used everywhere else for `LANGUAGE_VERSION_WITH_RESERVED_WORDS` and fixes the crash.

This was observed as a SIGSEGV in `ts_language_subtypes` reaching users via Zed loading tree-sitter grammars compiled to Wasm.

### AI Use

I found this bug by showing Claude Opus 4.7 the crash backtrace and directing it to inspect `ts_wasm_store_load_language`, since that Wasm loading logic has been problematic in the past.

### DIFF
--- a/lib/src/wasm_store.c
+++ b/lib/src/wasm_store.c
@@ -1325,7 +1325,7 @@ const TSLanguage *ts_wasm_store_load_language(
   }
 
   bool has_supertypes =
-    wasm_language.abi_version > LANGUAGE_VERSION_WITH_RESERVED_WORDS &&
+    wasm_language.abi_version >= LANGUAGE_VERSION_WITH_RESERVED_WORDS &&
     wasm_language.supertype_count > 0;
 
   int32_t addresses[] = {


amaanq (approved):

maxbrunsfeld:
@amaanq @WillLillis @clason Are you guys ok with doing a 0.26 patch release just for this?

clason:
We've been conscientious in backporting other fixes, so there's more than this in the pipeline.

Let me check if there are recently merged PRs missing a (merged) backport, then I can trigger the release if @WillLillis agrees.

WillLillis:
@maxbrunsfeld 

The NO_COLOR fix that was bundled in with #5594 could be done separately on the release branch (without the painting changes) but isn't super important.

#5569 might be nice to have but that's completely up to you, I'm not sure how that would impact Zed.

Beyond those two I think our release branch is in solid shape, I don't see any issue with another patch release 🙂

maxbrunsfeld:
Yes, #5569 Seems like it'd be good to get in too.

clason:
Luckily it backported cleanly, so I'll merge it once the CI is green, re-trigger the release checks, and publish once they pass, too.

clason:
While we wait, you could take a look at @WillLillis' last comment on the CSR PR ;)

PR #5614, opened 2026-05-21
generate: preserve word-token transitions when pruning keyword completions

bitbottrap:
I believe this behavior is a bug and am proposing a correction. It was encountered during a grammar development that triggered the issue. AI was used to help identify and correct the issue. The workaround without this
PR is to implement a scanner. It is fairly complex, both (what I believe to be) the error and the fix. Understanding might be best if you start with the test grammar and corpus and compare with and without the PR.

Problem:
When a grammar declares 'word: $ => $.identifier' tree-sitter promotes certain keyword tokens to a separate 'ts_lex_keywords' table so the runtime can use word-boundary checks to distinguish them from identifiers. However, some keyword candidates fail the promotion filter. For example, a token whose lexical conflict status differs from the plain identifier and they "leak" back into the main 'ts_lex' table. Before this fix, those leaked keywords caused a silent correctness bug. Any identifier that started with the leaked keyword's text (e.g. in our test grammar and corpus 'neqvbase' starts with 'neqv') was incorrectly split at the keyword boundary instead of being recognized as a single identifier.

During lexer DFA construction (build_lex_table.rs, 'populate_state'), when a token reaches a completion state with precedence N, any NFA transitions that represent continuing to read more characters are suppressed via 'prefer_transition(transition.prec, N)' if the transition's precedence is lower. This is correct and intentional for tokens in 'ts_lex_keywords', because the runtime will re-examine the completed text against the word token using word-boundary rules. But for leaked keywords (those that 'identify_keywords' excluded and remain only in 'ts_lex' with no runtime word-boundary safety net) the suppression is wrong. Once the DFA drops the identifier continuation branch, there is no second chance to recover the full identifier, and the parser silently accepts the shorter keyword match.

Fix:
'populate_state' now checks a 'leaked_keywords: Option<&TokenSet>' parameter (supplied only for the main lex table 'None' for 'ts_lex_keywords') before discarding a transition. If the completed token is in 'leaked_keywords' and the suppressed transition leads to an NFA state belonging to the word token, the transition is rescued and kept. Leaked keywords never go through the runtime word/boundary path, so there is no double counting. Promoted keywords and grammars without a word token are unaffected. A new 'word_token_leaked_keyword' fixture grammar and nine corpus tests cover the exact failure mode and the boundary cases introduced by the fix.

### DIFF
--- a/crates/generate/src/build_tables.rs
+++ b/crates/generate/src/build_tables.rs
@@ -321,18 +321,6 @@ fn populate_external_lex_states(parse_table: &mut ParseTable, syntax_grammar: &S
     }
 }
 
-/// Returns `(promoted_keywords, leaked_keywords)`.
-///
-/// `promoted_keywords` are the keyword candidates that pass every filter and
-/// are extracted by the runtime word-extraction path via `ts_lex_keywords`.
-///
-/// `leaked_keywords` are candidates that were dropped here (either because
-/// another candidate matches the same string, or because substituting the
-/// word token would introduce new lexical conflicts) and therefore remain in
-/// the main `ts_lex` table. They are the keywords whose word-boundary
-/// behavior cannot rely on the runtime word-extraction path and which must
-/// be considered by `LexTableBuilder` when deciding whether to keep a
-/// pruned word-token continuation transition.
 fn identify_keywords(
     lexical_grammar: &LexicalGrammar,
     parse_table: &ParseTable,
--- a/crates/generate/src/build_tables.rs
+++ b/crates/generate/src/build_tables.rs
@@ -58,7 +58,7 @@ pub fn build_tables(
     )?;
     let token_conflict_map = TokenConflictMap::new(lexical_grammar, following_tokens);
     let coincident_token_index = CoincidentTokenIndex::new(&parse_table, lexical_grammar);
-    let keywords = identify_keywords(
+    let (keywords, leaked_keywords) = identify_keywords(
         lexical_grammar,
         &parse_table,
         syntax_grammar.word_token,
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
Could you add a minimal repro for the bug in `test/fixtures/test_grammars`?

bitbottrap:
Looks like I'm going to need the corpus.

maxbrunsfeld:
I'm a little worried that it's not generally expected for the word token to take precedence over all other tokens, regardless of precedence.

bitbottrap:
> I'm a little worried that it's not generally expected for the word token to take precedence over all other tokens, regardless of precedence.

Yes, this is overly broad.

bitbottrap:
Apologies. What was submitted earlier was not ready for review. I've re-written the original post to better reflect the actual problem.

Basically, the fix should now be safe because it only preserves the path back to the identifier (word) token. An identifier can only extend past a keyword if more word characters follow - so when a separator comes next, the identifier can't grow any further and the keyword wins exactly as before. So here I think the concern about identifiers always being selected is being addressed. I've added tests to explicitly show precedence allowing non word tokens being chosen. Leaked keywords end up behaving like whole-word matches.

The grammar is also larger than I'd like for a test grammar. However, there are a number of potential cases that this change impacts and I wanted to cover enough to show it wasn't breaking.

From your source I can tell the project isn't a big fan of source comments. But I wanted to leave them in to help explain what's going on. I can strip them out before a merge.

Prana111025 (approved):
Ibsvvssg

Prana111025 (approved):
Cheek

bitbottrap:
I've stripped comments to more closely match the rarely commented code. If you want to review the comments they're in an earlier commit.

clason:
Needs a (manual) rebase and review from @maxbrunsfeld

WillLillis:
This has been stale for awhile, and I'm fairly certain what it aims to do is not correct in general. This can be discussed in an issue before re-implementing if desired.

PR #5620, opened 2026-05-23
fix(cli): place generate artifacts next to grammar file rather than cwd

WillLillis:
### The Problem

This is some weird behavior from tree-sitter-cli that I don't *think* is intentional. If you run `tree-sitter generate path/to/grammar.js`, the generated `src` directory will get placed in the current working directory of the tree-sitter invocation, rather than next to `grammar.js`. This is a bit unintuitive and leaves artifacts in unexpected places.

### The Solution

Instead, if we provide an _explicit_ path to a grammar file, ensure the generated files are placed in the proper place.

### DIFF
--- a/crates/generate/src/generate.rs
+++ b/crates/generate/src/generate.rs
@@ -251,6 +251,18 @@ where
             repo_path = path_buf;
             repo_path.join("grammar.js")
         } else {
+            // Given an _explicit_ path to an input file, derive the repo root from the
+            // conventional locations:
+            //     - grammar.js sits inside <root>/
+            //     - grammar.json sits inside <root>/src/
+            let repo_root = match path_buf.extension() {
+                Some(e) if e == "js" => path_buf.parent(),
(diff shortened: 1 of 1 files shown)


clason (approved):

maxbrunsfeld (approved):

asm123na (approved):

asm123na (commented):
https://console.cloud.google.com/cloud-setup?organizationId=1042636749976

WillLillis:
This is complicated a bit by passing `src/grammar.json`. Need some time to test out with some more grammars later tonight before merging this.

EDIT: Should be good now. If we get `path/to/grammar.js`, we use `path/to` as the repo root. If we get a path to grammar.json, we assume it's under `src/grammar.json` and place it in the _granparent_ directory, which should still be the proper directory.

clason:
> This is complicated a bit by passing src/grammar.json. 

Yeah, that part is important -- nvim-treesitter relies on that fundamentally ;)

PR #5629, opened 2026-05-26
Allow building with the system wasmtime

WhyNotHugo:
Allow compiling and linking the system wasmtime using the 'wasm-system' feature.

Have the existing 'wasm' feature and this new 'wasm-system' feature depend on a new '_wasm' feature, which gates all the portions of code which need to be enabled for wasm support.

The alternative to this was renaming the existing feature to 'wasm-bundled', and then making it and 'wasm-system' depend on a common 'wasm'. This is a bit cleaner, but would be a breaking change for all downstream consumers.

### AI Policy

- [x] If AI tools were used: I have disclosed the tool and extent of usage below:
  - _figuring out how the cargo-compiled wasmtime was being statically linked (i.e.: finding `DEP_WASMTIME_C_API_INCLUDE` and `fn use_wasmtime`)._

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1995,6 +1995,7 @@ dependencies = [
  "serde_json",
  "streaming-iterator",
  "tree-sitter-language",
+ "wasmtime",
  "wasmtime-c-api-impl",
 ]
 
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -30,9 +30,10 @@ harness = false
 name    = "benchmark"
 
 [features]
-default = [ "qjs-rt" ]
-qjs-rt  = [ "tree-sitter-generate/qjs-rt" ]
-wasm    = [ "tree-sitter/wasm", "tree-sitter-loader/wasm" ]
+default     = [ "qjs-rt" ]
+qjs-rt      = [ "tree-sitter-generate/qjs-rt" ]
+wasm        = [ "tree-sitter/wasm", "tree-sitter-loader/wasm" ]
+wasm-system = [ "tree-sitter/wasm-system", "tree-sitter-loader/wasm-system" ]
 
 [dependencies]
 ansi_colours.workspace          = true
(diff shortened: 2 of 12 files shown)


ObserverOfTime (commented):
1. What is "system wasmtime" and what's the benefit? All I see is a different dependency.
2. Why not keep the feature name as is and change the `cfg` conditions to use `any()`?

WhyNotHugo:
By "system wastime" I mean the host system's existing wasmtime. Typically: the one shipped by the current distribution.

Aside from a different dependency, `lib/binding_rust/build.rs` will link the system's wasmtime, rather than compiling and statically linking another build.

> Why not keep the feature name as is and change the cfg conditions to use any()?

Honestly, I initially thought that was going to result in a noisier diff and kind of ugly results… but clearly the current approach isn't better than that. I've updated the PR with this approach.

clason:
To be honest, the overall design strategy of tree-sitter is as a full-stack monolith. Distros will fight that (distro gonna distro) and work around it, but that does not necessarily coerce us. 

In particular, wasmtime makes frequent (monthly!) breaking releases, while we intentionally track LTS.

So I do not see any value (and quite the opposite) for us to take on the added maintenance burden.

WhyNotHugo:
> In particular, wasmtime makes frequent (monthly!) breaking releases, while we intentionally track LTS.

This is good to keep in mind. Packaging `wasmtime-lts` sounds reasonable, so I'll try and make that happen.

---

Is this realistically that much of a burden? You can ping me if it interferes with anything.

Most (all?) downstream distributions will want to use the system version, keeping builds simpler. For example, wasmtime needs patches to build on certain architectures, so relying on the distro package is much simpler than trying to apply those patches during tree-sitter's build process.

maxbrunsfeld:
I agree with @clason that Wasmtime moves quickly enough that this is going to be a bit maintenance-intensive. We really need to aggressively limit how much responsibility accretes to this repository. I get the motivation behind the feature, and it's a valid use-case, but I don't think we should take this on.

WhyNotHugo:
Ack. Do you have a different approach in mind, or should all distributions just carry patches downstream?

PR #5639, opened 2026-05-30
fix(lib): update doc comment for `ts_parser_parse`

### LINKED ISSUE #5638: Update documentation for `ts_parser_parse`
### AI Policy

### Problem

The documentation reads:

```
/**
 * Use the parser to parse some source code and create a syntax tree.
 *
 * If you are parsing this document for the first time, pass `NULL` for the
 * `old_tree` parameter. Otherwise, if you have already parsed an earlier
 * version of this document and the document has since been edited, pass the
 * previous syntax tree so that the unchanged parts of it can be reused.
 * This will save time and memory. For this to work correctly, you must have
 * already edited the old syntax tree using the [`ts_tree_edit`] function in a
 * way that exactly matches the source code changes.
 *
 * The [`TSInput`] parameter lets you specify how to read the text. It has the
 * following three fields:
 * 1. [`read`]: A function to retrieve a chunk of text at a given byte offset
 *    and (row, column) position. The function should return a pointer to the
 *    text and write its length to the [`bytes_read`] pointer. The parser does
 *    not take ownership of this buffer; it just borrows it until it has
 *    finished reading it. The function should write a zero value to the
 *    [`bytes_read`] pointer to indicate the end of the document.
 * 2. [`payload`]: An arbitrary pointer that will be passed to each invocation
 *    of the [`read`] function.
 * 3. [`encoding`]: An indication of how the text is encoded. Either
 *    `TSInputEncodingUTF8` or `TSInputEncodingUTF16`.
 *
 * This function returns a syntax tree on success, and `NULL` on failure. There
 * are four possible reasons for failure:
 * 1. The parser does not have a language assigned. Check for this using the
      [`ts_parser_language`] function.
 * 2. Parsing was cancelled due to the progress callback returning true. This callback
 *    is passed in [`ts_parser_parse_with_options`] inside the [`TSParseOptions`] struct.
 *
 * [`read`]: TSInput::read
 * [`payload`]: TSInput::payload
 * [`encoding`]: TSInput::encoding
 * [`bytes_read`]: TSInput::read
 */
```
This makes no reference to the `decode` member and refers to the removed `TSInputEncodingUTF16` value.

### Steps to reproduce

Check the documentation on HEAD.

### Expected behavior

The documentation mentions the up-to-date options for the `encoding` member and describes the use of the `decode` member.

0.26.9 and HEAD

### Operating system/version

N/A

WillLillis:
- Closes #5638


clason (approved):

wenkokke:
Thanks! It's not specified in the comment, but I assume decode should be NULL if either UTF8 or UTF16LE/BE is passed and non-NULL if Custom is passed?

WillLillis:
> Thanks! It's not specified in the comment, but I assume decode should be NULL if either UTF8 or UTF16LE/BE is passed and non-NULL if Custom is passed?

Pretty much. `decode` is only accessed if Custom is passed, otherwise it's never read.

PR #5641, opened 2026-05-30
fix(lib): Consider subtree lookahead bytes when determining whether the parser can reuse a node.

### LINKED ISSUE #5636: Incorrect incremental parsing result for markdown-inline
### AI Policy

### Problem

On a high-level, the problem is that the incremental parse tree differs from a from-scratch parse tree. Suppose you have this file that only contains a newline: `"\n"`, and create a markdown and markdown-inline parser, and type
```
`c `
```
one by one, and at each key stroke re-parse markdown parser, and set the range of markdown-inline according to markdown parser.

The expected parse tree from markdown-inline should be `(inline (code_span))` because of the backticks. But in reality we only get `(inline)`.

AI disclosure: We encountered this problem in Emacs, and I had Claude write the reproduce script in C.

### Steps to reproduce

```sh
git clone https://github.com/casouri/tree-sitter-repro-markdown-inline-codespan
cd tree-sitter-repro-markdown-inline-codespan
./run.sh
```

### Expected behavior

The script will do what I described above: type each character, re-parse markdown, set ranges for markdown-inline, re-parse markdown-inline, and print out the parse tree of markdown-inline, it looks like this:

```
=== initial \n ===
  src="\n"
  inline=(no inline node)

=== after insert '`' ===
  src="`\n"
  inline=(inline)

=== after insert 'c' ===
  src="`c\n"
  inline=(inline)

=== after insert ' ' ===
  src="`c \n"
  inline=(inline)

=== after insert '`' ===
  src="`c `\n"
  inline=(inline)

=== FRESH parse of final buffer ===
  src="`c `\n"
  inline=(inline (code_span (code_span_delimiter) (code_span_delimiter)))
```

As you can see, the final parser tree is just `(inline)`, but we expect a `code_span` in there.

master

### Operating system/version

macOS 15.6

WillLillis:
### The Problem

A cached external-scanner token whose lookahead reached the previous range boundary can be incorrectly reused, producing a tree that diverges from a fresh parse of the same buffer.

Stepping through the failing edit from #5636 ("`c \n" -> "`c `\n"):

  1. Before the edit, the inline included range was [0..2). The ` at byte 0 produced an UNCLOSED_SPAN token: size = 1, lookahead_bytes = 1 (the scanner advanced to byte 2 looking for a close, then lexer->eof() returned true).
  2. The text edit at byte 3 propagates through ts_subtree_edit. Its existing lookahead-aware check (subtree.c:734) correctly leaves the cached UNCLOSED_SPAN alone, since the edit is past its lookahead window.
  3. The reparse sets the inline included range to [0..4). ts_range_array_get_changed_ranges produces a diff of [2..4).
  4. ts_parser__reuse_node evaluates the cached UNCLOSED_SPAN and calls ts_parser__has_included_range_difference(self, byte_offset, end_byte_offset) with (0, 1). The intersection check asks "does [2..4) overlap [0..1)?" The answer is no, so the token is reused unchanged.

Step 4 is the bug. The cached token's content spans bytes [0..1), but its validity depends on bytes [0..2), because the scanner's lexer->eof() probe at byte 2 was a real decision input. The check ignores lookahead_bytes entirely.

### The Solution

Include lookahead_bytes in the range passed to `ts_parser__has_included_range_difference`. Now the markdown case checks [2..4) against [0..2), finds an intersection, and rejects the cached token. The scanner re-runs at byte 0, sees the matching ` at byte 3 before eof(), and emits the open/close pair - producing the expected code_span.

- Closes #5636

### AI Policy

Used Claude 4.7 Opus to investigate the issue and decide where the fix should be made.

### DIFF
--- a/crates/cli/src/tests/tree_test.rs
+++ b/crates/cli/src/tests/tree_test.rs
@@ -795,3 +795,44 @@ fn get_changed_ranges(
     *tree = new_tree;
     result
 }
+
+// Regression test for an incremental reparse bug where an external
+// scanner's choice depends on lexer->eof() (and thus on the parser's
+// current included ranges). The cached token at byte 0 was emitted when
+// the included range stopped just after the opener. Widening the range
+// to include a later matching delimiter must invalidate that cached
+// token so the scanner re-runs and emits the open form instead of the
+// unclosed form.
+#[test]
+fn test_reuse_invalidates_scanner_token_when_included_range_expands() {
+    let language = get_test_fixture_language("external_lookahead_eof_boundary");
+    let mut parser = Parser::new();
+    parser.set_language(&language).unwrap();
+
+    let source = "``";
+
+    parser
+        .set_included_ranges(&[Range {
+            start_byte: 0,
+            end_byte: 1,
+            start_point: Point::new(0, 0),
+            end_point: Point::new(0, 1),
+        }])
+        .unwrap();
+    let tree1 = parser.parse(source, None).unwrap();
+    assert_eq!(tree1.root_node().to_sexp(), "(document (unclosed_delim))");
+
+    parser
+        .set_included_ranges(&[Range {
+            start_byte: 0,
+            end_byte: 2,
+            start_point: Point::new(0, 0),
+            end_point: Point::new(0, 2),
+        }])
+        .unwrap();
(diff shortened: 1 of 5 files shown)


clason (approved):

maxbrunsfeld:
This makes perfect sense; nice work @WillLillis

PR #5646, opened 2026-05-31
fix(cli): avoid infinite loop on unmatched highlight assertion

### LINKED ISSUE #5645: tree-sitter test hangs instead of failing on a mismatched highlight assertion

dekobon:
Fixes #5645

A non-matching highlight test never advanced to the outer loop resulting in the actual highlights buffer growing until the process ran out of memory. We now break once the inner scan is done so that the assertion fails cleanly.

### AI Policy

- [X] I have read the [AI Policy](https://tree-sitter.github.io/tree-sitter/6-contributing.html#ai-policy) and this PR complies with it.
- [X] If AI tools were used: I have disclosed the tool and extent of usage below.

I used AI to develop a prototype to fix. Comments and all words are my own. Code is a mixture of my own and AI. Moreover, I used AI to help me learn what all the terms in tree-sitter mean like "highlight test". I also used AI to help me hunt down the root cause of the problem.

### DIFF
--- a/crates/cli/src/test_highlight.rs
+++ b/crates/cli/src/test_highlight.rs
@@ -174,6 +174,14 @@ pub fn iterate_assertions(
 
                 j += 1;
             }
+
+            // Here the inner loop has scanned every highlight that could match this
+            // assertion. If it fell through here (rather than breaking out of
+            // 'highlight_loop), then `i` was not advanced, so re-entering the outer
+            // loop would rescan the same highlights forever causing a hang. The scan
+            // is complete at this point, so we stop because a still-unmatched
+            // assertion is considered a failure.
+            break 'highlight_loop;
         }
 
         if !passed {
--- a/crates/cli/src/tests/test_highlight_test.rs
+++ b/crates/cli/src/tests/test_highlight_test.rs
@@ -4,7 +4,7 @@ use tree_sitter_highlight::{Highlight, Highlighter};
 use super::helpers::fixtures::{get_highlight_config, get_language, test_loader};
 use crate::{
     query_testing::{Assertion, Utf8Point, parse_position_comments},
-    test_highlight::get_highlight_positions,
+    test_highlight::{get_highlight_positions, iterate_assertions},
 };
 
 #[test]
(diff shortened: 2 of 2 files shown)


WillLillis:
Superseded by #5097

PR #4364, opened 2025-04-15
feat(bindings): implement `Clone` for `Query`

WillLillis:
Opening as a draft for now, as 

a) I'm not sure if this is something we want to expose in the public API, deferring to the powers that be to answer that question
b) potential issues with my implementation on the C side
c) not super sure about the wasm implementation

- Related issue #4337

PR #5649, opened 2026-05-31
Add `ts_query_copy` for cloning queries

the-mikedavis:
This allows implementing `Clone for Query` in Rust for example. It's a deep-copy of the entire query object. You may wish to clone a query if you use `ts_query_disable_capture`. One motivation is tags.scm code navigation queries. You may wish to have one query derived from tags.scm which covers definitions and one for references. References are typically significantly more common that definitions, so disabling those captures might save a a fair amount of work when only querying for definition symbols. Instead of initializing the same tags.scm query twice and then disabling, we can save work by cloning the already-built query.

### AI Policy

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -4268,6 +4268,54 @@ fn test_query_disable_pattern() {
 }
 
 #[test]
+fn test_query_deep_clone() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+        let query = Query::new(
+            &language,
+            "
+                (function_declaration
+                    name: (identifier) @name)
+                (function_declaration
+                    body: (statement_block) @body)
+            ",
+        )
+        .unwrap();
+
+        let mut clone = query.deep_clone();
+        clone.disable_pattern(1);
+
+        let source = "function foo() { return 1; }";
+        let mut parser = Parser::new();
+        parser.set_language(&language).unwrap();
+        let tree = parser.parse(source, None).unwrap();
+        let mut cursor = QueryCursor::new();
+
+        // The clone with pattern 1 disabled only produces the @name match.
+        let clone_matches = collect_matches(
+            cursor.matches(&clone, tree.root_node(), source.as_bytes()),
+            &clone,
+            source,
+        );
+        assert_eq!(clone_matches, &[(0, vec![("name", "foo")])]);
+
+        // The original is unaffected and still produces both @name and @body.
+        let original_matches = collect_matches(
+            cursor.matches(&query, tree.root_node(), source.as_bytes()),
+            &query,
+            source,
+        );
+        assert_eq!(
+            original_matches,
+            &[
+                (0, vec![("name", "foo")]),
+                (1, vec![("body", "{ return 1; }")]),
+            ]
+        );
+    });
+}
+
+#[test]
 fn test_query_alternative_predicate_prefix() {
     allocations::record(|| {
         let language = get_language("c");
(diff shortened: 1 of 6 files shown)


maxbrunsfeld:
I still worry that having Clone will lead people to use queries wrong, not realizing that they are shareable across cursors and threads. cloning should almost never be necessary, and none of the other clone operations of tree-sitter objects are this deep/slow.

Should we deprecate disabling captures in query, and make that a cursor level feature? Do you have any push back or alternative thoughts @WillLillis ?

clason:
> Should we deprecate disabling captures in query, and make that a cursor level feature?

Neovim is using that feature, and it addresses a real issue there, so I'd prefer not to throw the baby out with the bathwater.

maxbrunsfeld:
Fair. I guess this is fine. I just worry that the `Clone` impl in Rust will be a footgun, and should almost never be used. Maybe we should make it an inherent method called `deep_clone`, with docs that explain "You shouldn't need this, this object is shareable."

the-mikedavis:
Yeah a separate function (not `impl Clone for Query`) sounds good to me. We can document there that you should only need to clone a query when you mutate it, not when you want to reuse it across threads.

maxbrunsfeld:
@the-mikedavis Sounds good

WillLillis:
> Should we deprecate disabling captures in query, and make that a cursor level feature? Do you have any push back or alternative thoughts @WillLillis ?

@maxbrunsfeld  Sorry for the late reply, I'm usually not able to do much on public channels during the work day. 

I'd agree with @clason regarding disabling captures on the query-level.

> I still worry that having Clone will lead people to use queries wrong, not realizing that they are shareable across cursors and threads. cloning should almost never be necessary, and none of the other clone operations of tree-sitter objects are this deep/slow.

Yeah, I worry about the misuse here as well. I had a similar PR open here: https://github.com/tree-sitter/tree-sitter/pull/4364, however I closed it after discussing offline with @amaanq. He had an alternate approach in mind, but hasn't had time to full work it out/implement it yet.

I'll give this some more thought, but I think a `deep_clone` (or similar) method instead of the `Clone` trait with appropriate documentation warning the user seems like a decent compromise here.

maxbrunsfeld:
With the Rust changes, I see no problem with this. Thank you @the-mikedavis!

maxbrunsfeld (approved):

maxbrunsfeld:
Actually, @the-mikedavis could you add one test that exercises this method in Rust?

the-mikedavis:
What did you have in mind for making this a cursor level feature @maxbrunsfeld? For my purposes I think I want to disable entire patterns (which I believe is more impactful than disabling captures performance-wise?). I've been poking through the query engine and I think we can add a bitset to the query cursor for disabled patterns and skip exploring if a pattern index is in the disabled patterns set. I'm not sure yet how that would translate to disabling captures.

WillLillis (approved):

WillLillis:
i think this is good to merge as-is. If further discussion is desired for making disabling captures a cursor level feature can happen in a dedicated followup issue.

the-mikedavis:
Thanks @WillLillis ! I agree, changing the query engine to support this directly can be done totally independently. I may send a PR as time allows 🙂

WillLillis:
> Thanks @WillLillis ! I agree, changing the query engine to support this directly can be done totally independently. I may send a PR as time allows 🙂

I appreciate it! Just noting here so you don't waste any time, I'm not 100% sure that's a direction we want to go (i.e. https://github.com/tree-sitter/tree-sitter/pull/5649#issuecomment-4593906296). If there's a real need for it, a discussion in a new issue would be the right first step to ensure we don't break any existing usages. 🙂

PR #5655, opened 2026-06-01
fix(cli): highlight test did not properly check for spans on later rows

jannschu:
The early exit condition to skip remaining highlights was incorrectly checking the lexicographical order, which could lead to false passes if a highlight was on a later row with smaller column number.

### AI Policy

I used Claude Code to review my change locally, has been a while since I wrote this, wanted "a second opinion" :), this is the follow up mentioned in #5097.

### DIFF
--- a/crates/cli/src/test_highlight.rs
+++ b/crates/cli/src/test_highlight.rs
@@ -151,7 +151,9 @@ pub fn iterate_assertions(
                 i += 1;
                 continue;
             }
-            if highlight.0.row >= position.row && highlight.0.column > end_column {
+            if (highlight.0.row > position.row)
+                || (highlight.0.row == position.row && highlight.0.column > end_column)
+            {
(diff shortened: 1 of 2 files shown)


clason:
Please don't request Copilot review on public PRs; if you need that, open the PR against your own fork and do it there. You can open a PR here when you're ready.

jannschu:
This PR is ready, the review happens automatically for all my PRs.

clason:
Yes, so please turn that off before opening a PR. See the AI policy (which you checked that you have read).

jannschu:
I turned auto review off. Sorry, half read first rule as  „no PRs created by LLMs“, but the auto review is explicitly listed. Do you insist on a new PR?

clason:
Yes, that would be cleaner (you can use the existing branch, so it's just a single click).

PR #5671, opened 2026-06-07
feat(generate): collect warnings as structured data

WillLillis:
The generate crate is presented as a library, but was really written for just tree-sitter-cli. We've taken some incremental steps along the way to clean up some rough edges in this regard already, namely returning structured errors via `thiserror`  instead of `anyhow`'s loose display only info. 

Today, the generate crate prints warnings for 4 separate non-fatal issue points throughout parser generation. These warnings are unconditionally sent to stderr via `log`, and aren't accessible at all to library consumers. Furthermore, these prints to stderr clash unintuitively with the CLI's `--json` flag. 

Most of these errors are fairly easy to detect by examining the grammar definition, besides the unnecessary conflicts warning which requires a non-trivial amount of logic to run first. It makes sense to gather this warning information as structured data so that it can both be displayed regularly, displayed as JSON, or consumed programmatically by projects depending on `tree-sitter-generate`.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -33,7 +33,7 @@ use tree_sitter_cli::{
     wasm,
 };
 use tree_sitter_config::Config;
-use tree_sitter_generate::OptLevel;
+use tree_sitter_generate::{Diagnostic, GenerateError, OptLevel};
 use tree_sitter_highlight::Highlighter;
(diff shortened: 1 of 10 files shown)


wetneb:
Thanks for working on this! I haven't had a close look at the implementation but it does seem like a step in a very good direction.

clason (approved):

PR #5686, opened 2026-06-15
fix(lib): address strict aliasing violations in `TreeCursor` type

### LINKED ISSUE #5259: Strict Aliasing Violations in `tree_cursor.c`
Originally reported in https://github.com/tree-sitter/tree-sitter/issues/4960#issuecomment-3707523955:

>     3. Please also note that there is some real, uh, strict-aliasing-violating excitement going on over in `tree_cursor.c` that I do not feel qualified to address:
> 
> 
> ```
> lib/src/tree_cursor.c: In function ‘ts_tree_cursor_new’:
> lib/src/tree_cursor.c:157:37: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
>   157 |   ts_tree_cursor_init((TreeCursor *)&self, node);
>       |                                     ^~~~~
> lib/src/tree_cursor.c: In function ‘ts_tree_cursor_copy’:
> lib/src/tree_cursor.c:701:36: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
>   701 |   TreeCursor *copy = (TreeCursor *)&res;
>       |                                    ^~~~
> ```
> 
> The relevant types here are, flattened to one line and field names removed,
> 
> ```
> typedef struct { const TSTree *; Array(TreeCursorEntry); TSSymbol; } TreeCursor;
> typedef struct { const void *; const void *; uint32_t context[3]; } TSTreeCursor;
> ```
> 
> On machines where `alignof(void*) > 2*sizeof(uint32_t)` (say, CHERI architectures), these structures will be _of different size_ because `Array()` `struct`-ures will contain padding bytes to bring their size up to an integral multiple of `alignof(void*)`, which will be included in `TreeCursor` but not in `TSTreeCursor`.

WillLillis:
This addresses the remaining strict aliasing violations in the core C library. Instead of type casting pointers between the internal `TreeCursor` and public facing `TSTreeCursor`, we instead just zero initialize and `memcpy` over. A `_Static_assert` is added to ensure that `TSTreeCursor` is greater than or equal to `TreeCursor` in size. This ensures both `memcpy`s never read OOB, but still allows usage on architectures where the two aren't identical in size (i.e. m68k). In addition to the code fix, I also added some build flags to guard against future regressions in this area.

This will cause compilation to fail on targets where `alignof(void*) > 2*sizeof(uint32_t)` (see https://github.com/tree-sitter/tree-sitter/issues/4960#issuecomment-3707523955). While a build error isn't great, silent data corruption (the situation today on those targets) is much much worse.

If, sometime in the future, we're able to add said architectures to our CI matrix (or at the very least, we gain access to a machine to test builds on as needed), some fixes can be made to ensure compatibility across the board. One solution I saw (but did not implement for this PR, as we have no means of testing it) is to inline the `Array` fields for `TreeCursor`'s `stack` field. This would eliminate any problematic struct padding for the array, and should work with the existing `array_` macros we use today because they just duck-type the `contents`, `size`, and `capacity` fields.

- Closes #5259

### AI Policy

Used Clause Opus 4.8 to discuss solutions and look into the effects on unconventional architectures (i..e CHERI and m68k).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -211,7 +211,7 @@ jobs:
         make -j CFLAGS="$CFLAGS" CC=$CC AR=$AR
       env:
         PLATFORM: ${{ matrix.platform }}
-        CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
+        CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types -Werror=strict-aliasing -Wstrict-aliasing=2
 
     - name: Build C library (CMake)
       if: "!matrix.cross && !matrix.vm"
--- a/.github/workflows/wasm_exports.yml
+++ b/.github/workflows/wasm_exports.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Build C library (make)
         run: make -j CFLAGS="$CFLAGS"
         env:
-          CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types
+          CFLAGS: -g -Werror -Wall -Wextra -Wshadow -Wpedantic -Werror=incompatible-pointer-types -Werror=strict-aliasing -Wstrict-aliasing=2
 
       - name: Build Wasm Library
         working-directory: lib/binding_web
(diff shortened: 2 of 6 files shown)


clason:
If it's technically breaking (even on niche platforms), we probably shouldn't backport it?

WillLillis:
> If it's technically breaking (even on niche platforms), we probably shouldn't backport it?

Yeah fair enough. I could do a manual backport without the static assert so the release branch could still get the behavioral fix at least.

clason (approved):

PR #5692, opened 2026-06-16
feat(query): add #has-ancestor? / #capture-ancestor? predicates

Thorium:
#### Query predicates for ancestor context: `#has-ancestor?` / `#capture-ancestor?`

Hello again, yet another AI-assisted PR from weird-edge-case-issues of non-mainstream environments. ;-) 
This PR adds two query predicates that let a pattern constrain a capture by its **enclosing context** (can I avoid mentioning monad here...almost) — i.e. require that a captured node be nested, at any depth, inside a node of a given type, and optionally bind that ancestor so the existing text predicates can inspect it.

```scheme
;; gate: keep @x only when it is somewhere inside a `class_body`
((identifier) @x (#has-ancestor? @x class_body))

;; gate + bind: also expose the enclosing node as @ctx for #eq?/#match?
((identifier) @op
 (#capture-ancestor? @op call_expression @ctx)
 (#eq? @ctx "..."))
```

The feature lives entirely in the **core query matcher**, so every language binding inherits it with no per-binding work, and it uses only tree structure (node symbols) — never source text — so it needs no access to the document.

#### Compatibility

- **ABI:** no bump, unchanged.
- **Generated parsers / grammars:** unchanged. parser.c unchanged.
- **Bindings:** unchanged; the predicates work through the shared core matcher.
- **Existing queries:** unaffected — the enforcement path is a no-op when a query defines no ancestor constraints.

So no breaking changes of any kind...

#### The Problem

Tree-sitter's query language has no way to express *"match X only when it is inside Y."* Patterns match by structural nesting written out explicitly, and there is no descendant/ancestor axis. This makes a whole class of common
highlighting/analysis tasks impossible to express precisely:

- **Context-sensitive keywords/operators.**: Many languages give ordinary-looking identifiers special meaning *only inside a particular block*. The widest example is functional / monadic builder syntax, where the same word is a keyword in one block and a plain function elsewhere:
  - **Haskell** `do`-notation / list comprehensions.
  - **F#** computation expressions: `select`, `where`, `join`, … are keywords inside `query { … }` but are also normal names (`List.where`, `Seq.head`) everywhere else.
  - **OCaml** binding operators (`let*`, `let+`, `and*`, …).
  - **Scala** for-comprehension generators.
- **Scope-dependent identifiers.**: Highlight `self` / `this` only inside a class body; `yield` only inside a generator function; a label only inside the loop it belongs to.
- **Embedded sub-languages. (EDSL)**: Treat certain identifiers as keywords only inside a specific construct (e.g. an SQL/query builder block, a shader block, a format-string call).

Today the only options are 
- highlight the word everywhere — a flood of false positives, since those names are common — or 
- restructure the grammar to give the construct a distinct node type, which grows the generated parser and is often impossible (the distinction is semantic, not syntactic). Neither is acceptable, so these cases simply go un-highlighted or mis-highlighted.

#### A Solution

As usual, I have a problem, I don't care who fixes it and how, as long as I have a solution. What this PR adds:

Two predicates, recognised and enforced by the core matcher:

##### `#has-ancestor? @capture <type>...`

Keeps a match only if the captured node has an ancestor (at any depth) whose type is one of the listed node types (any-of). Pure structural gate; needs no source text. Covers every language where the enclosing construct is already its
own node type.

```scheme
; OCaml: highlight binding-operator keywords only inside an expression block
((value_name) @keyword.operator
 (#any-of? @keyword.operator "let*" "let+" "and*" "and+")
 (#has-ancestor? @keyword.operator do_expression))

; highlight `this` only inside a class
((this) @variable.builtin (#has-ancestor? @variable.builtin class_body))
```

##### `#capture-ancestor? @capture <type> @out`

Same gate, but also binds the matched (nearest) ancestor to a new capture `@out`, so the binding's existing `#eq?` / `#match?` / `#any-of?` text predicates can test the enclosing context — for example, *which* builder a block belongs to when several share one node type.

```scheme
; F#: the 40+ query custom operations, but ONLY inside `query { ... }`
; (not inside seq/async/task CEs, and not as List.head / Seq.where elsewhere)
((identifier) @keyword.operator
 (#any-of? @keyword.operator
   "select" "where" "join" "groupBy" "sortBy" "thenBy" "count" "distinct"
   "head" "last" "take" "skip" "exists" "all" "contains" "zip" ...)
 (#capture-ancestor? @keyword.operator ce_expression @_ce)
 (#match? @_ce "^query\\b"))
```

Because `#capture-ancestor?` rejects the match when no such ancestor exists, the downstream text predicate never sees a vacuous (empty-capture) pass.

##### How it works (all within `lib/src/query.c`):

1. **Lowering at compile time.** In `ts_query_new`, after a pattern is parsed, its predicate steps are scanned for `#has-ancestor?` / `#capture-ancestor?` and lowered into `AncestorConstraint`s on the `TSQuery` (anchor capture, accepted ancestor symbols as a slice of a shared symbol array, optional output capture). `#capture-ancestor?` registers its output capture on demand.

2. **Enforcement at match-finish.** `ts_query_cursor__push_finished_state` — the single point every finished match flows through — walks `ts_node_parent` from the anchor capture's node (nearest accepted ancestor wins, mirroring innermost-scope semantics). If none is found, the match is dropped; for `#capture-ancestor?`, the matched ancestor is appended to the capture list. Cost is O(depth) per candidate, only for constrained patterns.

3. **Streaming correctness.** A constrained pattern can fail *after* matching structurally, so it is no longer "guaranteed": its `root_pattern_guaranteed` / `parent_pattern_guaranteed` flags are cleared after analysis, preventing `ts_query_cursor_next_capture` from streaming "definite" captures before the constraint is checked. `next_match` advances in a loop, since a rejected candidate can leave `finished_states` empty.

4. **Validation.** `ts_query__parse_predicate` checks arity and argument order and resolves each ancestor type name to a grammar symbol, returning `TSQueryErrorNodeType` / `TSQueryErrorSyntax` with precise offsets instead of silently ignoring a malformed predicate.

Bindings that don't know these predicates still treat them as opaque general predicates (ignored at the binding layer) — but the core has already enforced the constraint during matching, so the results are correct everywhere with zero
binding changes.

##### Why ancestor matching (not a general descendant axis)

A full arbitrary-depth descendant combinator in the forward matcher would ripple through depth bookkeeping, the longest-match rule, quantifiers, the guaranteed-step analysis, and the capture-list pool — large and risky. Matching *upward* from a finished capture is a bounded parent walk that reuses existing node APIs, perturbs nothing in forward matching, and maps directly onto the "enclosing context" question these 

### AI Policy

Yes, I have used Claude Opus 4.8 to help with code and documentation, as neither Rust or English are my main languages.

Thanks for reading.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -418,6 +418,48 @@ fn test_query_errors_on_invalid_predicates() {
 }
 
 #[test]
+fn test_query_errors_on_invalid_ancestor_predicates() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+
+        // An unresolved ancestor type name is a NodeType error (precise offset),
+        // not a silently-ignored predicate.
+        assert_eq!(
+            Query::new(
+                &language,
+                "((identifier) @x (#has-ancestor? @x not_a_real_node))",
+            )
+            .unwrap_err()
+            .kind,
+            QueryErrorKind::NodeType,
+        );
+        assert_eq!(
+            Query::new(
+                &language,
+                "((identifier) @x (#capture-ancestor? @x bogus_node @c))",
+            )
+            .unwrap_err()
+            .kind,
+            QueryErrorKind::NodeType,
+        );
+
+        // Malformed arity / argument order is a Syntax error.
+        for q in [
+            "((identifier) @x (#has-ancestor? @x))", // no type
+            "((identifier) @x (#has-ancestor? statement_block @x))", // anchor not first
+            "((identifier) @x (#capture-ancestor? @x class_declaration))", // missing output
+            "((identifier) @x (#capture-ancestor? @x @c class_declaration))", // wrong order
+        ] {
+            assert_eq!(
+                Query::new(&language, q).unwrap_err().kind,
+                QueryErrorKind::Syntax,
+                "query should be a syntax error: {q}",
+            );
+        }
+    });
+}
+
+#[test]
 fn test_query_errors_on_impossible_patterns() {
     let js_lang = get_language("javascript");
     let rb_lang = get_language("ruby");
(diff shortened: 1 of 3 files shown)


Thorium:
This is a new feature, not a bug-fix. Besides that, this PR fixes the first part of the issue #880.

clason:
Instant question: Why does this need to be built-in (i.e., why do _we_ have to maintain it)? Downstream consumers can implement that themselves [very easily](https://github.com/neovim/neovim/blob/355d010ddf1dac50fe72ab36fbfdddef2a60a379/runtime/lua/vim/treesitter/query.lua#L579-L591).

Thorium:
This PR binds the matched ancestor as a new capture, so the standard `#eq?`/`#match?` can test its text.

Is there a way with current query constructs to highlight an identifier as a keyword only when it's inside a specific block — where the identifier is nested at arbitrary depth and the block shares its node type with look-alikes (distinguished only by a leading keyword)? Name-based `#any-of?` over-matches, and structural patterns can't reach a variable-depth node or tell same-typed blocks apart. If there's a way I missed, this PR isn't needed; if not, `#has-ancestor?` / `#capture-ancestor?` is the smallest fix that will also fix 880.

clason:
Again, why does the existing lower-level functionality not address your use case? I have linked an implementation for `#has-ancestor?` that is much (_much_) simpler than your PR.

(If you can't tell, I am _very_ skeptical about the value of this PR -- and also short on time to read walls of text -- so please address my questions precisely and concisely.)

Thorium:
I know the you don't know F#, but here are a few examples:
```fsharp
query { for x in xs do
        zip ys            // ← highlight: custom operation of the `query` builder
        select x }

Seq.zip xs ys             // not the same : qualified module function
xs |> List.zip ys         // Also not the same
seq { yield (x, y) }      // different builder; `zip` here would be a call, not operation
query { let zip = 3 in zip }  // zip is a local value, not the operation
```

clason:
Again -- and this is my final attempt -- why does this have to be **built in** in the low-level lib?

Thorium:
If there is an issue open with more than 30 upvotes, doesn't it signal either demand for this, or that the current documentation doesn't describe a solution that satisfies users' needs?

sogaiu:
The following may be off, but it seems to me that as clason hinted at in [this comment](https://github.com/tree-sitter/tree-sitter/pull/5692#issuecomment-4718789921), a variety of predicates are implemented by "downstream" consumers (e.g. editors) and this functionality is then available to the end users of that consumer.

For this specific functionality, going off of the above dialog it seems that it's possible to implement it outside of tree-sitter itself relatively easily using existing functionality (code [here](https://github.com/neovim/neovim/blob/355d010ddf1dac50fe72ab36fbfdddef2a60a379/runtime/lua/vim/treesitter/query.lua#L579-L591)) and it seems to have been done so already at least by one downstream consumer (i.e. neovim).

If I had to guess, as a resource-constrained project, I think the maintainers are not interested in taking on and having to maintain functionality that can be practically (and is in some cases already) implemented elsewhere.

Also, it doesn't appear that the direct questions in the following comments were answered:

* [1](https://github.com/tree-sitter/tree-sitter/pull/5692#issuecomment-4718789921)
* [2](https://github.com/tree-sitter/tree-sitter/pull/5692#issuecomment-4719032733)
* [3](https://github.com/tree-sitter/tree-sitter/pull/5692#issuecomment-4719160763)

ObserverOfTime:
Technically the implementation is [here](https://github.com/neovim/neovim/blob/355d010ddf1dac50fe72ab36fbfdddef2a60a379/src/nvim/lua/treesitter.c#L1250-L1282).

Thorium:
It walks the root→node path and checks each ancestor's type against a list of strings. That's it.

Why it still doesn't solve the F# case:

1. Type-only. It compares ts_node_type against type names. It cannot look at the ancestor's text. In F#, query {}, seq {}, async {}, task {} are all ce_expression — so (#has-ancestor? @op ce_expression) confines an op to "inside some computation expression," but cannot single out the query builder. The common-named ops (zip, head, …) would still light up inside seq/async/task. This is exactly the gap that needed the text-aware #capture-ancestor? + #eq?/#match?, which this builtin doesn't provide.
2. Editor-specific. It lives in src/nvim/lua/treesitter.c — Neovim's own binding, not tree-sitter core. Helix, Zed, the tree-sitter CLI, Emacs, the Rust/WASM bindings don't have it. A grammar's highlights.scm is shared across all of them, so it can't portably rely on #has-ancestor?.

So for this (now closed) PR, this changes nothing:
- It can't be used in the portable highlights.scm (non-portable), and even where it is available (Neovim), it's type-only and can't isolate query from the other CEs — so it wouldn't correctly scope the common-named ops anyway.
- It does confirm the ecosystem pattern: ancestor logic is implemented per-binding, type-only — which is consistent with why the core PR (a portable, text-aware, capture-producing version) was declined.

Net: The current PR (the unambiguous-names rules, app-head + bare-statement) would be the right portable fix. If you specifically wanted a Neovim-only improvement, you could add (#has-ancestor? @op ce_expression) to nvim-treesitter's own F# queries to confine the common-named ops to CE context — but it would still over-match across seq/async/task and wouldn't belong in this grammar repo. 

I get it, tree-sitter is kind of read-only maintenance mode, and LSP:s are the way to go, and maintainers are too busy. Guess what, that's totally fine, I can accept the limitations and live with the current implementation just fine.

clason:
Queries are not meant to be portable.

PR #5693, opened 2026-06-17
fix(lib): address strict aliasing violations in `TreeCursor` type

WillLillis:
Backport of #5686 (minus the static assert in tree_cursor.c)


4mzmwsm24f-wq (commented):
Y

PR #5702, opened 2026-06-19
Update D grammar to much newer canonical site.

gdamore:
The D grammar for tree-sitter referenced on this doc is very old (effectively unmaintained), and not in wide use.  This updates to my version which is updated regularly, and used by most mainstream editors that consume tree-sitter.

### AI Policy


clason:
This is not the tree-sitter parser for D but the D bindings for tree-sitter (even though the repo may contain both).

PR #5707, opened 2026-06-25
Update README.md

AndrewGibb:
On a fresh install of cargo, it seems that `cargo binstall` is no longer valid. `cargo install` should be used to install binaries.

### AI Policy

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -12,7 +12,7 @@ The Tree-sitter CLI allows you to develop, test, and use Tree-sitter grammars fr
 
 ### Installation
 
-You can install the `tree-sitter-cli` with [`cargo-install`](https://github.com/cargo-bins/cargo-binstall) :
+You can install the `tree-sitter-cli` with [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall):
 
 ```sh
 cargo binstall tree-sitter-cli
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -12,7 +12,7 @@ The Tree-sitter CLI allows you to develop, test, and use Tree-sitter grammars fr
 
 ### Installation
 
-You can install the `tree-sitter-cli` with `cargo-binstall`:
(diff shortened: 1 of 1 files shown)


wetneb:
Indeed, this installation method requires to install `cargo-binstall` first. Perhaps there could be a clearer way to document this in the README, so that people who are not familiar with `cargo-binstall` get a chance to discover it properly. Perhaps simply by linking `cargo-binstall` to https://github.com/cargo-bins/cargo-binstall?

AndrewGibb:
I can only really offer one data point: I have arrived at this document as a user of nvim following the breadcrumb trail of dependency installation for nvim-kickstart. I don't really know the Rust ecosystem at all. From my point of view, it would be helpful to call out explicitly to install `cargo-binstall`.

The link you suggest would be a helpful clue. It would show that `cargo-binstall` is not part of `cargo`.

wetneb:
Great! Then how about changing your PR so that it simply adds a link to https://github.com/cargo-bins/cargo-binstall#installation on the existing mention of `cargo-binstall`?

Since the `cargo install` method is mentioned just after that, and maintainers have explicitly asked `cargo-binstall` to be mentioned before it in #5533, I think it would give your PR greater chances of getting merged.

crates/cli/README.md
```diff
@@ -12,7 +12,7 @@ The Tree-sitter CLI allows you to develop, test, and use Tree-sitter grammars fr
 
 ### Installation
 
-You can install the `tree-sitter-cli` with `cargo-binstall`:
+You can install the `tree-sitter-cli` with [`cargo-install`](https://github.com/cargo-bins/cargo-binstall) :
```
WillLillis:
```suggestion
You can install the `tree-sitter-cli` with [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall):
```

WillLillis (changes requested):
Including the link here makes sense, should be good to merge after the corrections.

clason (approved):

WillLillis (approved):

PR #5719, opened 2026-06-29
fix(deps): zig manifest for wasmtime is missing windows hashes

clason:
Problem: `zig fetch` on 0.16 can't download zip archives, which wasmtime
uses for Windows. This makes `cargo xtask upgrade-wastime` fail silently
with empty hashes for these platforms.

Solution: Re-run xtask with Zig 0.17 nightly.

(This unblocks bumping `tree-sitter` for Neovim.)

PR #5722, opened 2026-06-30
fix(xtask): eliminate silent errors in zig fetch

WillLillis:
ref https://github.com/tree-sitter/tree-sitter/pull/5719#issuecomment-4830394109

CC @clason I wasn't able to reproduce the empty hash failure with zig 16.0 (maybe it's MacOS only?). Can you make sure this catches the silent error you were seeing earlier?


clason:
Hmm, now I cannot reproduce them either.

PR #5723, opened 2026-06-30
fix(query): various anchor fixes

### LINKED ISSUE #2889: query: adding a `?` quantifier changes meaning of trailing `.` anchor
This input parsed as C

```
#if defined(FOO)
#define BAR
// comment
#endif
```

matches the following query:

```
$ cat q.txt
(translation_unit
	(preproc_if
    	. condition: (_) @cond
        . (preproc_def)+ @def
        . (preproc_else)? @else
        .
    ) @body
)
$ tree-sitter query q.txt t.c 
t.c
  pattern: 0
    capture: body, start: (0, 0), end: (3, 6)
    capture: 0 - cond, start: (0, 4), end: (0, 16), text: `defined(FOO)`
    capture: def, start: (1, 0), end: (2, 0)
$ tree-sitter parse t.c 
(translation_unit [0, 0] - [4, 0]
  (preproc_if [0, 0] - [3, 6]
    condition: (preproc_defined [0, 4] - [0, 16]
      (identifier [0, 12] - [0, 15]))
    (preproc_def [1, 0] - [2, 0]
      name: (identifier [1, 8] - [1, 11]))
    (comment [2, 0] - [2, 10])))
```

I'm surprised by this, since the `comment` node at the end of the `preproc_if` should not be matched if I understand correctly how the `.` anchor works.

Removing the optional match leads to the query not matching:

```
$ cat q.txt 
(translation_unit
	(preproc_if
    	. condition: (_) @cond
        . (preproc_def)+ @def
        .
    ) @body
)
$ tree-sitter query q.txt t.c 
t.c
$ 
```

This is on `tree-sitter 0.20.8`. Seems like `0.20.9` hasn't been package on cargo, etc. so I can't test the latest release.

### LINKED ISSUE #5703: Clarify query anchor (`.`) documentation
### AI Policy

### Problem

For a query like `(list . (item) @first-item)`, it is unclear whether it will match `(item)` in `(list (not-an-item) (item))` or only `(list (item) (not-an-item))`.

> When `.` is placed before the first child within a parent pattern, the child will only match when it is the first named node in the parent.

Suggests that only _the first child_ will be matched, and if the first named node does not match (e.g., is not `(item)`), then there will just be no `@first-item`), but

> For example, the below pattern matches a given `array` node at most once, assigning the `@the-element` capture to the first `identifier` node in the parent `array`

Suggests that it’ll match on the first _`identifier`_ even if not necessarily the first node.

From testing, it seems like the first quoted bit is the actual behaviour. If possible, it would be wonderful with an example of how to capture the first of a given node, as the second quoted bit suggests, but I also get that that might be out of the scope of simply making the documentation less ambiguous. :)

### Expected behavior

If the first quoted bit is indeed how it’s supposed to work, maybe the second could be rephrased to something like:
> For example, the below pattern matches a given `array` node at most once, assigning the `@the-element` capture to the first node in the parent `array` if it is an `identifier` node

### Related issue(s)

- https://github.com/tree-sitter/tree-sitter/issues/5382

### LINKED ISSUE #5714: Incorrect quantified query behavior with anchor and nonmatching sibling
### AI Policy

### Problem

The following query

```query
(translation_unit (comment)* @c . (declaration) @d)
```

for this C code

```ruby
// doc
int a;
int b;
```

matches for `// doc` and `int a`, but fails to capture `int b`. the `b` decl should still be captured since its sibling `comment` is quantified as _optional_.

Extracted from https://github.com/tree-sitter/tree-sitter/issues/3567#issuecomment-2322784894

Note that https://github.com/tree-sitter/tree-sitter/issues/3567#issuecomment-2322784894 also references

```query
  (program
    ((comment)+ @comment.documentation .)?
    (class))
```

> That seems to just ignore the . anchor and match any comment that precedes (though not necessarily immediately precedes) the class.

As far as I can tell, the behavior of an anchor in this position is currently unspecified in the docs (related #5382 / #5703). We should decided on the intended behavior here and _then_ fix it + add a regression test.

### Steps to reproduce

- Clone tree-sitter-c
- Run the query on the code
- Observe inconsistent capture behavior

### Expected behavior

The queries should work as documented.

master (98c25b9)

### Operating system/version

N/A

WillLillis:
Fixes several closely related issues for query anchors. The commit messages and docs additions have more details, but here's a quick TLDR:

- Improve query.c's existing debug dump output
- Make nonsense anchors at the end of a group (i.e. `((comment)+ @c .)`) an error. On master today, these anchors are silently dropped. Note that _leading_ anchors here are already an error.
- For something like `(parent (comment)* @c . (decl))`, if `comment` fails to match, then the anchor is ignored
- For something like `(p (a)+ @A . (b)? @b .)`, if `b` fails to match, then `a` must fulfill the trailing anchor (it must be the last child of `p`)
- Document these behaviors

I think most of these fixes (besides the nonsense anchor error) make sense to backport. CC @clason if you think that's too breaking for a backport, I can drop the label and just cherry-pick the other commits myself.

- Fixes #2889
- Fixes #5714
- Fixes #5703
- Progress towards #5382 

### AI Policy

Used Claude Opus 4.8 to help determine the root cause and fix for each issue.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -1259,6 +1259,45 @@ function foo() {}
 }
 
 #[test]
+fn test_query_matches_with_last_child_anchor_after_optional() {
+    allocations::record(|| {
+        let language = get_language("c");
+        let query = Query::new(
+            &language,
+            "(preproc_if (preproc_def)+ @def . (preproc_else)? @else .)",
+        )
+        .unwrap();
+
+        // The optional `(preproc_else)?` is absent, so the trailing anchor's
+        // last-child requirement transfers to the last `preproc_def`. A trailing
+        // comment means the def is not the last child, so nothing matches.
+        assert_query_matches(
+            &language,
+            &query,
+            "
+#if X
+#define A
+// c
+#endif
+",
+            &[],
+        );
+
+        // With the def as the last child, the (else-less) match is allowed.
+        assert_query_matches(
+            &language,
+            &query,
+            "
+#if X
+#define A
+#endif
+",
+            &[(0, vec![("def", "#define A\n")])],
+        );
+    });
+}
+
+#[test]
 fn test_query_matches_with_last_named_child() {
     allocations::record(|| {
         let language = get_language("c");
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -257,6 +257,51 @@ fn test_query_errors_on_invalid_syntax() {
 }
 
 #[test]
+fn test_query_errors_on_anchor_at_group_edge() {
+    allocations::record(|| {
+        let language = get_language("javascript");
+
+        // Anchors between siblings, or at the first/last position of a *node*
+        // pattern, are valid.
+        assert!(Query::new(&language, "((_) . (_))").is_ok());
+        assert!(Query::new(&language, "(program (_) (_) .)").is_ok());
+        assert!(Query::new(&language, "(program (_)* @x . (_))").is_ok());
+
+        // A `.` at the edge of a *group* is rejected. A group is not a node, so it
+        // has no last child to anchor, and there is no sibling within the group to
+        // anchor to.
+        assert_eq!(
+            Query::new(&language, "((_) .)").unwrap_err(),
+            QueryError {
+                row: 0,
+                offset: 5,
+                column: 5,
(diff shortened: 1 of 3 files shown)


WillLillis:
CC @ribru17 if you have the time to take a look :)

clason (approved):

ribru17:
I’m afk until the 7th, but feel free to go ahead without me if that’s too long!

PR #5725, opened 2026-07-01
fix(query): short-circuit longest-match dedup for disjoint states

### LINKED ISSUE #40517: Regression: Ruby Tree-sitter scrolling becomes very slow on repeated `A #` lines
### Problem

After the tree-sitter dependency bump, scrolling a Ruby buffer with Tree-sitter enabled becomes very slow.

This reproduces with a small synthetic Ruby buffer containing repeated lines:

```ruby
A #
```

Pressing `<C-f>` is very slow. Holding `<C-f>` makes CPU usage reach 100%.

### Steps to reproduce

First, clone `nvim-treesitter`:

```sh
git clone --depth=1 https://github.com/nvim-treesitter/nvim-treesitter /tmp/nvim-treesitter
```

Install the Ruby parser:

```sh
nvim --clean +'set rtp^=/tmp/nvim-treesitter' +'runtime plugin/nvim-treesitter.lua' +'TSInstall ruby'
```

Wait until `:TSInstall ruby` finishes, then quit Nvim.

Create `repro.vim`:

```vim
set runtimepath^=/tmp/nvim-treesitter
execute 'set runtimepath^=' . fnameescape(stdpath('data') . '/site')
runtime plugin/nvim-treesitter.lua
call setline(1, repeat(['A #'], 1000))
set filetype=ruby
set number
set nomodified
lua vim.treesitter.start(0, 'ruby')
```

Then run:

```sh
nvim --clean -S repro.vim
```

After Nvim starts, press:

```text
<C-f>
```

### Actual behavior

Pressing `<C-f>` is very slow.

If `<C-f>` is held down, CPU usage reaches 100%.

### Expected behavior

Scrolling with `<C-f>` should remain responsive, and holding `<C-f>` should not cause continuous 100% CPU usage.

### Nvim version (nvim -v)

NVIM v0.13.0-dev-881+gc040f53dc1

### Vim (not Nvim) behaves the same?

no

### Operating system/version

Ubuntu 26.04

### Terminal name/version

gnome-terminal

### $TERM environment variable

xterm-256color

### Installation

build from repo

WillLillis:
## The Problem

An unanchored quantified sibling like `(program (comment)+ @doc (class))` keeps one match state per matching sibling. #5603 stopped over-pruning them, but the per-node longest-match dedup is pairwise, so with n live states matching became O(n^3).

## The Solution

Two states can only be capture subsets of one another if their captured byte ranges overlap, so keep each group ordered by first-capture position and stop the pairwise scan once the rest of the group is disjoint.

The problematic query from tree-sitter-ruby is:

```scm
 (program
   (comment)+ @comment.documentation
   (class))
```

which is probably a mistake. Assuming I'm reading it correctly, there should be an anchor between `comment` and `class` so that comments are captured as documentation only if they directly precede a class. However, we should still fix this in the query engine.

In order to get a clearer look at the regression, I made a throw-away rust project rather than test things interactively in neovim. The program parses a file with `A #` repeated a given number of times, and measures how long the query takes to run.

### The numbers:

**Full tree-sitter-ruby highlights query**

| Line Count     | `master` time (ms) | PR time (ms)  |
| :---             | :---   |:---      |
| 250      | 22.2       | 1.2   |
| 500   | 176.1        | 4.2      |
| 1000   | 1,412.8        | 16.6      |
| 2000   | 13,706.9        | 64.5      |
| 4000   | 89,088.4        | 248.0      |

**Problematic query itself**

| Line Count     | `master` time (ms) | PR time (ms)  |
| :---             | :---   |:---      |
| 250      | 22.9       | 1.1   |
| 500   | 174.3        | 3.9      |
| 1000   | 1,395.9        | 15.2      |
| 2000   | 11,101.6        | 59.6      |
| 4000   | 88,186.1        | 236.7      |

**Problematic query with anchor**

| Line Count     | `master` time (ms) | PR time (ms)  |
| :---             | :---   |:---      |
| 250      | 0.12       | 0.14   |
| 500   | 0.25        | 0.30      |
| 1000   | 0.50       | 0.64      |
| 2000   | 1.07        | 1.30      |
| 4000   | 2.19        | 2.70      |

We do pay a small cost in normal runs for this fix (it's the sorting done by `ts_query_cursor__sort_states_by_capture`. However this cost is small (typically around ~2%) and does not scale drastically with line  count.

- Fixes neovim/neovim#40517

### AI Policy

Opus 4.8 to track down where the regression was, and to find the interesting edge case that made sorting necessary.

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -1350,6 +1350,35 @@ fn test_query_matches_with_repeated_leaf_nodes() {
 }
 
 #[test]
+fn test_query_matches_optional_capture_before_uncaptured_required_sibling() {
+    allocations::record(|| {
+        let language = get_language("rust");
+        let query = Query::new(&language, "(block (line_comment)? @doc (line_comment))").unwrap();
+
+        // The optional `(line_comment)? @doc` before an *uncaptured* required
+        // `(line_comment)` yields two candidate completions at the block:
+        //     - one where the optional captured `// a` (the required node is then `// b`),
+        //     - one where the optional matched zero (the required node absorbs `// a`,
+        //     leaving `@doc` unbound).
+        // The zero-match completion's captures are a strict subset of the other's, so the
+        // longest-match rule must drop it (there is exactly one match). This regressed when
+        // the dedup pass gained an early-break. Without the accompanying capture-position
+        // sort, the break skips the subset "loser" and it leaks as an extra empty match.
+        assert_query_matches(
+            &language,
+            &query,
+            "
+            fn f() {
+                // a
+                // b
+            }
+            ",
+            &[(0, vec![("doc", "// a")])],
+        );
+    });
+}
+
+#[test]
 fn test_query_matches_with_optional_nodes_inside_of_repetitions() {
     allocations::record(|| {
         let language = get_language("javascript");
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -3687,6 +3687,46 @@ void ts_query_cursor__compare_captures(
   }
 }
 
+// Order two in-progress states for the longest-match dedup pass. Within a
+// (start_depth, pattern_index) group, states with no captures sort first, as they are a
+// subset of every other state (so the dedup pass must always compare them). The rest
+// sort by the start byte of their first capture.
+static bool ts_query_cursor__state_precedes(
+  const TSQueryCursor *self,
+  const QueryState *a,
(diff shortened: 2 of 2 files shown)


clason (approved):

clason:
Akshually, the cost is more like 20% in the last table 🤓

WillLillis:
> Akshually, the cost is more like 20% in the last table 🤓 

Oops, I think that 2% number was the total highlight query's time. 😅

PR #5733, opened 2026-07-07
fix(templates): update setup.py & Package.swift

ObserverOfTime:
### AI Policy

- `setup.py`: Ensure that C source files are added to the source distribution.
- `Package.swift`: Replace the deprecated dependency declaration method.

### DIFF
--- a/crates/cli/src/init.rs
+++ b/crates/cli/src/init.rs
@@ -836,6 +836,8 @@ fn generate_java_bindings(
     Ok(())
 }
 
+// TODO: remove old migrations
+
 fn update_package_json(path: &Path) -> Result<()> {
     let mut contents = fs::read_to_string(path)?
         .replace(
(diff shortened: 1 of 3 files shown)


crates/cli/src/init.rs
```diff
@@ -1241,8 +1249,13 @@ fn update_swift_package(path: &Path) -> Result<()> {
             "atPath: \"src/scanner.c\"",
             "atPath: \"\\(dir)/src/scanner.c\"",
         )
-        .replace("version: \"0.8.0\")", "version: \"0.9.0\")")
-        .replace("(url:", "(name: \"SwiftTreeSitter\", url:");
+        .replace("version: \"0.8.0\")", "version: \"0.10.0\")")
+        .replace("version: \"0.9.0\")", "version: \"0.10.0\")")
```
ObserverOfTime, line 1255: Replacing both versions for backwards compatibility but we should drop pre-0.26 migrations in 0.27.

crates/cli/src/init.rs
```diff
@@ -1209,18 +1209,26 @@ fn update_python_test_binding(path: &Path) -> Result<()> {
...
         info!("Updating Python free-threading support in setup.py");
-        contents = contents.replace(
+        let new_contents = contents.replace(
             r#"startswith("cp"):"#,
             r#"startswith("cp") and not get_config_var("Py_GIL_DISABLED"):"#,
         );
-        write_file(path, contents)?;
+        write_file(path, new_contents)?;
+    }
+    if !contents.contains("include(\"src/*.c\")") {
```
WillLillis, line 1227: If both this and the above `if` condition are true, then the second's replacement will overwrite the first. Instead we should leave `contents` as mutable and reassign it (rather than to `new_contents`).

crates/cli/src/init.rs
WillLillis: Want to leave a `TODO` comment while we're here?

WillLillis (changes requested):

WillLillis (approved):

PR #5737, opened 2026-07-09
feat(cli)!: don't load all languages via config when lib path is passed

### LINKED ISSUE #5729: make config file optional when parser is specified via command-line arguments
### AI Policy

### Problem

I'm using the following command:

```console
$ tree-sitter query --config-path "$configpath" --lib-path "$libpath" --lang-name "$langname" --captures "$querypath" <<<'...'
```

`--config-path "$configpath"` seems unnecessary in the presence of `--lib-path "$libpath"`. Unfortunately the config file is validated unconditionally. As a workaround I'm specifying a dummy path in __config.json__:

```json
{
  "parser-directories": [
    "XXX: Suppress warning about not having configured any parser directories"
  ]
}
```

### Expected behavior

I expect to be able to omit `--config-path` when using `--lib-path`.

I expect to only need to create __config.json__ when it serves a purpose.

WillLillis:
Also, make `--lang-name` without `--lib-path` a hard error

- Fixes #5729

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -210,7 +210,7 @@ struct Parse {
     pub lib_path: Option<PathBuf>,
     /// If `--lib-path` is used, the name of the language used to extract the
     /// library's language function
-    #[arg(long)]
+    #[arg(long, requires = "lib_path")]
     pub lang_name: Option<String>,
     /// Select a language by the scope instead of a file extension
     #[arg(long)]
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
can you mark the first commit as breaking (due to the new hard error)?

clason:
(and maybe pull out the second commit into a separate PR so we can easily backport it)

WillLillis:
> can you mark the first commit as breaking (due to the new hard error)?

Good call, done

> (and maybe pull out the second commit into a separate PR so we can easily backport it)

#5742

PR #5738, opened 2026-07-10
feat(web): expand `Language::load` to accept `URL` filepath

### LINKED ISSUE #5628: web-tree-sitter: Support URL objects directly in `Language.load`
### AI Policy

### Problem

`Language.load` currently accepts `string | Uint8Array`, but not a `URL` object.

In modern TypeScript/ESM code, it’s very common to resolve assets like this:

```ts
const WASM_URL = new URL('./tree-sitter-javascript.wasm', import.meta.url);
const language = await Parser.Language.load(WASM_URL);
```

Right now, TypeScript rejects this even though the runtime behavior appears to already support it.

The current implementation in `lib/binding_web/src/language.ts` looks like:

```ts
static async load(input: string | Uint8Array): Promise<Language> {
  let binary: Uint8Array | WebAssembly.Module;
  if (input instanceof Uint8Array) {
    binary = input;
  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
  } else if (globalThis.process?.versions.node) {
    const fs: typeof import('fs/promises') = await import('fs/promises');
    binary = await fs.readFile(input);
  } else {
    const response = await fetch(input);
    ...
  }
}
```

That makes this look like mostly a type signature limitation rather than a missing runtime capability.

### Evidence

On my side, in TypeScript, I forced-casted a `URL` value like this:

```ts
const language = await Language.load(WASM_URL as any as string);
```

and it worked perfectly fine at runtime as far as my usage could tell.

That makes me think this is likely just a signature expansion, or close to it.

### Expected behavior

It would be nice if `Language.load` accepted `URL` directly, for example:

```ts
static async load(input: string | URL | Uint8Array): Promise<Language>
```

### Why this would help

- Works naturally with `new URL(..., import.meta.url)` patterns
    - This is apparently common enough that bundlers will mutate the URL when they move wasm into the build, as in my case
- Unintuitively, `URL.toString()` actually breaks because TS isn't expecting the url scheme
    - In my integration, I am using `URL.pathname` to get around this issue
- Avoids unsafe TypeScript casts like `as any as string`
- Matches how web-platform APIs like `fetch` already accept `URL`

### Possible implementation note

This may be mostly a TypeScript/API-surface change.

The browser branch already does:

```ts
const response = await fetch(input);
```

which should already be fine with `URL`.

The only thing that may need attention is the Node.js branch:

```ts
binary = await fs.readFile(input);
```

If necessary, that branch could normalize `URL` inputs explicitly (for example, only for `file:` URLs), but from a user perspective this seems like a very natural input type for this API to support.

### Suggested scope

Even if full cross-environment `URL` handling needs some care, it would already be valuable if the API officially accepted `URL` where the current runtime behavior already supports it.

WillLillis:
- Closes #5628

### AI Policy

Used Opus 4.8 to talk through JS basics and ensure there weren't any unintended side effects with this change.

### DIFF
--- a/lib/binding_web/src/language.ts
+++ b/lib/binding_web/src/language.ts
@@ -227,9 +227,10 @@ export class Language {
 
   /**
    * Load a language from a WebAssembly module.
-   * The module can be provided as a path to a file or as a buffer.
+   * The module can be provided as a path to a file, a `URL` to a file, or as a
+   * buffer.
    */
-  static async load(input: string | Uint8Array): Promise<Language> {
+  static async load(input: string | URL | Uint8Array): Promise<Language> {
     let binary: Uint8Array | WebAssembly.Module;
     if (input instanceof Uint8Array) {
       binary = input;
--- a/lib/binding_web/test/language.test.ts
+++ b/lib/binding_web/test/language.test.ts
@@ -4,6 +4,7 @@ import { LookaheadIterator, Language } from '../src';
 import { Parser } from '../src';
 import { C } from '../src/constants';
 import { readFile } from 'fs/promises';
+import { pathToFileURL } from 'url';
 
 let JavaScript: Language;
 let Rust: Language;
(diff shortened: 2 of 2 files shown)


WillLillis:
Lint failure is unrelated, we just need to rebase once #5737 is merged.

clason (approved):

Issue #5564, opened 2026-05-01
prebuilt binaries: tree-sitter is not executable


loeffel-io:
### AI Policy

- [x] I have read the AI Policy and this issue complies with it.

### Problem

The zip file does include a non executable tree-sitter bin, that makes it hard to use in our bazel env and requires workaround:

```starlark
http_archive(
    name = "com_github_tree_sitter_tree_sitter_darwin_arm64",
    build_file_content = """
genrule(
    name = "tree_sitter_bin",
    srcs = ["tree-sitter"],
    outs = ["tree-sitter-cli"],
    cmd = "cp $< $@ && chmod +x $@",
    executable = True,
    visibility = ["//visibility:public"],
)
    """,
    sha256 = "9dc0dc3415a1cd30499750579defbf3f8e000a98f12a65cda8e25981f07e7b0f",
    urls = [
        "https://github.com/tree-sitter/tree-sitter/releases/download/v0.26.8/tree-sitter-cli-macos-arm64.zip",
    ],
)
```

### Expected behavior

Would be great to put already executable binaries into the archive.
Tree-sitter is currently the only tool from 50+ tools that requires this workaround.

Thank you 🙏

clason:
PR welcome (and please check whether this only applies to arm64 or all architectures).

(To my knowledge, this is a fundamental limitation of zip archives, so I'm a bit surprised that all other tools don't have this problem...)

loeffel-io:
@clason, thanks for your reply

do you think that `chmod +x $exe` would be enough in the next line of https://github.com/tree-sitter/tree-sitter/blob/master/.github/workflows/release.yml#L46?

i think this could be belong to all architectures for now.
i also thought this could be a zip limitation but for `ast-grep` for example it works:

```json
  "ast-grep": {
    "binaries": [
      {
        "kind": "archive",
        "url": "https://github.com/ast-grep/ast-grep/releases/download/0.40.5/app-aarch64-apple-darwin.zip",
        "file": "ast-grep",
        "sha256": "55c3a471a483daab49b6413972c3655087f2ce47335ab019067ed1fbf6672107",
        "os": "macos",
        "cpu": "arm64",
        "type": "zip"
      }
    ]
  },
```

greetings from innsbruck

clason:
[...]

PR #5745, opened 2026-07-11
fix(ci): re-set executable permission on release artifacts before zipping

WillLillis:
I spent some time poking around at this issue, and it seems like `zip` _does_ preserve executable permission bits. The upload artifact action that runs before our use of `zip` and `gzip` does not: https://github.com/actions/upload-artifact#permission-loss.

Setting the executable bit here _should_ fix zip archives, but not gz. At worst, it's harmless and essentially costs us nothing.

- References #5564


clason (approved):

clason:
hmm, backport action broken?

PR #5751, opened 2026-07-13
feat(lib): add state to missing nodes

jaboatman:
This allows them to return all possible missing values from a `LookaheadIterator`.

### AI Policy

### DIFF
--- a/lib/binding_rust/bindings.rs
+++ b/lib/binding_rust/bindings.rs
@@ -383,7 +383,7 @@ unsafe extern "C" {
     pub fn ts_node_is_error(self_: TSNode) -> bool;
 }
 unsafe extern "C" {
-    #[doc = " Get this node's parse state."]
+    #[doc = " Get this node's parse state.\n\n For a missing node, this is the state from the recovery path that was\n selected by the parser. It can be used with [`ts_lookahead_iterator_new`] to\n inspect the symbols that are valid in that state. This does not necessarily\n include every symbol that could be recovered by inserting a missing node."]
     pub fn ts_node_parse_state(self_: TSNode) -> TSStateId;
 }
 unsafe extern "C" {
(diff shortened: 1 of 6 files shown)


clason:
This is missing a strong justification (concrete real-world use case that is impossible without that change), especially since this touches the C API.

jaboatman:
Thanks - At my company we have a programming language which uses [rust-sitter](https://github.com/otonoma/rust-sitter) to define the grammar. We then use this in the runtime and the language server (which is a nice feature to have only one grammar definition).

Currently syntax errors are reported by using the tree-sitter lookahead iterator and presenting a list of possible next tokens that were expected.
When a node is determined as `MISSING`, tree-sitter will inject a suitable node there, which works fine for cases where the next node only has one option (e.g. trailing `;`). However, if there are multiple choices available, you will only receive one of them. But, with the parse state attached to the missing node (as in this PR), you can see all of the choices with the lookahead iterator.

Example without the state attached:
image

Best we could do in this case would be look at what the missing node was (`Ident`) and display that.

Example with the state attached:
image

The lookahead iterator can now display all of the possible choices for the `MISSING` node.

jaboatman:
Hi, any chance you were able to review? I understand if you are busy.
I have a handful of other small changes I have made to tree-sitter to support rust-sitter. My aim is to be able to get my rust-sitter changes into crates.io, but I currently have to depend on a github fork of tree-sitter, which isn't allowed for crates. (Alternatively, I could put the fork on crates as well under a different name, but then I have to maintain _that_ as it gets out of sync.)

Thanks for your time.

WillLillis:
I haven't had a chance to seriously consider this yet, sorry. Will get to it in due time if Max/Amaan aren't able to first.

WillLillis (approved):
This seems reasonable to me. I couldn't find any instance elsewhere on the lib that depended on the previous `0` value here, and this is a nice feature to have. I made a few tweaks:

- removed the uneeded `api.h` include
- updated the docs so this feature is discoverable
- tweaked the commit message, since it was somewhat misleading. This doesn't expose _every_ possible state for a MISSING node in general. If multiple stacks are valid, we only get the nodes from whichever stack the runtime selected.

jaboatman:
> This seems reasonable to me. I couldn't find any instance elsewhere on the lib that depended on the previous `0` value here, and this is a nice feature to have. I made a few tweaks:
> 
>     * removed the uneeded `api.h` include
> 
>     * updated the docs so this feature is discoverable
> 
>     * tweaked the commit message, since it was somewhat misleading. This doesn't expose _every_ possible state for a MISSING node in general. If multiple stacks are valid, we only get the nodes from whichever stack the runtime selected.

Excellent, thanks.

Issue #5607, opened 2026-05-18
Keyword extraction behaves inconsistently when keywords are case-insensitive


Tuupertunut:
### AI Policy

- [x] I have read the AI Policy and this issue complies with it.

### Problem

When using case-insensitive regexes for keywords, keyword extraction fails or succeeds to extract keywords seemingly randomly, based on what the keywords are. The generated parser.c also looks very different for different sets of keywords.

#### Case 1

Here is a minimal reproducible grammar:

```js
word: ($) => $.variable,

rules: {
    expression: ($) => choice($.variable, $.unary_expression),
    unary_expression: ($) => seq($.unary_command, $.expression),
    variable: ($) => token(prec(1, /[a-zA-Z_][a-zA-Z0-9_]*/)),
    unary_command: ($) => token(prec(2, choice(/count/i, /creatediarylink/i))),
},
```

There are two case insensitive unary commands or "keywords" in this language, `count` and `creatediarylink`.

Parse the file
```
count_foo
```
and see how it gives an incorrect parse tree:
```
$ tree-sitter generate && tree-sitter parse -c testfile 
0:0 - 1:0   expression
0:0 - 0:9     unary_expression
0:0 - 0:5       unary_command `count`
0:5 - 0:9       expression
0:5 - 0:9         variable `_foo`
```
Keyword extraction does not seem to work here.

<details>
<summary>Generated parser.c (spoiler)</summary>

```c
/* Automatically @generated by tree-sitter */

#include "tree_sitter/parser.h"

#if defined(__GNUC__) || defined(__clang__)
#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
#endif

#define LANGUAGE_VERSION 15
#define STATE_COUNT 6
#define LARGE_STATE_COUNT 3
#define SYMBOL_COUNT 5
#define ALIAS_COUNT 0
#define TOKEN_COUNT 3
#define EXTERNAL_TOKEN_COUNT 0
#define FIELD_COUNT 0
#define MAX_ALIAS_SEQUENCE_LENGTH 2
#define MAX_RESERVED_WORD_SET_SIZE 0
#define PRODUCTION_ID_COUNT 1
#define SUPERTYPE_COUNT 0

enum ts_symbol_identifiers {
  sym_variable = 1,
  sym_unary_command = 2,
  sym_expression = 3,
  sym_unary_expression = 4,
};

static const char * const ts_symbol_names[] = {
[...]

PR #5756, opened 2026-07-15
fix(generate): only strip case-folded ſ/K when the /i flag is set

### LINKED ISSUE #5755: Characters `K` and `ſ` are excluded from ranges accepting all characters
### AI Policy

### Problem

Consider a simple grammar, which parses single characters enclosed in double quotes:
```js
export default grammar({
  name: "unicode_bug",

  rules: {
    source_file: $ => seq('"', /[^"]/, '"'),
  }
});
```
This grammar successfully parses things like `"a"`, `"ß"`, `"丈"`. In other words, any character that is not a double quote, enclosed by double quotes. But it rejects `"ſ"` and `"K"`. This is unexpected to me because those characters are valid unicode characters that I would expect to be accepted by the grammar.

For the `K` character, this used to work with tree-sitter 0.25.5 and does not work with tree-sitter 0.26.11 anymore. Bisecting led me to the culprit: 07c4ed220e434165991ebf5254dc613c90e64a9c.

### Steps to reproduce

```sh
git clone https://github.com/wetneb/tree-sitter-unicode-bug
cd tree-sitter-unicode-bug
tree-sitter generate
tree-sitter parse test.txt
```

### Expected behavior

Those two characters are not excluded from unicode ranges which accept all characters.
I am not sure how to avoid the issue that motivated 07c4ed220e434165991ebf5254dc613c90e64a9c, though. Maybe @WillLillis has ideas? 

tree-sitter 0.26.11

### Operating system/version

debian testing

WillLillis:
The #5607 fix removed the long-s `ſ` (U+017F) and Kelvin sign `K` (U+212A) from any character class that contained the code point plus its ASCII pair. But those code points are also legitimate members of broad classes that never underwent case folding, so a pattern like `/[^"]/` or `/\p{L}/` (no `i` flag) had them silently punched out.

Case folding is the only thing that introduces these code points, and it only runs under `/i`. Gate the strip on the case-insensitive flag so folding artifacts are removed while a class that genuinely spans `ſ`/`K` keeps them.

A broad class combined with `/i` is still affected.

| Pattern | No strip | Long-s strip (#3831) | Long-s + Kelvin K strip (#5740) | This PR (case insensitive-only strip) |
|:--- |:--- |:--- |:--- |:--- |
| k/s in-class keywords               | Broken | Broken | Works | Works |
| s single-letter keyword            | Broken | Works | Works | Works |
| Broad class `[^"]`, `\p{L}`          | Works | Works | Broken | Works |
| Broad class insensitive `[^"]/i` | Works | Works | Broken | Broken |

With this change set, every reasonable case works, with the one exception being large character classes with the case insensitive flag attached. I'm currently looking into a much more involved fix that would work in _all_ cases, but I'm not sure it's going to be worth the LOC cost. This PR marked as a draft until then.

EDIT: More involved fix makes sense. Instead of letting the `regex-syntax` crate handle folding, we can do it ourselves and remove the added unicode points only when they were added via folding. All 4 cases from the above table now work for this PR.

- Fixes #5755

### AI Policy

Used Opus 4.8 to walk through each state master has been in for the various patterns.

### DIFF
--- a/crates/generate/src/prepare_grammar/expand_tokens.rs
+++ b/crates/generate/src/prepare_grammar/expand_tokens.rs
@@ -1,6 +1,6 @@
 use regex_syntax::{
     ParserBuilder,
-    hir::{Class, Hir, HirKind},
+    hir::{Class, ClassUnicode, ClassUnicodeRange, Hir, HirKind},
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -15,6 +15,7 @@ use crate::{
 struct NfaBuilder {
     nfa: Nfa,
     is_sep: bool,
+    case_insensitive: bool,
     precedence_stack: Vec<i32>,
 }
 
@@ -79,6 +80,7 @@ pub fn expand_tokens(mut grammar: ExtractedLexicalGrammar) -> ExpandTokensResult
     let mut builder = NfaBuilder {
         nfa: Nfa::new(),
         is_sep: true,
+        case_insensitive: false,
         precedence_stack: vec![0],
     };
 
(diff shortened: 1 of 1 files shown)


clason (approved):

wetneb:
Thanks a lot for the fix, it looks great to me :)

PR #5769, opened 2026-07-18
feat(cli)!: rework HTML output modes in highlight

### LINKED ISSUE #554: Feature request: add flag to output html w/o tables for CLI tree-sitter highlight
Hello,

this is closely related to https://github.com/tree-sitter/tree-sitter/issues/553.

The idea would be to have a flag to switch between the current table layout and, alternatively, a `<pre class="language-x"><code><spans…>` or `<div class="language-x"><pre><spans…>` layout w/o line-numbers like in the examples below:

E.g. from this code:
```javascript
function square(n) {
  return n * n;
}
```

GitHub currently outputs this in the issue preview from the markdown editor:

```html
<div class="highlight highlight-source-js">
  <pre>
    <span class="pl-k">function</span> 
    <span class="pl-en">square</span>(
    <span class="pl-smi">n</span>) {
    
    <span class="pl-k">return</span> n 
    <span class="pl-k">*</span> n;
  }
</pre>
</div>
```

Structure of output from `pygmentize -f html -l javascript`

```html
<div class="highlight">
  <pre>
    <span class="kd">function</span> 
    <span class="nx">square</span>
    <span class="p">(</span>
    <span class="nx">n</span>
    <span class="p">)</span> 
    <span class="p">{</span>

    <span class="k">return</span> 
    <span class="nx">n</span> 
    <span class="o">*</span> 
    <span class="nx">n</span>
    <span class="p">;</span>

    <span class="p">}</span>
  </pre>
</div>
```

Prism generates a structure like this:

```html
<pre class="language-js">
  <code class="language-js">
    <span class="token keyword">function</span> 
    <span class="token function">square</span>
    <span class="token punctuation">(</span>n
    <span class="token punctuation">)</span> 
    <span class="token punctuation">{</span>

    <span class="token keyword">return</span> n 
    <span class="token operator">*</span> n
    <span class="token punctuation">;</span>

    <span class="token punctuation">}</span>
  </code>
</pre>
```

What do you think about such a CLI flag for `tree-sitter highlight`? 

With some help on https://github.com/tree-sitter/tree-sitter/issues/553 I can also give this a try and send you a PR of an implentation

WillLillis:
Outputs a plain `<div class="highlight"><pre><code>` block instead of the default `<table>` with a line number column.

- Closes #554 

### AI Policy

Used Opus 4.8 to discuss the various output format options for the `--no-tables` format.

### DIFF
--- a/crates/cli/src/highlight.rs
+++ b/crates/cli/src/highlight.rs
@@ -12,6 +12,7 @@ use std::{
 use ansi_colours::{ansi256_from_rgb, rgb_from_ansi256};
 use anstyle::{Ansi256Color, AnsiColor, Color, Effects, RgbColor};
 use anyhow::Result;
+use clap::ValueEnum;
 use log::{info, warn};
 use serde::{Deserialize, Deserializer, Serialize, Serializer, ser::SerializeMap};
 use serde_json::{Value, json};
@@ -26,8 +27,9 @@ pub const HTML_HEAD_HEADER: &str = "
   <style>
     body {
       font-family: monospace
-    }
-    .line-number {
+    }";
+
+pub const HTML_LINE_NUMBER_STYLE: &str = "    .line-number {
       user-select: none;
       text-align: right;
       color: rgba(27,31,35,.3);
@@ -35,8 +37,7 @@ pub const HTML_HEAD_HEADER: &str = "
     }
     .line {
       white-space: pre;
-    }
-  </style>";
+    }";
 
 pub const HTML_BODY_HEADER: &str = "
 </head>
@@ -308,12 +309,36 @@ fn terminal_supports_truecolor() -> bool {
         .is_ok_and(|truecolor| truecolor == "truecolor" || truecolor == "24bit")
 }
 
+/// The kind of HTML emitted when highlighting to HTML.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
+pub enum HtmlOutput {
+    /// A complete, self-contained document wrapping a plain
+    /// `<div class="highlight"><pre><code>` block.
+    Document,
+    /// A complete document with a line-number column (a `<table>` layout).
+    #[value(name = "line-numbers")]
+    NumberedDocument,
+    /// Only the code markup, without the surrounding document.
+    Fragment,
+}
+
+/// How token colors are applied in HTML output.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
+pub enum HtmlStyling {
+    /// `class="..."` spans plus a generated `<style>` carrying the theme's colors.
+    Classes,
+    /// `style="..."` spans with the colors inlined.
+    Inline,
+    /// `class="..."` spans with no colors emitted (supply your own stylesheet).
+    Minimal,
+}
+
 pub struct HighlightOptions {
     pub theme: Theme,
     pub check: bool,
     pub captures_path: Option<PathBuf>,
-    pub inline_styles: bool,
-    pub html: bool,
+    /// `None` for regular output, `Some((layout, style))` when emitting HTML.
+    pub html: Option<(HtmlOutput, HtmlStyling)>,
     pub quiet: bool,
     pub print_time: bool,
     pub cancellation_flag: Arc<AtomicUsize>,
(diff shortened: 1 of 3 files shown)


clason (approved):

clason:
Hmm... This is a bit fine-grained to my taste -- going that route will net us a whole zoo of flags to fine-tune the output. Can we take a step back and discuss the actual use cases for the command (we want to support) and see what they need?

Maybe it's enough to have a (default?) "full" output that can be directly piped into a browser and a "porcelain" output that contains just the markup without _any_ wrapping ("bring your own")?

clason:
Or, put another way, is that plain span likely to serve _all_ non-CLI needs? If so, maybe we can use a more "intent-oriented" name for the flag (which will also cover the inevitable bikeshedding of the output afterwards)?

clason:
Or put yet another way: When would you _want_ a table with line numbers in your HTML output? Why not make this the default behavior? (Honest question.)

WillLillis:
> Maybe it's enough to have a (default?) "full" output that can be directly piped into a browser and a "porcelain" output that contains just the markup without _any_ wrapping ("bring your own")?

Yeah, I think this is the right way to go. I settled on the following for `--html`

- No additional args: A self contained document
- `--css-classes`: The existing option, css-classes rather than inline styles. (Minimal maintenance burden and its orthogonal to the other flags)
- `--fragment`: The "porcelain" mode, emits only an HTML fragment.
- `--line-numbers`: Off by default, adds a line table

`--line-numbers` and `--fragment` conflict, and all flags require `--html` to be passed. I think this provides small enough building blocks for people to use, while also giving a nice enough OOTB view.

clason:
If we rethink things from scratch, I feel that `css-classes` should be the default for HTML output. Is there even a need to have hard-coded colors in that case? (Honest question.) Then we could make it a `minimal` opt-in -- but maybe that would just be too much churn for the same number of flags?

In any case, I agree it's orthogonal to this change. But I think this (before 0.27) is a chance to rip some bandaids off; I'd rather make a single batch of breaking changes instead of piecemeal updates over multiple versions.

WillLillis:
> If we rethink things from scratch, I feel that `css-classes` should be the default for HTML output. Is there even a need to have hard-coded colors in that case? (Honest question.) Then we could make it a `minimal` opt-in -- but maybe that would just be too much churn for the same number of flags?

I was mostly leaning on Chesterton's Fence, but don't see any strong reason not to redesign things a bit here. It doesn't add much to include a `--minimal` mode here, but I think making `--css-classes` the default should be fine. I reorganized the two flags to take values (I was getting a bit lost in the soup of boolean switches).

Copied directly from `tree-sitter highlight --help`:

```
      --layout <LAYOUT>
          When generating HTML, the document structure to emit

          Possible values:
          - document:     A complete, self-contained document wrapping a plain `<div class="highlight"><pre><code>` block
          - line-numbers: A complete document with a line-number column (a `<table>` layout)
          - fragment:     Only the code markup, without the surrounding document

          [default: document]

      --style <STYLE>
          When generating HTML, how token colors are applied

          Possible values:
          - classes: `class="..."` spans plus a generated `<style>` carrying the theme's colors
          - inline:  `style="..."` spans with the colors inlined
          - minimal: `class="..."` spans with no colors emitted (supply your own stylesheet)

          [default: classes]
```

clason:
Thanks, that is much cleaner! Just for the record, what would be the use case for
1. `layout = line-numbers` _for HTML output_ (we shouldn't touch console output)?
2. `style = minimal` (is there sufficient demand for that over `classes`)?

(I agree that `minimal` is very low cost but in general I prefer to add things based on -- reasoned and reasonable -- demand rather than "just in case"; my own earlier suggestion is obsolete with the flag refactor.)

WillLillis:
> 1. `layout = line-numbers` for HTML output (we shouldn't touch console output)?

I don't feel super strongly about this, but I'm not super familiar with the use cases behind `tree-sitter highlight --html` besides what the linked issue discusses. My main thought here is that we used to include line numbers by default, so it makes some sense to leave the functionality in behind a flag. That being said, I wouldn't mind removing it if you think it's worth it. Agreed (I think) on not touching console output, mostly from a maintenance perspective.

>  2. `style = minimal` (is there sufficient demand for that over classes)?

The reasoning behind `minimal` is to allow users to supply their own style sheet, rather than to be tied to the theme in their local config (typically intended for console output). I think this serves as a nice building block if users want to build on top of the command's output, which seems close to/ is a parallel of the use case described in #554.

PR #5770, opened 2026-07-19
ci: update macos runners

clason:
THIS IS A TEST.

(Eventually we want to do a general bump to `latest`, but right now I just want to reproduce a local build failure with Xcode 27 beta.)


clason:
Hmm, local failure does not reproduce on the runners. Curiouser and curiouser!

PR #5782, opened 2026-07-23
Fix associativity resolution with mixed shift precedences

### LINKED ISSUE #1301: Adding a rule changes associativity from right to left
In my example, I want `addition` to be right-associative. This works by itself, but when I add another rule `superaddition`, `addition` becomes left associative. So adding one rule changes the behaviour of another rule.

This seems unexpected to me. Is this a bug in tree-sitter, or a limitation of LR1 parsers? Could tree-sitter at least warn that it changes the behaviour, or when it can't enforce the `prec.right` on a rule?

Grammar:

```
module.exports = grammar({
  name: 'superaddition',

  rules: {
    expression: $ => choice(
      $.addition,
      $.superaddition,
      $.number
    ),
    addition: $ => prec.right(1, seq($.expression, "+", $.expression)),
    superaddition: $ => prec.right(seq($.expression, "+", "+", $.expression)),
    number: $ => "1"
  }
});
```

Test:

```
================================================================================
Addition is right associative
================================================================================

1 + 1 + 1 + 1

--------------------------------------------------------------------------------

(expression
  (addition
    (expression
      (number))
    (expression
      (addition
        (expression
          (number))
        (expression
          (addition
            (expression
              (number))
            (expression
              (number))))))))
```

Sjord:
Only shift_is_less and shift_is_more cannot distinguish the situation where
both an equal-precedence shift and a lower-precedence shift exists. Add
shift_is_equal to detect whether any shift has equal precedence.
    
Also add test.
    
- Closes #1301

I am not particularly sure about this fix. I had difficulty running the unit tests myself, and I only half understand shift-reduce parsers.

### AI Policy

I used a local LLM agent and online LLM chat to identify the problem. Initially I also asked it to develop a fix, but then I read the AI policy and re-developed it myself.


clason:
That breaks some existing parsers (bash), so more care is needed. I don't think this is something that naive LLM use can fix.

WillLillis:
@Sjord Thanks for getting the ball rolling on this! I reworked your fix a bit to avoid breaking other conflict cases, and added a corpus grammar for the symmetric left precedence case.

clason (approved):

PR #5784, opened 2026-07-23
fix(query): don't emit captures when `ERROR` and `MISSING` steps never match

### LINKED ISSUE #5079: Query: `(MISSING)` and `(ERROR)` match despite no missing or error nodes
### Problem

Queries which check for `(MISSING)` and `(ERROR)` seem to match despite the parse tree not actually containing any missing or error nodes.

Might be specific to https://tree-sitter.github.io/tree-sitter/7-playground.html (respectively generally the JavaScript bindings). I wasn't able to reproduce this locally yet with java-tree-sitter (jtreesitter), but maybe my test setup is incorrect.

### Steps to reproduce

1. Open https://tree-sitter.github.io/tree-sitter/7-playground.html
2. Select "Java" as language, and check the "Query" and "Accessibility" checkboxes
3. Enter the following valid Java source code
    ```java
    class A extends B {
    }
    ```
4. Enter one of the following Tree-sitter queries
    ```
    (superclass
      (MISSING)
    ) @a
    ```
    ```
    (superclass
      (ERROR)
    ) @a
    ```
   :x: Bug: The queries match the `extends B` part of the source (highlighted in the source editor), despite there neither being a `MISSING` nor an `ERROR` node.

Side note: If you replace `(MISSING)` with `(MISSING type_identifier)`, then the query behaves as expected, having only a match if there is actually a `MISSING` node (e.g. for `class A extends {}`; note the missing superclass).

### Expected behavior

`(MISSING)` and `(ERROR)` should only have matches if there are actually missing or error nodes.

v0.25.10 (?, version of Playground)

### Operating system/version

WASM

chereshnyuk:
### The problem

On valid source with no errors, a captures query like `(superclass (ERROR)) @a` still emits the capture. Only captures are affected. In matches mode the same query correctly returns nothing, so you need `-c` to see it.

```console
$ cat A.java
class A extends B {}
$ cat q.scm
(superclass (ERROR)) @a
```

Before:
```console
$ tree-sitter query -c q.scm A.java
A.java
    pattern:  0, capture: 0 - a, start: (0, 8), end: (0, 17), text: `extends B`
```

After:
```console
$ tree-sitter query -c q.scm A.java
A.java
```

### The solution

Clear those flags on every step reachable only through an `ERROR` or `MISSING` step, stopping at the zero-match skip targets of `?` and `*`. This also stops a multi-step `ERROR`/`MISSING` match from being dropped under a match limit.

### AI Policy

Sonnet 5 wrote the regression tests.
Opus 4.8 validated the  solution.
Opus 5 back-to-back testing

### DIFF
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -2049,6 +2049,36 @@ static bool ts_query__analyze_patterns(TSQuery *self, unsigned *error_offset) {
     }
   }
 
+  // Mark as fallible every step reachable only through an ERROR or MISSING step.
+  for (unsigned i = 0; i < self->patterns.size; i++) {
+    QueryPattern *pattern = array_get(&self->patterns, i);
+    unsigned start = pattern->steps.offset;
+    unsigned end = start + pattern->steps.length - 1; // exclude DONE
+    unsigned fallible_end = start;
+    for (unsigned j = start; j < end; j++) {
+      QueryStep *step = array_get(&self->steps, j);
+      if (step->is_missing || step->symbol == ts_builtin_sym_error) {
+        unsigned limit = step->alternative_is_skip ? step->alternative_index : end;
+        if (limit > fallible_end) fallible_end = limit;
+      }
+      if (j < fallible_end) {
+        step->parent_pattern_guaranteed = false;
+        step->root_pattern_guaranteed = false;
+      }
+    }
+
+    // A dead end keeps its initialized flags, which would halt the chase below.
+    if (fallible_end > start) {
+      for (unsigned j = start; j < end; j++) {
+        QueryStep *step = array_get(&self->steps, j);
+        if (step->is_dead_end) {
+          step->parent_pattern_guaranteed = false;
+          step->root_pattern_guaranteed = false;
+        }
+      }
+    }
+  }
+
   // Propagate fallibility. If a pattern is fallible at a given step, then it is
   // fallible at all of its preceding steps.
   bool done = self->steps.size == 0;
--- a/lib/src/query.c
+++ b/lib/src/query.c
@@ -1650,6 +1650,46 @@ static void ts_query__dump_steps(const TSQuery *self, const char *label) {
 }
 #endif
 
+// Mark fallible ERROR/MISSING steps and what only follows one.
+static void ts_query__mark_fallible_steps(TSQuery *self) {
+  Array(bool) entered = array_new();
+  array_grow_by(&entered, self->steps.size);
+  for (unsigned i = 0; i < self->patterns.size; i++) {
+    *array_get(&entered, array_get(&self->patterns, i)->steps.offset) = true;
+  }
+  bool done = false;
+  while (!done) {
+    done = true;
+    for (unsigned i = 0; i < self->steps.size; i++) {
+      QueryStep *step = array_get(&self->steps, i);
+      if (!*array_get(&entered, i) || step->depth == PATTERN_DONE_MARKER) continue;
+      // A dead end or fallible step gets no sequence edge; alternative edges are unguarded.
+      if (
+        !step->is_dead_end &&
+        !(step->is_missing || step->symbol == ts_builtin_sym_error) &&
+        i + 1 < self->steps.size &&
+        !*array_get(&entered, i + 1)
+      ) {
+        *array_get(&entered, i + 1) = true;
+        done = false;
+      }
+      if (step->alternative_index != NONE && !*array_get(&entered, step->alternative_index)) {
+        *array_get(&entered, step->alternative_index) = true;
+        done = false;
+      }
+    }
+  }
+  for (unsigned i = 0; i < self->steps.size; i++) {
+    QueryStep *step = array_get(&self->steps, i);
+    if (step->depth == PATTERN_DONE_MARKER) continue;
+    if ((step->is_missing || step->symbol == ts_builtin_sym_error) || !*array_get(&entered, i)) {
+      step->parent_pattern_guaranteed = false;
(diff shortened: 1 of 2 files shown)


clason:
(please don't merge `master` into a PR branch; instead `git rebase` and then force-push your branch to keep history clean and the change sets meaningful)

chereshnyuk:
Got it, thanks. Fixed rebasing and force-pushing instead of merging now.

lib/src/query.c
```diff
@@ -2049,6 +2049,36 @@ static bool ts_query__analyze_patterns(TSQuery *self, unsigned *error_offset) {
     }
   }
 
+  // Mark as fallible every step reachable only through an ERROR or MISSING step.
+  for (unsigned i = 0; i < self->patterns.size; i++) {
+    QueryPattern *pattern = array_get(&self->patterns, i);
+    unsigned start = pattern->steps.offset;
+    unsigned end = start + pattern->steps.length - 1; // exclude DONE
+    unsigned fallible_end = start;
+    for (unsigned j = start; j < end; j++) {
```
WillLillis: This treats the `steps` list as a flat array when it's a graph. Later alternation branches past the index of `fallible_end` will be incorrectly marked off.

lib/src/query.c
```diff
@@ -2049,6 +2049,36 @@ static bool ts_query__analyze_patterns(TSQuery *self, unsigned *error_offset) {
...
+      QueryStep *step = array_get(&self->steps, j);
+      if (step->is_missing || step->symbol == ts_builtin_sym_error) {
+        unsigned limit = step->alternative_is_skip ? step->alternative_index : end;
+        if (limit > fallible_end) fallible_end = limit;
+      }
+      if (j < fallible_end) {
+        step->parent_pattern_guaranteed = false;
+        step->root_pattern_guaranteed = false;
```
WillLillis: I'm fairly certain this interacts poorly with the query cursor, causing in progress captures to not be emitted.

lib/src/query.c
```diff
@@ -2049,6 +2049,36 @@ static bool ts_query__analyze_patterns(TSQuery *self, unsigned *error_offset) {
...
+      }
+    }
+
+    // A dead end keeps its initialized flags, which would halt the chase below.
+    if (fallible_end > start) {
```
WillLillis: Similar issue with `steps` here.

WillLillis (changes requested):
I haven't had a chance to _fully_ review this, but I did find a few issues it stands now. I've left some comments on the source, but there are likely other unintended side effects (as with nearly any change to the query engine). One example I found some differing behavior from master: 

```scm
(if_statement [(_) @any (ERROR) @err]) @root
```

```c
int main(void) { if (1) { } }
```

clason:
> Sonnet 5 wrote the regression tests.
> Opus 4.8 validated the solution.

And what did _you_ do?

With all due respect (and please correct me if I'm wrong), it looks like you're just feeding review comments straight to your LLMs and pushing the updates it spits out in reply (at least squashed and not letting it push commits directly on top of your PR...) Such PRs violate our AI policy. (If it were enough to point an LLM at the issue and say "fix pls", we would have done so ourselves...)

Specifically, it is not very respectful to not reply (yourself!) to maintainer comments. If you can't explain the changes _and how they address the issues raised in the comment_, we can't really trust the changes. (If you're not ready to answer and need more time, mark the PR as "draft" until you are.)

chereshnyuk:
@WillLillis Thanks for the catch. The difference is confirmed. It comes from master emitting an extra capture rather than from this change dropping one:

```scm
(if_statement [(_) @any (ERROR) @err]) @root
```
```c
int main(void) { if (1) { } }
```
With this change an alternation branch whose `ERROR` or `MISSING` step cannot match does not change what the pattern reports. Master instead emits once per branch present. Substituting for the alternation in `(if_statement <alternation>) @root` on the same tree:

| alternation substituted | matches | `@root` emitted here | `@root` emitted on master |
| --- | :---: | :---: | :---: |
| `[(_) @any]` no dead branch | 2 | 1 | 1 |
| `[(_) @any (ERROR)]` | 2 | 1 ✓ | 2 ✗ |
| `[(_) @any (ERROR) (ERROR) (ERROR)]` | 2 | 1 ✓ | 4 ✗ |
| `[(_) @any (MISSING)]` | 2 | 1 ✓ | 2 ✗ |

`@any` is emitted twice in every row and is omitted. A tick marks agreement with the first row which is what the asserts. The first row also shows that neither side emits `@root` once per match.

lib/src/query.c
chereshnyuk: The hunk is gone.

chereshnyuk:
> > Sonnet 5 wrote the regression tests.
> > Opus 4.8 validated the solution.
> 
> And what did _you_ do?
> 
> With all due respect (and please correct me if I'm wrong), it looks like you're just feeding review comments straight to your LLMs and pushing the updates it spits out in reply (at least squashed and not letting it push commits directly on top of your PR...) Such PRs violate our AI policy. (If it were enough to point an LLM at the issue and say "fix pls", we would have done so ourselves...)
> 
> Specifically, it is not very respectful to not reply (yourself!) to maintainer comments. If you can't explain the changes _and how they address the issues raised in the comment_, we can't really trust the changes. (If you're not ready to answer and need more time, mark the PR as "draft" until you are.)

@clason Thank you for the feedback. No disrespect was intended. The PR has already been moved to draft. The list of models has also been updated.

clason:
> And what did you do?

> The list of models has also been updated.

Ok, this is such a flagrant violation of item 2 of our AI policy, you have earned yourself a (temporary) ban from this project.

PR #5801, opened 2026-07-28
fix(parser): avoid wide error nodes on unparseable input

### LINKED ISSUE #52390: Slow tree-sitter parsing on broken code
### Reproduction steps

1. Generate a file with broken sql 
```sh
$ (echo ""; for i in $(seq 1 9999); do echo "('$i'),"; done; echo "('10000')"; echo ') as x(ix);') > test_broken.sql
```
2. `zed test_broken.sql`

### Current vs. Expected behavior

Try scrolling down or editing the `test_broken.sql`, it's really slow because of the main thread being stalled.

I've fixed this same issue in gpui-component, the problem there was that tree-sitter parses the broken file into
```
[program][1, 0] - [10003, 0]
  [ERROR] [1, 0] - [10001, 11]
    [ERROR] [1, 1] - [1, 6]
    [ERROR] [2, 1] - [10001, 6]
      [keyword_as] [10001, 2] - [10001, 4]
```

So tree-sitter will try to iterate over all the nodes for highlighting on every render, even though they are outside of the visible range. I haven't looked at how Zed sets things up, but I solved it with trying to skip subtrees that are outside of the visible range.
I'm no tree-sitter expert though, perhaps there's a better way to detect this and bail early?

The broken sql thing is quite common for me, e.g. copying a big insert statement, and then editing the insert part into like `SELECT * FROM (VALUES`. Zed probably runs into the same issue with other languages when a large part of code becomes "broken", like a large array of any language. This is just a way to easily reproduce the issue

### Zed version and system specs

Zed: v0.228.0+stable.203.8421009ef8a022df1196d54bb42fd94366ec0988 (Zed) 
OS: Linux Wayland ubuntu 25.10
Memory: 60.7 GiB
Architecture: x86_64
GPU: AMD Radeon RX 5700 XT (RADV NAVI10) || radv || Mesa 25.2.8-0ubuntu0.25.10.1

### Attach Zed log file

<details><summary>Zed.log</summary>

```log

```

</details>

### Relevant Zed settings

<details><summary>settings.json</summary>

```json

```

</details>

### Relevant Keymap

<details><summary>keymap.json</summary>

```json

```

</details>

### (for AI issues) Model provider details

_No response_

### If you are using WSL on Windows, what flavor of Linux are you using?

None

timvermeulen:
Changes error recovery to nest the error children found so far in an `_ERROR` node rather than combining them all in a long flat list, which
- makes error recovery linear in the number of error children instead of quadratic
- ultimately results in a balanced tree of error nodes which can make queries much faster

This outright fixes https://github.com/zed-industries/zed/issues/52390 which is what lead me to investigate this in the first place. Parsing that pathological example went from 2255ms to 77ms with this change, and querying from 227ms to 0.2ms.

### AI Policy

Opus 5 to explore what was causing these wide error nodes, and to write a first version of this change.

### DIFF
--- a/lib/src/parser.c
+++ b/lib/src/parser.c
@@ -1250,10 +1250,17 @@ static bool ts_parser__recover_to_state(
       Subtree error_tree = *array_get(&error_trees, 0);
       uint32_t error_child_count = ts_subtree_child_count(error_tree);
       if (error_child_count > 0) {
-        array_splice(&slice.subtrees, 0, 0, error_child_count, ts_subtree_children(error_tree));
+        SubtreeArray nested = array_new();
+        array_reserve(&nested, error_child_count);
         for (unsigned j = 0; j < error_child_count; j++) {
-          ts_subtree_retain(*array_get(&slice.subtrees, j));
+          Subtree child = ts_subtree_children(error_tree)[j];
+          ts_subtree_retain(child);
+          array_push(&nested, child);
         }
+        Subtree nested_error = ts_subtree_from_mut(ts_subtree_new_node(
+          ts_builtin_sym_error_repeat, &nested, 0, self->language
+        ));
+        array_insert(&slice.subtrees, 0, nested_error);
       }
       ts_subtree_array_delete(&self->tree_pool, &error_trees);
     }
--- a/lib/src/subtree.c
+++ b/lib/src/subtree.c
@@ -335,6 +335,15 @@ void ts_subtree_compress(
   }
 }
 
+// The part of an error node's cost that penalizes the extent it spans, as
+// opposed to the cost of its contents.
+static inline uint32_t ts_subtree__error_extent_cost(Length size) {
+  return
+    ERROR_COST_PER_RECOVERY +
+    ERROR_COST_PER_SKIPPED_CHAR * size.bytes +
+    ERROR_COST_PER_SKIPPED_LINE * size.extent.row;
+}
+
 // Assign all of the node's properties that depend on its children.
 void ts_subtree_summarize_children(
   MutableSubtree self,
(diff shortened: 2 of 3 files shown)


clason:
Can this be backported (or can the fix be extracted for `release-0.26`)?

Do you have proper benchmarks for the performance improvement (both for the "no error" and the "error" case)? (Anecdata is nice but can often miss the full picture ;))

timvermeulen:
Yes, there's no conflicts with `release-0.26`. Here's the error benchmark results:

| language | normal master | normal PR | Δ | error master | error PR | Δ |
|---|---:|---:|---:|---:|---:|---:|
| bash | 9346 | 10072 | +7.8% | 8362 | 8514 | +1.8% |
| c | 10843 | 10714 | -1.2% | 1770 | 1778 | +0.5% |
| cpp | 8494 | 8476 | -0.2% | 2191 | 2210 | +0.9% |
| embedded-template |  | |  | 98257 | 100600 | +2.4% |
| go | 11940 | 11930 | -0.1% | 1578 | 1626 | +3.0% |
| html | 242 | 241 | -0.2% | 18794 | 18601 | -1.0% |
| java |  | |  | 2820 | 2908 | +3.2% |
| javascript | 13402 | 13916 | +3.8% | 2204 | 2192 | -0.5% |
| jsdoc |  | |  | 24452 | 24986 | +2.2% |
| json |  | |  | 22145 | 22554 | +1.8% |
| php |  | |  | 95276 | 94818 | -0.5% |
| php_only |  | |  | 1578 | 1574 | -0.3% |
| python | 8167 | 8074 | -1.1% | 1460 | 1448 | -0.8% |
| ruby |  | |  | 2074 | 2126 | +2.5% |
| rust | 11368 | 11529 | +1.4% | 1372 | 1384 | +0.9% |
| tsx |  | |  | 2786 | 2791 | +0.2% |
| typescript |  | |  | 2380 | 2389 | +0.4% |
| **OVERALL** | 9085 | 9308 | +2.4% | 17772 | 17973 | +1.1% |

This is averaged over many runs so I'm not entirely sure why it still shows an improvement for the no error case. Either way, the numbers in the PR description were obviously from an extremely contrived example, though a few individual error cases benefit significantly from this change as well (again averaged over many runs):

```
java/deeply-nested.html         +78.7%
java/deeply-nested-custom.html  +68.1%
jsdoc/proc.go                   +47.4%
go/clean-old.sh                 +43.1%
jsdoc/ast.rs                    +24.8%
```

jqsdy5mvph-creator (approved):

jqsdy5mvph-creator (approved):

clason:
Ok, I looked at this a bit, and I have the following minor concerns:
1. Commits should be atomic; the `refactor` one is pointless in itself, so please squash it into the one where it's actually used.
2. What precisely does the `fix` commit fix? Can you link an issue and/or add a regression test? Ideally, this should be a separate PR so we can backport it.
3. Taking the `fix` commit as a basis, what is the performance delta for the `perf` commit in itself?

(Given that this PR addresses some catastrophic slowdowns in specific situations while keeping the base performance roughly similar (and even regressing slightly), I would classify this as a `fix`, not a `perf`.)

timvermeulen:
> * Commits should be atomic; the `refactor` one is pointless in itself, so please squash it into the one where it's actually used.

👍 

> * What precisely does the `fix` commit fix? Can you link an issue and/or add a regression test? Ideally, this should be a separate PR so we can backport it.

It fixes an error cost calculation bug: `error_repeat` children were skipped when calculating their parent's cost to avoid double counting their extent penalty, which usually results in the correct cost, but not necessarily when they contain nested `_ERROR` nodes (the cost of which would be skipped). I included that fix here because the other commit makes nesting such `_ERROR` nodes a very common occurrence, and without that fix it would cause these now nested trees of errors to have most of their cost discarded when compared to their original flat list version. Let me know if you'd still like me to make a separate PR for it, though!

> * Taking the `fix` commit as a basis, what is the performance delta for the `perf` commit in itself?

I did more benchmarking, and tbh for both the `fix` commit vs `master`, as well as for the `perf` commit vs `fix`, I didn't get any per-language score change that exceeded the noise when comparing `master` with itself. Only for some individual error cases did I get some statistically significant scores:
```
go/clean-old.sh +74.9%           // fix
java/deeply-nested.html +493%    // perf
deeply-nested-custom.html +282%  // perf
jsdoc/proc.go +131%              // perf
```

> * (Given that this PR addresses some catastrophic slowdowns in specific situations while keeping the base performance roughly similar (and even regressing slightly), I would classify this as a `fix`, not a `perf`.)

👍

WillLillis (approved):
Thanks!

PR #5805, opened 2026-07-30
fix(generate): keep skipped extras out of token start positions

chereshnyuk:
### The problem

A token can swallow the whitespace in front of it. States with compatible token sets share a lexer, so if any one of them has a token that can match a separator character. That lexer consumes the space instead of skipping it and the token start stays put.

```console
$ cat example.nu
error make {msg: "foo"}
```

**Before:**
```console
$ tree-sitter parse example.nu --cst
```
```diff
  0:0  - 1:0    nu_script
  0:0  - 0:23     ctrl_error
  0:0  - 0:5        "error"
- 0:5  - 0:10       "make"
```

**After:**
```console
$ tree-sitter parse example.nu --cst
```
```diff
  0:0  - 1:0    nu_script
  0:0  - 0:23     ctrl_error
  0:0  - 0:5        "error"
+ 0:6  - 0:10       "make"
```

### The solution

Both passes now keep token sets apart when they disagree on which separator characters their tokens can consume. 

### AI Policy

- [ x] I have read the [AI Policy](https://tree-sitter.github.io/tree-sitter/6-contributing.html#ai-policy) and this PR complies with it.
- [ x] If AI tools were used: I have disclosed the tool and extent of usage below.

Sonnet 5 wrote the edge and corner tests.
Opus 5 validated the solution.


clason:
Be aware that there's a massive refactor of the generation code going on behind the scenes, so it's probably better to park this until that has landed (because I predict massive conflicts).

Also, this PR seems to be pretty large given the functional change (as expected from LLMs...)

chereshnyuk:
Sounds good, parking it until the refactor lands makes sense. No rush on our side since this has been running in PROD for a while. Most of the diff is tests. Happy to take another pass and trim it so the functional change reads cleaner.

chereshnyuk:
@clason Trimmed it down. The test coverage was excessive so the redundant cases are gone. The functional change should read cleanly now. Waiting on the refactor before moving this forward.

WillLillis:
Unfortunately, this issue isn't isolated to token set merging. Using the `token_with_separator_prefix` corpus grammar from this PR, an input of `   error make {}` (3 leading spaces) still reports incorrect ranges:

image

You can even completely disable merging in the generate pipeline and this bug still persists. The fix requires both a generate side patch (with more care regarding performance, ideally) as well as a patch to the C runtime with an ABI bump. Such a fix will also have to be built on top of the dod-ify branch (#5811). I have an in-progress patch I'm working on locally, and plan to move forward with that, as it will require coordination with other potential ABI-breaking changes.

I appreciate your help in determining the root cause for part of this class of bug 🙂

PR #5809, opened 2026-08-02
Eliminate recursion in node traversal and dot-graph output

### LINKED ISSUE #5807: Unbounded recursion on nested parse may cause crashes
### Problem

Four functions in the core C library recurse once per level of tree nesting.
A tiny, deeply nested document will overflow the stack and crash.

The functions are: `ts_subtree__print_dot_graph`, `ts_node_child_with_descendant`, `ts_subtree_has_trailing_empty_descendant`, `ts_node_child_by_field_id`.

`ts_node_string` has already been fixed in #5622.

For default stack depth of 8MiB this is unlikely,
but my search script found four real-world examples that have 2-3k nesting levels:

* 3333 levels in [intel-ucode-defs.h](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/microcode/intel-ucode-defs.h)
* 2819 levels in [_win32.cpp](https://github.com/vosen/ZLUDA/blob/master/ext/detours/samples/traceapi/_win32.cpp)
* 2099 levels in [test_bpf.c](https://github.com/torvalds/linux/blob/master/lib/test_bpf.c)

### Steps to reproduce

Build&run:

```rust
use std::fs::File;
use tree_sitter::{Parser, Language};

fn main() {
    let language = Language::new(tree_sitter_javascript::LANGUAGE);
    parser.set_language(&language).unwrap();

    // ~200 KB of nested parentheses.
    let depth = 100_000;
    let source = format!("{}1{}", "(".repeat(depth), ")".repeat(depth));
    let tree = parser.parse(&source, None).unwrap();

    // stack overflow in print_dot_graph
    let file = File::create("/dev/null").unwrap();
    tree.print_dot_graph(&file);

}
```

### Expected behavior

Printing the dot graph of any document should complete without overflowing the stack.

master as of 2026-08-02 (`963b5a5`), version 0.27.0. Present in v0.26.11 as
well.

### Operating system/version

Linux (kernel 7.0.0-28-generic, x86_64).

### Disclosure

This issue was written by human based on LLM-assisted PR.
Numbers and links were generated by scripts generated by AI.

mgajda:
### Summary

Replaces unbounded recursion by iteration in four functions to prevent
potential crashes on automatically generated deeply nested input.

- `ts_subtree__print_dot_graph`
- `ts_node_child_with_descendant`
- `ts_subtree_has_trailing_empty_descendant`
- `ts_node_child_by_field_id`

### Verification

- New test to verify crash was removed after change:
  `test_deeply_nested_dot_graph_does_not_overflow_stack`.
- Full `tree-sitter-cli` test suite passes (302 tests).
- `Node::to_string` and `Tree::print_dot_graph` output is byte-identical to
  before on nested and balanced trees.
- `valgrind` was used to confirm the memory occupancy did not change.

### Related

Closes #5807.

### AI Policy

This change was produced with the assistance of an AI coding agent (DeepSeek V4 Flash
managed by OpenCode). Over several iterations, agent was instructed through several iterations to diagnose the crash on a test suite, ensure to fix unbounded recursion, and propose a unit test.

After initial PR was prepared, and reviewed it and proceeded with several iterations
of simplification and redundancy elimination.

I ran the full test suite manually (including crash-reproduction).

I take responsibility for the change.

### DIFF
--- a/crates/cli/src/tests/node_test.rs
+++ b/crates/cli/src/tests/node_test.rs
@@ -1248,3 +1248,55 @@ fn parse_json_example() -> Tree {
     parser.set_language(&get_language("json")).unwrap();
     parser.parse(JSON_EXAMPLE, None).unwrap()
 }
+
+/// A tiny source of deeply nested parenthesized expressions.
+fn deeply_nested_source(depth: usize) -> String {
+    let mut source = "(".repeat(depth);
+    source.push('1');
+    source.push_str(&")".repeat(depth));
+    source
+}
+
+/// `Node::to_string` calls `ts_node_string`. It used to recurse once per
+/// level of nesting. A 100k-deep document crashed with a C stack overflow.
+/// The stringifier now walks the tree with an explicit stack. This input
+/// crashes the old code. It passes with the new one.
+#[test]
+fn test_deeply_nested_node_to_string_does_not_overflow_stack() {
+    let depth = 100_000;
+    let mut parser = Parser::new();
+    parser.set_language(&get_language("javascript")).unwrap();
+    let tree = parser.parse(deeply_nested_source(depth), None).unwrap();
+
+    let string = tree.root_node().to_string();
+    assert!(string.starts_with(
+        "(program (expression_statement (parenthesized_expression (parenthesized_expression"
+    ));
+    assert_eq!(string.matches('(').count(), string.matches(')').count());
+}
+
+/// `Tree::print_dot_graph` calls `ts_tree_print_dot_graph`. It used to
+/// recurse once per node. A deeply nested document crashed with a C stack
+/// overflow. It now walks the tree with an explicit stack.
+#[test]
+fn test_deeply_nested_dot_graph_does_not_overflow_stack() {
+    let depth = 100_000;
+    let mut parser = Parser::new();
+    parser.set_language(&get_language("javascript")).unwrap();
+    let tree = parser.parse(deeply_nested_source(depth), None).unwrap();
+
+    let path =
+        std::env::temp_dir().join(format!("tree-sitter-deep-dot-{}.dot", std::process::id()));
+    {
+        let file = std::fs::File::create(&path).unwrap();
+        tree.print_dot_graph(&file);
+    }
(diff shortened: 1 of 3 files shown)


WillLillis:
This is a large change to several functions in the the core library all lumped together. While the recursive to iterative transformation is desired, this level of change requires a lot of care. To make this kind of change, ideally, this PR would have:

- A performance analysis both for the functions in isolation as well as their use in other code paths (i.e. the query engine)
- Any discussion of approach, decisions made, etc.
- (Optionally) a real world failure motivating this change

If you're open to doing the leg work necessary, feel free to reopen. In the meantime, without any motivating failure (outside of tests specifically designed to trigger this failure), I'd prefer not to essentially redo all of the work necessary to ensure this PR is ready to merge.

mgajda:
@WillLillis Can you please confirm to ensure we are on the same page:
1. Do you want change to each function implemented as a separate PR.
2. Do you need to see and read every design choice, which did not seem to be apparent.

Note that #5807 contain examples from. While these use 6x less stack than current Ubuntu 26.04 default limit (8MiB), I fully expect that a similar code could grow 6x in nesting.

WillLillis:
> @WillLillis Can you please confirm to ensure we are on the same page:
> 
> 1. Do you want change to each function implemented as a separate PR.

Having each change in the same PR is fine. Ideally though, each change _would_ be kept in a separate, buildable commit. This makes reviewing much easier, and helps us when bisecting issues later on.

> 2. Do you need to see and read every design choice, which did not seem to be apparent.

Some judgement is required here, but some discussion is warranted here. For changes to the C library like this, we need more than the typical LLM bullet list of "tests pass" + "output is byte identical".

> Note that #5807 contain examples from. While these use 6x less stack than current Ubuntu 26.04 default limit (8MiB), I fully expect that a similar code could grow 6x in nesting.

Yes, I understand that using an iterative approach uses less stack space than a recursive approach when presented with the handpicked examples from your issue. The problem is _most_ uses of these functions aren't adversarially-nested inputs. We need to make sure the common case is still reasonably performant. This is why we need evidence of actual measurements made and the methodology used.

PR #5811, opened 2026-08-03
perf(generate): embrace data oriented design

WillLillis:
This is a change I've been itching to try out for awhile since observing the benefits of an arena/pool-based rule representation during work on another feature. I've alluded to this other feature several times at this point, and I promise it's coming soon(ish!). Data-oriented design is a great programming paradigm that I'm still learning a lot about. A few great references for those who are interested:

- https://www.youtube.com/watch?v=rX0ItVEVjHc (great foundational DoD talk from Mike Acton)
- https://www.youtube.com/watch?v=IroPQ150F6c (excellent talk from Andrew Kelley that builds on Mike Acton's talk)
- https://floooh.github.io/2018/06/17/handles-vs-pointers.html (relevant blog post on arena-based data structures from [Andre Weissflog](https://github.com/floooh))

## Impact

The big picture impact for those uninterested in the details: `tree-sitter generate` is now significantly faster and uses much less memory. All `parser.c` files should be identical to those emitted by `master` today. Running across the nvim-treesitter corpus, `tree-sitter generate` is roughly 2.7x faster and uses 4.2x less memory. The largest/"heaviest" grammars benefit the most. A few outliers:

- tree-sitter-tlaplus −75.0% wall / −90.7% RSS (5.44GB -> 507MB), biggest wall win
- tree-sitter-nim −69.6% / −81.0% (7.01GB -> 1.33GB), biggest absolute RSS win
- tree-sitter-kotlin −69.9% / −84.8%
- tree-sitter-systemverilog −72.8% / −82.7%
- tree-sitter-ada 85.0s -> 31.1s (biggest grammar by wall time)
- tree-sitter-sql 5.93GB -> 1.44GB 
- tree-sitter-cpp −66.7% / −80.4%

Note that I'm currently working off of a much slower laptop, so these numbers will appear skewed compared to the figures stated in my previous performance PRs. The deltas are the important bits here

This should help tighten the development loop for grammar authors, and also greatly aid in heavy CI workflows (i.e. nvim-treesitter). Note that for nvim-treesitter's case, it should benefit both from faster generation times as well as increased parallelism unlocked by the lower per-grammar memory usage. 

### The "big" ideas

- If you have a bunch of something, make it as small as possible (a few caveats here, sometimes smaller representations raise access cost, which can outweigh the cache-friendliness savings)
- Precompute things into a cheap map when it matters
- Holding an index into a large backing vector is often much better than holding a pointer

### `Rule` rewrite

The current grammar IR is a recursive `Rule` tree. Recursive indirection is handled via heap allocations (`Vec`s and `Box`es). In addition, each `Rule` holds owned `String`s for every symbol name, pattern, and field. Each node is its own heap allocation, each `prepare_grammar` pass _rebuilds_ the tree, and the same names get re-allocated as `Strings` over and over.

 This branch replaces that with a flat, pooled arena. Rule becomes a 12-byte `Copy` enum that holds children as `RuleId` or `RuleIdRange` indices:

```rust
pub enum Rule {
    Blank,
    String(StrId),
    Pattern(StrId, StrId),
    NamedSymbol(StrId),
    Sym { kind: SymbolType, index: u32 },
    Seq(RuleIdRange),
    Choice(RuleIdRange),
    Repeat(RuleId),
    Metadata { params: ParamsId, rule: RuleId },
    Reserved { rule: RuleId, ctx: StrId },
}
```

where `StrId` is a newtype wrapper of a `NonZeroU32`, `RuleId` a newtype wrapper of a `u32`, and `RuleIdRange` a struct holding a `start` and `len` (both `u32`s).

All nodes live in one RulePool, where storage is append only. Each pass in `prepare_grammar` rewrite nodes in place and orphan the subtrees they replace. Thin metadata wrappers are given out of each pass, referencing the underlying pool.

### `Production` rewrite

After `prepare_grammar`, the grammar compiles into a `SyntaxGrammar` of variables -> productions -> steps. On master that's a three-level nested `Vec`:

```rust
Vec<
    SyntaxVariable { 
        productions: Vec<
            Production {
                steps: Vec<ProductionStep>
            }
        > 
    }
>
```
This deeply-nested representation is then passed to `build_tables`, which spends a lot of time pointer-chasing through during LR item set construction (the hottest phase of generation).

 This branch flattens it into range-based arenas on `SyntaxGrammar`:

```rust
    pub steps:  Vec<ProductionStep>,     // single backing store of all steps
    pub productions: Vec<Production>,    // each = (steps_start, steps_len, dyn_prec) into `steps`
    pub var_prods:  Vec<(u32, u32)>,     // each variable's [start, end) production range
```

Access is via production(id) -> ProdRef and variable_prod_ids(var) -> Range. `ProductionStep` is a 20-byte Copy struct that packs kind / precedence / associativity / alias-named into a single flags byte. Item-set construction now walks contiguous slices instead of three levels of Vec indirection.

### `String` pool

 - String interning (`StrPool`). Every grammar-level name/pattern/field is interned to a `StrId`. Owned `Strings` survive only at the serialization boundary. This helps shrink the size of various `Rule` nodes, and makes string equality checks a single integer comparison (`StrIds` are equal iff the corresponding pooled `Strings` are equal)

### Commit by commit walkthrough

Each commit message does a decent job explaining the change and performance impact, but I've added some brief summaries (and some sequencing details) here for convenience:

- Precompute order-preserving u32 item identity keys:
	- Performance neutral by itself, _sets up_ the following commit
- Flat pooled RulePool + StrPool + flat production storage:
	- Builds on the previous commit. 10% faster overall, but notably makes `prepare_grammar` ~70% faster in isolation. This will aid greatly in future "interactive" uses of the `generate` pipeline.
- `Symbol::index` usize→u32 (16→8B):
	- 2-7% wall time, 7-10% peak rss
- Coincident index: without-word bitset vs per-pair `Vec<StateId>`:
    - 7% wall time on small grammars, 20-25% wall time on large grammars, 20-50% peak rss
 - Stop cloning kernel item sets into parse_state_info:
    - 1-3% wall time and peak rss
 - Intern lookahead `TokenSets` via `LookaheadSetPool`:
    - 10-15% walltime, 3-5% peak rss
 - Memoize `get_production_id`:
    - 0-3% walltime, neutral peak rss
- Inline single-action `ActionList`:
    - greater than 20% wall time, 30% peak rss
- Memoize get_auxiliary_node_info per state:
    - 1-4% wall time  
- Route lex ids through the `LexStateId` typedef:
    - cleanup, sets up the next commit
- Shrink parse/lex/prod ids `usize`->`u32`:
    - 2-5% wall time, 9-12% peak rss
- Pool parse-table action lists via `ActionListId`:
    - 7% wall time, 12% peak rss

### Other considerations

There's no rush to get this in, I'll spend some time reading back over the diff to see if I missed anything. With the big part of this rework done, though, I'll have a lot more free time to review other PRs and work on bug fixes. Really appreciate everyone's patience on this front. 🙂 

### AI Policy

I usually just leave a quick blurb here, but I think the branch justifies a larger explanation.

When Fable was first released a few weeks ago, I directed it to implement the pool-based `Rule` representation. I let it work mostly unattended, and it failed somewhat horrendously (an even larger diff than this PR, and at least 20% slower than master).

After that result, I started with a clean slate and coached it through the rough shape of the changes over a few days As protype scaffolding, I built up conversion utilities between the pool-based `Rule` representation and master's form. Between each pass, the pool-based transformation was "materialized" into master's repr and checked for correctness. This pass-by-pass checking was necessary for the prototype, where there were more unknowns. For my own transcription (see below), I instead leaned on the nvim-treesitter corpus.

Once the pool-based prototype was done, the prototype branch moved onto `build_tables`, looking for both new optimizations unlocked by our new data layout, as well as old wins that I had missed in the pass. A combination of fable's and I's ideas landed (each change was benchmarked against the repo's corpus grammars, and later the nvim-treesitter corpus).

Once the prototype branch was in a somewhat decent state, I then moved over to Opus 4.8 (and later Opus 5). I used these lighter models to walk through the prototype branch's high level state, settle on ordering, and other logistics. With sequencing settled, I used the Opus models to walk through the prototype branch and transcribe the changes by hand to a new branch. With such a large diff, I didn't feel comfortable PRing without typing (and arguing with the models about) every line myself. Using this approach, I found several more performance boosts, cleanups, and other improvements off of the prototype branch. I also found a few things from the prototype branch to be unacceptable during my transcription, and used gpt5.6-sol default to refine redesign passes.

### DIFF
--- a/.github/workflows/nvim_ts.yml
+++ b/.github/workflows/nvim_ts.yml
@@ -73,7 +73,7 @@ jobs:
 
       - if: matrix.type == 'generate'
         name: Generate and compile parsers
-        run: $NVIM -l ./scripts/install-parsers.lua --generate --max-jobs=3
+        run: $NVIM -l ./scripts/install-parsers.lua --generate --max-jobs=10
         working-directory: ${{ env.NVIM_TS_DIR }}
         shell: bash
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ similar_names            = "allow"
 string_lit_as_bytes      = "allow"
 struct_excessive_bools   = "allow"
 too_many_lines           = "allow"
+tuple_array_conversions  = "allow"
 
 [workspace.lints.rust]
 mismatched_lifetime_syntaxes = "allow"
--- a/crates/generate/src/build_tables.rs
+++ b/crates/generate/src/build_tables.rs
@@ -18,6 +18,7 @@ use self::{
     build_lex_table::build_lex_table,
     build_parse_table::{ParseStateInfo, build_parse_table},
     coincident_tokens::CoincidentTokenIndex,
+    item::ItemKeyMap,
     item_set_builder::ParseItemSetBuilder,
     minimize_parse_table::minimize_parse_table,
     token_conflicts::TokenConflictMap,
@@ -52,7 +53,9 @@ pub fn build_tables(
     optimizations: OptLevel,
     diagnostics: &mut Vec<Diagnostic>,
 ) -> BuildTableResult<Tables> {
-    let item_set_builder = ParseItemSetBuilder::new(syntax_grammar, lexical_grammar, inlines);
+    let item_key_map = ItemKeyMap::new(syntax_grammar, inlines);
+    let item_set_builder =
+        ParseItemSetBuilder::new(syntax_grammar, lexical_grammar, inlines, &item_key_map);
     let following_tokens =
         get_following_tokens(syntax_grammar, lexical_grammar, inlines, &item_set_builder);
     let (mut parse_table, parse_state_info) = build_parse_table(
--- a/crates/generate/src/build_tables.rs
+++ b/crates/generate/src/build_tables.rs
@@ -30,7 +30,7 @@ use crate::{
     node_types::VariableInfo,
     rules::{AliasMap, Symbol, SymbolType, TokenSet},
     strpool::StrPool,
-    tables::{LexTable, ParseAction, ParseTable, ParseTableEntry},
+    tables::{ActionList, LexTable, ParseAction, ParseTable, ParseTableEntry},
 };
 
 pub struct Tables {
@@ -204,7 +204,7 @@ fn populate_error_state(
 
     let recover_entry = ParseTableEntry {
         reusable: false,
-        actions: vec![ParseAction::Recover],
+        actions: ActionList::One(ParseAction::Recover),
     };
 
     // Exclude from the error-recovery state any token that conflicts with one of
--- a/crates/generate/src/build_tables.rs
+++ b/crates/generate/src/build_tables.rs
@@ -29,6 +29,7 @@ use crate::{
     nfa::{CharacterSet, NfaCursor},
     node_types::VariableInfo,
     rules::{AliasMap, Symbol, SymbolType, TokenSet},
+    strpool::StrPool,
     tables::{LexTable, ParseAction, ParseTable, ParseTableEntry},
 };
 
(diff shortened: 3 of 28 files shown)


clason (approved):

clason:
Can you try raising the `max_jobs` to 10 here? https://github.com/tree-sitter/tree-sitter/blob/308aee0c90b0527b185b3100046781d69935c5c5/.github/workflows/nvim_ts.yml#L76

(Ideally, we could drop the limit altogether, but then GH can't handle the parallel network load.)

maxbrunsfeld:
This is fantastic. Great explanation. Seems very straightforward, in terms of not adding much complexity, but hugely impactful.

maxbrunsfeld (approved):

WillLillis:
There's a few minor changes I'd like to push up this weekend but after that it should be good to go!

WillLillis:
> Can you try raising the `max_jobs` to 10 here?
> 
> https://github.com/tree-sitter/tree-sitter/blob/308aee0c90b0527b185b3100046781d69935c5c5/.github/workflows/nvim_ts.yml#L76
> 
> (Ideally, we could drop the limit altogether, but then GH can't handle the parallel network load.)

Ran clean! It doesn't look like it sped up the workflow much, but a lot of that could be download time variability/other factors.

clason:
> It doesn't look like it sped up the workflow much

Oh, it most certainly did! The generate workflow is now as fast as the build workflow, which is a huge improvement!

maxbrunsfeld:
Awesome job.

WillLillis:
Thanks! Really appreciate you taking a look 🙂

PR #5816, opened 2026-08-04
build(deps): bump wasmtime to v36.0.13

clason:


WillLillis:
Don't we need to bump the versions/hashes in build.zig.zon as well for this?

clason:
But that always takes ages... 😭

WillLillis (approved):

Issue #5818, opened 2026-08-05
Query behaves different on two identical trees


Newosko:
### AI Policy

- [x] I have read the AI Policy and this issue complies with it.

### Problem

Very simple reproduction below (3 rule grammar showing issue).

(from repro repo)

Two identical trees
> tree-sitter generate && tree-sitter parse input.txt
```
(source_file [0, 0] - [3, 0]
  (trailing_sep [0, 0] - [0, 7]
    (name [0, 2] - [0, 3])
    (name [0, 4] - [0, 5])
    (name [0, 6] - [0, 7]))
  (leading_sep [1, 0] - [1, 7]
    (name [1, 2] - [1, 3])
    (name [1, 4] - [1, 5])
    (name [1, 6] - [1, 7]))
  (trailing_named [2, 0] - [2, 9]
    (name [2, 2] - [2, 3])
    (tag [2, 3] - [2, 4])
    (name [2, 5] - [2, 6])
    (tag [2, 6] - [2, 7])
    (name [2, 8] - [2, 9])))
```

Three query's
```
(trailing_sep (name) @bug .)
(leading_sep (name) @correct1 .)
(trailing_named (name) @correct2 .)
```

Running 
> tree-sitter generate && tree-sitter query query.scm input.txt
```
input.txt
  pattern: 0
    capture: 0 - bug, start: (0, 2), end: (0, 3), text: `a`
  pattern: 0
    capture: 0 - bug, start: (0, 4), end: (0, 5), text: `b`
  pattern: 0
    capture: 0 - bug, start: (0, 6), end: (0, 7), text: `c`
  pattern: 1
    capture: 1 - correct1, start: (1, 6), end: (1, 7), text: `c`
  pattern: 2
    capture: 2 - correct2, start: (2, 8), end: (2, 9), text: `c`
```

As you can see the two identical paterns on two identical trees don't behave consistently

### Steps to reproduce

I made the simplest possible reproduction in
```
git clone https://github.com/Newosko/tree-sitter-bug-repro
cd tree-sitter-bug-repro
make
```

### Expected behavior

same public visible tree + same query => same matches for query

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.26.10

### Operating system/version

macOS/26.5.2

WillLillis:
Fixed by #5824

PR #5819, opened 2026-08-05
Fix parsing inconsistency #5818

Newosko:
### AI Policy

- [ X] I have read the [AI Policy](https://tree-sitter.github.io/tree-sitter/6-contributing.html#ai-policy) and this PR complies with it.
- [ X] If AI tools were used: I have disclosed the tool and extent of usage below.

AI tool (claude code) was used for initial root cause + produce very simple reproduction of the issue.
https://github.com/Newosko/tree-sitter-bug-repro

### Summary

Two identical trees (using tree-sitter parse) + two identical queries give different matches, which is not expected.
This PR solves https://github.com/tree-sitter/tree-sitter/issues/5818

The block that computes both has_later_siblings and has_later_named_siblings was gated on has_later_siblings being unknown.

An anonymous . sets has_later_siblings but not has_later_named_siblings.
=> has_later_named_siblings is not set even when later named siblings exist one level up.

Gating on has_later_named_siblings instead lets the walk keep going until the value the anchor needs is resolved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


WillLillis:
Hmmm, looks like I can't push to this PR because you opened it from your fork's master branch. Could you re-open it from a different branch name? The fix is correct, I'd just like to add some test cases and do some cleanup in the area. 🙂

Newosko:
Hey @WillLillis , thanks for taking a look! I made a new PR using feature breanches:
https://github.com/tree-sitter/tree-sitter/pull/5824

PR #5820, opened 2026-08-06
Expose `OptLevel` on `generate_parser_for_grammar`

jaboatman:
As per https://github.com/tree-sitter/tree-sitter/issues/4711, I need the ability to disable this optimization. Previously this function had them disabled by default, but it was (sensibly) updated to enable them by default.

### AI Policy


WillLillis (approved):
Seems reasonable, thanks! I squashed your two commits and marked it as breaking. We can merge this once GH's outage ends and we get CI green.

PR #5824, opened 2026-08-07
fix(parser): fix parsing bug repeat operator

Newosko:
Re-open of https://github.com/tree-sitter/tree-sitter/pull/5819 using feature branch

Two identical trees (using tree-sitter parse) + two identical queries give different matches, which is not expected.
This PR solves https://github.com/tree-sitter/tree-sitter/issues/5818

The block that computes both has_later_siblings and has_later_named_siblings was gated on has_later_siblings being unknown.

An anonymous . sets has_later_siblings but not has_later_named_siblings.
=> has_later_named_siblings is not set even when later named siblings exist one level up.

Gating on has_later_named_siblings instead lets the walk keep going until the value the anchor needs is resolved.

# AI Policy
[ X] I have read the [AI Policy](https://tree-sitter.github.io/tree-sitter/6-contributing.html#ai-policy) and this PR complies with it.
[ X] If AI tools were used: I have disclosed the tool and extent of usage below.
AI tool (claude code) was used for initial root cause + produce very simple reproduction of the issue.
https://github.com/Newosko/tree-sitter-bug-repro

### DIFF
--- a/crates/cli/src/tests/query_test.rs
+++ b/crates/cli/src/tests/query_test.rs
@@ -6485,3 +6485,64 @@ export default grammar({
 
     assert_query_matches(&language, &query, source, &[(0, vec![("tuple", "()")])]);
 }
+
+#[test]
+fn test_last_child_anchor_looks_past_hidden_repeat() {
+    let language = get_test_fixture_language("last_child_anchor_past_hidden_repeat");
+
+    let source = "T a.b.c\nL a.b.c\nN a!.b!.c\n";
+
+    let query = Query::new(
+        &language,
+        "
+        (trailing_sep (name) @last .)
+        (leading_sep (name) @last .)
+        (trailing_named (name) @last .)
+        ",
+    )
+    .unwrap();
+
+    assert_query_matches(
+        &language,
+        &query,
+        source,
+        &[
+            (0, vec![("last", "c")]),
+            (1, vec![("last", "c")]),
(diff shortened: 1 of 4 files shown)


WillLillis (approved):

WillLillis:
Thanks so much for the great reproduction and minimal fix!

WillLillis:
A quick note here, this fix is correct, but it does cause `ts_tree_cursor_current_status` to do more work in the query path. I measured this to be performance neutral to ~1% slower. If this ends up being a bigger slow down for some other queries, we can look into gating the computation of `has_later_named_siblings` here based on whether the function is called for a query step with a trailing anchor.

Newosko:
Perfect, thank you @WillLillis  for taking a look at the performance impact of the change also.

I am unfamiliar with the testing suite for this repo but maybe it makes sense to add a test for this case? (this grammar style + tailing anchor query)
This way we're sure we don't regress later

WillLillis:
> Perfect, thank you @WillLillis for taking a look at the performance impact of the change also.
> 
> I am unfamiliar with the testing suite for this repo but maybe it makes sense to add a test for this case? (this grammar style + tailing anchor query) This way we're sure we don't regress later

Performance testing is tricky. It's something we'd probably like to add some guards around, but will require a lot of design, discussion, etc. to make it useful. For now, we'll monitor the issues here (and in Neovim) for any unexpected regressions.

Issue #3966, opened 2024-11-26
Inaccurate starting point of terminal tokens, including unwanted whitespaces


blindFS:
### Problem

Related grammar:

```js
  extras: ($) => [/[ \t]/, $.comment],
  ...
  ctrl_error: ($) =>
    seq(
      "error",
      "make",
      optional($._flag),  // <--- works fine without this line
      field("error_record", $.val_record),
    ),
```

<img width="381" alt="image" src="https://github.com/user-attachments/assets/11ce70cf-e2f4-4c17-9685-b34cfd74089c">

You can see that the "make" token starts early at [0, 5] instead of [0, 9].

If I remove the line of `optional($._flag)`, it outputs correct region again.

The code of the `_flag` rule is [here](https://github.com/nushell/tree-sitter-nu/blob/74e6037383ce3a77ed6fdb8281bbd69316c723a4/grammar.js#L1286), I think it has nothing to do with whitespaces.

### Steps to reproduce

```sh
git clone https://github.com/nushell/tree-sitter-nu
cd tree-sitter-nu
tree-sitter generate
```

### Expected behavior

The "make" token should start at [0, 9] as expected.

### Tree-sitter version (tree-sitter --version)

tree-sitter 0.24.3

### Operating system/version

macOS 15.1.1

blindFS:
Another one:

![image](https://github.com/user-attachments/assets/17e1a39e-b217-4a92-9863-2bb39c1b698b)

![image](https://github.com/user-attachments/assets/a3f5dba3-5324-4a5c-a8f9-677b1dd98168)

PS. incremental parsing from `true` to `(true)` generates the correct result:

![image](https://github.com/user-attachments/assets/627f865c-649a-4725-aeba-f5b1c8080239)


[if statement rule](https://github.com/nushell/tree-sitter-nu/blob/74e6037383ce3a77ed6fdb8281bbd69316c723a4/grammar.js#L1646)

If I change the optional repeating [`token.immediate`](https://github.com/nushell/tree-sitter-nu/blob/74e6037383ce3a77ed6fdb8281bbd69316c723a4/grammar.js#L1232) suffix of `(true)` to just `PUNC().dot`, it works well again.

BobbieCarroll:
Alrigth

blindFS:
Found some work-around solutions, possibly helpful for debugging this issue.
https://github.com/nushell/tree-sitter-nu/pull/167
[...]

PR #5833, opened 2026-08-09
feat(test): allow corpus tests to use cst outputs

WillLillis:
This is already exposed for consumers via the CLI, and is a natural way to express some test expectations over the sexp form.

Also clean up some repeated logic in the internal test code, and narrow the cst rendering return type to `std::io::Result` rather than `anyhow::Result`.

This fell out as a natural extension for the work on https://github.com/tree-sitter/tree-sitter/issues/3966. While that problem has proven to be fairly involved (and will likely require an ABI bump), this is a nice internal improvement we can land now.

### AI Policy

### DIFF
--- a/crates/cli/src/fuzz.rs
+++ b/crates/cli/src/fuzz.rs
@@ -25,7 +25,7 @@ use crate::{
         random::Rand,
     },
     parse::perform_edit,
-    test::{DiffKey, TestDiff, TestEntry, TestExpectation, parse_tests, strip_sexp_fields},
+    test::{DiffKey, TestDiff, TestEntry, TestExpectation, parse_tests, render_test_output},
 };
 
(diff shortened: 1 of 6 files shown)


clason:
I wouldn't backport this as it's a behavior change? Or am I missing something (e.g., that otherwise we couldn't easily backport upcoming fixes)?

(Or is this actually a `fix`, not a `feat`?)

WillLillis:
> I wouldn't backport this as it's a behavior change? Or am I missing something (e.g., that otherwise we couldn't easily backport upcoming fixes)?
> 
> 
> 
> (Or is this actually a `fix`, not a `feat`?)

It's a behavior change (new "feature") but only for our internal rust tests. My initial thought is that this should help the backport with the fix to the issue I linked (not done yet, just to be clear), but actually we won't be packporting that because it's ABI breaking. Good catch :)

clason (approved):

PR #5837, opened 2026-08-10
refactor(highlight): minimize unsafe usage via iterator API and self-referential structs

Wybxc:
In the implementation of `tree-sitter-highlight`, several patterns employ `unsafe`, including:
- A transmute between `QueryCaptures` and `_QueryCaptures` (this is strictly unsafe because the Rust compiler makes no guarantees about the memory layout of structs without a specified `#[repr]`).
- A transmute to extend lifetimes in order to construct self-referential structs (this may be safe).

This PR refactors part of the `highlight.rs` implementation to eliminate these `unsafe` patterns as much as possible. It directly uses the streaming iterator API of `QueryCaptures` introduced in #3504, thereby avoiding the need to access its private fields, and it employs `ouroboros` to implement safe self-referential structs.

This PR does not change the underlying algorithm. However, as discussed in #3504, the way `tree-sitter-highlight` uses `QueryCaptures` is theoretically incompatible with the streaming iterator API. To preserve the existing algorithm, this PR retains some `unsafe` code to implement it within the stream iterator API.

### AI Policy

I discussed the fix plan with GPT 5.6 Sol, and all the code was written manually by me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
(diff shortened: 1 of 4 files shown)


clason:
Requesting LLM code review violates the AI policy. You can do that on your own fork.

PR #5841, opened 2026-08-10
refactor(highlight): minimize unsafe usage via iterator API and self-referential structs

Wybxc:
In the implementation of `tree-sitter-highlight`, several patterns employ `unsafe`, including:
- A transmute between `QueryCaptures` and `_QueryCaptures` (this is strictly unsafe because the Rust compiler makes no guarantees about the memory layout of structs without a specified `#[repr]`).
- A transmute to extend lifetimes in order to construct self-referential structs (this may be safe).

This PR refactors part of the `highlight.rs` implementation to eliminate these `unsafe` patterns as much as possible. It directly uses the streaming iterator API of `QueryCaptures` introduced in #3504, thereby avoiding the need to access its private fields, and it employs `ouroboros` to implement safe self-referential structs.

This PR does not change the underlying algorithm. However, as discussed in #3504, the way `tree-sitter-highlight` uses `QueryCaptures` is theoretically incompatible with the streaming iterator API. To preserve the existing algorithm, this PR retains some `unsafe` code to implement it within the stream iterator API.

### AI Policy

I discussed the fix plan with GPT 5.6 Sol, and all the code was written manually by me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -276,7 +282,7 @@ version = "4.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d012d2b9d65aca7f18f4d9878a045bc17899bba951561ba5ec3c2ba1eed9a061"
 dependencies = [
- "heck",
+ "heck 0.5.0",
  "proc-macro2",
  "quote",
  "syn 3.0.3",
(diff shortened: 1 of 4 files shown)


WillLillis:
I'm a bit hesitant to go down this path without a stronger argument besides eliminating (currently) benign instances of `unsafe`. This adds a new dependency, alters the inner workings of a crate that's largely under-tested, and lacks any measurement for the runtime cost for this change.

Ideally, an improvement for the highlight crate would get rid of the `_QueryMatch` and friends adapters and restructure to use a non-peeking algorithm so the normal streaming iterators from the lib can be exercised.

Wybxc:
> Ideally, an improvement for the highlight crate would get rid of the `_QueryMatch` and friends adapters and restructure to use a non-peeking algorithm so the normal streaming iterators from the lib can be exercised.

Does the Tree-sitter development team have any thoughts on the non-peeking algorithm?

WillLillis:
> > Ideally, an improvement for the highlight crate would get rid of the `_QueryMatch` and friends adapters and restructure to use a non-peeking algorithm so the normal streaming iterators from the lib can be exercised.
> 
> 
> 
> Does the Tree-sitter development team have any thoughts on the non-peeking algorithm?

This is still largely an open question, but it's probably going to be a large and risky refactor. There haven't been any reported bugs as a result of the unsafe usage/legacy adapters, so it hasn't been enough of a priority to look into yet.

PR #5845, opened 2026-08-12
rust: trim unused imports

WillLillis:
Importing `char` brings deprecated methods into scope in some cases. Importing `str` isn't necessary in many other cases.

Just some minor cleanup here, no behavior changes. While working on #5844, I noticed that `char::from_u32` was deprecated in favor of...`char::from_u32`. Using the non-deprecated version just requires removing the import so the function resolves to the one implemented on the builtin type, rather than the standard library's `char` module. This method is marked deprecated with TBD, meaning it's silently allowed for now. There's no reason not to clean this up when we notice it to avoid CI failures in the future. The removed `str` imports don't remove any deprecated methods, but are unnecessary.

### AI Policy

Used gpt5.6-sol to scan for other soft deprecated/unused imports after noticing the afforementioned `char` issue myself.

### DIFF
--- a/crates/cli/benches/benchmark.rs
+++ b/crates/cli/benches/benchmark.rs
@@ -2,7 +2,6 @@ use std::{
     collections::BTreeMap,
     env, fs,
     path::{Path, PathBuf},
-    str,
     sync::LazyLock,
     time::Instant,
 };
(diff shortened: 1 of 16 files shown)


clason (approved):

clason:
Doesn't backport cleanly, but doesn't seem to be "fix" anything that merits the manual effort?

(I'd like to put out a fixup release to iron out the regressions, but otherwise we should concentrate on  0.27.)

PR #5847, opened 2026-08-13
Upgrade to latest wasmtime, improve robustness and perf of wasm-based parsing

maxbrunsfeld:
This PR upgrades Tree-sitter to use Wasmtime 47.0.0, the latest stable version. It also improves error handling during Wasm loading - avoiding leaking space inside of the Wasm store's memory and function tables. It also optimizes wasm parsing a bit further, by storing lexer function handles instead of reading them from a function table on every invocation.

| Metric | Wasm before | Wasm after | Native | Wasm change | Wasm after vs native |
|---|---:|---:|---:|---:|---:|
| Average normal-code throughput | 5,089 bytes/ms | 9,708 bytes/ms | 14,872 bytes/ms | **+90.8% / 1.91×** | 65.3% |
| Worst normal-code throughput | 323 bytes/ms | 333 bytes/ms | 318 bytes/ms | **+3.0% / 1.03×** | 104.6% |
| Average error-code throughput | 4,298 bytes/ms | 11,224 bytes/ms | 27,721 bytes/ms | **+161.1% / 2.61×** | 40.5% |
| Worst error-code throughput | 21 bytes/ms | 74 bytes/ms | 216 bytes/ms | **+250.8% / 3.51×** | 34.1% |

It looks like most of the benefit came from upgrading Wasmtime, but function caching also sped things up by about 10%.

### AI Policy

All of this code was written by GPT-5.6 Sol and reviewed by me. My initial prompt was just to upgrade to the latest version of Wasmtime. Then I asked it to review `wasm.c` for correctness and look for optimization opportunities. GPT-5.5 found some error handling bugs and also a key optimization opportunity - storing function handles rather than reading them from the table on each call.


maxbrunsfeld:
The windows neovim CI job failure seems unrelated to these changes.

WillLillis (approved):
Awesome!

clason:
That is great, but could we please wait for 48.0.0 (next week), which is the [next LTS](https://docs.wasmtime.dev/stability-release.html)? Chasing monthly major releases is not sustainable, and normal releases only get (often security) fixes backported for a month or so.

(I expect that this was the hard work, so bumping to 48.0.0 from here is going to be small beans.)

This is important because we force downstream to migrate as well, requiring potentially breaking changes on their side. We _can_ merge this now on master, but we should _not_ think about making a release without an LTS.

clason:
(And nvim-treesitter failures are normal Github Thursday; thanks, LLMs.)

maxbrunsfeld:
> That is great, but could we please wait for 48.0.0 (next week), which is the [next LTS](https://docs.wasmtime.dev/stability-release.html)?

That sounds fine to me.

clason:
(I need to write this down somewhere.)

Assuming all goes well, things seem to align for a 0.27 release by end of August.

clason:
If I may make two suggestions:
1. We should add the benchmark (especially how to benchmark WASM) to the documentation (docs/src/6-contributing.md).
2. It would be good to squash the commits so they are atomic (i.e., every commit builds and passes tests), e.g., one each for adding benchmarks, bumping wasmtime (including adaptations for breaking changes), performance tweaks on top. This makes a huge difference for when we need to bisect for regressions.

maxbrunsfeld:
> It would be good to squash the commits so they are atomic (i.e., every commit builds and passes tests), e.g., one each for adding benchmarks, bumping wasmtime (including adaptations for breaking changes), performance tweaks on top. This makes a huge difference for when we need to bisect for regressions.

Thoughts on just squash-merging the entire PR?

clason:
That's certainly the lazy option, but would lose useful information.

maxbrunsfeld:
Alright; I have rebased them.

maxbrunsfeld:
Ah god sorry - I still had the merge mode set to "squash", so my rebasing was lost anyway.

clason:
Doesn't that require bumping the minimal Rust version (our Cargo.toml specifies 1.90, but wasmtime v48 raised it to 1.95)?

Also, the Dockerfile is now severely out of date -- as you are the only one who has ever touched it, do you still need it, or can we remove it here to reduce maintenance burden?)

PR #5849, opened 2026-08-13
build(swift)!: drop Package.swift

### LINKED ISSUE #5523: Xcode 26 / Swift 6.3: 'import TreeSitter' yields empty module (TSLanguage not in scope)

clason:
Problem: `Package.swift` build script is unmaintained and leads to build
errors.

Solution: Remove `Package.swift`; downstream tools should rely on the
swift-tree-sitter bindings (or, if they want a custom bare-metal build,
handle this in their own build scripts).

Closes #5523


WillLillis:
I may be misreading the Swift docs, but I believe deleting this file will cause the downstream swift-tree-sitter bindings to fail because they rely on consuming tree-sitter core as a swift package. Given that swift-tree-sitter is already broken though (judging by the discussion in https://github.com/tree-sitter/tree-sitter/pull/5524), I think it seems reasonable to remove this for now. swift-tree-sitter can either move to depending on the C-lib directly, or we can reintroduce this file when someone with the know-how has a chance to try.

WillLillis (approved):

clason:
swift-tree-sitter seems to be relatively actively but not seriously maintained (their `Package.swift` is still stuck on 0.25.5...). So that project looks like it needs a general renovation, as part of which it can (hopefully) work around the dropped file (by including what's necessary in their build scripts).

@mattmassicotte

mattmassicotte:
Deleting this package file will break downstream dependencies that try to pull in tags after this is merged. You will not break any packages that are pinned to existing tags, which SwiftTreeSitter does. I was not in favor of adding a Package.swift in the first place, so I'm looking forward to this being removed.

SwiftTreeSitter is stuck on an older version because it depends on a runtime API that was removed in more recent versions. I cannot update to 0.26+ without putting in some work, and I'm not yet sure how complex the need changes will be.

I was intending on tackling the version upgrade at the same time as switching back to a git-based dependency.

clason:
Thank you, that is very helpful (and I didn't mean to throw any shade; my wording was ill-chosen)!

And, yes, that would be a breaking change in 0.27 -- which, as a zerover minor version bump, is intended to be breaking. Your strategy of pinning to tags is absolutely the right one (and bindings should mirror library versioning exactly -- at least the minor version!) In the past, we have not been good about semantic versioning but we are taking this pretty seriously now (so patch releases are always safe to upgrade to, accidental regressions notwithstanding). 

It would be worth discussing what needs to be done to make this process more frictionless to the binding maintainers (who -- PSA -- are separate from the core team, even though they are under the same org). Various bindings are in various states of brokenness, and that is... not great. At the same time, we cannot centralize more responsibility in the core project, especially for languages nobody here is using.

clason:
(Merging this now since we all are agreed here, but this need not shut down the discussion.)

mattmassicotte:
No worries!

Why do you think it is important for bindings to match runtime versions? The runtime version *typically* only matters to the bindings internals, it seems like it should not actually make a difference to binding consumers. Or are you worried about parser ABI compatibility?

clason:
Oh, it absolutely matters to consumers due to breaking _API_ changes! (As well as other behavior changes, which may be beneficial or not.) Of course, ABI compatibility matters, too, because of upper and lower bounds.

It bears repeating explicitly: We do **not** intend different minor versions of tree-sitter (lib) to be interchangeable! Only patch versions are (meant to be) pure bug fixes and should be always safe to update. (So if you have "0.26 bindings", these should work with any 0.26.x tag without changes.)

It is therefore very important for consumers to have full control over the library (minor) version, and that means that -- if they access it through bindings -- the bindings need to pass this through. Of course, details may vary: depending on the language and the structure of the bindings, they may depend explicitly on tree-sitter in their manifest in addition to the bindings, or they may only pull in the bindings and then have to rely on _that_ pulling in the correct lib. 

So if you move back to tracking git rather than individual tags, you should follow the corresponding release branch instead of `master`.

mattmassicotte:
Hmm, perhaps SwiftTreeSitter is atypical of other bindings then? It does not expose the runtime's API, but instead an API of its own. The consumers of SwiftTreeSitter are completely unaffected by the runtime and with very little exception do not need to even be aware of it.

So, for example, when I complete the migration from 0.25 to 0.27, I intend on there being zero source compatibility issues for SwiftTreeSitter clients.

I will move back to tracking a branch. But I still do not see a technical reason to mirror the runtime library's version.

clason:
Sorry, but I don't understand: If your bindings internally make use of new (C-lib) API introduced in 0.27, how can they work with a 0.25 library (e.g., system-wide distro installation)? And vice versa (as is the case!) if an API is removed, bindings based on the 0.25 state of the API will not work with 0.27?

(It's also a matter of clearly signaling to consumers and users about the compatibility matrix -- which is complex enough for tree-sitter as it is, due to its many moving parts...)

mattmassicotte:
SwiftTreeSitter does not support linking with externally-installed dylibs for the runtime. It must be linked statically as part of the SwiftPM build process. That's how it works today, and also how it works for all consumers of the current (but now just removed) Package.swift implementation.

This is less true for the parsers themselves. SwiftTreeSitter supports most possible configurations. However, the vast majority of consumers (myself included) rely on static linking to make sure the final artifact is deployable to users in a way that is independent of what they might have installed.

clason:
That's good :) My point is that it would be helpful (for the wider tree-sitter ecosystem) to be _very_ public about which version of the library is linked (and, ideally, have some quality control).

mattmassicotte:
Via the package dependency statement, in its Package.swift, which is currently set to "0.25.x". Most users (and I suspect virtually all) use the same technique for managing parser versions. Historically, parser version matters to consumers much more than runtime, since SwiftTreeSitter exists to abstract the runtime anyways, and parsers tend to impact the end user experience a lot more.

I think I understand your point, and I do not disagree! However, if I make a material change to SwiftTreeSitter that does *not* involve changing the runtime version, I must change the version. So, keeping it in lock-step actually presents a technical challenge for me.

Thinking more: I don't think it is feasible for me to have SwiftTreeSitter match the runtime version.

clason:
Yeah, I agree. (I was exactly in the same boat regarding parser versioning, which used to be coupled to the main lib version.)

My thinking was to just mirror the minor version (e.g., 0.26) and leave the patch version to be bumped independently. But I guess that would not necessarily work either. Anyway, just a suggestion; as long as the compatibility is made clear in some way, that's fine.

PR #5850, opened 2026-08-13
build(deps): bump wasi-sdk to v34-rc2

clason:
This prepares for the upcoming v34 release of wasi-sdk that will be based on LLVM 23, in the hope that we can include this for 0.27.

(If the release is delayed or there are significant regressions in the wild, I'll revert before our release.)


clason:
Looks like the scripts can't handle the RC version string.

PR #5851, opened 2026-08-13
Allow using Tree-sitter rust lib in multi-threaded web apps compiled to wasm32-unknown-unknown

maxbrunsfeld:
### Problem

The Tree-sitter Rust library already supports compiling to `wasm32-unknown-unknown` target, since PR https://github.com/tree-sitter/tree-sitter/pull/4820. But there are several blockers to using Tree-sitter in a production Rust Wasm application:

1. **Sending Trees Between Workers** - A tree normally retains its `TSLanguage`, which contains lexer and external-scanner function pointers. In WebAssembly, those function pointers are local to a particular module instance and cannot safely be used from another worker.

2. **Memory Allocator** - The C core previously linked a custom malloc implementation when targeting `wasm32-unknown-unknown`. But it was using an allocator that was designed for a very different purpose - the tiny resettable allocator that we introduced for running external WASM scanners under the native library, via Wasmtime. The allocator was only designed for cheaply managing heap allocations that are scoped to a single `parse` invocation, and is not general-purpose.

3. **libc functions for external scanners** - There are a short list of functions from `libc` that external scanners are allowed to depend on. When they are compiled to wasm and loaded via Tree-sitter's native wasmtime support, they are provided via a pre-compiled wasm blob that is included with the library. But until now, when Tree-sitter itself is compiled to wasm32-unknown-unknown, they were provided via a different system, where custom libc functions were compiled into the languages themselves. This divergent linking strategy was a bit confusing I think, and had the concrete downside that in this regime, we used an ASCII-only version of the `wctype` APIs.

### Change

1. The `Language` struct no longer implements `Send` on Wasm. The `Tree` type of course still implements `Send`. Here is how we preserve the Tree's `language()` accessor:

When compiling to Wasm, we introduce the concept of a *language context id*. This is a lazily initialized value that is local to a *wasm instance*, implemented using inline assembly. Whenever we create a Tree, we record the language context id of the instance on which Tree was parsed. Then, when accessing a Tree's language using (`Tree::language`/`ts_tree_language`), we check if we are on the same wasm instance as the tree was originally created on. If not, we lazily give the tree a callback-free copy of its language. The copied language retains all of the grammar tables needed for tree inspection and queries, but cannot be assigned to a parser. No `TSLanguage` ABI layout change is required.

The language has a new `is_parseable` method, which allows you to check if a language is one of these callback-free copies.

2. For allocation:

The Rust binding now uses C-compatible `malloc`, `calloc`, `realloc`, `free`, and `abort` functions that forward to the application's Rust global allocator. This way the Tree-sitter state shares a heap with the rest of the Rust crates that are being used. This allocator is also multi-threading aware. The adapter stores the allocation size in an 8-byte prefix so that C's unsized `free(void *)` API can reconstruct the `Layout` required by Rust. Eight-byte payload alignment is sufficient for Tree-sitter's C allocations and supported external scanners.

3. For `libc`:

This PR changes the `wasm32-unknown-unknown` target to work much more similarly to the existing Wasm build modes. The core library vendors a subset of `wasi-libc`, and provides a fixed list of functions from it as exports, so that languages compiled to `wasm32-unknown-unknown` can import them. We use the full implementation of `wctype` APIs from wasi-libc, so external scanners will  have the same unicode support in all build targets.

### Web worker integration test

This adds a standalone `test-rust-wasm-web` xtask and a Rust `wasm32-unknown-unknown` fixture.

The test:

1. Compiles the Rust binding as a shared-memory Wasm module.
2. Separately compiles the real JavaScript grammar, including its external scanner, as a dynamically linked Wasm side module.
3. Instantiates the same Rust runtime module on the UI thread and a Node.js `worker_thread`, sharing one `WebAssembly.Memory`.
4. Loads the JavaScript grammar only on the parsing worker.
5. Requests an initial parse from Rust UI code.
6. Transfers the owned tree pointer through shared memory coordination.
7. Inspects the returned tree on the UI runtime.
8. Repeatedly edits the Rust source `String` and `Tree`.
9. Sends the source address, source length, and old-tree pointer back to the worker without copying or encoding the source in JavaScript.
10. Incrementally reparses on the worker and returns the new tree.
11. Retains and inspects a clone of the old tree on the UI runtime while the worker owns and uses the other clone.
12. Repeats this for 100 edits, appending integers to a JavaScript array and verifying the resulting child count after every parse.

The test also verifies that the callback-free language obtained from a transferred tree is rejected by `Parser::set_language`. 

### Compatibility

This PR should only make observable changes to the `wasm32-unknown-unknown` target, in both `tree-sitter` and `tree-sitter-language` crates. For other targets, the only observable change is the addition of an `is_parseable` method on languages, which will always return true on non-wasm platforms.

### AI Policy

This code was written by GPT-5.5 Sol, and reviewed by me.


maxbrunsfeld:
@clason Soon after this lands, I would like to get the v0.27.0 release going. Are there blockers to that I should be aware of?

crates/language/wasm/unsupported/tree-sitter-0.26/README.md
clason, line 1: I don't understand why we would put this here in a branch where it will never apply? (`tree-sitter-language` bumps that ensure this are handled as part of the release process.)

maxbrunsfeld: It's just a precaution, in case someone upgrades `tree-sitter-language` but does *not* upgrade `tree-sitter` (leaves it at v0.26 or earlier). That configuration doesn't work, so we explicitly notify them to upgrade Tree-sitter via a build failure.

Cargo.toml
```diff
@@ -134,4 +134,4 @@ tree-sitter-highlight = { path = "./crates/highlight", version = "0.27.0" }
 tree-sitter-loader    = { path = "./crates/loader", version = "0.27.0" }
 tree-sitter-tags      = { path = "./crates/tags", version = "0.27.0" }
 
-tree-sitter-language = { path = "./crates/language", version = "0.1" }
+tree-sitter-language = { path = "./crates/language", version = "0.1.8" }
```
clason, line 137: If it's a breaking change, it should be `0.2.0`.

maxbrunsfeld: Unfortunately, it's incredibly disruptive to do such a bump - every existing grammar becomes incompatible with the new library. The incompatibility only occurs for the `wasm32-unknown-unknown` target, which didn't really work well in the first place (because of mistakes noted in the PR description). So we're going to keep `tree-sitter-language` on the v0.1 line, and just prompt people to upgrade `tree-sitter` if they do happen to use that target already, and bumped one crate but not the other.

clason:
> @clason Soon after this lands, I would like to get the v0.27.0 release going. Are there blockers to that I should be aware of?

Yes, the wasmtime LTS, as mentioned before. Also, a thorough review, as this touches the C-lib in a way that affects _all_ consumers.

@WillLillis may have some necessary follow-up PRs to his recent changes in the wings.

crates/language/wasm/unsupported/tree-sitter-0.26/README.md
clason: The dependency resolution would not allow that (unless they force things, in which case it's on them)?

I don't think carrying such an ephemeral (over-)defensive guard is worth the maintenance cost.

maxbrunsfeld: I think it would be allowed by cargo; you could just upgrade `tree-sitter-language` right?

Cargo.toml
clason: I'd rather treat wasm32 target as inherently broken then and trust people to upgrade voluntarily. There's few enough users that education is the more efficient strategy.

maxbrunsfeld: Yeah, that's the approach this PR is taking. There's just an explicit error message to instruct people to upgrade `tree-sitter`.

crates/language/wasm/unsupported/tree-sitter-0.26/README.md
clason: I'm not an expert, but if tree-sitter itself depends on `^0.1.8`, the solver should throw a hissy fit? If not, then this is best documented (*cough* changelog *cough*).

Cargo.toml
clason: And my point is that the cost is too high for that. (I don't believe anybody would voluntarily upgrade `tree-sitter-language` by itself; only if forced to do so by a `tree-sitter` bump.)

nineff:
Hi,
currently you can't compile `tree-sitter` together with `tree-sitter-c` for the target `wasm32-unknown-unknown`, even though `tree-sitter` 0.27 and `tree-sitter-language` 0.1.8 are released.
You can easily test this by simply running: `cargo build --target wasm32-unknown-unknown` on an empty crate with nothing but those mentioned dependencies.

Is this intentional or a bug?

maxbrunsfeld:
What error are you getting?

nineff:
I get the following:
```
cargo new test_ts
//Output ommited
cargo add tree-sitter tree-sitter-c
//Output ommited
cargo build -r --target wasm32-unknown-unknown
   Compiling serde_json v1.0.151
   Compiling regex-automata v0.4.18
   Compiling tree-sitter-c v0.24.2
warning: tree-sitter-c@0.24.2: /home/niko/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tree-sitter-language-0.1.8/wasm/unsupported/tree-sitter-0.26/stdio.c:1:2: error: "tree-sitter 0.26 is incompatible with this version of tree-sitter-language on wasm32-unknown-unknown; upgrade tree-sitter to 0.27 or newer"
warning: tree-sitter-c@0.24.2:     1 | #error "tree-sitter 0.26 is incompatible with this version of tree-sitter-language on wasm32-unknown-unknown; upgrade tree-sitter to 0.27 or newer"
warning: tree-sitter-c@0.24.2:       |  ^
warning: tree-sitter-c@0.24.2: 1 error generated.
error: failed to run custom build command for `tree-sitter-c v0.24.2`
```

I'm compiling on Arch  with rust 1.98.1

PR #5872, opened 2026-08-22
chore: remove Dockerfile

clason:
Problem: The committed Dockerfile is not used by the standard tooling in
this repo and therefore bitrots freely, which increases maintenance
burden on the (non-Docker-using) maintainers.

Solution: Drop the Dockerfile from the repo.

Part of some necessary spring-cleaning. If we do this, we should do this now (before the 0.27 release), in order to avoid shipping a broken file.


clason:
(same goes for the Nix flake, to be honest, but that is currently not broken afaik)

WillLillis:
> (same goes for the Nix flake, to be honest, but that is currently not broken afaik)

As I understand it, the Nix flake is used for local development only, and shouldn't affect anyone downstream. It can be patched up once a Nix user that understand the project has the time/interest.

WillLillis (approved):

clason:
Same goes for the Dockerfile; the question is whether individual local tooling should incur global longterm maintenance cost (small as it may be).

PR #5874, opened 2026-08-23
fix(generate): further node-types.json cleanup

### LINKED ISSUE #1654: Aliasing to a Grammar Symbol without a Production Produces an AST without a Corresponding Node in node-types.json

WillLillis:
See commit messages for details. This fixes #1654, as well as several other issues I discovered along the way. The last commit adds a new error variant to `tree-sitter-generate`. I think this is the proper direction in this case (as opposed to adding a warning), because _not_ catching this issue can lead to panics, nondeterministic output, and other unfun things. No grammar in the nvim-treesitter corpus currently exercises this error path.

- Fixes #1654 
- Supersedes #5271

After the changes were made, I had an agent go through the entire nvim-treesitter corpus, comparing each grammar's node-types.json file as generated on master vs. this PR. Some commit hashes may be old compared to nvim-treesitter's current main branch, but I don't think anything has drifted too far. I did a quick check through of the agent's findings myself (Note this wasn't exhaustive, that would take a very long time and I'm fairly confident in both the results here as well as the source code changes, CI run TBD). Here's a summary of all changes found (feel free to ignore, pasting here in case there's unintended side effects that we need to track down):

### Missing supertype-alias entries

These are the fix for #1654. The parser can produce a concrete alias wrapper, but master omitted its top-level schema entry.

  - Python-derived grammars:
      - Python
      - BitBake
      - Starlark and its vendored Python
      - Snakemake’s vendored Python
      - as_pattern_target, named, is added as a regular concrete node.
      - It has one required child drawn from the expression alternatives.
      - It does not inherit the canonical supertype’s subtypes.

  - Smali:
      - Adds named field_type.
      - Its required child is one of array_type, class_identifier, or primitive_type, matching the aliased type supertype.

  - Wing:
      - Adds named keyword_argument_value.
      - Its required child is one of the expression alternatives such as call, reference, binary_expression, or a literal.

  - V:
      - Adds named element and interpolation_expression.
      - Both wrap one required _expression alternative.

  - Fortran:
      - Adds named concurrent_mask.
      - It wraps one required _expression alternative.

### Supertype aliases merged into existing identities

Here the alias name already had an entry, so adding another object would be wrong. The existing object is widened to describe every possible appearance.

- Snakemake:
      - as_pattern_target, named, was previously represented as a leaf.
      - It now has fields: {} and the possible expression children.
      - children.required is false because this identity still has another leaf appearance.
      - The result remains one as_pattern_target entry.

  - PHP and PHP-only:
      - Named name previously had only fields: {}.
      - It now also permits an optional expression child.
      - The child alternatives include assignment_expression, binary_expression, primary_expression, yield_expression, and the other PHP expression forms.
      - The child is optional because ordinary leaf-like name appearances still exist.

  - HLSL and Slang:
      - Named semantics already permitted an optional identifier child.
      - Its child union now also includes the expression alternatives contributed by aliases such as alias(seq(":", $.expression), $.semantics).
      - It remains one entry with optional children.

### Named and anonymous identities separated

Master keyed regular aggregation by text alone. In these files, two runtime identities had the same text but different named values. The branch emits both.

  - Just:
      - Adds anonymous expression alongside the existing named expression.
      - The anonymous entry has optional if_expression or value children.

  - Djot:
      - Retains anonymous class.
      - Adds the distinct named class rule entry.

  - TypoScript:
      - Adds anonymous identifier alongside named identifier.

  - Hurl:
      - Master incorrectly attached the anonymous alias’s structure to named json_key_string.
      - The structured entry is now correctly named: false.
      - A separate named json_key_string entry represents the named rule.

  - Fortran:
      - Adds anonymous method_name alongside named method_name.
      - Adds named statement_label_reference alongside its anonymous alias.
      - This file also contains the concurrent_mask supertype fix above.

  - Surface:
      - Emits both named and anonymous comment.
      - Emits both named and anonymous expression_value.
      - Master collapsed each pair into a single entry whose namedness depended on aggregation order.

### Duplicate anonymous identities merged

Master appended anonymous token entries separately from structured aliases. That could produce two objects with exactly the same (type, named) identity. The branch routes both through the same map.

  - Cylc: anonymous empty-string identity, 2 → 1.
  - HTML-Django: anonymous empty-string identity, 2 → 1.
  - Perl: anonymous >, 2 → 1.
  - Haskell: anonymous (#, 2 → 1.
  - Idris: anonymous empty-string identity, 2 → 1.
  - KCL: anonymous null_assignment, 2 → 1.
      - The surviving structured entry’s child becomes optional because the identity also has a leaf-token appearance.

  - Nushell: anonymous _prefix, 2 → 1.
  - Swift: anonymous ?, 2 → 1.
  - Typst: anonymous ( and ), each 2 → 1.
  - Razor: 23 anonymous identities, each 2 → 1:
      - `at_attribute`, `at_await`, `at_block`, `at_do`, `at_for`, `at_foreach`, `at_if`, `at_implements`, `at_inherits`, `at_inject`, `at_layout`, `at_lock`, `at_model`, `at_namespace`, `at_page`, `at_preservewhitespace`, `at_rendermode`, `at_section`, `at_switch`, `at_try`, `at_typeparam`, `at_using`, `at_while`

(Except for KCL’s requiredness correction, these diffs simply remove redundant objects. No shape information is discarded.)

### extra metadata separated by namedness

  - Enforce:
      - Anonymous else loses `extra: true`.
      - The distinct named else identity remains the extra.

  - Kconfig:
      - Anonymous comment loses `extra: true`.
      - The distinct named comment identity remains the extra.

### AI Policy

Used gpt5.6-sol to discuss each issue and the proper approach. Each fix hand typed and tested by me. The initial sweep over the nvim-treesitter corpus was also performed by this model, results spot checked by me.

### DIFF
--- a/crates/generate/src/node_types.rs
+++ b/crates/generate/src/node_types.rs
@@ -49,7 +49,7 @@ struct NodeInfoJSON {
     subtypes: Option<Vec<NodeTypeRef>>,
 }
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg(feature = "load")]
 struct NodeTypeRef {
     kind: StrId,
@@ -644,7 +644,7 @@ fn generate_node_types(
         &extra_names,
     );
 
-    let mut result = node_types_json.into_iter().map(|e| e.1).collect::<Vec<_>>();
+    let mut result = node_types_json.into_values().collect::<Vec<_>>();
     result.extend(anonymous_node_types);
     result.sort_unstable_by(|a, b| {
         b.subtypes
@@ -779,7 +779,7 @@ fn collect_extra_names(
 /// Add one JSON entry per supertype and build the supertype-to-subtypes map.
 #[cfg(feature = "load")]
 fn build_supertype_entries(
-    node_types_json: &mut BTreeMap<StrId, NodeInfoJSON>,
+    node_types_json: &mut BTreeMap<NodeTypeRef, NodeInfoJSON>,
     syntax_grammar: &SyntaxGrammar,
     lexical_grammar: &LexicalGrammar,
     default_aliases: &AliasMap,
@@ -795,7 +795,10 @@ fn build_supertype_entries(
         }
         let variable = &syntax_grammar.variables[i];
         let node_type_json = node_types_json
-            .entry(variable.name)
+            .entry(NodeTypeRef {
+                kind: variable.name,
+                named: true,
+            })
             .or_insert_with(|| NodeInfoJSON {
                 kind: variable.name,
                 named: true,
@@ -837,7 +840,7 @@ fn build_supertype_entries(
     reason = "all parameters are required to build the entries"
 )]
 fn build_regular_entries(
-    node_types_json: &mut BTreeMap<StrId, NodeInfoJSON>,
+    node_types_json: &mut BTreeMap<NodeTypeRef, NodeInfoJSON>,
     syntax_grammar: &SyntaxGrammar,
     lexical_grammar: &LexicalGrammar,
     default_aliases: &AliasMap,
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
I think all of these grammars are at tip in nvim-treesitter (with the exception of Djot, which moved to codeberg and archived their GH repo in April).

PR #5875, opened 2026-08-23
generate: use a single buffer for the string pool's backing store

WillLillis:
This representation slightly lowers walltime and allocations, but also (more importantly!) makes `StrPool` `Send`.

Note that while `hashbrown` is not a _direct_ dependency of the generate crate, it is already used indirectly. No new dependencies in the tree with this change. I also used this as a chance to cleanup/improve some of the `StrPool` documentation.

### AI Policy

Fable pitched this representation and performed the memory measurements while conducting an unrelated review. I refined the implementation and added all relevant comments for the invariants used.

### DIFF
--- a/crates/generate/src/strpool.rs
+++ b/crates/generate/src/strpool.rs
@@ -1,6 +1,7 @@
-use std::{num::NonZeroU32, rc::Rc};
+use std::{hash::BuildHasher as _, num::NonZeroU32};
 
-use rustc_hash::FxHashMap;
+use hashbrown::{HashTable, hash_table::Entry};
+use rustc_hash::FxBuildHasher;
 
 /// Interned string id, a 1-based index into the pool's string table
 #[derive(Clone, Copy, PartialEq, PartialOrd, Ord, Hash, Debug, Eq)]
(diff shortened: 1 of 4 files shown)


crates/generate/src/strpool.rs
```diff
@@ -20,6 +21,11 @@ impl StrId {
     }
 
     /// Inverse of [`Self::raw`]. Caller must pass a value produced by `raw`.
+    ///
+    /// Panics
```
amaanq:
```suggestion
    /// # Panics
```

crates/generate/src/strpool.rs
```diff
@@ -32,17 +38,31 @@ impl Default for StrId {
...
+/// of unique strings.
 #[derive(Clone, Debug)]
 pub struct StrPool {
-    strs: Vec<Rc<str>>,
-    str_ids: FxHashMap<Rc<str>, StrId>,
+    /// Bytes of _every_ interned string, concatenated in intern order.
+    buf: String,
+    /// Byte ranges into `buf` for each string, indexed by [`StrId::index`]
+    spans: Vec<StrSpan>,
+    /// Ids of every interned string
```
amaanq:
```suggestion
    /// [`StrId`]s of every interned string
```
maybe?

amaanq (approved):

crates/generate/src/strpool.rs
WillLillis: Yeah that's better

PR #5878, opened 2026-08-23
build(deps): cargo update

clason:
We probably want to park this until the other PRs are merged, but we should schedule a final dependency update before the 0.27 release.


WillLillis (approved):

clason:
what happened there? o.O

clason:
Ah, looks like all the cargo updates were from wasmtime dependencies, which the wasmtime bump PR already pulled in, so rebasing made this an empty change -- which auto-closes the PR. TIL.

PR #5879, opened 2026-08-23
feat(generate): add token.sameLine(rule) DSL function

ManasSakthivel:
Adds token.sameLine(rule) to the Tree-sitter Grammar DSL.

A same-line token matches only when its first character appears on the same source line as the preceding token. Horizontal whitespace (spaces and tabs) is permitted in the separator gap; any vertical whitespace character (LF, CR, CRLF, VT U+000B, FF U+000C) causes the token to not match.

Implementation
--------------
The constraint is enforced entirely at NFA-build time in expand_tokens.rs, mirroring the architecture of token.immediate():

  Plain token:          separator = (extras | whitespace)*
  token.immediate:      separator = <none>
  token.sameLine:       separator = [ \t]*   (horizontal-only)

A new build_same_line_separator() function constructs the [ \t]* NFA subtree. Because LF/CR/VT/FF have no transition in those NFA states, the lexer naturally returns no match when a vertical whitespace character precedes the token. No changes to parser.c or the C runtime are required.

Also fixes a latent bug in get_implicit_precedence(): same-line tokens were not receiving the +1 implicit precedence boost that token.immediate tokens receive, which would have caused token.sameLine("=>") to lose conflict resolution to a plain "=>" at the same lexer position.

Files changed
-------------
  crates/generate/src/dsl.js                        DSL entry point
  crates/generate/src/parse_grammar.rs              JSON deserialization
  crates/generate/src/rules.rs                      MetadataParams, builder, hash
  crates/generate/src/grammars.rs                   LexicalVariable field
  crates/generate/src/prepare_grammar/expand_tokens.rs  NFA, separator, tests
  crates/generate/src/node_types.rs                 test fixture struct literal
  docs/src/assets/schemas/grammar.schema.json       schema enum
  docs/src/creating-parsers/2-the-grammar-dsl.md    documentation

Tests
-----
4 new test functions, 22 NFA-level scenarios:
  - allows horizontal whitespace (6 cases)
  - rejects vertical whitespace: LF, CR, CRLF, VT, FF (10 cases)
  - plain token regression: still crosses newlines (5 cases)
  - implicit precedence: same-line beats plain at same position (3 cases)
  - token.immediate regression (1 case)

All 78 tests pass.

Design note: grammar extras (e.g. inline comments) are intentionally not consumed before a same-line token — only bare horizontal whitespace is permitted. This conservative default can be revisited in a follow-up.

### AI Policy

- [ ] I have read the [AI Policy](https://tree-sitter.github.io/tree-sitter/6-contributing.html#ai-policy) and this PR complies with it.

### DIFF
--- a/crates/generate/src/dsl.js
+++ b/crates/generate/src/dsl.js
@@ -201,6 +201,14 @@ token.immediate = function (value) {
   };
 }
 
+token.sameLine = function (value) {
+  checkArguments(arguments, arguments.length, token.sameLine, 'token.sameLine', '', 'literal');
+  return {
+    type: "SAME_LINE_TOKEN",
+    content: normalize(value)
+  };
+}
+
 function normalize(value) {
   if (typeof value == "undefined")
     throw new Error("Undefined symbol");
--- a/crates/generate/src/grammars.rs
+++ b/crates/generate/src/grammars.rs
@@ -61,6 +61,10 @@ pub struct LexicalVariable {
     pub kind: VariableType,
     pub implicit_precedence: i32,
     pub start_state: u32,
+    /// True for `token.sameLine(rule)` — this token uses a horizontal-only
+    /// separator in the NFA, so the lex function will not match if a newline
+    /// appears in the leading whitespace before the token's first character.
+    pub is_same_line_token: bool,
 }
 
 #[derive(Debug, Default, PartialEq, Eq)]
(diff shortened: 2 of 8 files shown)


clason:
Fair warning: We do not consider pointing an LLM at a random open issue and letting it make a PR for open source credit as a welcome contribution. Please consider this before reopening this PR (or a similar new one).

ManasSakthivel:
Hey, Thanks for the clarification. I apologize for approaching my first contribution this way. I used LLM assistance more extensively than I should have during the implementationv and I recognize that this does not align with the project's expectations for open-source contributions.

I have closed the PR and will not reopen it or submit a similar PR.I’ll make sure any future contributions I make to the project are based on my own understanding and work and follow the project's contribution expectations.

PR #5885, opened 2026-08-25
build(deps): bump wasi-sdk to v34

clason:
Breaking change: `wasm32-wasi` is now `wasm32-wasip1`.


WillLillis (approved):

WillLillis:
> Breaking change: `wasm32-wasi` is now `wasm32-wasip1`.

Should we mark the commit as breaking as well?

clason:
Hmm, it's not really _our_ breaking change; things should continue working as before (unless you deviate from the "happy path"...)

PR #5896, opened 2026-08-28
chore: spring clean unused local dev files

clason:
- **chore: remove Zed editor config** (editor-specific config should not be in the global repo)
- **chore: remove .dockerignore** (fixup for #57872)


WillLillis (changes requested):
I still use the  .taplo.toml file and would like it to stay. I can add some CI jobs for this and fix up the xtask if you'd like. :slightly_smiling_face:

sogaiu:
Perhaps it should just be summer cleaning then :)

clason:
> I still use the .taplo.toml file and would like it to stay. I can add some CI jobs for this and fix up the xtask if you'd like. 🙂

Yes, although that can still add maintainer friction. Also, the file was snuck in hidden in an unrelated PR, so I would like to talk about the defaults we enforce (especially the alignment, which I am not a fan of).

WillLillis (approved):

PR #5897, opened 2026-08-29
fix(cli): correct zero-based positions for --edits

cuishuang:
### AI Policy

Fix an off-by-one error in `offset_for_position`.

The previous implementation incremented the offset before resolving row 0, causing zero-based positions on the first line to be shifted by one byte. It also consumed the first newline while locating row 0, which could allow columns past the end of that line.

The implementation now tracks the start byte of the current line explicitly. Regression tests cover the first-line origin, interior columns, the end of the first line, the start of the second line, and out-of-bounds coordinates.


WillLillis (approved):
Thanks!

PR #5898, opened 2026-08-29
feat(lib): add ranged S-expression display

### LINKED ISSUE #4982: Hidden rules with regex tokens silently match empty strings instead of reporting errors
### Problem

When using hidden rules (rules starting with `_`) with regex tokens that require at least one character (like `/[A-Za-z0-9_]+/`), the parser silently matches empty strings at zero-width positions instead of reporting syntax errors. This behavior is inconsistent with non-hidden rules and string literal tokens.

### Steps to reproduce

**Grammar 1: Hidden rule with regex (problematic).**

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $.id),
    id: $ => $._id,
    _id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

**Input file `1.txt`:**

```plaintext
.
```

**Command:**

```sh
tree-sitter generate
tree-sitter parse 1.txt
```

**Actual Output:**

```plaintext
(source_file [0, 0] - [1, 0]
  (id [0, 1] - [0, 1]))
```

**Expected Behavior:** Should report a MISSING error for `id` since there's no content after the dot.

---

**Grammar 2: Non-hidden rule with regex (correct).**

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $.id),
    id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

**Output:**

```plaintext
(source_file [0, 0] - [1, 0]
  (id [0, 1] - [0, 1]))
1.txt   Parse:    0.87 ms            2 bytes/ms (MISSING id [0, 1] - [0, 1])
```

**This correctly reports MISSING!**

---

**Grammar 3: Direct hidden rule reference.**

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $._id),
    _id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

**Output:**

```plaintext
(source_file [0, 0] - [1, 0])
```

**Again, silently fails without error!**

---

**Grammar 4: Hidden rule with string literal.**

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $._id),
    _id: _ => token('a'),
  },
});
```

**Output:**

```plaintext
(source_file [0, 0] - [1, 0])
1.txt   Parse:    0.02 ms           89 bytes/ms (MISSING "a" [0, 1] - [0, 1])
```

**String literals correctly report errors!**

### Expected behavior

Provide consistent error reporting behavior

tree-sitter 0.25.10 (da6fe9beb4f7f67beb75914ca8e0d48ae48d6406)

### Operating system/version

macOS Sequoia 15.5 arm64

WillLillis:
### The Problem

For the grammars in #4982, the parser correctly inserts a missing node for a hidden rule.  The CLI however reports a successful parse because its error-finding code walked the tree with a `TreeCursor`. `TreeCursor`s _intentionally_ skip hidden nodes, so the problematic node isn't directly unobservable through the normal walker API. Various hacks can be made to the CLI's detection logic to make a best guess for the node's range, but I don't think it's possible for a good solution here.

### The Solution

The good news is that `ts_node_string` does _not_ use a `TreeCursor`, and can see these hidden MISSING nodes. All we need to do is add ranges, and we fix the CLI's problem. I'd like to note here that the CLI crate re-implements this `ts_node_string_with_range` logic in several places. Instead, we can simply re-use the existing core functionality with ranges attached. When a tree contains an ERROR or MISSING node, it descends to said node, or the largest visible node containing it. This is what we then display on a failed parse. From the issue, copied over here for ease of reading, all parse outputs take in an input `.\n` (literal newline at the end):

#### Grammar 1: Hidden rule with regex (fixed):

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $.id),
    id: $ => $._id,
    _id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

```
(source_file [0, 0] - [1, 0]
  (id [0, 1] - [0, 1]
    (MISSING _id [0, 1] - [0, 1])))
stdin   Parse:    0.03 ms           62 bytes/ms (id [0, 1] - [0, 1] (MISSING _id [0, 1] - [0, 1]))
```

#### Grammar 2: Non-hidden rule with regex (still works):

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $.id),
    id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

```
(source_file [0, 0] - [1, 0]
  (MISSING id [0, 1] - [0, 1]))
stdin   Parse:    0.03 ms           67 bytes/ms (MISSING id [0, 1] - [0, 1])
```

#### Grammar 3: Direct hidden rule reference (fixed):

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $._id),
    _id: _ => token(/[A-Za-z0-9_]+/),
  },
});
```

```
(source_file [0, 0] - [1, 0]
  (MISSING _id [0, 1] - [0, 1]))
stdin   Parse:    0.04 ms           45 bytes/ms (source_file [0, 0] - [1, 0] (MISSING _id [0, 1] - [0, 1]))
```

#### Grammar 4: Hidden rule with string literal (still works):

```js
module.exports = grammar({
  name: 'test',
  rules: {
    source_file: $ => seq('.', $._id),
    _id: _ => token('a'),
  },
});
```

```
(source_file [0, 0] - [1, 0]
  (MISSING "a" [0, 1] - [0, 1])
stdin   Parse:    0.03 ms           70 bytes/ms (MISSING "a" [0, 1] - [0, 1])
```

- Closes #4982

Marking as a draft for now until after 0.27 is released.

### AI Policy

Used gpt5.6-sol to discuss the various solutions before settling on `ts_node_string_with_ranges`.

### DIFF
--- a/crates/cli/src/parse.rs
+++ b/crates/cli/src/parse.rs
@@ -14,8 +14,8 @@ use log::info;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use tree_sitter::{
-    InputEdit, Language, LogType, ParseOptions, ParseState, Parser, Point, Range, Tree, TreeCursor,
-    ffi,
+    InputEdit, Language, LogType, Node, ParseOptions, ParseState, Parser, Point, Range, Tree,
+    TreeCursor, ffi, format_sexp,
 };
 
 use crate::{fuzz::edits::Edit, paint::paint, util};
@@ -278,6 +278,35 @@ pub struct ParseResult {
     pub duration: Option<Duration>,
 }
 
+#[must_use]
+pub fn first_error_or_container(root: Node<'_>) -> Option<Node<'_>> {
+    if !root.has_error() {
+        return None;
+    }
+
+    let mut cursor = root.walk();
+    'descend: loop {
+        let node = cursor.node();
+        if node.is_error() || node.is_missing() {
+            return Some(node);
+        }
+
+        if cursor.goto_first_child() {
+            loop {
+                if cursor.node().has_error() {
+                    continue 'descend;
+                }
+                if !cursor.goto_next_sibling() {
+                    cursor.goto_parent();
+                    break;
+                }
+            }
+        }
+
+        return Some(node);
+    }
+}
+
 pub fn parse_file_at_path(
     parser: &mut Parser,
     language: &Language,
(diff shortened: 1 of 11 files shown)


clason (approved):

WillLillis:
Marking as a draft for now, considering not adding a new `with_ranges` function and instead adding a boolean parameter to the original when we have our next ABI break scheduled.

maxbrunsfeld:
This is, I think, the first place where hidden rules would be exposed in the public interface of a syntax tree. I think that instead, we might want to continue to honor their "hidden-ness", and just call the missing hidden node a generic zero-byte `ERROR`. It's a bit less informative, but it preserves the invariant that a hidden node is private.

clason:
I think the main issue here is the disparity between the internal parser and the `tree-sitter parse` output. The tree-sitter CLI output is primarily a diagnostic tool for the internal lib, and I would consider _any_ divergence as a real issue that needs to be fixed. (There are too many duplicate implementations already, and we should make sure to keep the Rust layer over the lib as thin as possible.)

Whether and how hidden rules are exposed is secondary to that (and whatever we choose should apply to all paths equally).

PR #5912, opened 2026-09-02
fix: decode UTF-16 surrogate pairs with input endianness

ydah:
### AI Policy

Parsing UTF-16BE source containing supplementary-plane characters, such as `let emoji = "😀";`, decoded the emoji as two isolated surrogates on little-endian hosts, causing incorrect lexer lookahead and potentially shifted token boundaries.

This PR fixes the trailing surrogate byte-order conversion in both UTF-16LE and UTF-16BE decoders and adds a regression test for U+1F600.

### DIFF
--- a/crates/cli/src/tests/parser_test.rs
+++ b/crates/cli/src/tests/parser_test.rs
@@ -254,6 +254,21 @@ fn test_parsing_with_custom_utf16_be_input() {
 }
 
 #[test]
+fn test_utf16_decodes_surrogate_pairs() {
+    let mut parser = Parser::new();
+    let language = get_test_fixture_language("utf16_surrogate_oob");
+    parser.set_language(&language).unwrap();
+
+    let le = [0xD83D_u16.to_le(), 0xDE00_u16.to_le()];
+    let tree = parser.parse_utf16_le(le, None).unwrap();
+    assert_eq!(tree.root_node().to_sexp(), "(program (supplementary))");
+
+    let be = [0xD83D_u16.to_be(), 0xDE00_u16.to_be()];
+    let tree = parser.parse_utf16_be(be, None).unwrap();
+    assert_eq!(tree.root_node().to_sexp(), "(program (supplementary))");
+}
+
+#[test]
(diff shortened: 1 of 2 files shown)


clason:
Hmm, looks like we're rewriting our own utf8 parser -- poorly.

Maybe it would make more sense to switch to a proper library like https://github.com/JuliaStrings/utf8proc?

Although of course that would come with performance concerns -- which also goes for this change here. Have you benchmarked the effect?

WillLillis:
> Hmm, looks like we're rewriting our own utf8 parser -- poorly.

Definitely something to look into optimizing, but for now (for all the inputs I've tested at least), this portion of parsing is a very small part of the cost.

> Maybe it would make more sense to switch to a proper library like [JuliaStrings/utf8proc](https://github.com/JuliaStrings/utf8proc?rgh-link-date=2026-09-02T16%3A56%3A43Z)?

It is tempting to push this complexity off to a library, but I'd like to keep the "no dependencies" property of the C library as much as possible. If this are of the project proves to be an issue going forward, we can spend some development time bringing it up and/or re-evaluating this decision.

> Although of course that would come with performance concerns -- which also goes for this change here. Have you benchmarked the effect?

Benchmarked this myself locally and found everything to be within noise, which makes sense. This should only  generate additional instructions for non-native-endian UTF-16 containing surrogate pairs, and this portion of the code is a relatively small part of tree-sitter's parsing. CC @ydah, going forward, some basic measurements/methodology description for similar changes would be very appreciated. :slightly_smiling_face:

WillLillis (approved):
Thanks! Nice use of the existing corpus grammar in the test.

clason:
Agreed, the "dependency-free" is a prominent selling point so we should keep it.

My concern was mostly with robustness and edge cases (which may change as the standard evolves), but as we don't have to _edit_ text (e.g., case folding), the surface area is much smaller.

It may still be worth a look (but very low priority) whether we can extract the equivalent of this header from utf8proc and keep it synced somehow to profit from their ongoing work without cost.

PR #5917, opened 2026-09-03
fix(cli)!: fix query test capture range matching across lines

cuishuang:
### AI Policy

`query --test` can report an invalid query test as passing when an assertion is on a later row than the end of a multiline capture.

`assert_expected_captures` compared the assertion's end column with the capture's end column without requiring both points to be on the same row.

Compare the assertion's inclusive end point with the capture's exclusive end point using `Utf8Point` ordering.


WillLillis (changes requested):
The fix itself looks good here. A few things to address though. Could you

- Delete the added  test module and move the new unit test into `query_testing.rs`.
- Make `assert_expected_captures` pure, either pulling the file read into the caller or making a simple wrapper helper. This removes the need for  the need for the temp dir in the unit test.
- For your unit test, make the capture end something in bounds, so we don't introduce other confounding  factors into the test ((1, 10) doesn't correspond to any valid position in your input text)

For future PRs, could you use a branch other than your fork's master? This breaks a lot of GH's tooling and adds unneeded friction.

cuishuang:
> The fix itself looks good here. A few things to address though. Could you
> 
> * Delete the added  test module and move the new unit test into `query_testing.rs`.
> * Make `assert_expected_captures` pure, either pulling the file read into the caller or making a simple wrapper helper. This removes the need for  the need for the temp dir in the unit test.
> * For your unit test, make the capture end something in bounds, so we don't introduce other confounding  factors into the test ((1, 10) doesn't correspond to any valid position in your input text)
> 
> For future PRs, could you use a branch other than your fork's master? This breaks a lot of GH's tooling and adds unneeded friction.

Thanks. I moved the regression test into query_testing.rs, made assert_expected_captures pure by moving file reading and assertion parsing to the caller, and changed the test capture end point to a valid in-bounds position.

I also noted the branch naming issue and will use a dedicated branch for future PRs. 

Please review it again.

crates/cli/src/tests/query_testing_test.rs
```diff
@@ -0,0 +1,25 @@
+use std::fs;
```
WillLillis: This file still needs to be deleted.

cuishuang: Thanks for your review.  

Deleted the test module and pushed the update. The unit test now lives in query_testing.rs.

WillLillis (approved):

WillLillis:
Marked as breaking since this changes a public function for the CLI library, but since that realistically isn't consumed that way it should be fine to backport.

PR #5933, opened 2026-09-10
build(deps): bump wasmtime to v48.0.2

clason:


lib/Cargo.toml
```diff
@@ -51,7 +51,7 @@ default-features = false
 features = [ "cranelift", "gc", "gc-null" ]
 optional = true
 package = "wasmtime-c-api-impl"
-version = "48.0.1"
+version = "48.0"
```
WillLillis: We want an explicit `"48.0.2"` here right?

clason: No, that was deliberate; this will match any `48.0.x` so `cargo update` will just pull in the latest.

(As backported bugfix releases of an LTS, I assume that it's always safe to bump without needing code changes.)

But I'm happy to revert to specifying the full version if you prefer!

WillLillis: Good point, I think either should be fine here.

WillLillis (approved):

PR #5937, opened 2026-09-12
feat(generate)!: mirror `regex_syntax` error kinds

WillLillis:
`regex_syntax` errors are not serializable, so `ExpandRuleError::Parse`
had a `String` instead of structured data.

Add `RegexError`, a serializable mirror carrying the pattern, the error
kind, and the spans `regex_syntax` reports.

BREAKING CHANGE: Alters public error types for the generate crate.

------

This is a continuation of the same "improve generate as a library" effort from prior PRs. Since we're manually mirroring the error types from this crate, I added a CI job that ensures they're in sync anytime a related file is modified. The maintenance burden for this _should be_ small. If other complications (or upstream changes) complicate this, we can reconsider. The payoff will be evident in an upcoming feature.

While I was working on generate's error types, I found that several weren't publicly exported, so that has now been resolved. In addition to the visibility issue, I changed several to hold `Box<str>` instead of `String` for their payload types. This shrinks `Result` sizes across the crate, which is neutral to marginally better perf-wise.

### AI Policy

Used Opus 5 to help design and implement the check job.

### DIFF
--- a/.github/workflows/regex_error_kinds.yml
+++ b/.github/workflows/regex_error_kinds.yml
@@ -1,0 +1,33 @@
+name: Check Regex Error Kinds
+
+on:
+  pull_request:
+    paths:
+      - Cargo.lock
+      - crates/generate/src/prepare_grammar/pattern.rs
+      - crates/xtask/src/**
+  push:
+    branches: [master]
+    paths:
+      - Cargo.lock
+      - crates/generate/src/prepare_grammar/pattern.rs
+      - crates/xtask/src/**
+
+jobs:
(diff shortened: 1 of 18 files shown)


clason (approved):

clason:
Hmm, can we move that job into a different workflow, e.g., the `checks` one? (Try to common minimize setup time etc.) Same for the  `check-wasm-stdlib`.

Low-priority and out of scope for this PR, but now would be a good time to refactor the CI workflow to allow more load balancing and cleaner separation of concerns (and naming).

PR #5942, opened 2026-09-13
ci: consolidate consistency checks into checks job

WillLillis:
Run the regex error-kind and Wasm stdlib checks in CI’s existing `checks` job so they share checkout and runner setup.

Addresses https://github.com/tree-sitter/tree-sitter/pull/5937#issuecomment-5645145479

### AI Policy

Used gpt5.6-sol to  review things after moving.


.github/workflows/ci.yml
```diff
@@ -41,11 +41,23 @@ jobs:
       - name: Lint web files
         run: make lint-web
 
+      # rustdoc's JSON output is nightly-only and unstable
+      - name: Set up pinned nightly Rust toolchain
+        run: rustup toolchain install nightly-2026-09-03 --profile minimal
```
clason, line 46: Since we need to re-install a different toolchain, I would keep these as a different (parallel) job -- although still in the same file.

Name could be something like "consistency"?

clason (approved):

clason (changes requested):
Sorry, wanted to fix this up myself, but any change needs to be reflected in the required workflow settings so this needs you either way.

PR #5944, opened 2026-09-13
feat(generate): implement the native grammar DSL

WillLillis:
### What is this?

This is a custom DSL designed solely for tree-sitter grammar development. It's fast, designed to support external tooling, and eliminates the need for a JS runtime at any point in the grammar generation pipeline. This is a completely optional (JS isn't going anywhere) alternative _frontend_ for grammar development. To be clear, **support for JS grammars is not going anywhere.** Everything after `load_grammar_file` is an identical code  path  to the existing JS frontend. There is a large wall of text below outlining this feature, it's motivation, and implementation. I'm happy to discuss any and  all points here, but please  read through the PR description first. I'll expand on these points (and the language itself) below. 

#### No  guarantee of stability

The CJS to ESM migration work is still a (relatively) recent and painful memory. Breaking changes in the JS ecosystem and/or toolchain necessitate grammars to be updated. There is no guarantee there won't be further breaking changes in the future.

- Can't we just require an older version of node to generate?
Sure, but this is not a sustainable long-term solution. People, systems, and distros move on. We shouldn't be stuck to old technology due to changes that we don't care about or even need.
- Most changes are trivial to fix. Is it really that big of a deal?
**Yes**. Many tree-sitter grammars are developed and maintained by the community. After a grammar is "done", many grammar authors understandably don't have the time or interest to make compatibility changes ad infinitum. If the maintainer is absent, the solution is to fork. We now have two copies of the grammar.  And if it happens again? etc. The bottom line is, once a tree-sitter grammar works, it should _always_ work. Only carefully considered changes made in this repo should change that, if absolutely necessary. *How can we ensure tree-sitter is usable 5 years from now? 10 years? Beyond?* If the collection of community grammars decays into a mess of "this requires node 0.x.y because because of z, repeated hundreds of times, the project  will die.

Beyond JS the language, we also have  to consider JS the runtime. Node has been the defacto standard, and tree-sitter's default runtime for many years now. The landscape however, isn't settled. There is already special handling for separate runtimes in the current DSL's JS files, and some of these have already bitrotted. It's not clear which runtime will "win", and such a development shouldn't affect _grammar development_.  

#### npm

The last year or so has been a near constant stream of security incidents. Tree-sitter grammars currently depend on eachother via npm when using inheritance, so this is an attack  surface we're exposed to. No grammar author, downstream consumer, or tree-sitter maintainer should have to worry about supply chain attacks when generating a parser. 

#### The user experience isn't that great

Small mistakes such as typos lead to long JS stack traces that are difficult to condense into the actual problem at hand:

image

This friction isn't necessary, and we absolutely can do better:

image

More on this below.

####  It's a bit much

Many of the JavasScript-isms used in grammars today simply aren't necessary for grammar development. This raises the barrier to entry for contribution and wastes development time. Requiring JS tooling for in-editor docs on the grammar builtins is an incredibly heavy solution made necessary only by our reliance on JS.

### The language

Ok, so if JS is so awful, what's the alternative? I've spent the last 5ish months developing this Tree-sitter grammar (TSG) DSL, designed only for grammar development. The language is implemented entirely in Rust, *inside the generate crate*. I designed this language to make things as familiar as possible to the JS patterns evident  in the current ecosystem, while also staying as minimal as was reasonable. We own all of the code, so any and all breakage is our own. I want to re-emphasize the _minimal_ aspect here. The current implementation is  just under 11k LOC, ~17k LOC with tests. Any changes must be weighted heavily with the maintenance burden they add. I spent a lot of time trying out a lot of different implementations, feature sets, and options. While I'm open to discuss any aspect of the  language design, I'm fairly confident in the choices made in this branch currently. A few points I'd like to highlight

#### Performance

TSG evaluation is fairly fast. While actual time when called from within the tree-sitter-cli may vary (due to icache pressure and other variables), every grammar I've translated evaluates in  under a millisecond. Here's a graph showing evaluation times after a recent optimization  effort:

image

It's important to note here that evaluation times are already a small part of total grammar generation time. So why care about performance at all? The answer (besides fun) is that it matters for tooling (see below). I  put as much effort as possible into the core language implementation, so that other pieces built on top could offer a solid user experience on every keystroke.

#### Tooling

In parallel with the core language implementation, I've also been developing a companion LSP: https://github.com/WillLillis/ts_grammar_ls. Because the language is written as a library, the LSP can just use it as a dependency, offering things like diagnostics, goto definition, references, etc. using the same building blocks. 

The long bet here is that we'll be able to develop significantly better tooling for grammar development with these fundamental pieces in place. Heuristics on grammar quality, lints for anti-patterns, and other aids for grammar authors can live in one place intended for grammar authors alone. The tree-sitter-cli currently serves a dual role for both grammar development by authors, and generation by downstream consumers. The LSP, on the other hand, can be solely developer focused, serving as both an in-editor aid as well as a CLI validation tool similar to ts_query_ls. 

As an example for what this unlocks, here is  an in-editor repl. This works both for the entire grammar, as well  as _any_ other arbitrary rule you select as the start rule.

rec-20260913-153723-compressed

Important to note here, the LSP's development was largely LLM driven. During development  while the core language implementation was still changing rapidly, it was too much of a burden to keep it up myself. Its purpose was largely to  see _what_ needed to be exposed from the core language as a library. Before 0.28 is released, the LSP will either be completely rewritten or largely reworked to a much higher quality standard.

#### Grammar for the grammar

I've written a tree-sitter grammar, for grammar, here: https://github.com/WillLillis/tree-sitter-tsg.  Following this PR getting merged, this repo should be moved  into the tree-sitter-grammars org.

### Migration

I currently have over 100 grammars translated from their JS definitions to TSG. We plan to merge these translations as applicable  to the tree-sitter-grammars org after 0.28 is released. These translations have been verified to generate functionally identical parsers, but still are largely LLM-driven, and will require a human to review/refine stylistically before merging. To be completely clear, the JS DSL is _not going anywhere_. If you want to keep your JS grammar definition, you can. We would like to encourage people to  switch based on the TSG's smaller  dependency chain, as well as the tooling we can develop around it. The full list of translated grammars  is here:

<details>
  <summary>Translated grammars</summary>

- c (tree-sitter-c): https://github.com/tree-sitter/tree-sitter-c/tree/native-dsl-translation
- cpp (tree-sitter-cpp): https://github.com/tree-sitter/tree-sitter-cpp/tree/native-dsl-translation
- javascript (tree-sitter-javascript): https://github.com/tree-sitter/tree-sitter-javascript/tree/native-dsl-translation
- json_grammar (tree-sitter-json): https://github.com/tree-sitter/tree-sitter-json/tree/native-dsl-translation
- query (tree-sitter-query): https://github.com/tree-sitter-grammars/tree-sitter-query/tree/native-dsl-translation
- rust_lang (tree-sitter-rust): https://github.com/tree-sitter/tree-sitter-rust/tree/native-dsl-translation
- python (tree-sitter-python): https://github.com/tree-sitter/tree-sitter-python/tree/native-dsl-translation
- go (tree-sitter-go): https://github.com/tree-sitter/tree-sitter-go/tree/native-dsl-translation
- gomod (tree-sitter-go-mod): https://github.com/WillLillis/tree-sitter-go-mod/tree/native-dsl-translation
- vim (tree-sitter-vim): https://github.com/tree-sitter-grammars/tree-sitter-vim/tree/native-dsl-translation
- haskell (tree-sitter-haskell): https://github.com/tree-sitter-grammars/tree-sitter-haskell/tree/native-dsl-translation
- php_only (tree-sitter-php): https://github.com/tree-sitter/tree-sitter-php/tree/native-dsl-translation
- php (tree-sitter-php): https://github.com/tree-sitter/tree-sitter-php/tree/native-dsl-translation
- ocaml (tree-sitter-ocaml): https://github.com/tree-sitter/tree-sitter-ocaml/tree/native-dsl-translation
- ocaml_interface (tree-sitter-ocaml): https://github.com/tree-sitter/tree-sitter-ocaml/tree/native-dsl-translation
- ocaml_type (tree-sitter-ocaml): https://github.com/tree-sitter/tree-sitter-ocaml/tree/native-dsl-translation
- typescript (tree-sitter-typescript): https://github.com/tree-sitter/tree-sitter-typescript/tree/native-dsl-translation
- tsx (tree-sitter-typescript): https://github.com/tree-sitter/tree-sitter-typescript/tree/native-dsl-translation
- bash (tree-sitter-bash): https://github.com/tree-sitter/tree-sitter-bash/tree/native-dsl-translation
- lua (tree-sitter-lua): https://github.com/WillLillis/tree-sitter-lua/tree/native-dsl-translation
- ruby (tree-sitter-ruby): https://github.com/tree-sitter/tree-sitter-ruby/tree/native-dsl-translation
- java (tree-sitter-java): https://github.com/tree-sitter/tree-sitter-java/tree/native-dsl-translation
- html (tree-sitter-html): https://github.com/tree-sitter/tree-sitter-html/tree/native-dsl-translation
- toml (tree-sitter-toml): https://github.com/WillLillis/tree-sitter-toml/tree/native-dsl-translation
- nix (tree-sitter-nix): https://github.com/WillLillis/tree-sitter-nix/tree/native-dsl-translation
- dart (tree-sitter-dart): https://github.com/WillLillis/tree-sitter-dart/tree/native-dsl-translation
- dockerfile (tree-sitter-dockerfile): https://github.com/WillLillis/tree-sitter-dockerfile/tree/native-dsl-translation
- elixir (tree-sitter-elixir): https://github.com/WillLillis/tree-sitter-elixir/tree/native-dsl-translation
- elm (tree-sitter-elm): https://github.com/WillLillis/tree-sitter-elm/tree/native-dsl-translation
- zig (tree-sitter-zig): https://github.com/tree-sitter-grammars/tree-sitter-zig/tree/native-dsl-translation
- sql (tree-sitter-sql): https://github.com/WillLillis/tree-sitter-sql/tree/native-dsl-translation
- clojure (tree-sitter-clojure): https://github.com/WillLillis/tree-sitter-clojure/tree/native-dsl-translation
- julia (tree-sitter-julia): https://github.com/tree-sitter-grammars/tree-sitter-julia/tree/native-dsl-translation
- scala (tree-sitter-scala): https://github.com/tree-sitter/tree-sitter-scala/tree/native-dsl-translation
- c_sharp (tree-sitter-c-sharp): https://github.com/tree-sitter/tree-sitter-c-sharp/tree/native-dsl-translation
- asm (tree-sitter-asm): https://github.com/WillLillis/tree-sitter-asm/tree/native-dsl-translation
- agda (tree-sitter-agda): https://github.com/tree-sitter/tree-sitter-agda/tree/native-dsl-translation
- nim (tree-sitter-nim): https://github.com/WillLillis/tree-sitter-nim/tree/native-dsl-translation
- puppet (tree-sitter-puppet): https://github.com/WillLillis/tree-sitter-puppet/tree/native-dsl-translation
- beancount (tree-sitter-beancount): https://github.com/WillLillis/tree-sitter-beancount/tree/native-dsl-translation
- numbat (tree-sitter-numbat): https://github.com/WillLillis/tree-sitter-numbat/tree/native-dsl-translation
- tlaplus (tree-sitter-tlaplus): https://github.com/WillLillis/tree-sitter-tlaplus/tree/native-dsl-translation
- perl (tree-sitter-perl): https://github.com/WillLillis/tree-sitter-perl/tree/native-dsl-translation
- gdscript (tree-sitter-gdscript): https://github.com/WillLillis/tree-sitter-gdscript/tree/native-dsl-translation
- latex (tree-sitter-latex): https://github.com/WillLillis/tree-sitter-latex/tree/native-dsl-translation
- vue (tree-sitter-vue): https://github.com/tree-sitter-grammars/tree-sitter-vue/tree/native-dsl-translation
- ada (tree-sitter-ada): local-only (no upstream), used as a  stress test.
- xml (tree-sitter-xml): https://github.com/tree-sitter-grammars/tree-sitter-xml/tree/native-dsl-translation
- dtd (tree-sitter-xml): https://github.com/tree-sitter-grammars/tree-sitter-xml/tree/native-dsl-translation
- fortran (tree-sitter-fortran): https://github.com/WillLillis/tree-sitter-fortran/tree/native-dsl-translation
- nu (tree-sitter-nu): https://github.com/WillLillis/tree-sitter-nu/tree/native-dsl-translation
- pem (tree-sitter-pem): https://github.com/tree-sitter-grammars/tree-sitter-pem/tree/native-dsl-translation
- printf (tree-sitter-printf): https://github.com/tree-sitter-grammars/tree-sitter-printf/tree/native-dsl-translation
- properties (tree-sitter-properties): https://github.com/tree-sitter-grammars/tree-sitter-properties/tree/native-dsl-translation
- diff (tree-sitter-diff): https://github.com/tree-sitter-grammars/tree-sitter-diff/tree/native-dsl-translation
- gitattributes (tree-sitter-gitattributes): https://github.com/tree-sitter-grammars/tree-sitter-gitattributes/tree/native-dsl-translation
- kdl (tree-sitter-kdl): https://github.com/tree-sitter-grammars/tree-sitter-kdl/tree/native-dsl-translation
- kconfig (tree-sitter-kconfig): https://github.com/tree-sitter-grammars/tree-sitter-kconfig/tree/native-dsl-translation
- luadoc (tree-sitter-luadoc): https://github.com/tree-sitter-grammars/tree-sitter-luadoc/tree/native-dsl-translation
- gstlaunch (tree-sitter-gstlaunch): https://github.com/tree-sitter-grammars/tree-sitter-gstlaunch/tree/native-dsl-translation
- ungrammar (tree-sitter-ungrammar): https://github.com/tree-sitter-grammars/tree-sitter-ungrammar/tree/native-dsl-translation
- xcompose (tree-sitter-xcompose): https://github.com/tree-sitter-grammars/tree-sitter-xcompose/tree/native-dsl-translation
- po (tree-sitter-po): https://github.com/tree-sitter-grammars/tree-sitter-po/tree/native-dsl-translation
- uxntal (tree-sitter-uxntal): https://github.com/tree-sitter-grammars/tree-sitter-uxntal/tree/native-dsl-translation
- pymanifest (tree-sitter-pymanifest): https://github.com/tree-sitter-grammars/tree-sitter-pymanifest/tree/native-dsl-translation
- luap (tree-sitter-luap): https://github.com/tree-sitter-grammars/tree-sitter-luap/tree/native-dsl-translation
- hyprlang (tree-sitter-hyprlang): https://github.com/tree-sitter-grammars/tree-sitter-hyprlang/tree/native-dsl-translation
- cpon (tree-sitter-cpon): https://github.com/tree-sitter-grammars/tree-sitter-cpon/tree/native-dsl-translation
- ron (tree-sitter-ron): https://github.com/tree-sitter-grammars/tree-sitter-ron/tree/native-dsl-translation
- gn (tree-sitter-gn): https://github.com/tree-sitter-grammars/tree-sitter-gn/tree/native-dsl-translation
- udev (tree-sitter-udev): https://github.com/tree-sitter-grammars/tree-sitter-udev/tree/native-dsl-translation
- yuck (tree-sitter-yuck): https://github.com/tree-sitter-grammars/tree-sitter-yuck/tree/native-dsl-translation
- requirements (tree-sitter-requirements): https://github.com/tree-sitter-grammars/tree-sitter-requirements/tree/native-dsl-translation
- tcl (tree-sitter-tcl): https://github.com/tree-sitter-grammars/tree-sitter-tcl/tree/native-dsl-translation
- firrtl (tree-sitter-firrtl): https://github.com/tree-sitter-grammars/tree-sitter-firrtl/tree/native-dsl-translation
- tablegen (tree-sitter-tablegen): https://github.com/tree-sitter-grammars/tree-sitter-tablegen/tree/native-dsl-translation
- re2c (tree-sitter-re2c): https://github.com/tree-sitter-grammars/tree-sitter-re2c/tree/native-dsl-translation
- qmldir (tree-sitter-qmldir): https://github.com/tree-sitter-grammars/tree-sitter-qmldir/tree/native-dsl-translation
- go_sum (tree-sitter-go-sum): https://github.com/tree-sitter-grammars/tree-sitter-go-sum/tree/native-dsl-translation
- cyberchef (tree-sitter-cyberchef): https://github.com/tree-sitter-grammars/tree-sitter-cyberchef/tree/native-dsl-translation
- chatito (tree-sitter-chatito): https://github.com/tree-sitter-grammars/tree-sitter-chatito/tree/native-dsl-translation
- doxygen (tree-sitter-doxygen): https://github.com/tree-sitter-grammars/tree-sitter-doxygen/tree/native-dsl-translation
- readline (tree-sitter-readline): https://github.com/tree-sitter-grammars/tree-sitter-readline/tree/native-dsl-translation
- bicep (tree-sitter-bicep): https://github.com/tree-sitter-grammars/tree-sitter-bicep/tree/native-dsl-translation
- squirrel (tree-sitter-squirrel): https://github.com/tree-sitter-grammars/tree-sitter-squirrel/tree/native-dsl-translation
- hare (tree-sitter-hare): https://github.com/tree-sitter-grammars/tree-sitter-hare/tree/native-dsl-translation
- pony (tree-sitter-pony): https://github.com/tree-sitter-grammars/tree-sitter-pony/tree/native-dsl-translation
- odin (tree-sitter-odin): https://github.com/tree-sitter-grammars/tree-sitter-odin/tree/native-dsl-translation
- kotlin (tree-sitter-kotlin): https://github.com/tree-sitter-grammars/tree-sitter-kotlin/tree/native-dsl-translation
- meson (tree-sitter-meson): https://github.com/tree-sitter-grammars/tree-sitter-meson/tree/native-dsl-translation
- func (tree-sitter-func): https://github.com/tree-sitter-grammars/tree-sitter-func/tree/native-dsl-translation
- cairo (tree-sitter-cairo): https://github.com/tree-sitter-grammars/tree-sitter-cairo/tree/native-dsl-translation
- bitbake (tree-sitter-bitbake): https://github.com/tree-sitter-grammars/tree-sitter-bitbake/tree/native-dsl-translation
- make (tree-sitter-make): https://github.com/tree-sitter-grammars/tree-sitter-make/tree/native-dsl-translation
- arduino (tree-sitter-arduino): https://github.com/tree-sitter-grammars/tree-sitter-arduino/tree/native-dsl-translation
- cuda (tree-sitter-cuda): https://github.com/tree-sitter-grammars/tree-sitter-cuda/tree/native-dsl-translation
- nqc (tree-sitter-nqc): https://github.com/tree-sitter-grammars/tree-sitter-nqc/tree/native-dsl-translation
- glsl (tree-sitter-glsl): https://github.com/tree-sitter-grammars/tree-sitter-glsl/tree/native-dsl-translation
- ispc (tree-sitter-ispc): https://github.com/tree-sitter-grammars/tree-sitter-ispc/tree/native-dsl-translation
- svelte (tree-sitter-svelte): https://github.com/tree-sitter-grammars/tree-sitter-svelte/tree/native-dsl-translation
- starlark (tree-sitter-starlark): https://github.com/tree-sitter-grammars/tree-sitter-starlark/tree/native-dsl-translation
- hlsl (tree-sitter-hlsl): https://github.com/tree-sitter-grammars/tree-sitter-hlsl/tree/native-dsl-translation
- slang (tree-sitter-slang): https://github.com/tree-sitter-grammars/tree-sitter-slang/tree/native-dsl-translation
- commonlisp (tree-sitter-commonlisp): https://github.com/tree-sitter-grammars/tree-sitter-commonlisp/tree/native-dsl-translation
- luau (tree-sitter-luau): https://github.com/tree-sitter-grammars/tree-sitter-luau/tree/native-dsl-translation
- objc (tree-sitter-objc): https://github.com/tree-sitter-grammars/tree-sitter-objc/tree/native-dsl-translation
- hcl (tree-sitter-hcl): https://github.com/tree-sitter-grammars/tree-sitter-hcl/tree/native-dsl-translation
- netrc (tree-sitter-netrc): https://github.com/tree-sitter-grammars/tree-sitter-netrc/tree/native-dsl-translation
- csv (tree-sitter-csv): https://github.com/tree-sitter-grammars/tree-sitter-csv/tree/native-dsl-translation
- tsv (tree-sitter-csv): https://github.com/tree-sitter-grammars/tree-sitter-csv/tree/native-dsl-translation
- psv (tree-sitter-csv): https://github.com/tree-sitter-grammars/tree-sitter-csv/tree/native-dsl-translation
- cst (tree-sitter-cst): https://github.com/tree-sitter-grammars/tree-sitter-cst/tree/native-dsl-translation
- linkerscript (tree-sitter-linkerscript): https://github.com/tree-sitter-grammars/tree-sitter-linkerscript/tree/native-dsl-translation
- poe_filter (tree-sitter-poe-filter): https://github.com/tree-sitter-grammars/tree-sitter-poe-filter/tree/native-dsl-translation
- capnp (tree-sitter-capnp): https://github.com/tree-sitter-grammars/tree-sitter-capnp/tree/native-dsl-translation
- thrift (tree-sitter-thrift): https://github.com/tree-sitter-grammars/tree-sitter-thrift/tree/native-dsl-translation
- smali (tree-sitter-smali): https://github.com/tree-sitter-grammars/tree-sitter-smali/tree/native-dsl-translation
- ssh_config (tree-sitter-ssh-config): https://github.com/tree-sitter-grammars/tree-sitter-ssh-config/tree/native-dsl-translation
- gpg_config (tree-sitter-gpg-config): https://github.com/tree-sitter-grammars/tree-sitter-gpg-config/tree/native-dsl-translation
- zsh (tree-sitter-zsh): https://github.com/WillLillis/tree-sitter-zsh/tree/native-dsl-translation
- markdown (tree-sitter-markdown): https://github.com/tree-sitter-grammars/tree-sitter-markdown/tree/native-dsl-translation
  - feature-gating: https://github.com/tree-sitter-grammars/tree-sitter-markdown/tree/native-dsl-feature-gating
- markdown_inline (tree-sitter-markdown): https://github.com/tree-sitter-grammars/tree-sitter-markdown/tree/native-dsl-translation
  - feature-gating: https://github.com/tree-sitter-grammars/tree-sitter-markdown/tree/native-dsl-feature-gating
- wgsl (tree-sitter-wgsl): https://github.com/WillLillis/tree-sitter-wgsl/tree/native-dsl-translation
- wgsl_bevy (tree-sitter-wgsl-bevy): https://github.com/tree-sitter-grammars/tree-sitter-wgsl-bevy/tree/native-dsl-translation
- css (tree-sitter-css): https://github.com/tree-sitter/tree-sitter-css/tree/native-dsl-translation
- scss (tree-sitter-scss): https://github.com/tree-sitter-grammars/tree-sitter-scss/tree/native-dsl-translation
- verilog (tree-sitter-verilog): https://github.com/tree-sitter/tree-sitter-verilog/tree/native-dsl-translation

  </details>

#### One commit?

I primarily developed  this on one branch, and it ended up being just short of 500 commits. Many of these changes were experiments, refactors, backtracks, and other things not relevant to our main tree. 

#### Why now?

This is a large feature, and while I've given it as much care and attention as possible, it still needs a significant amount of time to bake on master before we include  it in a release. I could have easily spent another 5 months on this, but it's in a good state now and ready for other people to take a look and try it out. I encourage anyone who's interested to build this for yourself locally (or get the nightly build after this is merged) and try it out!

####  That's a lot of code, is it slop?

LLMs were used to aid in the development of this feature, but I've carefully read every line, and written a large chunk of this myself. This was developed in a little over 5 months, any mistakes made  are my own.

#### How was it tested

- There is a thorough test suite committed in tree. 
- I used rust bindings to LLVM's libfuzz to eliminate panics
- The large number of already-translated grammars (listed above)
- *Lots* of manual testing

#### Anything else I missed

This was a large undertaking, and it forced to me to learn a lot. While I've been as thorough as possible, it's impossible not to miss things at this scale when working solo. A large portion of 0.28's development time will be spent on  finalizing the language, the LSP, and other related bits. I will be spending a significant amount of time cleaning up this branch further before anything is merged.

#### Post-merge enhancements

- I spent a lot of time working on the error ergonomics, but I'm sure there will be a long tail of error messages  that can be improved as users report them. Missing "Defined here" notes, actionable suggestions, etc. For example, one pending improvement is validating regexes in the DSL  rather than in generate's shared backend, yielding significantly improved error messages with source locations. For example:

image

- The DSL currently terminates after its _first_ error. After merging this, I plan to work on an "error sink" design that will collect all errors within a given "pass", so that multiple actionable items can be presented as once. This will be most helpful to LSP users.

- Settle discussion on a stdlib of common functionality (i.e.  a `comma_sep` macro, common regexes, etc.) that we can extend in this  and further releases.

- Settle on available vendoring options/guidance for inherited grammars
	- submodules (please no)
	- "just copy/paste the code" (works but too adhoc)
	- certain blessed fetch/pdate workflows performed by the LSP. A grammar lockfile and/or additions to tree-sitter.json

### AI Policy

AI tools were  used extensively during development of this branch. Development spanned several months, so it used every Claude Opus from 4.7 to 5, Fable, and gpt5.6-sol. LLMs were used to bounce ideas off of, generate code in some  places,  and to review. PR description (and  all of the typos I've surely missed) entirely written by me.


clason:
Note for downstream users: This only replaces the "backend" that converts a grammar definition to the "IR", which is `grammar.json`. As long as that file is committed to the repo (which is recommended), the `parser.c` can be generated from that via `tree-sitter generate src/grammar.json` (which, again, is recommended) regardless of whether the grammar is specified as JS or TSG. The grammar language thus _only_ matters for grammar developers and contributors, not for users!

maxbrunsfeld (changes requested):
We should have a check-in about this synchronously. Please don't proceed with this yet; it's an absolutely massive commitment to introduce a custom surface language for every grammar.

My initial reaction is -
* I agree we should stop using the NPM package registry for dependencies between grammars, and finish decoupling the high-level tooling from `package.json` and NPM in general.
* I don't agree that JS is an unstable surface language in general. I think that's approximately the *exact opposite* of the truth - JavaScript is objectively by far the most well-known, rigorously-standardized and invested-in programming language that has ever existed. And that's only becoming *more* true over time.
* Because JavaScript is the world's most standardized and popular programming language, every programming tool in existence can open a Tree-sitter grammar, syntax highlight it usefully, and do things like display a structural outline of the symbols in the grammar. Every coding LLM, large or small, can fluently read and write Tree-sitter grammars. There's an industry-standard documentation format (typescript) for the formal behavior of every function in the Tree-sitter DSL.
* JS grammar evaluation is not a significant part of parser generation time, there's no performance reason to switch.
* On developer experience - the fact that you can easily extract constants and helper functions in JavaScript is super useful. I see you created a typed `macro` system to preserve this functionality in TSG. That would mean that everyone who wants to hack on a Tree-sitter grammar needs to go learn your weird custom typed macro system!

What types of errors can we catch with a custom LSP that can't be caught as TypeScript type errors by the standard LSP that's built into the TypeScript compiler, and already integrates with all programming tools?

wetneb:
I find the overall idea nice as I've been somewhat annoyed by the dependency on the Node ecosystem myself, but I also find @maxbrunsfeld's points convincing.

Coming more from a process perspective, I wonder if the improvement was discussed and designed in any public discussion space? It's fairly common in open source projects to expect some preliminary discussions in the open before such large changes are implemented. It's often helpful to avoid disappointment after massive efforts are put in a direction which ends up not being consensual (I don't know if it's the case here).

Whether or not open discussions are expected before large changes, I would find it useful to document that fact, and other aspects of the governance of this project (which roles there are, who has them, how do people get in and out of those roles). I'd be happy to help document the current status quo if that's helpful.

I'm aware that @WillLillis is a maintainer and that I'm not part of the team, so apologies if it feels like an inappropriate meddling.

WillLillis:
Sorry for the delayed response, I'm not able to post here while at my day job.

> We should have a check-in about this synchronously.

That would be great. I'm free most weekends, as well as after 5:30PM EST most weekdays.

> * I agree we should stop using the NPM package registry for dependencies between grammars, and finish decoupling the high-level tooling from `package.json` and NPM in general.

The problem with this, though, is that there has to be a concrete reason in order for people to spend the time migrating existing grammars. I think TSG as it is now (there's a few more things I'd like to push up as I've discussed with clason) in addition to the tooling it unlocks for us down the  road is a strong enough  argument. I don't think simply eliminating `package.json`/NPM is a strong enough lure for most grammar authors. This is just my opinion, however.

> * I don't agree that JS is an unstable surface language in general. I think that's approximately the _exact opposite_ of the truth 

I'd disagree here. I brought up the ESM/CJS difficulties above. tree-sitter-vim's migration was a particularly hacky mess, see https://github.com/tree-sitter-grammars/tree-sitter-vim/pull/47. I'm sure Clason can list many others that we've dealt with/had to punt on during his maintenance of nvim-treesitter and the tree-sitter-grammars org.  

>  - JavaScript is objectively by far the most well-known, rigorously-standardized and invested-in programming language that has ever existed. And that's only becoming _more_ true over time.

This is true, and this point is more of an opinion, but why do  we need a fully-fledged programming language to  generate a grammar description? A large part of the work here was spent studying grammars and seeing what common  patterns emerged. Most are very simple, but a few outliers (vim, haskell, verilog, etc.) have become much more  than necessary for grammar descriptions. I think there's a lot of value in pushing people towards the [pit of success](https://www.computerenhance.com/p/pit-of-success).

> * Because JavaScript is the world's most standardized and popular programming language, every programming tool in existence can open a Tree-sitter grammar, syntax highlight it usefully, and do things like display a structural outline of the symbols in the grammar. 

The  0-30 tooling-wise here is a decent argument, but I seriously think we're losing out on that remaining 30-60 for grammar author facing tooling. Evaluating a JS module into a JSON blob loses all source provenance, and makes it impossible to develop LSP-like tooling that handles every grammar without handling all of JavaScript first. And again, to access this tooling, we are requiring grammar authors to pull in _all_ of JS's tooling. It's a very heavy dependency and deters potential contributors. Javascript tooling helps with the _easy_ part of writing the grammar, but offers almost nothing regarding tree-sitter specifics (outside of the TS signatures for our builtin combinators). In addition, the vast majority of grammar developers have an investment in their language/grammar alone, with no interest in JS. 

> Every coding LLM, large or small, can fluently read and write Tree-sitter grammars. 

A large part of the effort for TSG was to make it as simple and minimal as possible. Extract the common patterns from all the grammars out there, and condense it into a simple language that you can pick up in an hour tops. I guarantee that any LLM capable of working with the JS DSL is perfectly capable of working with TSG. This goes for people as well. 

> * JS grammar evaluation is not a significant part of parser generation time, there's no performance reason to switch.

Agreed, the generation time performance here is not a reason to switch, and I'm sorry if my initial description of the PR sounded like it was  making  that case. The performance of the DSL here is a nice-to-have, but not a requirement. The motivation for optimizing it was so that we can develop quality tooling _on top of_ the language. Having near instant errors,  warnings, references, and other points of introspection make grammar writing a much more pleasant activity. Down the line, we'd like to develop better linting tools to improve grammar quality. If this tooling isn't easy to use, then it won't be used.


> * On developer experience - the fact that you can easily extract constants and helper functions in JavaScript is super useful. I see you created a typed `macro` system to preserve this functionality in TSG. That would mean that everyone who wants to hack on a Tree-sitter grammar needs to go learn your weird custom typed macro system!

The entire language is statically typed, it's just that types were easily inferred everywhere besides macro arguments. I would really encourage you to take a look at the macro system before dismissing it  like this. It's incredibly minimal, takes 5 minutes to learn, and fits the vast majority of uses that helper functions  currently fill in the JS  grammars. Besides macros, `let` bindings can also be re-used to share constants.

> What types of errors can we catch with a custom LSP that can't be caught as TypeScript type errors by the standard LSP that's built into the TypeScript compiler, and already integrates with all programming tools?

I don't think the type safety argument here is a particularly strong point for TSG. Types have to exist at _some_ point, so I made the language  statically typed so users could have better errors earlier on in the pipeline. The benefits I'm trying to emphasize with this DSL are its minimalism, tighter integration with tree-sitter focused tooling, and a much lighter dependency chain.

----

I thought carefully about all of these questions before starting this work. I also discussed the plans for this with both @clason and @amaanq who both agreed this was a worthwhile  direction to pursue. I want to emphasize again, the majority of the work done here was in condensing the common patterns that tree-sitter grammars have settled on into a small language that expresses those ideas directly with less footguns for larger grammars to fall  into. This was the large open question when  I started, and if it wasn't doable I wouldn't have pursued this/opened this PR.

Something I want to re-state here, for emphasis. This does not affect generation of JS-based grammars at all. TSG is intentionally very minimal, and if it's not expressive enough for an author's wants/needs, the JS DSL will always  be there.

Happy to discuss this further with you here and/or offline. :slightly_smiling_face:

amaanq:
My initial thoughts to Max's concerns are outlined below. Note that I haven't looked at this too carefully or been following along all too closely, but Will has given me updates throughout the development of this large feature, and to me, it seems like a huge net positive for the ecosystem.

> I don't agree that JS is an unstable surface language in general. I think that's approximately the exact opposite of the truth - JavaScript is objectively by far the most well-known, rigorously-standardized and invested-in programming language that has ever existed.

I do agree, objectively this is true. JavaScript is one of the most widely used languages out there, and everyone knows it to an extent, humans or LLMs alike. Anyone can kind of look at a grammar and sorta get what's going on - one could argue there is real value in that.

However, what I disagree on is how tree-sitter uses JavaScript. That is not what I would consider typical JavaScript code; if you were to take a JS dev and drop them into a grammar to look at and review, they would be very lost for some time. The JavaScript _we_ use is really more like its own mini-DSL piggybacking off the JS runtime. It takes time and effort for any developer to learn and be effective with it. The grammar object is self referential in an unorthodox manner, and this reflects in the tooling...

> Because JavaScript is the world's most standardized and popular programming language, every programming tool in existence can open a Tree-sitter grammar, syntax highlight it usefully, and do things like display a structural outline of the symbols in the grammar. 

...which brings me to this point. No JS LSP out there can properly show you references or definitions of grammar rules, which IMO, is a bare minimum these days (what language doesn't have a good enough, open source LSP? besides kotlin :laughing:). Additionally, I even worked with some well-known TypeScript field experts/wizards to get recursive self-referential methods to work in the `.d.ts` so the TypeScript LSP could be better. It is just not possible, unfortunately! 

To illustrate, in a grammar like this:

```js
rules: $ => ({
  expression: $ => choice($.identifier)
})
```

a JavaScript or TypeScript LSP can understand the syntax, but it does not understand that `$.expression` refers to the grammar rule named `expression`. It cannot reliably provide definitions or references for grammar rules. [(see upstream issue)](https://redirect.github.com/microsoft/TypeScript/issues/47599)

Not only can it not do the bare minimum of easing the cognitive load of understanding the shape of the grammar for the user, but also it misses out on the additional plethora of features we _could_ have, such as the REPL, which exists, and a state count/graph breaking down each individual rule, thus helping grammar authors more easily understand which rules or parts of a rule are causing significant issues. So:

> What types of errors can we catch with a custom LSP that can't be caught as TypeScript type errors by the standard LSP that's built into the TypeScript compiler, and already integrates with all programming tools?

there are clearly a lot of errors or problems that just can't be caught with JS LSPs, as I've outlined above. Additionally, suppose we decided to instead work on a "major version" update to the JS DSL....that just doesn't make sense to me if we can do far better in our own DSL.

> JS grammar evaluation is not a significant part of parser generation time, there's no performance reason to switch.

I also agree. The perf gains, in my eyes, are just a nice cherry on top; to be clear, the performance gains are totally moot! I think the biggest sell for this PR is the massive upgrade in developer experience. I also find the DSL to be simpler than the JS "DSL", and my prediction is that eventually when this becomes the standard, it will end up being easier to read and write for both humans and LLMs.

---

I know I've reiterated a lot of points from Will's comment above, however, I really wanted to highlight the ones that particularly matter to me. I've probably authored the most amount of grammars out of anyone here, and I will say, the current DX with writing grammars is not great, to put it nicely, and is pretty abysmal if I'm being honest. Sure, we could add some stopgaps here and there to marginally improve it, but that's not what I would be looking for if I were approaching the project with fresh eyes. 

I like fast, instant, feedback in the editor as I'm typing (hint, hint) or as I save the file; it really matters to the end user, and it matters more so for those nowadays using AI to write most, if not all of their grammars - it's hard for them to get good feedback on the grammar itself without digging deep into some of the CLI's debugging commands, or even the grammar.json/parser.c itself.

maxbrunsfeld:
Hey @amaanq @WillLillis, this is a bit of a busy week at Zed, so not able to spend as much time on this as I'd like, but I do want to say, I agree 100% about the importance of some of the problems you are raising: grammars should 100% support go-to-definition, find-all-references, rename, type info on hover, diagnostics, and auto-format.

That said, I think there's a very large conceptual and annoyance cost to introducing a new language for this. Personally, as a user, I hate having to learn "some guy's random made up language" just to solve a problem that could have been solved by a well documented set of functions in a more standard language. The amount of feedback over the years that I've received to the effect of "It was so approachable and delightful that I could just write the grammar using some JS functions" is more than I can count. It's an important part of Tree-sitter's minimalism and economy of concepts that has lead to its adoption over more than a decade. It's not just some random decision.

However, the DSL as it exists now was designed before LSP came out, before the JS language developed a number of features that are now universally standardized across browsers and basically all JS runtimes. JS now has a standardized module system, imports and named exports. My belief is that there's a very simple "v2" we could make of the grammar DSL that would not rely on Proxies (or anything exotic) at all, and would completely eliminate all of these problems that you're citing, allowing for all of these IDE features to work, using existing language servers that most programming tools already ship with support for out of the box.

I can prototype something so that we have something concrete to discuss, but it might take me a couple of days to do it.

maxbrunsfeld:
Ok, here's an initial prototype:

core: https://github.com/tree-sitter/tree-sitter/commits/prototype/module-grammar-dsl/
rust grammar: https://github.com/tree-sitter/tree-sitter-rust/blob/prototype/module-grammar-dsl/grammar.mjs
c grammar: https://github.com/tree-sitter/tree-sitter-c/blob/prototype/module-grammar-dsl/grammar.mjs
c++ grammar (via inheritance): https://github.com/tree-sitter/tree-sitter-cpp/blob/prototype/module-grammar-dsl/grammar.mjs

In this rev of the DSL, grammar rules are just normal variables, and they're assigned names using standard export syntax. So if you open these in an editor like Zed or VS Code, you get go-to-definition, find-all-references, rename, auto-format, diagnostics on every keystroke, type-info-on-hover, etc, out of the box, via either of the good TypeScript language servers: the new official one, or vtsls.

clason:
Sorry, but this does not address the two core issues here, which (having ~wasted~ spent years on full-stack maintenance from Neovim, nvim-treesitter, tree-sitter-grammars, to here) I could not agree with more and would like to reiterate more forcefully:

> the vast majority of grammar developers have an investment in their language/grammar alone, with no interest in JS.

I cannot stress enough that outside a certain bubble, people consider JS a cancer to be avoided rather than a success to be emulated. Just from a single glance, the custom DSL is _much_ cleaner and easier to reason about for grammars than your JS DSL v2. Having to learn a new DSL for my grammar, I don't know why I would ever choose the latter over the former.

> the current DX with writing grammars is [...] pretty abysmal

This is by far the biggest (but not only) pain point with tree-sitter in the real-world. While JS is easy to pick up and write "baby's first grammar" in, there is virtually _zero_ help -- be it documentation or tooling -- on how to write a _good_ grammar for a reasonably complex language. And you can observe that in the wild (if you actually engage with the wider ecosystem): Out of the ~400 grammars I've had the (dis)pleasure to look at and work with, the vast majority got abandoned after hitting the first or second cliff beyond the "easy 40%". And of the remainder, about half have brute-forced their way through, producing veritable Rube Goldberg machines of staggering ugliness -- often by resorting to needless JS features ("MOAR MODULES!") to manage the complexity instead of structuring the _grammar_ itself properly (which would have also fixed the concomitant performance and/or size issues with these grammars that people keep complaining about).

Again, this has _nothing to do with JS_ but with how tree-sitter works internally (and opaquely) **and no generic JS/TS tooling can help with that**. On the other hand, as @WillLillis has already demonstrated, his DSL allows **tree-sitter specific** tooling that can.

So his (not your) DSL would be a _huge_ boost to the wider tree-sitter community, and (on a personal note) your unwillingness or inability to engage with the substance of these arguments (which two of your fellow maintainers have spent a lot of effort to make more eloquently) leaves quite a bad taste in my mouth. (And I will have to think very hard about possible downstream consequences.)

clason:
This is certainly an improvement over the status quo, but it does not address the two core issues here, which (having spent years on full-stack maintenance from Neovim, nvim-treesitter, tree-sitter-grammars, to here) I could not agree with more and would like to reiterate:

> the vast majority of grammar developers have an investment in their language/grammar alone, with no interest in JS.

I cannot stress that enough. The popularity of JS is limited to a specific segment of software developers, which tree-sitter has transcended long ago. (In fact, many if not most grammars are written by enthusiasts or researchers with no professional software development background at all. I consider that a strength of tree-sitter.)

> the current DX with writing grammars is [...] pretty abysmal

This is by far the biggest (but not only) pain point with tree-sitter in the real-world. While JS is easy to pick up and write "baby's first grammar" in, there is virtually _zero_ help -- be it documentation or tooling -- on how to write a _good_ grammar for a reasonably complex language. And you can observe that in the wild (if you actually engage with the wider ecosystem): Out of the ~400 grammars I've had the (dis)pleasure to look at and work with, the vast majority got abandoned after hitting the first or second cliff beyond the "easy 40%". And of the remainder, about half have brute-forced their way through, producing veritable Rube Goldberg machines of staggering ugliness -- often by resorting to needless JS features ("MOAR MODULES!") to manage the complexity instead of structuring the _grammar_ itself properly (which would have also fixed the concomitant performance and/or size issues with these grammars that people keep complaining about).

Again, this has _nothing to do with JS_ but with how tree-sitter works internally (and opaquely) **and no generic JS/TS tooling can help with that**. On the other hand, as @WillLillis has already demonstrated, his DSL allows **tree-sitter specific** tooling that can.

maxbrunsfeld:
> and no generic JS/TS tooling can help with that. On the other hand, as WillLillis has already demonstrated, his DSL allows tree-sitter specific tooling that can.

I'm all for building Tree-sitter specific tooling. But you're making it sounds as if that's impossible (or prohibitively difficult) to do, without developing our own *custom programming language*. That part is factually not true. I would like us to get on the same page about that.

The JS DSL doesn't need to have the dynamism that it currently has. We don't need to allow path-based imports. With the fairly small change that I proposed on the branches above, the DSL becomes quite statically analyzable. Statically analyzing modern JavaScript is, again, one of the most thoroughly solved compiler problems in the software industry. There is a huge set of mature tools available.

clason:
We are talking past each other. I do not disagree that the current DSL can be improved in order to facilitate better JS tooling and that some people would welcome this. My points are that

1. this would be completely orthogonal to better _grammar_ tooling: "grammar analysis" features like the REPL or live state count neither require nor profit from standard JS LSP features like static type analysis or goto definition;
2. a large segment (arguably the majority) of grammar developers need/want the former and not (or at least much, _much_ more than) the latter.

@WillLillis proposal directly addresses these points, which is why I think it is worth the (admittedly large) cost (while "just" better JS LSP/static analysis features are not, because they are not the real bottleneck).

With this, I think I have said all I can usefully say and will step away now.

maxbrunsfeld:
> We are talking past each other.

I'm trying to keep the discussion a bit more concrete. Here is how I see it. Let know where you disagree:

"Developer experience" can be divided into several categories. Here are four important ones:
1. documentation - explaining how to write grammars in prose
2. simple syntax-only features (syntax highlighting, symbol outlines)
3. standard IDE features (rename, go-to-definition, find-all-references, diagnostics, type info)
4. tree-sitter-specific editor integration (showing parse state count inline, integration with a REPL)

Claim: We should provide the best support possible, in **all** of these four categories, to **as many users as possible**.

Now, there are at least two possible changes we could make in order to achieve this goal:
1. Change the JS DSL to use modern static JS constructs instead of dynamic ones
2. Create a custom formal language, *TSG*, that is specific to Tree-sitter with its own syntax and type system

We can consider the pros and cons of the two proposals, with respect to each kind of "tooling" separately:

1. For documentation
   * Either way, we have to explain the *concepts* that make up grammars
   * If we evolve the JS DSL, we need to explain how those concepts map to Tree-sitter's JS functions. The language itself is standard, and explained quite thoroughly elsewhere. Again, it's objectively the most standardized, commonly known programming language.
   * If we create our own language, we need to have some kind of specification of how that language works.
       * What can you do in the type system?
       * What are the semantics of macros?
       * How exactly do `for` loops work?
       * What are the semantics of the `::` operator?
       * What are the built-in data types?
   * **Comparison** - On the documentation front, it's not even close. A custom programming language has orders of magnitude more documentation surface than a set of ~10 JS functions. The up-front cost of trying Tree-sitter grows massively.
2. For syntax highlighting and code outlines:
   * If we stay with JS, all tools support tree-sitter grammars automatically
   * If we move to a custom DSL, approximately no tool will ever support Tree-sitter grammars out of the box. But we control our own syntax, so we have the option of making slightly more terse syntax for certain constructs. For example, you can have a `rule` keyword. To me that's quite minor, but it I'm including it for completeness.
   * **Comparison** - For this simple category of tooling, it is clearly better to use JS. Many important tools will **never** add syntax-highlighting support for TSG, but every tool has JS syntax highlighting. 
3. For the standard LSP features (go-to-definition, diagnostics, etc):
   * If we stay with a JS DSL, all editors will basically get mature implementations of these basic features out of the box
   * If we move to a custom language, you will have to install a custom language server to get these in any editor.
   * **Comparison** - For standard IDE features, more users will experience them if we stay with JS. Some users will install and configure a custom LSP for Tree-sitter. Many will not.
4. For custom Tree-sitter-specific tooling (which I agree is valuable!)
   * If we stay with JS, we can still build a custom Tree-sitter grammar language server. It would simply use standard, existing JS analysis tools. If the grammar DSL is based on static features of JS, as in the proposal I linked above, there would be 100% reliable ability to resolve a Rule by name to a given range of source code that contains that rule's definition. And we wouldn't need to build our own parser, as there are already a variety of mature options.
   * If we move to a custom language, we can build whatever tooling we want, by implementing our own parser and type system.
   * **Comparison** - There are no blockers to building this category of tooling **either way**. The editor integration would be slightly different, because in one case, it would trigger solely off of the file extension `.tsg`, but in the JS case, there would need a slightly more specific activation mechanism, activating when you open a `grammar.mjs` file, or perhaps checking for the presence of `tree-sitter.json`. But that's trivial in most editors.

**Overall conclusion** - We should simply make the JS DSL use static constructs, and then focus on building awesome Tree-sitter-specific tools. In the short, medium, and in the very long term, it continues to be a clearly better solution for end users. There are a few more punctuation characters in grammars, but that doesn't really matter very much, compared to the overwhelming benefits of *not* using a custom formal language.

Again, please let me know what part of this you disagree with, or what you think I have left out.

lewis6991:
Sorry about adding another cook to the kitchen, but I do want to chime in here because I'm pretty strongly in favour of a dedicated grammar language.

I think we're putting too much weight on the conceptual cost of a custom language. Dedicated languages are pretty normal for parser generators. We already have a grammar DSL with its own semantics, JS is just how we currently express it. When I first started using Tree-sitter, the JS was actually the confusing part. I understood it as a C library with a Rust parser generator, so my reaction was basically "what does JavaScript have to do with this?" I don't know JS particularly well either, and I would rather learn a small language designed for this problem than learn enough JS to understand an embedded DSL.

I also don't find ~10k LOC particularly alarming by itself. The maintenance burden matters more than the raw size, and having over 100 real grammars translated seems like pretty good validation of the language surface. I also see Tree-sitter owning the whole stack as a benefit. There are fewer external dependencies, we control the compatibility boundary, and the language and tooling can evolve together.

I don't put much weight on generic editor support being available everywhere out of the box. Grammar authors are already a specialised audience. Good tooling is extremely important, but the valuable tooling here is tooling that helps people write good Tree-sitter grammars. Things like conflicts, ambiguity, state counts, grammar structure and the effects of changes matter much more than generic JS support.

This is also where I think JS becomes a disadvantage. Grammar authors are currently free to use arbitrary JS functions, helpers, imports and metaprogramming. We cannot really assume a nice static subset when building grammar tooling. We could try to restrict what JS is allowed, but then we either break existing freedom or effectively define a Tree-sitter-specific subset of JS anyway.

A dedicated language gives us that restricted and analyzable model directly. If we need to build Tree-sitter-specific tooling regardless, I'd rather build it around a small language whose semantics we completely control than try to recover those semantics from arbitrary JavaScript.

maxbrunsfeld:
> This is also where I think JS becomes a disadvantage. 

Again, there are real concrete advantages to this. Helper functions and things like `Array.map` are [used](https://github.com/tree-sitter/tree-sitter-bash/blob/master/grammar.js#L1199-L1201) [widely](https://github.com/tree-sitter/tree-sitter-c/blob/b780e47fc780ddc8da13afa35a3f4ed5c157823d/grammar.js#L1382) in grammars. They're great! Parser combinator libraries are widely popular for a reason, and many of the LR-style parser generators before Tree-sitter are not well loved! Tree-sitter gives you some of the abstraction ability that parser combinators give you, in an LR generator, and that's been an important part of its design since the beginning.

Meanwhile, nobody has yet named a concrete disadvantage of allowing this kind of thing. That doesn't mean there isn't one, but if you are going to weigh in, please consider trying to name a *concrete* feature that we'd like to be able to implement that we can't implement with a JS DSL, and that we can implement with a hand-rolled language. Because right now, there's a lot of vague frustration with JS in general that is reading as just somewhat misinformed about what is and isn't possible with JS, and unrepresentative of the general user base.

wetneb:
Let me try to give concrete examples of some static checks I've been thinking about implementing on tree-sitter grammars, where I've been held back by the Javascript layer.

### Unreachable rules

If you define a rule that's not reachable from the root `source_file`, tree-sitter will prune it at the parser generation stage. It would be convenient to be able to highlight such rules in an IDE, or at least point to the line number where they are defined when emitting a warning about those rules being pruned out, as the fact that the rule is being pruned out might be a symptom of an oversight.

Something like this:
```bash
$ tree-sitter generate
WARN: grammar.js:79: rule `list_comprehension` is unreachable from `source_file`
```

If I had to do this today, I'm not quite sure how. The evaluation of the Javascript file would need to record the location of the definition of rules, or parse the `grammar.js` after the fact to find the said rules, but a priori I can't assume that they are written as the usual `export default grammar({ … })`.

Maybe if each rule is exported as its own symbol as in @maxbrunsfeld's new prototype it can somehow be traced more efficiently? You'd still need to be able to handle [a lot of possible ways to `export` such symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export), so you'd need to make tree-sitter depend on some Javascript static analysis tool to obtain those line numbers, I guess? Or can I somehow use the off-the-shelf NodeJS runtime to determine on what line a symbol exported in a JS file is defined?

### Warnings for constructs incurring large state counts

If I've understood things well, expressions such as
```
seq(
  optional($.foo),
  optional($.bar),
  optional($.baz),
  optional($.ouch)
)
```
can be detrimental to the state count of the parser, and the following laxer variant can be preferable:
```
repeat(choice($.foo, $.bar, $.baz, $.ouch))
```
(apologies if that's an inaccurate description - it's just meant to be an example of a possible optimization inside the definition of a rule)

It would be really nice if tree-sitter could warn me of places where this optimization makes sense. Maybe with a dedicated command such as:
```
$ tree-sitter analyze
WARN: grammar.js:132: expensive `seq` with optional contents requiring 64 states, consider replacing it with `repeat(choice(…))`
```

For a feature like this (with the reporting of the line number) you'd need to be able to locate the `seq(…)` expression, which is possibly deep inside a rule declaration (or an auxiliary Javascript function?), and probably do so at a relatively late stage of the parser generation if you want to display state counts. It's probably not so easy to implement even with a custom DSL, but with a grammar obtained by interpreting Javascript, I am a bit at a loss…

If the number of states isn't required to be reported, I guess such warnings could be implemented in Javascript directly, with a sort of `console.warn()` called inside the implementation of `seq`, assuming such calls are able to emit their stack traces that could be used to trace this back to the rule calling `seq`… but isn't that a bit hacky?

### Linting

It could be nice to have a linter which could flag inelegant constructs and fix them, applying some simplifications for you. For instance:
`choice(choice('foo', 'bar'), 'baz')` -> `choice('foo', 'bar', 'baz')`
or
`seq('foo')` -> `'foo'`

I could probably hack something that would work okay-ish with a tool like `astgrep`, writing down a list of such replacements to be applied on a Javascript file.
But because those `choice(…)` and `seq(…)` are actually function calls, I imagine this would be brittle: those functions could have been redefined by the user in some scope, changing their meaning and rendering my patterns incorrect.
Maybe not a huge deal? But not a very sound basis to build reliable tooling on, no?

aheber:
I don't maintain this project and I'm grateful for everyone who puts in significant time and effort. I understand this was a large effort and thank you for caring. If you all decide to go forward with this then I can support that.

I particularly like the [ELM style error messages](https://elm-lang.org/news/compiler-errors-for-humans). I do see that as a significant improvement.




I'm just a grammar author. I don't have a problem with Javascript and when I wrote my first grammar I was very excited to be able to use JS. I felt it made it easier for me. That is coming from someone who already had experience with JS but as was said earlier, there is no programming language more common than JS even if everyone likes to complain about it.

If you've gotten significant feedback that JS is making it hard for people to be successful I haven't seen it. I'm not saying nobody mentioned it but it isn't what comes up often. I've read issues here and been in the Discord chat for years. I think the common complaints on those channels are about different things. If you collected feedback from grammar authors I'm sorry I missed providing my feedback until now.



The idea that there will be two ways to write a grammar worries me. When I've seen other software do this one of those ways always falls behind in documentation, capabilities, or both. I anticipate a future where the docs talk about TSG but there is a secret "hack" where you can write your grammar in JS if you need to. That doesn't appeal to me and I think it would be more confusing. Especially if over time the "hack" gets more and more hacky.

I do a few weird things with my grammars. I would have to look further into the DSL to know if I would run into any problems converting.

WillLillis:
I apologize that I haven't been very responsive in here the last few days, if I had known how controversial this was outside of the normal maintainer team I would have picked another time to propose this and  gone about it quite differently. That being said, others have done an excellent job expressing why this work was pursued and the benefits it would bring.

I still believe that:

- Having a dependency on JS/ a JS runtime is a liability, whether that becomes realized or not is yet to be seen.
- While JS is a very popular language and this makes grammar authoring easier to begin for those already familiar with it, I would still hold that allowing _all_ of JS also allows poor uses of it when the task at hand is authoring grammars. I do not understand the draw of its universality in relatively trivial pieces of tooling such as basic highlighting, symbol views, etc.
- I'm not sure why these two separate approaches can't coexist, leaving the choice up the grammar author.

However, the _main_ argument for  this work was always improved tree-sitter _specific_ tooling. The grammar ecosystem is widely community driven, which is great! There is no way a single team could produce the wide breadth of grammars that the incredible tree-sitter community has over the past years. The issue we continue to run into, however, is that a lot of grammars have serious issues. Excessively large state counts, poorly structured constructs, and half-baked build time config selection being a few examples. I'm sure @clason could share more from his tireless years spent working closely with said ecosystem via nvim and nvim-treesitter.

This is not a condemnation or dig at any of these grammars, their authors, or anyone else involved. The main issue I and others have observed is that there is no clear means for improving said issues. Outside of corpus tests and reading `parser.c` (which can be incredibly difficult to grok due to its large size), what is an author expected to do? How are they even supposed to know that there's an issue? If no such tooling aids had been developed in the 13+ years of the project, why is that?

The main technical issue I saw was the separation between the JS frontend of the crate and the Rust backend, which today are connected by the `grammar.json` "IR". These two domains have no means of communicating outside of JSON serialization, which I saw to be the main barrier. Any tooling that wants to provide actionable information from the backend needs information from the frontend which isn't provided (and based on @amaanq's [earlier comment](https://github.com/tree-sitter/tree-sitter/pull/5944#issuecomment-5672868239)), cannot be provided in the DSL's current form.

I have my doubts about how well a JS DSL 2.0 could be rolled out without breaking the [vast quantity](https://github.com/tree-sitter/tree-sitter/wiki/List-of-parsers) of grammars already in the wild. However, assuming such a thing could be done (and it would be great if that was the case!), the main issue I still see if this process boundary/JSON serialization layer between compiler stages. The motivation behind developing the TSG within the same crate as `generate`'s backend was removing this boundary so that _any_  arbitrary tooling could be developed.

Any functionality the `generate` crate uses would be a function call away. Everything is kept in sync trivially because a mismatch means a trivially caught  compiler error. The alternative I see for this in the current approach is using the same JSON serialization layer as we do today, which opens us up to a large amount of potential runtime issues (panics, silent corruption, or otherwise) that hopefully we could prevent in testing. I do not see a way for a more direct, in-process connection to be made with the current architecture. On the other hand, with TSG any new functionality is at most a `pub(crate)` away. In addition, any information communicated over this process boundary has a real cost in memory, walltime, and of course code maintenance (in a project historically very short on maintainer time).

All of that being said (I appreciate anyone who's  been willing to humor my ramblings up to this point), the final call is with you @maxbrunsfeld. I don't see this conversation continuing in a productive direction, and in fact it only seems to be frustrating everyone involved (including myself). If you really think this kind of feature is unworkable and would be net negative for the project, feel free to close this PR out. The only thing I'd ask is if you could be a little more active here (or a private channel outside of GH) to give your thoughts and opinions more readily. I would sincerely prefer to not waste this amount of time again.

-----

Thank you to everyone who shared their thoughts here, especially @clason, @lewis6991, and @amaanq.

maxbrunsfeld:
Will and I had a video call about this. TLDR - we're not going to switch language at this time.

---

This is tough because I don't like being an obstruction to creative, careful, good engineering work by people who care a lot about Tree-sitter and its users.

The decision to use an embedded JavaScript DSL for grammar development has some unique benefits whose importance have been validated by substantial experience - writing the foundational grammars, hand-onboarding the first other grammar authors, and fostering the grammar ecosystem for the first ten years. Switching to a custom language would give these up, in a way that we couldn't get back.

In this thread, people have raised a lot of very valid pain points with the current grammar-authoring experience, and specific features that would help address them. Thanks @wetneb for the very specific things you've mentioned. Fortunately, I really think that all of the problems that have been raised in this thread are *very solveable* within a V2 of the JS DSL.

Some key properties of the new JS DSL that I prototyped on the branch that I linked:
* Because rules are named exports, all rules can be statically resolved to a known declaration location.
* Node.js is no longer a supported way of evaluating grammar JS files. When using the new edition, you must use the QuickJS in-process evaluation.
* No IO is permitted in grammars.
* Imports are limited to local helper files and other grammars that are explicitly declared in `tree-sitter.json`

I'll land this initially behind a `generate --unstable` feature flag or something so that we can experiment with it and see if it introduces any issues. Any grammar that switches over to the new edition will get full IDE support via the TypeScript language server, without requiring any action on the part of the user.

Then we can start experimenting with Tree-sitter specific IDE features. I'm personally fine with shipping a `tree-sitter lsp` subcommand built into the CLI, to provide some of this.

clason:
> I'll land this initially behind a generate --unstable feature flag or something so that we can experiment with it and see if it introduces any issues.

I think we don't need a flag with this; it's enough to dispatch on the filename argument if "new-style" grammars are named `grammar.mjs` (i.e., the "unstable opt-in" would be just `tree-sitter generate grammar.mjs` instead of `tree-sitter generate [grammar.js]` -- possibly with a fallback so `tree-sitter generate` would Do The Right Thing if `grammar.mjs` exists but `grammar.js` doesn't).

I'm also a bit wary of shoving even more into the CLI, but of course code reuse could make things sufficiently simpler so that it would pay off.

maxbrunsfeld:
> I think we don't need a flag with this

I do want to keep the idea of an explicit `edition` declared in your `tree-sitter.json` though, not just inferring it from the filenames that are present.

My only thing that was driving the idea of the CLI flag is, I'd like to be able to land this on master but be explicit about the fact that it's a new unstable feature for now. Maybe we could just print a warning to stderr initially, if we don't like the flag?

> I'm also a bit wary of shoving even more into the CLI

Yeah, I don't have a strong opinion here - just saying I'm not opposed to adding dev tooling into the main repo, if it's broadly applicable.

clason:
> My only thing that was driving the idea of the CLI flag is, I'd like to be able to land this on master but be explicit about the fact that it's a new unstable feature for now. 

That should be assumed; we have structured the development process precisely to allow for that. (`master` reports its version as v0.28, and you need to build from source or download nightlies to use that). We are also not in any hurry to release so we can iterate on `master` as long as we want.

If you want, you should add "experimental" to the docs and the CLI help; that should be more than enough. (The CLI will be used in CI, so `stderr` is very noisy and can trip up workflows so should be used only as a matter of last resort.)

(No objection to gating it behind a `treesitter.json` field, though! I'm just wary of adding yet more flags that will later be removed.)

maxbrunsfeld:
I see, fair. Let's just have `edition` be explicit, but no stderr warning, no flag.

WillLillis:
As Max summarized above, we had a very productive chat over how we can move forward with a new version of the JS-based DSL that captures that vasty majority of the benefits I was targeting here, without the pains of such a large switch. I am very optimistic about what we'll be able to do here with those changes.

I want to apologize for the extended absence here. Without going into too much detail, my family had an emergency and I haven't had the the time or energy to focus on anything else. I will return to regularly working on the project soon, but cannot say how soon yet.

Thanks again to everyone who contributed to the discussion here. I'm going to close this out, and (I assume) further discussion for the JS DSL and other concerns can continue in their own PRs/issues as needed.

